PR #106, opened 2019-03-24
Add tests for claris_randomizer._process_command

henriquegemignani:
These tests should reproduce the errors in #105.


SpaghettiToastBook:
Fixes the first two bullet points of #105, and I'm pretty sure it also fixes the third one (i.e. I believe #107 is caused by something else).

PR #129, opened 2019-04-08
Fix logic wanting to place extra starting items instead of taking a r…

henriquegemignani:
…andom action


henriquegemignani:
This is affecting #127 . This made logic unable to create a seed starting at ship with no items.

PR #218, opened 2019-05-10
Increase precision when showing average ammo per expansion

SpaghettiToastBook:


henriquegemignani:
Could you add an entry to CHANGELOG.md, please?

PR #285, opened 2019-06-15
Add trick for blind jumps

### LINKED ISSUE #282: Add trick for doing things without the matching visor
For example, using Dark Visor platforms without Dark Visor.

SpaghettiToastBook:
Fixes #282.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


SpaghettiToastBook:
I guess I need to update the permalink version?

henriquegemignani:
> 
> 
> I guess I need to update the permalink version?

Huh, since you can configure each trick, adding a new trick affects the permalink. Interesting.
Increasing the permalink version is mostly so we can easily detect that the permalink is from another Randovania version, in case we change the format. Since we also add a hash of the database, I think you can just replace the Permalinks used in the tests with a new one.

SpaghettiToastBook:
Okay, sounds good.

PR #308, opened 2019-06-18
Invisible objects trick

### LINKED ISSUE #282: Add trick for doing things without the matching visor

SpaghettiToastBook:
Fixes #282.

Does this count as a major change?


henriquegemignani:
I call things "major" when I want to highlight that change over others in the changelog. Nothing really special about it.

SpaghettiToastBook:
✓

PR #326, opened 2019-06-30
Add/change Windchamber Gateway requirements

gollop:


SpaghettiToastBook:
I think the SJ-only method should be Easy, but other than that this looks good.

PR #327, opened 2019-07-06
Add major/minor split option

### LINKED ISSUE #133: Option to shuffle items in major/minor pools
This is a feature of some Super Metroid randomizers. For this game we could shuffle major items and energy tanks in one pool and expansions in the other.

SpaghettiToastBook:
Fixes #133.

Probably needs a test or two, but other than that this should work!

The permalink version and the options file version were both bumped up.


randovania/game_description/resources/pickup_index.py
```diff
@@ -30,3 +30,68 @@ def __eq__(self, other):
     @property
     def index(self) -> int:
         return self._index
+
+    @property
+    def is_major_location(self) -> bool:
+        return self._index in _MAJOR_LOCATIONS
+
+_MAJOR_LOCATIONS = [
```
henriquegemignani: This should come from the data somehow.

A new field on PickupNodes that describes if they're major/minor? Better yet would be the actual original pickup there, but that can come later.

randovania/generator/filler/retcon.py
```diff
@@ -138,15 +140,18 @@ def action_report(message: str):
         if isinstance(action, PickupEntry):
             assert action in pickups_left
 
-            if num_random_starting_items_placed >= minimum_random_starting_items and current_uncollected.indices:
+            uncollected_indices = current_uncollected.indices
+            if randomization_mode is RandomizationMode.MAJOR_MINOR_SPLIT:
+                uncollected_indices = list(filter(lambda pickup_index: pickup_index.is_major_location, uncollected_indices))
```
henriquegemignani:
```python
uncollected_indices = [pickup_index for pickup_index in uncollected_indices if pickup_index.is_major_location]
```

randovania/generator/filler/retcon.py
```diff
@@ -138,15 +140,18 @@ def action_report(message: str):
...
+            if num_random_starting_items_placed >= minimum_random_starting_items and uncollected_indices:
```
henriquegemignani: Add a line break before this to add some vertical space

randovania/generator/generator.py
```diff
@@ -188,21 +189,34 @@ def _assign_remaining_items(rng: Random,
     """
 
     unassigned_pickup_indices = [
-        pickup_node
+        pickup_node.pickup_index
         for pickup_node in filter_unassigned_pickup_nodes(world_list.all_nodes, pickup_assignment)
     ]
 
-    if len(remaining_items) > len(unassigned_pickup_indices):
+    num_ETMs = len(unassigned_pickup_indices) - len(remaining_items)
```
henriquegemignani: `num_nothing` or `num_etm`.

SpaghettiToastBook:
In Super Metroid randomizer, there's a major location and a minor location (Hi-Jump Energy Tank and right side Supers in Wrecked Ship, respectively) that the randomizer treats as minor and major, respectively. I was thinking that we may want to do this here, although it's not a big deal either way.

henriquegemignani:
If you add a new PickupNode field for this, we can freely configure if a location is major/minor.

SpaghettiToastBook:
That's my plan!

SpaghettiToastBook:
Currently this doesn't work because `UncollectedState.resources` doesn't work.

SpaghettiToastBook:
Would it be a good idea to have a list of major item locations in the Welcome tab?

PR #340, opened 2019-07-10
Distinguish between Aether and Dark Aether in hints

### LINKED ISSUE #199: Hints: WORLD_ONLY should refer to sub regions
"Upper Torvus"/"Lower Torvus".
Also Light/Dark world.

SpaghettiToastBook:
Fixes #199 (maybe). I don't think any more detail should be added to the world-only hints, not only for balance, but because there isn't a clear way to divide all of the worlds further.


henriquegemignani:
There's Lower/Upper Torvus, Lower/Upper Sanctuary.

Agon could be Pirate base?
Temple Grounds I can't think of anything.

But for Balance, might be enough just dark/light worlds

SpaghettiToastBook:
Pirate Base is a bit nebulous, since the mining equipment seems to belong to them, and none of the Pirate stuff is on Dark Aether. I also can't think of anything for Temple Grounds.

I played a bit of a seed with just light/dark added and I thought it was the right balance of detail.

PR #371, opened 2019-08-25
Add three-word hashes to title screen

SpaghettiToastBook:
The "- " at the beginning of the seed hash string in the patcher file is admittedly a bit hacky.


randovania/layout/layout_description.py
```diff
@@ -24,6 +27,10 @@ def _playthrough_list_to_solver_path(playthrough: List[dict]) -> Tuple[SolverPat
         for step in playthrough
     )
 
+@lru_cache(maxsize=1)
+def _shareable_hash_words():
+    with (get_data_path() / "hash_words" / "hash_words.json").open() as hash_words_file:
```
henriquegemignani: Could the additional words be added to the database, and then this list be built on-the-fly from that database?

SpaghettiToastBook: I'm not sure what you mean...do you mean extracting the room name words?

henriquegemignani: The list of possible words could be built by going over all the rooms and resources in the database during runtime. Since you added enemies, you could add a new field to the database with these.

But if you think this is too much for this, I'm fine either way. I just dislike data duplication.

SpaghettiToastBook: I totally get where you're coming from, but I don't think it's a good idea in this case:
- For the item names, that would make the available word list change with the item pool, which I'm not a fan of
- Since not all of the room name words are used, there would need to be a list of words not to use, which somewhat defeats the point anyway

PR #395, opened 2019-09-06
chore (randomizer): hand-compiled libzo2 for 64-bit OSX Mojave

ursineasylum:
Mono on OSX Mojave can't handle loading the currently-included lzo2.dll library to run Randomizer.exe. Recompiling LZO natively from source (http://www.oberhumer.com/opensource/lzo/) and dropping the resulting dylib in next to the executable resolves this issue, at least on this platform.


henriquegemignani:
Thanks!

PR #399, opened 2019-09-08
Remove pointless gitignore folders

henriquegemignani:


SpaghettiToastBook:
We should remove all but one of the virtual environment directory names. My preference would be `venv/` or `env/`.

henriquegemignani:
Sometimes I use `.venv` but I'm fine either way.

SpaghettiToastBook:
I mostly just don't want it to be `ENV` :P. `.venv` is fine.

PR #403, opened 2019-09-14
Joke hint adjustments

SpaghettiToastBook:
Fixes #403.

Unfortunately the "Metroids are not pets" joke hint is a little bit too long to fit in the text box. We could make the hint go over multiple pages, but I don't know if it's worth bothering to do for just that.

### DIFF
--- a/randovania/games/prime/patcher_file_lib/item_hints.py
+++ b/randovania/games/prime/patcher_file_lib/item_hints.py
@@ -128,6 +128,9 @@ def create_hints(patches: GamePatches,
 
     for asset, hint in patches.hints.items():
         if hint.precision.is_joke:
+            if not joke_hints:
+                joke_hints = sorted(_JOKE_HINTS)
+                rng.shuffle(joke_hints)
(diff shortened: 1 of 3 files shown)


randovania/games/prime/patcher_file_lib/item_hints.py
```diff
@@ -43,15 +43,13 @@
 ]
 
 _JOKE_HINTS = [
```
henriquegemignani: We should leave the rules for a joke hint to be added as a comment here.

henriquegemignani:
We need more joke hints :(

randovania/games/prime/patcher_file_lib/item_hints.py
SpaghettiToastBook: Agreed. Are the ones I put in #403 okay?

henriquegemignani: I think you meant the issue, not a circular link. But yeah.

PR #404, opened 2019-09-14
Database adjustments

SpaghettiToastBook:


CHANGELOG.md
```diff
@@ -193,6 +193,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+-   Added: Method to unblock the portal in Mining Station B without Scan Visor (Normal and above).
+
+-   Added: Method to get to the Darkburst location in Mining Station B with just Space Jump Boots and Screw Attack,
+    and without using slope jumps or bomb space jumps (Hypermode difficulty).
+
+-   Added: Method to manipulate Power Bomb Guardian into opening the Power Bomb Blast Shield on the door to
+    Undertemple Access, using Boost Ball (Normal and above).
+
+-   Fixed: The method to open the Hydrodynamo Station Seeker door using Screw Attack without Seeker Launcher now
```
henriquegemignani: You can only use SA via Air Underwater when you don't have Gravity Boost?

SpaghettiToastBook: You can Screw Attack underwater via Air Underwater while you have Gravity Boost, but that particular trick isn't possible if you have it. Or at least, I've never managed to pull it off with Gravity. The reduced falling speed makes it easier to fire the missile in time before Screw Attacking, and the arc of the Screw Attack is best without Gravity Boost.

PR #405, opened 2019-09-14
Config files with invalid JSON are now correctly dealt with.

henriquegemignani:


SpaghettiToastBook:
Maybe there should be a test for this?

henriquegemignani:
> Maybe there should be a test for this?

I did take a look but we didn't have any easy test to quickly modify. But since codecov is complaining...

PR #406, opened 2019-09-14
Make the Options.save_to_disk more robust

henriquegemignani:


randovania/interface_common/options.py
```diff
@@ -139,10 +140,25 @@ def _save_to_disk(self):
...
+        # Windows doesn't, so we need to erase the current config before renaming
+        # This opens a small chance of data loss. We could instead rename to `config_old`,
+        # and when loading we would also check that path. But that's way overkill for this
+        config_path = self._data_dir.joinpath("config.json")
+        if platform.system() == "Windows" and config_path.exists():
+            config_path.unlink()
+
+        # Place the new, complete, config to the desired path
+        new_config_path.rename(config_path)
+
```
SpaghettiToastBook: I think we could just use `Path.replace`.
https://docs.python.org/3/library/pathlib.html#pathlib.Path.replace

PR #596, opened 2020-02-14
Probability offset improvements

cdoggers:


henriquegemignani:
For logic, probability isn't a number that range from 0 to 1: that items weight is compared to the sum of all option's weight. Maybe a better name for this would be "weight_offset".
What it means is that these items are indeed used for progression. And you clearly can see that, otherwise they would always be last for progression.

As for the temple keys, while the offset is high, it is also the only thing that gives the first two keys a non-zero weight. Not only that, placing all keys does not mean something is placed at the boss.

Given play testing I've observed and played myself, I'm pretty happy with how keys are placed and how often bosses are required.

cdoggers:
What I have observed is that only items with a >0 weight can be placed, unless there is only one progression option. -.99 max gives power bomb and light suit a chance to still be used when there are multiple progression options.

kHFRAB2qBYAAFiTkUCDFGHKLMOPQTUVWXYaegDDDDBCZQmUIAAQwwwwwwwQgsMMMAADn, is an extreme example, but shows the severe problems with the fill algorithm. Lowering the offsets of keys is just a band-aid; the whole algorithm needs to be completely redone, as it's demonstrably crappy.

henriquegemignani:
Indeed only something with >0 can be placed.
Power Bombs in particular unlocks a considerable number of things that it still has high weight despite the big negative offset.

As for Light Suit, I sincerely hasn't played that much without Progressive Suits and we've changed a few things about the weights since then. It was essential to make it not appear at the start of every single game.

In any case, band-aid fixes are boring and I'm interested in seeing your improved algorithm.

cdoggers:
Bombs  space jump and dark beam unlock lots of items too. Bombs and space jump can unlock even more than power bombs do. Why specifically wanting to make power bombs unlikely? What's your goal with the fill generator here?

Percent-based weight reductions for power bombs may be helpful.

An improved algorithm for room selection could use an identical system to item selection. Weights could be decayed based on how long the room has been available, prioritizing newly accessible rooms while still allowing older accessible rooms to be chosen. Exact numbers for decay and such would require testing.

cdoggers:
Key weight reduction is not a band-aid. My hope with that change is to have keys show up in each region about the same amount, when not required, instead of right now, where Agon keys typically are only found in Agon and Temple Grounds.

henriquegemignani:
> 
> 
> Bombs space jump and dark beam unlock lots of items too. Bombs and space jump can unlock even more than power bombs do. Why specifically wanting to make power bombs unlikely? What's your goal with the fill generator here?

Because otherwise Power Bombs are almost always placed right at the start, which is bad for multiple reasons.

> 
> An improved algorithm for room selection could use an identical system to item selection. Weights could be decayed based on how long the room has been available, prioritizing newly accessible rooms while still allowing older accessible rooms to be chosen. Exact numbers for decay and such would require testing.

What a nice idea. Shame that it's already implemented.

> 
> Key weight reduction is not a band-aid. My hope with that change is to have keys show up in each region about the same amount, when not required, instead of right now, where Agon keys typically are only found in Agon and Temple Grounds.

I've considered forcing keys to be placed on their own worlds, but haven't done that because they're being placed on very appropriate places. The placement of the keys being unrelated to the actual contents of the temple is pretty neat. Collecting all keys to a temple means the boss is potentially useful, but not guaranteed. So there's some risk in making the play, or you can look for the hint.

cdoggers:
You didn't answer this..

What's your goal with the fill generator here?

henriquegemignani:
1. Provide a good gameplay experience, from a game design perspective
2. Support starting from any location, with any items, with any item pool
(no bombs, no light beam, etc), with customizable elevators, door types and
room connections
3. An item should not favor certain locations more than others (only if
possible to be there)


Power Bombs is more 3, with a touch of 1.
Keys is mostly 1.

PR #655, opened 2020-04-16
Wall boost difficulty changes

gollop:
Changes many of the wall boosts to a more appropriate difficulty, instead of Hard for every wall boost.


henriquegemignani:
Could you please add an entry to the changelog, please?
If you could also add entries for your last PR as well.

PR #657, opened 2020-04-18
Hall of Honored Dead item with Power Bombs

gollop:
Trivial difficulty trick. Fixes https://github.com/randovania/randovania/issues/527.


henriquegemignani:
Too bad that logic doesn't have 3 PBs right now :X

PR #658, opened 2020-04-20
Logic fixes, additions

gollop:
Fixes https://github.com/randovania/randovania/issues/474. Fixes https://github.com/randovania/randovania/issues/535. Fixes https://github.com/randovania/randovania/issues/645.


henriquegemignani:
In Main Reactor, you replaced a `Space Jump Boots ≥ 1, Screw Attack ≥ 1` with a `Screw Attack ≥ 1, Screw Attack without Space Jump ≥ 1`.
Is the older one invalid? Because other it makes sense to keep (in case someone disables Screw Attack without Space Jump).

henriquegemignani:
Sand Processing:
You replaced the Screw Attack path with a BSJ. It's reasonable to keep both. Maybe also simplify the nodes in this room?

PR #791, opened 2020-08-03
Logic Changes

dyceron:
Added logic for Torvus Plaza without Boost Ball. Has two variants: one with just SA + SJ for Hard, and one with SJ + SA + Bombs + Spider for  Normal. Updated the E-Tank -> Door node to account for this change.

https://youtu.be/I81EA1e6zmg


dyceron:
New logic for the Controller Access rooms to be used without Bombs. Hypermode difficulty (could be lowered to hard because these are the easier ones, but hypermode is probably fine).

dyceron:
Updated the changelog to account for the recent logic changes.

PR #818, opened 2020-08-16
Added existing tricks from Transit Tunnels to Room Center instead of …

JaggerTSG:
…Training Access


henriquegemignani:
Missing CHANGELOG entries :)

PR #858, opened 2020-08-24
Meditation vista boost jump method

Belattor:
Added: Method of reaching Meditation vista item with Boost jump (Veteran and above)


dyceron:
Changelog update?

PR #909, opened 2020-09-10
Minor Updates

dyceron:
Small updates for #103 and #333

### DIFF
--- a/randovania/data/json_data/prime2.txt
+++ b/randovania/data/json_data/prime2.txt
@@ -40,6 +40,13 @@ Templates
               Shoot Darkburst
               Shoot Sonic Boom
 
+* Bomb Slot without Bombs (No Space Jump):
+      All of the following:
+          Morph Ball and Bomb Slot without Bombs (Hypermode)
+          Any of the following:
+              Shoot Darkburst
+              Shoot Sonic Boom
+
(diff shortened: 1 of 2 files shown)


randovania/data/json_data/prime2.txt
```diff
@@ -2756,8 +2752,8 @@ Asset id: 1272952761
       After Transport Center Gate
   > Transport Center/Event - Transport Center Gate
       Any of the following:
-          Bomb Slot without Bombs (Space Jump)
           Morph Ball and Morph Ball Bomb
+          No Space Jump Boots and Bomb Slot without Bombs (Space Jump)
```
henriquegemignani, line 2756: No Space Jump Boots?

dyceron: Yes because space jump isn't needed. I did this because the nsj template includes sunburst, which isn't viable for most slots. Better solution might be to make other template or something.

henriquegemignani: `No Space Jump Boots` implies that you *can't* have SJ.

dyceron: So what do you recommend?

henriquegemignani (approved):

PR #1028, opened 2020-09-30
Added item nodes for mp3 and very basic no trick requirements

kirbymastah:


randovania/data/json_data/prime3.txt
```diff
@@ -719,23 +797,28 @@ Asset id: 10553270987934706862
...
 ----------------
 Crash Site
 Asset id: 15814834557214288722
-> Dock to G.F.S. Theseus; Heals? False
-  * Normal Door to G.F.S. Theseus/Dock to Crash Site
+> Dock to G.F.S. Theseus (Door); Heals? False
+  * Normal Door to G.F.S. Theseus/Dock to Crash Site (Door)
 
 > Dock to Gateway; Heals? False
   * Normal Door to Gateway/Dock to Crash Site
```
dyceron, line 813: This should have a "Grapple Swing  or SA” requirement for crossing the gap

randovania/data/json_data/prime3.txt
```diff
@@ -719,23 +797,28 @@ Asset id: 10553270987934706862
...
 
-> Dock to G.F.S. Theseus; Heals? False
-  * Normal Door to G.F.S. Theseus/Dock to Crash Site
+> Dock to G.F.S. Theseus (Tunnel); Heals? False
+  * Morph Ball Door to G.F.S. Theseus/Dock to Crash Site (Tunnel)
+  > Pickup (Missile Expansion)
+      Trivial
+
+> Pickup (Missile Expansion); Heals? False
+  * Pickup 24
```
dyceron, line 821: Same as above. No connections it looks like, and the pickup doesn't connect to any nodes. Applies to all the pickup nodes it seems?

randovania/data/json_data/prime3.txt
```diff
@@ -751,48 +834,83 @@ Grand Court Path
 Asset id: 8468936149308094087
 > Dock to Gateway; Heals? False
   * Normal Door to Gateway/Dock to Grand Court Path
+  > Pickup (Missile Expansion)
+      Trivial
```
dyceron, line 838: Shouldn't this require PED Suit? I assume it's being shuffled currently but unsure.

randovania/data/json_data/prime3.txt
kirbymastah: Readjusted and fixed

randovania/data/json_data/prime3.txt
kirbymastah: Too lazy to implement PED suit requirements for now - not really sure how we're handling that

henriquegemignani: Is there something that explicitly requires PED? Or is it just that you can enter hypermode?
If the later, there's an item explicitly for hypermode.

kirbymastah: Cool, just gonna use that then. Also gonna add it alongside any hyper ball/missile/grapple requirements.

Not entirely sure if we should make hypermode required for rundas/ghor/gandrayda though? I'm leaning towards no

dyceron: Probably not, because its not required on them. The less Hypermode requirements the better imo, so better kept to things that actually need it.

PR #1171, opened 2020-10-31
Light Suit visual is no more overwritten by Dark Suit

UltiNaruto:


henriquegemignani:
I personally like this side effect :P

But other than that, I have some unpushed changes to how the game hook works that conflict with this, dislikes the idea of reading extra addresses like this and more importantly, accidentally fixes this.

PR #1245, opened 2020-11-16
Feature: Starting area quick fill by world checkbox

EthanArmbrust:
On starting locations tab, replaced the world group titles with tristate checkboxes. Clicking this checkbox will select or unselect all areas in that world. Selecting only some of the areas in that world will set the world checkbox to partially selected.

### DIFF
--- a/randovania/gui/dialog/logic_settings_window.py
+++ b/randovania/gui/dialog/logic_settings_window.py
@@ -468,11 +468,10 @@ class LogicSettingsWindow(QDialog, Ui_LogicSettingsWindow):
         world_areas = [world_list.area_to_area_location(area)
                        for area in world.areas if check.is_dark_world == area.in_dark_aether]
         with self._editor as editor:
-            for area in world_areas:
-                editor.set_layout_configuration_field(
-                    "starting_location",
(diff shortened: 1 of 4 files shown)


randovania/gui/dialog/logic_settings_window.py
```diff
@@ -432,6 +460,20 @@ def _on_check_starting_area(self, check, _):
...
+    def _on_check_starting_world(self, check, _):
+        if self._during_batch_check_update:
+            return
+        world_list = self.game_description.world_list
+        world = world_list.world_by_asset_id(check.world_asset_id)
+        world_areas = [world_list.area_to_area_location(area)
+                       for area in world.areas if check.is_dark_world == area.in_dark_aether]
+        with self._editor as editor:
+            for area in world_areas:
+                editor.set_layout_configuration_field(
```
henriquegemignani, line 472: `set_layout_configuration_field` does extra boiler plate which is a waste inside a for loop like this

henriquegemignani:
Do you think you can write a test that calls `_on_check_starting_world`? Ideally one by simulating a click on one of these new checkboxes.

You're also missing an entry in the CHANGELOG.

PR #1251, opened 2020-11-18
Added Basic Prime 1 Tricks to Logic

samuelroy21:
Added lots of basic tricks to logic across the whole game (except sunken frigate/deep tallon and Frigate Orpheon). Most of them will be very easy tricks. I performed every trick here myself to ascertain its difficulty and documented them appropriately.

### DIFF
--- a/randovania/data/json_data/prime1.txt
+++ b/randovania/data/json_data/prime1.txt
@@ -1041,13 +1041,18 @@ Asset id: 1282373491
 > Pickup (Missile); Heals? False
   * Pickup 51; Major Location? False
   > Dock to Save Station C
-      Missile and Space Jump Boots and Gravity Suit
+      All of the following:
+          Missile and Space Jump Boots
+          Gravity Suit or Slope Jump (Beginner)
   > Dock to Frost Cave Access
-      Space Jump Boots and Gravity Suit
+      All of the following:
+          Space Jump Boots
+          Gravity Suit or Slope Jump (Beginner)
   > Dock to Upper Edge Tunnel
       All of the following:
-          Space Jump Boots and Gravity Suit
+          Space Jump Boots
           Missile or Grapple Beam
+          Gravity Suit or Slope Jump (Beginner)
 
 ----------------
 Hunter Cave
@@ -1241,7 +1246,9 @@ Asset id: 1423896872
   > Dock to Lower Edge Tunnel
       Trivial
   > Dock to Storage Cave
-      Power Bomb and Grapple Beam and Space Jump Boots and Morph Ball
+      All of the following:
+          Power Bomb and Grapple Beam and Space Jump Boots and Morph Ball
+          X-Ray Visor or Invisible Objects (Beginner)
   > Dock to Security Cave
       Grapple Beam and Space Jump Boots and Morph Ball
 
@@ -1249,13 +1256,19 @@ Asset id: 1423896872
   * Plasma Door to Storage Cave/Dock to Phendrana's Edge
   > Dock to Upper Edge Tunnel
       Power Bomb and Morph Ball
+  > Dock to Security Cave
+      All of the following:
+          Power Bomb and Space Jump Boots and Morph Ball
+          Grapple Beam or Standable Terrain (Beginner)
 
 > Dock to Security Cave; Heals? False
   * Morph Ball Door to Security Cave/Dock to Phendrana's Edge
   > Dock to Upper Edge Tunnel
       Morph Ball
   > Dock to Storage Cave
-      Power Bomb and Morph Ball
+      All of the following:
+          Power Bomb and Morph Ball
+          X-Ray Visor or Invisible Objects (Beginner)
 
 ----------------
 Gravity Chamber
@@ -1275,7 +1288,11 @@ Asset id: 1226265714
 > Pickup (Gravity Suit); Heals? False
   * Pickup 54; Major Location? True
   > Dock to Chamber Access
-      Space Jump Boots and Gravity Suit
+      All of the following:
+          Gravity Suit
+          Any of the following:
+              Space Jump Boots
+              Morph Ball Bomb and Morph Ball and Double Bomb Jump (Beginner)
 
 > Pickup (Missile); Heals? False
   * Pickup 55; Major Location? False
--- a/randovania/data/json_data/prime1.txt
+++ b/randovania/data/json_data/prime1.txt
@@ -2117,7 +2117,9 @@ Asset id: 4126087266
           Morph Ball Bomb and Morph Ball and Heated Rooms Damage ≥ 319
           Space Jump Boots and Heated Rooms Damage ≥ 277
   > Pickup (Missile)
-      Morph Ball Bomb and Morph Ball and Heated Rooms Damage ≥ 358
+      Any of the following:
+          Morph Ball Bomb and Morph Ball and Heated Rooms Damage ≥ 358
+          Space Jump Boots and Standable Terrain (Beginner) and Heated Rooms Damage ≥ 227
 
 > Dock to Transport Tunnel B; Heals? False
   * Normal Door to Transport Tunnel B/Dock to Fiery Shores
@@ -2126,7 +2128,9 @@ Asset id: 4126087266
           Heated Rooms Damage ≥ 324
           Space Jump Boots and Heated Rooms Damage ≥ 284
   > Pickup (Missile)
-      Morph Ball Bomb and Morph Ball and Heated Rooms Damage ≥ 246
+      Any of the following:
+          Morph Ball Bomb and Morph Ball and Heated Rooms Damage ≥ 246
+          Space Jump Boots and Standable Terrain (Beginner) and Heated Rooms Damage ≥ 158
 
 > Dock to Warrior Shrine; Heals? False
   * Morph Ball Door to Warrior Shrine/Dock to Fiery Shores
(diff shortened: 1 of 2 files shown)


dyceron:
Did you rebase the branch before making this PR? All the previous commits from the last PR are still on this new PR.

samuelroy21:
> 
> 
> Did you rebase the branch before making this PR? All the previous commits from the last PR are still on this new PR.

I think I might've done that by accident when I was still figuring out how to keep my fork current with the main one, I'm not very good at Git sorry.

randovania/data/json_data/prime1.txt
```diff
@@ -405,8 +417,16 @@ Asset id: 3006924320
 
 > Pickup (Power Bomb); Heals? False
   * Pickup 38; Major Location? False
+  > Middle-Left Rooftop
+      Trivial
```
dyceron, line 421: Leaving from the PB seems wrong. You either need Space Jump or a Bomb Jump to escape the pit

randovania/data/json_data/prime1.txt
```diff
@@ -1209,21 +1246,29 @@ Asset id: 1423896872
...
 
 > Dock to Security Cave; Heals? False
   * Morph Ball Door to Security Cave/Dock to Phendrana's Edge
   > Dock to Upper Edge Tunnel
       Morph Ball
   > Dock to Storage Cave
-      Power Bomb and Morph Ball
+      All of the following:
+          Power Bomb and Morph Ball
+          X-Ray Visor or Invisible Objects (Beginner)
```
dyceron, line 1271: I really don't know/think Xray should be required for this. But i'm not sure.

randovania/data/json_data/prime1.txt
```diff
@@ -1718,6 +1775,7 @@ Asset id: 2758909034
           Missile ≥ 2 and Grapple Beam and Heat-Resisting Suit
           Missile ≥ 2 and Space Jump Boots and Heated Rooms Damage ≥ 161
           Space Jump Boots and Heated Rooms Damage ≥ 145 and Shoot Super Missile
+          Missile ≥ 2 and Morph Ball Bomb and Morph Ball and Heat-Resisting Suit
```
dyceron, line 1778: Not sure I understand what Morph/Bombs are for here, also the suit. Can be grabbed with a slope jump nsj.

randovania/data/json_data/prime1.txt
```diff
@@ -2171,6 +2237,7 @@ Asset id: 3835971118
       Any of the following:
           Space Jump Boots and Heated Rooms Damage ≥ 165
           Morph Ball and Spider Ball and Heat-Resisting Suit
+          Morph Ball Bomb and Morph Ball and Heated Rooms Damage ≥ 135 and Lava Damage ≥ 147
```
dyceron, line 2240: Missing Bomb Jump trick

randovania/data/json_data/prime1.txt
```diff
@@ -2183,12 +2250,16 @@ Asset id: 1282952170
 > Dock to North Core Tunnel; Heals? False
   * Wave Door to North Core Tunnel/Dock to Twin Fires
   > Dock to Twin Fires Tunnel
-      Grapple Beam or Space Jump Boots or Gravity Suit
+      Any of the following:
+          Grapple Beam or Space Jump Boots or Gravity Suit
+          Morph Ball Bomb and Morph Ball
```
dyceron, line 2255: Bomb Jump needed?

randovania/data/json_data/prime1.txt
```diff
@@ -2183,12 +2250,16 @@ Asset id: 1282952170
...
 
 > Dock to Twin Fires Tunnel; Heals? False; Spawn Point
   * Normal Door to Twin Fires Tunnel/Dock to Twin Fires
   > Dock to North Core Tunnel
-      Grapple Beam or Space Jump Boots or Gravity Suit
+      Any of the following:
+          Grapple Beam or Space Jump Boots or Gravity Suit
+          Morph Ball Bomb and Morph Ball
```
dyceron, line 2262: Bomb Jump needed?

randovania/data/json_data/prime1.txt
```diff
@@ -2402,8 +2477,12 @@ Asset id: 1681720207
       After Main Quarry Barrier Deactivated
   > Pickup (Missile)
       All of the following:
-          Wave Beam and Scan Visor and Space Jump Boots and Morph Ball and Spider Ball
+          Wave Beam and Scan Visor
           Thermal Visor or Invisible Objects (Beginner)
+          Any of the following:
+              Space Jump Boots and L-Jump (Beginner) and Slope Jump (Intermediate) and Standable Terrain (Beginner)
+              Morph Ball and Spider Ball
+          Space Jump Boots or Slope Jump (Intermediate)
```
dyceron, line 2485: This line seems wrong. It's part of the all of the following grouping, so it invalidates the Spider path.

randovania/data/json_data/prime1.txt
samuelroy21: I have it required because someone might forget where the rock wall is, much in the way they might forget the location of power conduits, even though they'd be obvious for a veteran player. That said, it is not marked as a requirement for leaving the cave, as the location of the wall would then be quite obvious.

randovania/data/json_data/prime1.txt
samuelroy21: This was done with single bomb jumps, no DBJ was used. Should single bomb jumps be considered a trick as well when done in a place the player would not expect to be required to use morph ball?

randovania/data/json_data/prime1.txt
samuelroy21: You can do a set of single bomb jumps to scale it from one side NSJ, as an alternative to said slope jump. I will however have to catalogue the slope jump later.

dyceron: Should add in the bomb jump trick then if keeping it.

randovania/data/json_data/prime1.txt
dyceron: Personally, calling the trick double bomb jump is bad imo, since it would have to be a dbj. if the trick is called "Bomb Jump", then it can cover bomb jumps that arent intended, whether single/double, as well as include higher level bomb jumps like uber/high/halfpipe at higher difficulties, rather than separate tricks.

randovania/data/json_data/prime1.txt
samuelroy21: I went and double checked, and I'm fairly sure I got this right, but here's my thinking so you can hopefully tell me where I went wrong.

Scan Visor - Required to scan terminal
Wave Beam - to activate conduit
Thermal Visor || Invisible Objects - to know where conduit is
Space Jump || Slope Jump(Intermediate) - to get to the part with the terminal/crane. There are some rocky jumps that require slope jumps just to get up to the section with the crane if you're NSJ.

The OR block, which can select either spider and morph or a bunch of tricks, is just for getting from the base of the crane to the pickup.

From the base of the crane, you can slope double jump off the slope of the terminal housing to stand on a lip at the top of the crane. From there you can double L-Jump to the crane's load, thus skipping the requirement for morph ball or spider ball.

randovania/data/json_data/prime1.txt
samuelroy21: Good catch. I couldn't manage a DBJ out, but I also might just be bad at the game.

randovania/data/json_data/prime1.txt
dyceron: I just double checked and the Space Jump or Slope Jump path is part of the "All of the Following" so it does invalidate the morph/spider path
image

randovania/data/json_data/prime1.txt
samuelroy21: I think you make sort of a good point. Since they're both "vanilla" techniques in a sense, I'll merge them together, though I'll probably have future unintended bomb techs like BSJ's, 3BJ's, &c. have their own tricks to give the user a degree of flexibility in controlling the tricks which the randomizer uses.

randovania/data/json_data/prime1.txt
samuelroy21: As it should be - there is no way to traverse the path without one of the two (obviously barring any tricks I haven't tried). 

This pair of jumps is the reason for that requirement
image


It is arranged in a way that is graphically quite odd, I'll try and organize it in a way that looks a little better.

randovania/data/json_data/prime1.txt
dyceron: BSJ would be its own trick yes, as seen for Echoes and Corruption

randovania/data/json_data/prime1.txt
dyceron: ah i see it then. remove the space jump and slope jump requirements from the "Any of the following part." that isnt needed to be stated since its already covered with the "OR"

samuelroy21: They should already be gone since it'll now be quarry access -> top platform -> missiles instead of just quarry access -> missiles.

randovania/data/json_data/prime1.txt
```diff
@@ -2500,12 +2496,29 @@ Asset id: 1681720207
   * Pickup 73; Major Location? False
   > Dock to Quarry Access
       Trivial
+  > Crane Platform
+      Morph Ball and Boost Ball
```
dyceron, line 2500: Make this path be trivial, since you can access the item with either morph or space jump depending on the path.

samuelroy21: Ah yeah fair. Perspective threw me off a bit.

randovania/data/json_data/prime1.txt
dyceron: @henriquegemignani thoughts?

henriquegemignani: It seems like a suitable case of Knowledge, or Invisible Objects in this case, in the lowest difficulty possible.
Leaving these tricks off means you can have no idea of where things are and still be fine.

dyceron: Sounds good. I'll go ahead and accept this pr then.

PR #1298, opened 2020-11-28
Update Linux Support for Focal (fixes #1248)

### LINKED ISSUE #1248: Various Issues with Bundled Mono Applications on Linux
In trying to get Randovania to work on my Ubuntu 20.04 machine, I've found the following issues:

* A native `liblzo2.so` library needs to be shipped alongside `Randomizer.exe`

* After resolving that, patching still dies at 37% for reasons I can't debug

---

The randomizer will successfully generate a valid seed, but when patching is attempted, I get stacktrace barf and an error in the statusbar about "tool did not send 'Randomized!'":

```
Invoking external tool with:  ['mono', '/home/salt/Projects/randovania/randovania/data/ClarisPrimeRandomizer/Randomizer.exe', '/home/salt/.local/share/Randovania/extracted_game']
[2020-11-17 09:05:45,563] ERROR in qt: Task exception was never retrieved
Traceback (most recent call last):
  File "/home/salt/Projects/randovania/randovania/games/prime/claris_randomizer.py", line 48, in _read_data
    line = await stream.readuntil(b"\r")
  File "/usr/lib/python3.8/asyncio/streams.py", line 629, in readuntil
    raise exceptions.IncompleteReadError(chunk, None)
asyncio.exceptions.IncompleteReadError: 1743 bytes read on a total of None expected bytes
```

According to what I get when I manually invoke the application, it appears to be unable to find `lzo2.dll`:

```
$ mono /home/salt/Projects/randovania/randovania/data/ClarisPrimeRandomizer/Randomizer.exe /home/salt/.local/share/Randovania/extracted_game

Unhandled Exception:
System.DllNotFoundException: lzo2.dll assembly:<unknown assembly> type:<unknown type> member:(null)
  at (wrapper managed-to-native) Retro.Compression.LZO.__lzo_init_v2(uint,int,int,int,int,int,int,int,int,int)
  at Retro.Compression.LZO..ctor () [0x00016] in <69a65e8e3c4c4fb59efe7a26fd6fa95c>:0 
[...]
```

Which is very odd because it sits alongside it in the same directory. I did a bit of digging and found that Mono is having issues because the library shipped with it is compiled for Windows, not Linux (thus the "bad ELF header" error below):

```
Mono: DllImport error loading library '/home/salt/Projects/randovania/randovania/data/ClarisPrimeRandomizer/lzo2.dll': '/home/salt/Projects/randovania/randovania/data/ClarisPrimeRandomizer/lzo2.dll: invalid ELF header'.
Mono: DllImport error loading library '/home/salt/Projects/randovania/randovania/data/ClarisPrimeRandomizer/liblzo2.dll.so': '/home/salt/Projects/randovania/randovania/data/ClarisPrimeRandomizer/liblzo2.dll.so: cannot open shared object file: No such file or directory'.
```

This should be a relatively easy fix: include a prebuilt `liblzo2.so` alongside the application. I added a symlink to my system's `liblzo2.so.2` and it worked just fine:

```
$ ls -al
total 2.4M
-rw-rw-r-- 1 salt salt    5 Nov 17 08:37 .gitignore
-rwxrwxr-x 1 salt salt 209K Nov 17 08:37 liblzo2.dylib*
lrwxrwxrwx 1 salt salt   42 Nov 17 09:35 liblzo2.so -> /usr/lib/x86_64-linux-gnu/liblzo2.so.2.0.0
-rwxrwxr-x 1 salt salt 121K Nov 17 08:37 lzo2.dll*
-rw-rw-r-- 1 salt salt 147K Nov 17 08:37 RandomizerData.json
-rw-rw-r-- 1 salt salt 1.8M Nov 17 08:37 Randomizer.exe
-rw-rw-r-- 1 salt salt  26K Nov 17 08:37 readme.txt
-rw-rw-r-- 1 salt salt 105K Nov 17 08:37 zlib1.dll
```

But after that, the randomizer will proceed to patch the game and fail at 37% with no descriptive error. Here's the log leading up to the failure:

```
Modifying NoARAM...
Modifying FrontEnd...
Removing Attract videos...
Modifying Tweaks...
Reading Tweaks...
Modifying GGuiSys...
Modifying Logbook...
[2020-11-17 09:49:56,915] ERROR in qt: Task exception was never retrieved
[...]
RuntimeError: External tool did not send 'Randomized!'. Did something happen?
```

And then I'm stumped. Mono debug logging doesn't yield anything helpful, just entries about shifting data around and opening/closing file handles.

I'm staying up way too late debugging this. If I get some time tonight I'll try to press into the issue further, but we're probably hitting C# cross-platform bullshittery that might be more properly resolved by just reimplementing the patcher in Python. That'd open the door for better, more descriptive error handling, too.

It should be pretty easy to spin up a 20.04 VM and reproduce; setup for the environment was done exactly as the README specified (with the exception of "activating" the venv via `source venv/bin/activate`, of course). Installing requirements will complain about `discord-py` being incompatible with `aiohttp` or something, but I doubt that has any bearing on this issue.

rehashedsalt:
This PR contains two minor commits to fix a couple issues with Linux support, mostly on Ubuntu Focal but also for any Debian derivatives and distros that ship Python 3.8, and fixes #1248 . Namely:

* `e368f47` - On Debian derivatives, `liblzo2.so.2` is not actually mapped to `lzo2.dll` in the system-wide Mono configs by default, and since `/usr/lib/x86_64-linux-gnu` isn't in Mono's library search path, it was wholly unable to find the library. This caused the randomizer and the menu mod patcher to die on startup. By adding .config files for those two applications, we add those library mappings ourselves, fixing the issue. This will require that the user have `liblzo2-2` installed system-wide, but this is a common dependency of most desktop environments and should not be a serious issue.

* `b1dd229` - In Python 3.8, one of the asyncio exceptions was moved from asyncio.streams to asyncio.exceptions. A separate code path was added for Python 3.8 in claris_randomizer.py that uses the new exception name, fixing basic patching functionality.

Between the two of these, I'm now able to use Randovania natively on my machine!

I was unable to do any testing on Windows; you'll want to verify that there are no regressions in that regard, especially with regard to the Mono `.config` files. Testing Python 3.8 on Windows may also be of interest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,9 @@ target/
 # macOS
 .DS_Store
 
+# Vim
+*.swp
+
 # Usual dev server db
 /tools/test.db
 /network-storage
--- a/randovania/data/ClarisEchoesMenu/EchoesMenu.exe.config
+++ b/randovania/data/ClarisEchoesMenu/EchoesMenu.exe.config
@@ -1,0 +1,3 @@
+<configuration>
+	<dllmap dll="lzo2.dll" target="liblzo2.so.2" os="!windows"/>
+</configuration>
--- a/randovania/data/ClarisPrimeRandomizer/Randomizer.exe.config
+++ b/randovania/data/ClarisPrimeRandomizer/Randomizer.exe.config
@@ -1,0 +1,3 @@
+<configuration>
+	<dllmap dll="lzo2.dll" target="liblzo2.so.2" os="!windows"/>
+</configuration>
--- a/randovania/games/prime/claris_randomizer.py
+++ b/randovania/games/prime/claris_randomizer.py
@@ -3,6 +3,7 @@ import json
 import platform
 import re
 import shutil
+import sys
 from asyncio import StreamWriter, StreamReader
 from pathlib import Path
 from typing import Callable, List, Union, Optional
@@ -44,10 +45,17 @@ async def _write_data(stream: StreamWriter, data: str):
 
 async def _read_data(stream: StreamReader, read_callback: Callable[[str], None]):
     while True:
-        try:
-            line = await stream.readuntil(b"\r")
-        except asyncio.streams.IncompleteReadError as incomplete:
-            line = incomplete.partial
+        # 3.7 uses a different exception name than 3.8
+        if sys.version_info[1] > 7:
+            try:
+                line = await stream.readuntil(b"\r")
+            except asyncio.exceptions.IncompleteReadError as incomplete:
+                line = incomplete.partial
+        else:
+            try:
+                line = await stream.readuntil(b"\r")
+            except asyncio.streams.IncompleteReadError as incomplete:
+                line = incomplete.partial
         if line:
             try:
                 decoded = line.decode()
--- a/randovania/games/prime/claris_randomizer.py
+++ b/randovania/games/prime/claris_randomizer.py
@@ -3,8 +3,11 @@ import json
 import platform
 import re
 import shutil
-import sys
 from asyncio import StreamWriter, StreamReader
+try:
+    from asyncio.exceptions import IncompleteReadError
+except ImportError:
+    from asyncio.streams import IncompleteReadError
 from pathlib import Path
 from typing import Callable, List, Union, Optional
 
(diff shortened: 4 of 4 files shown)


henriquegemignani:
A better way of solving this changing the imports in the top of the file:

```python
try:
    from asyncio.exceptions import IncompleteReadError
except ImportError:
    from asyncio.streams import IncompleteReadError
```

rehashedsalt:
Yeah, definitely. If you couldn't tell, I'm not normally a Python dev.

How's this?

henriquegemignani:
👍

PR #1322, opened 2020-12-02
Scrapworks Spawn Point Fix and Transit Station 0204 Event Fix

dyceron:


henriquegemignani:
To fix the test, you'll need to edit the `test/generator/test_generator_reach.py` and remove 145 from that list in the line with "Corruption Preset" on it.

That list means "these events aren't collectable", but since you made of them collectable it's now complaning.

PR #1332, opened 2020-12-04
Corruption Fixes/Updates/Changes

dyceron:
- Damage updates
- Event updates
- Node updates
- Tricks added in more places
- Logic fixed
- Minor fixes/updates


henriquegemignani:
Renaming things would've been better left to another PR, but oh well.

PR #1351, opened 2020-12-08
Feature: Prime 1 basic seed generation

EthanArmbrust:
The following changes were made to add support for making Prime 1 seeds:
-Created item database for Prime 1
-Create a default Prime 1 preset called "Prime Preset"
-Create an item pool generator for Chozo Artifacts
-UI: Added entry for Prime 1 in game drop down on the preset selection tab
-UI: Added layout configuration with template strings for Prime 1 (mostly copied from corruption)
-Add test for Prime 1 in test_generator_reach.py

I also made the following changes and fixes to the Prime 1 database:
-Added event for incinerator drone
-Fixed incorrect dock type for Main Plaza <=> Nursery Access
-Added connection from dock to east atrium to gathering hall access in gathering hall
-Add connection from left of half pipe to dock to great tree hall in life grove tunnel
-Fix check for draining water event in reflecting pool
-Remove connection from research core event node to dock to research core access
-Change training chamber e-tank pickup node type from generic to pickup
-Change victory condition from "Impossible" to "After Meta Ridley" for now


randovania/data/json_data/prime1.json
```diff
@@ -171,7 +171,7 @@
                 "extra": null
             },
             {
-                "index": 24,
+                "index": 42,
```
henriquegemignani, line 174: 24 is the correct value for Prime 1. Gotta adjust the rest of the code to handle having different ids depending on the game, which is a pending thing still for these resource_database checks

randovania/data/json_data/prime1.json
```diff
@@ -288,6 +288,27 @@
                 "short_name": "Newborn",
                 "max_capacity": 1,
                 "extra": null
+            },
+            {
+                "index": 47,
```
henriquegemignani, line 293: Same as Energy Tank, the correct thing is support a different id.

Though Prime 1 doesn't actually have a "Item Percent" item, but it should be fine to create a fake one with a >1000 id as well.

randovania/gui/lib/preset_describer.py
```diff
@@ -327,6 +360,29 @@ def _corruption_format_params(configuration: CorruptionConfiguration) -> dict:
     return format_params
 
 
+def _prime_format_params(configuration: PrimeConfiguration) -> dict:
+    major_items = configuration.major_items_configuration
+
+    format_params = {"energy_tank": f"{configuration.energy_per_tank} energy",
+                     "include_final_bosses": _bool_to_str(not configuration.skip_final_bosses),
+                     "elevators": configuration.elevators.value,
+                     "progressive_missile": _bool_to_str(has_shuffled_item(major_items, "Progressive Missile")),
```
henriquegemignani, line 369: Prime 1 doesn't have these progressive settings

henriquegemignani (changes requested):

randovania/data/json_data/prime1.json
EthanArmbrust: Should be fixed with the additions to resource_database

randovania/gui/lib/preset_describer.py
EthanArmbrust: Removed

henriquegemignani (approved):

PR #1427, opened 2020-12-27
[WIP] Use Python 3.9

henriquegemignani:


henriquegemignani:
Segfault on Linux (related to PySide updated, still gotta figure out that).
Failure on macOS, related do pytest-asyncio?

PR #1429, opened 2020-12-27
Echoes/Corruption Database Updates

### LINKED ISSUE #1339: Alpha Splinter requires Power Beam
As the title says. Data visualizer makes Power Beam a requirement for Dark Alpha Splinter (likely due to not wanting to softlock the battle with Alpha Splinter), but might be able to work around this via weapons that aren't Power Bombs or Screw Attack.

Also causes seeds where Power Beam isn't a starting item, to make Power Beam a very early item. Fixing this would open up possibilities for more locations to find the Power Beam.

dyceron:
Fixes #1339


henriquegemignani:
Not really sure, but something in the node connections in Gel Hall is causing issues with the tests.

randovania/data/json_data/prime2.txt
```diff
@@ -2067,7 +2067,22 @@ Asset id: 2731245106
...
       After Alpha Splinter
   > Event - Alpha Splinter
-      Power Beam
+      Any of the following:
+          Power Beam or Use Screw Attack (Space Jump)
+          All of the following:
+              Morph Ball
+              Boost Ball or Morph Ball Bomb or Power Bomb
+          All of the following:
+              Light Beam ≥ 15
```
henriquegemignani, line 2076: Light Beam should be >= 1

henriquegemignani (changes requested):

dyceron:
> Not really sure, but something in the node connections in Gel Hall is causing issues with the tests.

I mostly wasn't sure how ensure that the path could only be done once, unless I'm overthinking it. There's the single room oob which is only doable one time only. So it's Corrupted Pool door -> oob (as an event, so it's done once) -> Gel Hall Door. I set the event path to have before the event so thats probably breaking it. Probably wasn't needed.

henriquegemignani:
> > Not really sure, but something in the node connections in Gel Hall is causing issues with the tests.
> 
> I mostly wasn't sure how ensure that the path could only be done once, unless I'm overthinking it. There's the single room oob which is only doable one time only. So it's Corrupted Pool door -> oob (as an event, so it's done once) -> Gel Hall Door. I set the event path to have before the event so thats probably breaking it. Probably wasn't needed.

It's less what the actual requirements are and more "maybe there's two nodes with same name" or "two connections to the same node" or something weird of this style. The broken test is the one that converts the database to the binary format, so there's something unexpected in the structure itself.

PR #1438, opened 2021-01-02
Phazon Mines Tricks and Fixes

EthanArmbrust:
Details in commit messages


randovania/data/json_data/prime1.txt
```diff
@@ -2738,7 +2758,7 @@ Asset id: 2325199700
   > Pickup (Artifact of Warrior)
       Power Bomb and Morph Ball and After Security Drone Defeated
   > Top Floor
-      Scan Visor and Space Jump Boots
+      Scan Visor and Morph Ball Bomb and Morph Ball and Bomb Jump (Intermediate)
```
dyceron, line 2761: No trickless connection to the top level?

randovania/data/json_data/prime1.txt
```diff
@@ -2852,7 +2876,12 @@ Asset id: 2423298732
 > Dock to Omega Research; Heals? False
   * Ice Door to Omega Research/Dock to Ventilation Shaft
   > Dock to Elite Control
-      Morph Ball and Boost Ball
+      All of the following:
+          Morph Ball
+          Any of the following:
+              Boost Ball
+              Space Jump Boots and Combat/Scan Dash (Expert) and Standable Terrain (Intermediate)
```
dyceron, line 2883: You don't actually need morph for the dash method.

randovania/data/json_data/prime1.txt
```diff
@@ -3032,9 +3069,7 @@ Asset id: 4272124642
   > Dock to Save Station Mines B
       After Security Drone Defeated
   > Event - Security Drone
-      Any of the following:
-          X-Ray Visor
-          Combat (Beginner) and Invisible Objects (Beginner)
+      Energy Tank or Combat (Beginner)
 
 > Dock to Quarantine Access A; Heals? False
   * Ice Door to Quarantine Access A/Dock to Central Dynamo
```
dyceron, line 3075: Should update the requirements below for the drone event to match the requirements above, with added pbs of course and the methods to reach the event in the first place

randovania/data/json_data/prime1.txt
```diff
@@ -3050,7 +3085,11 @@ Asset id: 4272124642
...
   > Dock to Dynamo Access
-      Space Jump Boots and After Security Drone Defeated
+      All of the following:
+          After Security Drone Defeated
+          Any of the following:
+              Space Jump Boots
+              Morph Ball Bomb and Morph Ball and Bomb Jump (Expert)
   > Dock to Quarantine Access A
       Power Bomb and Morph Ball and After Security Drone Defeated
   > Event - Security Drone
```
dyceron, line 3095: Same as above, for the below drone fight requirements

randovania/data/json_data/prime1.txt
```diff
@@ -3191,22 +3247,29 @@ Asset id: 4211416922
...
 ----------------
 Quarantine Access B
 Asset id: 340985721
 > Dock to Metroid Quarantine B; Heals? False
   * Plasma Door to Metroid Quarantine B/Dock to Quarantine Access B
   > Dock to Fungal Hall B
-      Space Jump Boots
+      Any of the following:
+          Space Jump Boots
+          Morph Ball Bomb and Morph Ball and Phazon Damage ≥ 150
```
dyceron, line 3265: Bomb jump missing

randovania/data/json_data/prime1.txt
```diff
@@ -3191,22 +3247,29 @@ Asset id: 4211416922
...
 
 > Dock to Fungal Hall B; Heals? False; Spawn Point
   * Plasma Door to Fungal Hall B/Dock to Quarantine Access B
   > Dock to Metroid Quarantine B
-      Space Jump Boots
+      Any of the following:
+          Space Jump Boots
+          Morph Ball Bomb and Morph Ball and Phazon Damage ≥ 150
```
dyceron, line 3272: Bomb jump missing

randovania/data/json_data/prime1.txt
```diff
@@ -3297,18 +3362,22 @@ Asset id: 3153742515
...
-          Power Bomb and Morph Ball
-          Morph Ball Bomb or Boost Ball
+      Morph Ball Bomb and Power Bomb and Morph Ball
   > Pickup (Artifact of Newborn)
-      Morph Ball Bomb and Morph Ball and Phazon Suit
+      All of the following:
+          Morph Ball Bomb and Morph Ball
+          Any of the following:
+              Phazon Damage ≥ 2000
+              Boost Ball and Phazon Damage ≥ 1600
```
dyceron, line 3371: Probably worth adding a trick for this, like movement. Also the damage values are too high. They both assume you need phazon suit. If you want one too require phazon, set it to damage 1500. The other I assume you don't, so you'd put 1400

randovania/data/json_data/prime1.txt
```diff
@@ -3297,18 +3362,22 @@ Asset id: 3153742515
...
   > Dock to Fungal Hall B
       All of the following:
           Power Bomb and Morph Ball
-          Morph Ball Bomb or Boost Ball
+          Any of the following:
+              Morph Ball Bomb
+              Boost Ball and Movement (Beginner)
 
 > Pickup (Artifact of Newborn); Heals? False
   * Pickup 88; Major Location? True
```
dyceron, line 3383: Related to above, but maybe the damage required without phazon should be split among the two nodes. So reaching without phazon on the previous node needs X depending on the difficulty, then damage required to leave from the pickup node. At least that one will be consistent without phazon suit, so only one lowest level is needed there

randovania/data/json_data/prime1.txt
```diff
@@ -3322,16 +3391,22 @@ Asset id: 3734860277
   * Plasma Door to Fungal Hall A/Dock to Fungal Hall Access
   > Dock to Elevator B
       Space Jump Boots
+  > Pickup (Missile)
+      Morph Ball and Phazon Damage ≥ 100
```
dyceron, line 3395: Might be worth having a suit requirement here as the no trick path, and without the suit as beginner movement or something

randovania/data/json_data/prime1.json
```diff
@@ -16727,7 +16824,7 @@
                     "in_dark_aether": false,
                     "asset_id": 907887024,
                     "default_node_index": 1,
-                    "valid_starting_location": true,
+                    "valid_starting_location": false,
```
dyceron: This could be left as a valid start since it requires an event to leave anyway. Good for a potential room rando if that ever happens.

randovania/data/json_data/prime1.json
```diff
@@ -17564,7 +17661,7 @@
                     "in_dark_aether": false,
                     "asset_id": 902158134,
                     "default_node_index": 0,
-                    "valid_starting_location": true,
+                    "valid_starting_location": false,
```
dyceron: Same as above comment

randovania/data/json_data/prime1.txt
EthanArmbrust: I've added a drone fight trigger node to central dynamo to simplify the room. There should only be one connection to climb the room and one connection to fight the drone now.

EthanArmbrust: Also, what do you think about the difficulty of the HBJ(?) to escape Central Dynamo? I can't actually do this trick so I took a guess at how hard it is. Same for the 21% Vent Shaft half pipe bomb jump

dyceron: Hmm to me it looks like the old paths for the drone still exist, with "Xray or invisible objects." Unless thats been changed outside the PR already.

As for the bomb jumps, I think that is fair for now.

EthanArmbrust: e6d536da should have removed all of the old paths to the drone, unless I missed something

randovania/data/json_data/prime1.txt
```diff
@@ -3070,36 +3070,26 @@ Asset id: 4272124642
...
 
 > Dock to Save Station Mines B; Heals? False
   * Ice Door to Save Station Mines B/Dock to Central Dynamo
   > Dock to Dynamo Access
-      All of the following:
-          After Security Drone Defeated
-          Any of the following:
-              Space Jump Boots
-              Morph Ball Bomb and Morph Ball and Bomb Jump (Expert)
+      Trivial
```
dyceron: Since the Save Station door connects to the Drone Trigger node and that node connects to Dynamo Access, the connection to Dynamo Access from this node should be removed. Outside of that simplification, listing it as trivial is wrong anyway.

EthanArmbrust: Good catch. I've removed the connection

randovania/data/json_data/prime1.txt
EthanArmbrust: In my latest commit, I've put all the phazon suit/damage requirements into the FHB door -> item connection and left out any of requirements for the return trip. I didn't want to split up the damage requirements because the non-linear increase of phazon damage is going to make it difficult to figure out the precise damage requirements.

randovania/data/json_data/prime1.txt
EthanArmbrust: Fixed

randovania/data/json_data/prime1.json
EthanArmbrust: I don't see how this one can ever be valid. Since the default spawn point is the save station, you'd be trapped in the little space on the wrong side of the gate

randovania/data/json_data/prime1.json
EthanArmbrust: Fixed

randovania/data/json_data/prime1.txt
dyceron: I don't think it should be too hard really. regardless of the difficulty except no tricks, the damage for returning would be the same. Could look at 100% runs to get a rough estimate.

randovania/data/json_data/prime1.json
dyceron: Maybe not for the Save Station, unless the spawn is changed. But this is probably fine.

randovania/data/json_data/prime1.txt
EthanArmbrust: I've split up the damage based on [this video](https://www.youtube.com/watch?v=zpyhbOerMT4). For the path back, I've added paths for movement levels of 2-4, with a 100 energy decrease at each trick level.

dyceron: Seems good for now. It can always be fixed/adjusted later on as needed.

PR #1451, opened 2021-01-09
Open folder for previous game history on non-Windows plaforms

EthanArmbrust:
Clicking on "Previous Game History" in "Open" menu opens the folder in Finder on macOS or the default file manager for Linux based systems. Falls back to showing the textbox with the path if this fails.

### DIFF
--- a/randovania/gui/main_window.py
+++ b/randovania/gui/main_window.py
@@ -544,9 +544,9 @@ class MainWindow(WindowManager, Ui_MainWindow):
             if platform.system() == "Windows":
                 os.startfile(path)
             elif platform.system() == "Darwin":
-                subprocess.Popen(["open"], path)
+                subprocess.run(["open", path])
             else:
-                subprocess.Popen(["xdg-open", path])
+                subprocess.run(["xdg-open", path])
         except OSError:
             box = QtWidgets.QMessageBox(QtWidgets.QMessageBox.Information, "Game History",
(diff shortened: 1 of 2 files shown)


randovania/gui/main_window.py
```diff
@@ -539,9 +540,14 @@ def _open_auto_tracker(self):
...
     def _on_menu_action_previously_generated_games(self):
         path = self._options.data_dir.joinpath("game_history")
-        if platform.system() == "Windows":
-            os.startfile(path)
-        else:
+        try:
+            if platform.system() == "Windows":
+                os.startfile(path)
+            elif platform.system() == "Darwin":
+                subprocess.Popen(["open"], path)
```
henriquegemignani, line 547: This is likely supposed to be `["open", path]`, no?

henriquegemignani:
This is missing some test improvements as well, which is why the tests failed.
Maybe use `subprocess.run` instead of `subprocess.Popen`? Not sure in this precise case.

randovania/gui/main_window.py
EthanArmbrust: Yep. I'm going to try to test this on an old macbook to make sure the macOS stuff works, as well as make sure the test covers macOS

EthanArmbrust:
I've tested this works on macOS, and tests have been updated to include macOS and OSErrors

PR #1463, opened 2021-01-14
Echoes: Minor Adjustments

dyceron:


randovania/data/json_data/prime2.txt
```diff
@@ -10501,7 +10501,7 @@ Asset id: 2838429875
           Morph Ball
           Any of the following:
               Morph Ball Bomb
-              Power Bomb ≥ 2 and Knowledge (Intermediate)
+              Power Bomb ≥ 2 and Knowledge (Advanced)
```
henriquegemignani, line 10504: Shouldn't this be Bomb slots without Bombs? Despite not being one that uses a charge combo, it fits in the name

dyceron: Personally no. It's not the same type of trick as that. I get the idea but now you'd have to enable that trick to get this in logic, which I'm not a fan of.

PR #1555, opened 2021-03-03
Potential NSJ SA Fix

dyceron:
Ensures that logic paths that use NSJ Screw Attacks are ignored if Space Jump is on the logical path prior to it.

Example: Having Space Jump but attempting to leave Sanctuary Temple Access in reverse, which can be done on Beginner Z-Axis, but impossible with Space Jump. Logic assumes it is possible to cross regardless.


henriquegemignani:
I'll generate some thousand seeds with this change to have an idea of what kind of impact does this have on generation.

henriquegemignani (approved):

PR #1776, opened 2021-05-26
Show correct game name in description label of GameInputDialog

EthanArmbrust:
image

Show the correct game name in the description label above the input file entry.
Don't show the text about randovania saving a copy of the game if the patcher uses input file directly.  
Minor grammar fix changing "a ISO" to "an ISO"


henriquegemignani:
Thanks!
I think long term the Patcher class should have a game attribute? Though right now I'm unsure if a given patcher class should support only a single game. Might be the best.

PR #1778, opened 2021-05-26
Remember prime 1 iso input path

EthanArmbrust:
Once more games are supported, it may be better to store each game's input and output paths in a JSON, but this should be ok for now.


henriquegemignani:
I haven't done this yet because I want to do it properly. Like, even saving the path for Echoes if the internal copy is deleted.
With each hurdle solved while implementing Prime 1, I want the code base to be more flexible and ready for other games, not less.

So no, I'd prefer to implement something that keeps track of a path for each game. Maybe the same could also be used for output?

PR #1810, opened 2021-05-30
Twin Fires Logic

Draconif:


randovania/data/json_data/prime1.txt
```diff
@@ -1224,7 +1227,7 @@ Asset id: 1282373491
       Any of the following:
           Slope Jump (Advanced)
           All of the following:
-              Space Jump Boots
+              Space Jump Boots and No Gravity Suit and Enabled Allow Underwater Movement without Gravity Suit
```
henriquegemignani, line 1230: The only path without tricks enabled has a `No Gravity Suit`. Is that correct?

Draconif: that looks like a mistake I made, should just say space jump

randovania/data/json_data/prime1.txt
```diff
@@ -1059,13 +1059,16 @@ Asset id: 3617515525
...
               Gravity Suit
               Any of the following:
                   Space Jump Boots
                   Morph Ball Bomb and Morph Ball and Bomb Jump (Beginner)
-          Space Jump Boots and Slope Jump (Intermediate)
+          All of the following:
+              No Gravity Suit and Enabled Allow Underwater Movement without Gravity Suit
+              Any of the following:
+                  Slope Jump (Expert)
+                  Space Jump Boots and Slope Jump (Intermediate)
```
dyceron, line 1071: No need for no gravity for space jump method

dyceron (changes requested):

PR #1875, opened 2021-06-07
Life Grove Fix

EthanArmbrust:
Require that you can actually reach the item after using the spinner.
This seed is impossible because of the lack of this check: [Prime Randomizer - Radion Transport Warrior.txt](https://github.com/randovania/randovania/files/6606945/Prime.Randomizer.-.Radion.Transport.Warrior.txt)

I wasn't really sure where to put the bomb jump difficulty. I put it at advanced since it needs to be laddered, but it's not very hard

### DIFF
--- a/randovania/data/json_data/prime1.txt
+++ b/randovania/data/json_data/prime1.txt
@@ -5347,6 +5347,9 @@ Asset id: 2263559682
               All of the following:
                   Boost Ball or Spinners without Boost (Advanced)
                   Any of the following:
+                      Space Jump Boots
+                      Morph Ball Bomb and Bomb Jump (Advanced)
+                  Any of the following:
                       Morph Ball Bomb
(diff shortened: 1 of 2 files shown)


randovania/data/json_data/prime1.txt
```diff
@@ -5346,6 +5346,9 @@ Asset id: 2263559682
               Space Jump Boots and Standable Terrain (Advanced) and Single Room Out of Bounds (Advanced)
               All of the following:
                   Boost Ball or Spinners without Boost (Advanced)
+                  Any of the following:
+                      Space Jump Boots
+                      Morph Ball Bomb and Bomb Jump (Advanced)
```
dyceron, line 5351: This is fine. I'll take a look at to see if it should be adjusted.

PR #1894, opened 2021-06-09
Echoes Hints: Add Ice Beam and Ice Spreader to words that should be preceded by "an"

EthanArmbrust:


randovania/games/prime/patcher_file_lib/item_hints.py
```diff
@@ -22,6 +22,8 @@
     "Emerald Translator",
     "Energy Transfer Module",
     "Energy Tank",
+    "Ice Beam",
+    "Ice Spreader"
```
dyceron, line 26: X-Ray Visor should also be added here

PR #1910, opened 2021-06-13
Support multiple output formats for supported patchers

EthanArmbrust:
Since dol patching is handled by randomprime now, randovania should be able to save .ciso and .gcz for prime 1 now with no issue.  
If a patcher has multiple output formats supported, a set of radio buttons will appear to let the user select which format they want.  
image
Output format preference is saved per-game via the Options class. As a note, I had to bump the options file version to 15.

### DIFF
--- a/randovania/interface_common/options.py
+++ b/randovania/interface_common/options.py
@@ -82,7 +82,7 @@ class PerGameOptions:
         return PerGameOptions(
             cosmetic_patches=cosmetic_patches,
             input_path=decode_if_not_none(value["input_path"], Path),
-            output_format=decode_if_not_none(value["output_format"], str),
+            output_format=value["output_format"],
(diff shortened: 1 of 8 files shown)


randovania/interface_common/options.py
```diff
@@ -80,6 +82,7 @@ def from_json(cls, value, game: RandovaniaGame) -> "PerGameOptions":
         return PerGameOptions(
             cosmetic_patches=cosmetic_patches,
             input_path=decode_if_not_none(value["input_path"], Path),
+            output_format=decode_if_not_none(value["output_format"], str),
```
henriquegemignani, line 85: This can just be `value["output_format"]`

henriquegemignani:
I was planning on just allowing the other names in the output dialog and the user could name the iso differently. This way seems better :)

Just fix the tests though!

PR #2084, opened 2021-08-21
added charge beam requirements with combat tricks to some prime 1 rooms

blh2666:


henriquegemignani (changes requested):
You'll need to update with randovania's main. 

There's a section on the changelog for logic changes now. Could you be explicit on the rooms that were changed? Maybe something like "Charge Beam requirements were added to the following rooms:" , with a list of room names that also mentions the Combat or damage alternative.

Reminder that the damage values in the database are multiplied by the strictness. So 300 damage ends up being 450 if you run with the Medium (1.5x) strictness. If you believe the values are still fine, no need to change.

blh2666:
Accounted for damage strictness, merged the branches, and expanded the changelog as requested.

randovania/data/json_data/prime1/Chozo Ruins.txt
```diff
@@ -1395,12 +1397,16 @@ Asset id: 4216627403
               Any of the following:
                   Space Jump Boots
                   Morph Ball Bomb and Morph Ball and Bomb Jump (Beginner)
-          Wave Beam and After Hall of the Elders Bomb Slots
+          All of the following:
+              Wave Beam and After Hall of the Elders Bomb Slots
+              Charge Beam or Combat (Intermediate)
```
dyceron, line 1402: This change is unnecessary since the ghosts dont spawn when you use the wave slot, only after you reload the room. And even then, they arent required to fight and dont need to be accounted for.

randovania/data/json_data/prime1/Chozo Ruins.txt
```diff
@@ -1395,12 +1397,16 @@ Asset id: 4216627403
...
           Morph Ball Bomb and Morph Ball and Complex Bomb Jump (Expert)
   > Door to Elder Hall Access
       Trivial
   > Pickup (Energy Tank)
-      Ice Beam and After Hall of the Elders Bomb Slots
+      All of the following:
+          Ice Beam and After Hall of the Elders Bomb Slots
+          Charge Beam or Combat (Intermediate)
```
dyceron, line 1409: This change is also unnecessary. The change should be applied from Crossway Access South to bomb slots event, since thats when you have to fight the ghosts.

blh2666: Ghosts are required to be fought in order to use the bowling statue. Not the bomb slots. Only the wave and ice bomb slots need bowling statue in order to be used. Plasma bomb slot can be used while ghosts are still alive as it just moves the statue. Hence why I tied the charge requirement to the individual bomb slots instead of the room in general.

henriquegemignani (approved):

PR #2086, opened 2021-08-22
Database - New Tricks and Fixes

toasterparty:
These are all things brought up by the community in the discord server. I'll let the `.txt` diff speak for itself.

### DIFF
--- a/randovania/data/json_data/prime1/Phendrana Drifts.txt
+++ b/randovania/data/json_data/prime1/Phendrana Drifts.txt
@@ -356,27 +356,23 @@ Ruined Courtyard
 Asset id: 421627757
 > Door to Courtyard Entryway; Heals? False
   * Normal Door to Courtyard Entryway/Door to Ruined Courtyard
+  > Door to Specimen Storage
+      All of the following:
+          Morph Ball and Spider Ball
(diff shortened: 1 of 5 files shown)


randovania/data/json_data/prime1/Phendrana Drifts.txt
```diff
@@ -59,6 +59,10 @@ Asset id: 4146616697
   * Normal Door to Ruins Entryway/Door to Phendrana Shorelines
   > Door to Shoreline Entrance
       Trivial
+  > Door to Temple Entryway
+      All of the following:
+          Combat/Scan Dash (Intermediate)
+          Scan Visor or X-Ray Visor
```
dyceron, line 65: Cant this be done in any visor? If so, could be an invisible objects intermediate or something added on to this

toasterparty: The only viable scan point is the flicker bats. They cannot be locked onto with combat visor, but they can be locked onto with x-ray.

dyceron:
Make the appropriate changelog entries as well.

toasterparty:
> Make the appropriate changelog entries as well.

Done

PR #2090, opened 2021-08-24
fixes #392 - use different game ID with menu mod

### LINKED ISSUE #392: Use a different Game ID if Menu Mod is enabled
Save files are absolutely incompatible, so it would be nice to use different Game IDs so they don't share a memory card save.

duncathan:
echoes uses a different game ID (GM2E1R) when saving ISOs with menu mod enabled, so that different save file formats aren't shared

### DIFF
--- a/randovania/games/patchers/claris_patcher_file.py
+++ b/randovania/games/patchers/claris_patcher_file.py
@@ -402,9 +402,9 @@ def create_patcher_file(description: LayoutDescription,
     result = {}
     _add_header_data_to_result(description, result)
 
-    result["game_id"] = "G2ME0R"
+    result["publisher_id"] = "0R"
     if configuration.menu_mod:
-        result["game_id"] = "G2ME1R"
+        result["publisher_id"] = "1R"
 
     result["convert_other_game_assets"] = cosmetic_patches.convert_other_game_assets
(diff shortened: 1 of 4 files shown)


randovania/games/patchers/claris_patcher_file.py
```diff
@@ -402,6 +402,10 @@ def create_patcher_file(description: LayoutDescription,
     result = {}
     _add_header_data_to_result(description, result)
 
+    result["game_id"] = "G2ME0R"
```
henriquegemignani, line 405: This breaks non-NTSC, as `G2ME` is the NTSC id.

henriquegemignani (changes requested):

randovania/games/patchers/claris_patcher_file.py
duncathan: good catch, will change in the morning so it only modifies the publisher ID

duncathan:
the region support has been fixed. I see that I've failed a few checks - how should I go about resolving them?

henriquegemignani:
> 
> 
> the region support has been fixed. I see that I've failed a few checks - how should I go about resolving them?

The macOS fails occasionally. The coverage one you'd need to improve the tests and you can check the codecov report for more details on where.
I'm not that strict on code coverage: new code I usually prefer, not necessarily up to the 84+% target it has right now, but I'm fine with not touching it too.

PR #2095, opened 2021-08-24
log analyzer improvements

toasterparty:
Adds additional insights to the log analyzer.

New usage example:
```
python tools\log_analyzer.py D:\Libraries\Downloads\10K_batch-gen-seeds-odd --use-percentage D:\test.json --csv-dir D:\test --major-progression-only
```


tools/log_analyzer.py
```diff
@@ -184,26 +233,87 @@ def item_creator():
...
+            location_progression_count[location] = 0
+        
+        for item in items:
+            if location not in items[item].keys():
+                items[item][location] = 0
+
+    if use_percentage:
+        for item in items:
+            for room in items[item]:
+                items[item][room] = (float(items[item][room])/float(seed_count))
```
henriquegemignani, line 292: no need to cast everything to floats, in Python 3 `/` is float division (`//` is for integer division)

tools/log_analyzer.py
```diff
@@ -8,7 +8,7 @@
 import statistics
 from pathlib import Path
 from statistics import stdev
-from typing import Dict, Tuple, Optional, List, Iterable, Set
+from typing import Dict, Tuple, Optional, List, Iterable, Set, final
```
henriquegemignani, line 11: This seems unused.

tools/log_analyzer.py
```diff
@@ -17,6 +17,45 @@
 from randovania.games.game import RandovaniaGame
 from randovania.layout.layout_description import LayoutDescription
 
+NON_PROGRESSION = [
```
henriquegemignani, line 20: the code previously was using item names that show up in the item order as progression things. The only things that wouldn't catch is Artifacts, but that's being inconsistent with Sky Temple Keys.

Is there any particular reason for this?

henriquegemignani (changes requested):

tools/log_analyzer.py
toasterparty: It was needed for `regions_weighted`, because that looked at `game_modifications`. I think I can filter without this list. Let me try something.

toasterparty: I was able to remove the `--progression-only` flag and this `NON_PROGRESSION` list.

henriquegemignani (approved):

PR #2132, opened 2021-09-03
removed shorelines reposition trick

toasterparty:
removed trick as discussed, patcher change will come later


dyceron:
I'd also delete the event node for the spider track item collection

PR #2188, opened 2021-09-26
Adds template for killing Quad MBs and uses it as needed

duncathan:
```
* Kill Quad MB:
      Any of the following:
          All of the following:
              Morph Ball
              Any of the following:
                  Boost Ball
                  All of the following:
                      Combat (Beginner)
                      Morph Ball Bomb or Power Bomb
          Knowledge (Beginner) and Use Screw Attack (Space Jump)
```


duncathan:
in HoCM I added this: `After Spider Guardian or Kill Quad MB`
I know it's technically redundant, as Dark pointed out, so I can remove the spider guardian req if desired. I just like being explicit

PR #2191, opened 2021-09-27
removed overly-conservative modal hudmemos

toasterparty:
for multiworld quality of life


henriquegemignani:
Yes, there's a test that expects the precise patcher json for a sample prime1+prime2 multiworld seed, it's the `test_create_patch_data`.

PR #2200, opened 2021-09-29
Added a "Foldable" widget to make blocks that can be collapsed

Dinopony:
Also contains a small fix for the "as_json" problem in ItemCategory

### DIFF
--- a/test/gui/lib/test_foldable.py
+++ b/test/gui/lib/test_foldable.py
@@ -1,21 +1,25 @@
 import pytest
-
 from randovania.gui.lib.foldable import Foldable
 
 
-def test_foldable_initial_state():
+def test_foldable_initial_state(skip_qtbot):
     # Setup & Run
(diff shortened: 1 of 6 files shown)


Dinopony:
I don't get what's wrong with the new testfile I added. It works fine on my machine, but both Windows executable & macOS executable checks trip on this.

test/gui/lib/test_foldable.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+
+from randovania.gui.lib.foldable import Foldable
+
+
+def test_foldable_initial_state():
```
henriquegemignani, line 6: Check the other tests that involve GUI. I think all you need is to add the `skip_qtbot` fixture, but might be a good idea to add your widget to the qtbot thingy so they're cleaned up.

henriquegemignani (approved):

PR #2211, opened 2021-09-30
Add Prime 3 Minimal Logic

dyceron:
Still missing Energy Cells at the moment


randovania/resolver/bootstrap.py
```diff
@@ -55,6 +79,14 @@
         # Light Ammo
         46: 100
     },
+    RandovaniaGame.METROID_PRIME_CORRUPTION: {
+        # Missile
+        4: 100,
+        # Energy Tank
+        20: 14,
+        # Ship Missile
+        45, 11
```
henriquegemignani, line 88: `45: 11`

dyceron: Would it make sense to add Energy Cells into the list here?

PR #2213, opened 2021-10-01
Changed the LocationPoolTab to make it use Foldables

Dinopony:
This comes with a rework of how the location pool is presented, with additionnal radio buttons added for future implementations (force vanilla item into location, set fixed item in location...)

For some reason, commits from the last PR are still added inside this one, but cause no changes since they were effectively merged.


henriquegemignani:
> For some reason, commits from the last PR are still added inside this one, but cause no changes since they were effectively merged.

I squash-merge your old PR, that's the reason.

henriquegemignani:
Can you add some screen shots of how it looks?

Dinopony:
I don't know how I was expecting you to guess how it looks without screenshots... Here we go:

## Initial state



## Tooltip text for major/minor split



## One world unfolded



## Minor location inputs disabled when checking major/minor splits

henriquegemignani:
Not sure if I prefer to description of the setting with tooltip. I feel it's not really obvious. Dolphin kinda uses it for the graphics settings, but they're also using a custom component that displays the tooltip a lot more obviously.

Also, I'd prefer to keep the future settings invisible instead of disabled. You could use the preview flag to keep it hidden only if that's off.

Dinopony:
OK, I can revert back the minor/major presentation in a box.
As for the disabled instead of hidden matter, I chose to do this because I noticed a few "NYI" options in the multiworld interface and thought that was the way to go, but I guess it's indeed wiser to keep it hidden not to give false hopes in case we couldn't be able to implement it for some reason.
Thanks for the feedback, I'll fix this soon.

henriquegemignani:
I should likely hide these in the multiworld stuff. I was _planning_ on doing it right after, but I actually never went back to it at all.

Dinopony:
There we go! Locations are now put as one-liners as long as we don't have more items to show.

image

Dinopony:
Why is CodeCov complaining on files like randovania/cli/prime_database.py whereas I didn't change them at all?
On top of that, the check was passing properly on the previous commits, and by changing 3 trivial lines of code, we would have lost coverage? :/

PR #2309, opened 2021-11-16
Basic support for Dread, turning database format more general

henriquegemignani:
All worlds must have an unique name, all areas inside a world must have an unique name.
Asset id and similar other things are no longer a part of the object and must be placed in the "extra".


randovania/game_description/world/area.py
```diff
@@ -23,19 +23,29 @@ def __repr__(self):
         return "Area[{}]".format(self.name)
 
     def __hash__(self):
-        return self.area_asset_id
+        return hash(self.name)
+
+    @property
+    def in_dark_aether(self) -> Optional[bool]:
+        return self.extra.get("in_dark_aether")
```
duncathan, line 30: rather than being optional, would it make more sense to do `return self.extra.get("in_dark_aether', False)`?

randovania/game_description/world/area.py
```diff
@@ -23,19 +23,29 @@ def __repr__(self):
...
+
+    @property
+    def area_asset_id(self) -> int:
+        return self.extra["asset_id"]
```
duncathan, line 34: this should either be optional, and use `get()`, or raise a more detailed error if called on an area with no asset ID

randovania/game_description/world/dock.py
```diff
@@ -74,8 +75,12 @@ def get_by_type_and_index(self, dock_type: DockType,
 
 @dataclass(frozen=True, order=True)
 class DockConnection:
-    area_asset_id: int
+    area_name: str
     dock_index: int
 
+    @property
+    def area_asset_id(self):
+        return
```
duncathan, line 83: what's this for?

randovania/game_description/world/world.py
```diff
@@ -1,24 +1,33 @@
...
     def __repr__(self):
         return "World[{}]".format(self.name)
 
+    @property
+    def dark_name(self) -> Optional[str]:
+        return self.extra.get("dark_name")
+
+    @property
+    def world_asset_id(self) -> int:
+        return self.extra["asset_id"]
```
duncathan, line 29: see above

duncathan (changes requested):
While we're in here, making similar changes, another thing we ought to do rather than split it into two cumbersome schema upgrades is to drop the indices from entries in the resource database. Similar to using names as keys for worlds/areas instead of asset IDs, we should use names as keys for resource types/IDs instead of the numerical indices. `short_name` would be appropriate for this purpose, I think. I imagine it will mostly be a matter of using DictAdapter for them, right? It would be a huge improvement for DB readability, especially when editing templates (which we don't have a UI for yet).

randovania/games/binary_data.py
```diff
@@ -228,93 +222,112 @@ def _build_resource_info(**kwargs):
...
             "translator_gate": Struct(
+                **NodeBaseFields,
                 gate_index=VarInt,
             ),
             "logbook": Struct(
+                **NodeBaseFields,
                 string_asset_id=VarInt,
                 lore_type=ConstructLoreType,
-                extra=VarInt,
+                lore_extra=VarInt,
```
duncathan, line 288: `lore_extra` and `string_asset_id` should be moved to the `extra` field as well imo

randovania/game_description/world/area.py
henriquegemignani: I was thinking of removing this getter, but I can't see a way to decently do this for a while, sadly.

PR #2337, opened 2021-11-20
Add Padding and Hack to add EVNT id to ANIM file

Unknownforce351:
This is hacky, but it works.  This should really be created at build time of the asset file.

Padding seems to work in all cases at least.  So that's a plus.

You can fix my garbage code ;)

### DIFF
--- a/randovania/patching/prime/asset_conversion.py
+++ b/randovania/patching/prime/asset_conversion.py
@@ -299,11 +299,27 @@ def convert_prime2_pickups():
                 for asset in converter.converted_assets.values()
             ],
         }, meta_out, indent=4)
-
+    
+    unique_anim = []
     for asset in converter.converted_assets.values():
+        assetdata = format_for(asset.type).build(asset.resource, target_game=Game.PRIME)
+        if len(assetdata) % 32 != 0:
+            assetdata += b"\xFF" * (32-(len(assetdata) % 32))
         Path("converted").joinpath(f"{asset.id}.{asset.type.upper()}").write_bytes(
-            format_for(asset.type).build(asset.resource, target_game=Game.PRIME)
+            assetdata
         )
+        if asset.type == "ANIM":
+            for assetb in converter.converted_assets.values():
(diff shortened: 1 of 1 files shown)


randovania/patching/prime/asset_conversion.py
```diff
@@ -299,11 +299,27 @@ def id_generator(asset_type):
...
+            assetdata
         )
+        if asset.type == "ANIM":
+            for assetb in converter.converted_assets.values():
+                if assetb.type == "ANCS" and any(asset.id in x for x in converted_dependencies[assetb.id]):
+                    for dep in converted_dependencies[assetb.id]:
+                        if asset.id in unique_anim:
+                            continue
+                        if dep.type == "EVNT":
+                            with open(Path("converted").joinpath(f"{asset.id}.{asset.type.upper()}"),"r+b") as anim:
```
henriquegemignani, line 318: Instead of hacking up the files after they're written, you could do a for loop _before_ they're written and edit there, while they're still in the decoded dict format so no magic needed.

henriquegemignani (changes requested):

duncathan:
is it just me, or is this better suited for a retro-data-structures PR?

Unknownforce351:
I'm sure there's a better way to do it, but I don't know python well enough to make it happen in the other repo.

This is a unique case in the conversion process.  It's the only instance where an asset type doesn't exist at all in one game and needs to be created and consumed by other files in order to function in the other.  This should be in that repo for sure, because once we dive into MP3 items, this function will definitely be needed.

The ANIM file in MP2 does not have an asset ID for EVNT as the format of it changed and they removed the EVNT files completely, and instead embedded their content into to the ANCS file directly.

To make them work in MP1, we have to split out the EVNT content and create a completely new asset file/id.  The current process gives it a new ID, but those details are not currently exposed in the currently written converter.

The asset converter as it's written right now, doesn't expose new dependency asset ids to other dependencies.  I think that is what we would need at this point.

randovania/patching/prime/asset_conversion.py
Unknownforce351: Fixed.

PR #2338, opened 2021-11-20
convert resource database to keys instead of indices

duncathan:
oh god it's breaking everything


randovania/game_description/schema_migration.py
```diff
@@ -1,8 +1,12 @@
+from typing import Callable, Optional, Union
+from build.lib.randovania.game_description.resources.resource_type import ResourceType
```
henriquegemignani, line 2: This seems like a mess

randovania/game_description/schema_migration.py
```diff
@@ -69,19 +73,147 @@ def _migrate_v1(data: dict) -> dict:
...
+            red["name"] = find_resource(ResourceTypeLatest.ITEM, red.pop("index"))
+        return reduction
+    
+    db["damage_reductions"] = [migrate_reduction(red) for red in db["damage_reductions"]]
+
+    def migrate_requirement(requirement: dict) -> dict:
+        data = requirement["data"]
+
+        if requirement["type"] == "resource":
+            data["type"] = ResourceTypeLatest.from_index(data["type"]).value
```
henriquegemignani, line 104: A best practice to the migration functions is to not use project functionality which can then be changed and would break the migration in unexpected ways.

But I think that for the database migration, we can keep this around just for a short while and remove it so it doesn't really matter.

randovania/games/prime1/patcher/prime_items.py
```diff
@@ -2,7 +2,20 @@
 Some constants related to specific Prime 1 items that the logic needs to deal with.
 """
 
-ARTIFACT_ITEMS = [29 + i for i in range(12)]
+ARTIFACT_ITEMS = [
```
henriquegemignani, line 5: List list is quite redundant with the `ARTIFACT_NAMES` below

randovania/games/prime3/patcher/corruption_items.py
```diff
@@ -3,5 +3,5 @@
 """
 
 ENERGY_CELL_MODEL = "Energy Cell"
-ENERGY_CELL_ITEMS = [23 + i for i in range(9)]
-ENERGY_CELL_TOTAL_ITEM = 22
+ENERGY_CELL_ITEMS = ["Fuse"+str(i+1) for i in range(9)]
```
henriquegemignani, line 6: `f"Fuse{i + 1}"`

henriquegemignani (changes requested):

randovania/game_description/migration_data.py
```diff
@@ -1,5 +1,6 @@
 import functools
 import json
+from build.lib.randovania.game_description.resources.resource_type import ResourceType
```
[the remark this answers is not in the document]
duncathan: whoops that's wrong

randovania/layout/base/trick_level_configuration.py
```diff
@@ -39,7 +39,7 @@ def bit_pack_encode(self, metadata) -> Iterator[Tuple[int, int]]:
         encodable_levels = list(LayoutTrickLevel)
         encodable_levels.remove(LayoutTrickLevel.DISABLED)
 
-        for trick in sorted(_all_tricks(resource_database)):
+        for trick in _all_tricks(resource_database):
```
henriquegemignani, line 42: Any reason it doesn't sort anymore? The sort was there because this code is highly dependant on the order and having a sort makes to not care so much whatever happen on the other end

henriquegemignani (approved):

PR #2343, opened 2021-11-21
DockNode connect to node

henriquegemignani:
Also adds a description field to nodes


randovania/game_description/editor.py
```diff
@@ -65,37 +73,68 @@ def sub(n: Node):
...
                             )
 
                     if new_node is not None:
                         self.replace_node(area, node, new_node)
 
-        new_area = dataclasses.replace(current_area, name=new_name)
-        current_world.areas[current_world.areas.index(current_area)] = new_area
+    def replace_references_to_node_identifier(self, old_identifier: NodeIdentifier, new_identifier: NodeIdentifier):
+        if old_identifier == new_identifier:
+            return
```
duncathan, line 123: you've got some redundant checks for the same condition before calling this function, could just remove those and let the function handle it

randovania/game_description/world/node.py
```diff
@@ -78,15 +80,26 @@ class GenericNode(Node):
 
 @dataclasses.dataclass(frozen=True)
 class DockNode(Node):
-    dock_index: int
-    default_connection: DockConnection
+    """
+    Represents a connection to another area via something similar to a door and it's always to another DockNode.
+    The dock weakness describes the types of door the game might have, which could be randomized separately from where
+    the door leads to.
+    Alias is used when there are multiple pairs of dock nodes between the same two areas.
```
duncathan, line 87: Alias?

duncathan (changes requested):

randovania/game_description/world/node.py
henriquegemignani: Oh, something I gave up on having. good catch

duncathan (approved):

PR #2353, opened 2021-11-22
update changelog

duncathan:
I called the description field major because requirement comments were major in 3.2.0 but I am not at all attached to that


henriquegemignani:
There's a considerable chance next feature release is 4.0 with Dread, so
the major/minor lost is definitely changing lol

PR #2436, opened 2021-12-07
Initial Logic for Cataris. No walljumps. Heat/Lava damage are placeholders

kirbymastah:


henriquegemignani:
I'm listing these issues, but I'm fine with merging the PR first and then having another for fixes.

- Kraid Eyedoor Room: missing a path to StartPoint.
- Kraid Room: missing a way of crossing the room by fighting Kraid?
- Kraid Room: missing a path to the morph ball tunnel in the lava
- 043: Dock to Above Kraid is "Not Determined".
- Rooms before Kraid are missing a Varia Suit requirement.
- Teleport to Dairon: Dock to Diffusion Beam Room is Not Determined
- Teleport to Dairon: Missing Wave (and Diffusion?) way for destroy the missile tank blob.
- 054: block to east dock has cold damage
- 053: The wave door isn't marked as wave
- 053: I think we can have a hacky connection from the top-left dock to the wave door with wave as requirement, until there's a proper implementation of blast shields and remote activation.
- 052: back of the wave door from be Access Denied
- 019: Missing Varia Suit requirement
- 001: Use the "Simple IBJ" template (not a trick if has cross bombs, has the requirement for bombs+morph)
- 001: Spin boost doesn't reach the top? O.o
- Moving Magnet Walls (Tall): why does bottom left to top left door require Spider Magnet and speed booster?
- Teleport to Artaria (Blue):
    - Varia Suit requirements
    - Uppler-Left Room missing connection to the blocks just to it's right
    - Centter Room to Tile Group Bomb 1 should use the Simple BSJ template (missing bombs requirement, etc)
    - Same with Bomb 3 to 2
    - Teleporter is missing path to Lava Room
    - Right Room is missing connections to the SA Blocks
- 045: lower part is heated before some event (hallway thermal device?)
- 010: lower door to morph ball door should use Simple BSJ template
- Emmi Zone: morph ball tunnels only after EMMI is defeated. Maybe via a new type of dock weakness "Cataris EMMI Vent"?

henriquegemignani:
- 046: Top Left door is an Access Open, not Phase Shift Door.
- 059: Same thing with left door
- Teleporter to Artaria (Red):
    - Bottom-right door should go to teleporter just with morph
    - Teleporter to pickup with just morph
- Teleporter to Ghavoran:
   - Missing color
   - Lower Center -> Beam 3 should use simple ibj
   - Teleporter -> Top room needs morph ball
   - Top Room -> Pickup should use simple ibj

henriquegemignani:
- Speedboost Maze before Transport to Dairon
    - Maybe just "... near Dairon"?
    - Lower right dock is Not Determined
    - Right dock is also not Determined - maybe make a helper node in front to handle the thermal device event

PR #2444, opened 2021-12-09
Prime - Fix 2-way Elevator Shuffle

toasterparty:


henriquegemignani (changes requested):
You also need to add these to the preset migration, or everyone's existing presets will also show this issue.

toasterparty:
@henriquegemignani Like this?

henriquegemignani (approved):

PR #2461, opened 2021-12-16
Metroid Dread: Elun Logic

EthanArmbrust:
Initial work for Elun area


henriquegemignani (approved):
After you enter Elun, the path out locks behind you until _something_. It'd be nice to handle that.

PR #2529, opened 2022-01-06
Video comments for early Phendrana and Ice Ruins West logic update.

AlphaRage:
Added video comments for Phendrana Shorelines, Ice Ruins West and Ruined Courtyard. Changed and appended parts of the logic for Ice Ruins West.


Uncle-Reggie:
👍

PR #2530, opened 2022-01-06
Added more Tallon videos, and changed some logic in the database

Firemetroid:
Added some videos for Tallon, changed the frigate climb NSJ video with one that doesn't use boxes.
Added an SJ L-Jump Standable frigate climb to the logic.
Lowered the difficulty for frigate climb using Slope Space Jump and Standable, the Standable part is now (Beginner).
Added 2 entries to the changelog.


Uncle-Reggie:
👍

PR #2532, opened 2022-01-06
Prime - Item Position Rando

toasterparty:
I promise this is the last PR with the oops commit in the history.

Note: This PR establishes "Multi-Pickup Placement" as non-dangerous.


randovania/layout/base/base_configuration.py
```diff
@@ -60,7 +60,7 @@ def dangerous_settings(self) -> List[str]:
             if hasattr(f, "dangerous_settings"):
                 result.extend(f.dangerous_settings())
 
-        if self.multi_pickup_placement:
-            result.append("Multi-pickup placement enabled")
+        if self.shuffle_item_pos:
```
henriquegemignani, line 63: This should be on such a method in the prime configuration class instead of the base configuration.

We seem to have 0 tests for this, otherwise it'd have hit this.

henriquegemignani (changes requested):

randovania/layout/base/base_configuration.py
toasterparty: See latest change. Did I do inheritance good?

henriquegemignani (approved):

PR #2533, opened 2022-01-06
Early Phazon Mines videos and logic updates.

AlphaRage:
Added video comments and changed/added trick logic for Main Quarry, Elite Research, Ore Processing and Elite Control.


Uncle-Reggie:
👍

PR #2535, opened 2022-01-07
Added some videos to chozo tricks, changed some logic

Firemetroid:
So basically tower of light has 4 ways you can climb it, there's a combat dash method (which requires you to do 1 kinda hard combat dash and 2 more), I changed this method from beginner to advanced. The reason I changed it to advanced is that the first combat dash is kind of annoying and if you fail any of them you might need to do the first one again. There is an L-Jump method that still requires 2 combat dashes but it's much more forgiving since you can skip the first one. There is the slope method in which the slope itself is not hard but where you need to stand for it is. And there is the R-Jump method where you don't combat dash anything but still requires you to do optimal R-Jumps.


randovania/games/prime1/json_data/Chozo Ruins.txt
```diff
@@ -668,16 +672,32 @@ Extra - asset_id: 225636855
...
               Any of the following:
                   Space Jump Boots
                   Morph Ball Bomb and Morph Ball and Bomb Jump (Expert)
-          Space Jump Boots and Combat/Scan Dash (Beginner) and Standable Terrain (Intermediate)
+          All of the following:
+              Space Jump Boots
+              Any of the following:
+                  All of the following:
+                      # https://youtu.be/2jyvwogPSDI
+                      Combat/Scan Dash (Advanced) and Standable Terrain (Beginner)
```
Uncle-Reggie, line 691: Combat/Scan Dash (Advanced) -> Combat/Scan Dash (Intermediate)
Agree it is more difficult than Beginner, but is not as difficult as other Advanced category dashes.

Uncle-Reggie (changes requested):
691- Combat/Scan Dash (Advanced) -> Combat/Scan Dash (Intermediate)

randovania/games/prime1/json_data/Chozo Ruins.txt
Firemetroid: I think you are correct on this, I've made the change.

Uncle-Reggie (approved):
👍

PR #2569, opened 2022-01-12
CLI - Export Database Videos to html pages

toasterparty:
I think I have good ideas sometimes
[exported_db_videos.zip](https://github.com/randovania/randovania/files/7856091/exported_db_videos.zip)


henriquegemignani:
Now that I look at the exported stuff, it looks like a great thing to have in github pages.

henriquegemignani (approved):

henriquegemignani:
Oh, though this would be a lot better using actual randovania APIs to parse the JSONs, so this wouldn't break on db format changes.

duncathan (changes requested):
I'd suggest this be rewritten to use the randovania API, as you're recreating much of its functionality here. particularly, I'd make use of `RandovaniaGame` enum and its properties (such as long name), `default_database` in order to get the actual dbs rather than using the raw json, and then looping through the world, area, node, connection fields in there rather than working with it raw.

tools/export_db_videos.py
```diff
@@ -0,0 +1,241 @@
...
+        return
+
+    data = item["data"]
+    if data["comment"] is not None:
+        comment = data["comment"]
+        if "youtu" in comment:
+            video_id = comment.split("/")[-1].split("watch?v=")[-1].split(" ")[0].split("?t=")[0]
+            start_time = 0
+            if "?t=" in comment:
+                start_time = int(comment.split("?t=")[-1])
```
toasterparty, line 106: This will fail in cases where additional text follows a timestamped URL

toasterparty:
@duncathan @henriquegemignani 

```sh
python -m randovania database export-videos # Export all game dbs to /exported_videos/
python -m randovania database export-videos --game prime1 --output-dir foobar # Export Metroid Prime to /foobar/
```

randovania/cli/prime_database.py
```diff
@@ -570,7 +601,6 @@ def create_subparsers(sub_parsers):
         "--game",
         type=str,
         choices=[game.value for game in iterate_enum(RandovaniaGame)],
-        default=RandovaniaGame.METROID_PRIME_ECHOES.value,
```
toasterparty, line 573: @henriquegemignani Will this break anything? In hindsight, I don't think I needed to make this change.

henriquegemignani: I don't really care about the stability of the cli, change freely.

toasterparty:
[Magmoor Caverns.zip](https://github.com/randovania/randovania/files/7929561/Magmoor.Caverns.zip)

PR #2570, opened 2022-01-12
several cave story fixes

duncathan:


henriquegemignani:
No preset migration? Fine by me since it hasn't been in a release, but it's
annoying to anyone who already has a preset.

Lgtm otherwise

PR #2575, opened 2022-01-13
Guarantee unique prime patcher rng per MW player

toasterparty:
- Item anywhere should have separate
- No other randomprime RNG is used by randovania (except for file select FMV but that's being changed in next randomprime release)
- Maintains game export/import coherency


henriquegemignani (approved):

toasterparty:
@henriquegemignani Because solo seeds are index = 0, so seed + 0 is the same as before.

PR #2609, opened 2022-01-19
Code Cleanup

henriquegemignani:


randovania/cli/commands/export_db_videos.py
```diff
@@ -58,9 +58,11 @@
 </html>
 '''
 
+
```
toasterparty, line 61: @henriquegemignani Why is 2 empty lines the gold standard? I've been going out of my way to reduce to 1 when I'm making changes.

PR #2610, opened 2022-01-19
Fix minor empty dict field check in dread generator

aki-cat:
Testing generating Dread seeds and came across a KeyError exception in this line.


duncathan:
I have a proper fix for this, will open a PR momentarily

PR #2652, opened 2022-01-25
Update Artaria & Elun Logic. Add some boss logic.

kirbymastah:
Update Artaria and Elun room logic. Add fight requirements for Kraid, Corpius, Golzuna, and Escue.


ScorelessPine:
Speed booster is incorrectly listed as a way to damage/kill Corpius, when it does not deal damage to him.
Corpius can fire projectiles that can be shot at to give ammo refills, so minimum missile requirements can be more lenient.
After testing, Corpius can be damaged by Bombs and Cross Bombs, however the former is quite hard to hit his head with. Cross Bombs deal a fair bit of damage to him though, so it can be a reasonable method of dealing damage to him if missiles/charge are unavailable.

aki-cat:
adding to @ScorelessPine 's comment, Corpius is easily killable with vanilla starter items (no e-tanks, no missile upgrades, no charge beam). this is done in any% runs all the time and it's not particularly hard to pull off

PR #2671, opened 2022-01-29
Dread: Fix Artaria Dock to David Jaffe room logic

aki-cat:
Speed Booster+Morph Ball entrance is only possible when either Charge Beam door or Thermal door are open.

### DIFF
--- a/randovania/games/dread/json_data/Artaria.txt
+++ b/randovania/games/dread/json_data/Artaria.txt
@@ -1950,7 +1950,7 @@ Extra - asset_id: collision_camera_022
       All of the following:
           Morph Ball
           Any of the following:
-              Bomb and Infinite Bomb Jump (Intermediate)
+              Infinite Bomb Jump (Intermediate) and Lay Bomb
               All of the following:
                   Speed Booster
                   Charge Beam or After Artaria - Thermal Device
--- a/randovania/games/dread/json_data/Artaria.txt
+++ b/randovania/games/dread/json_data/Artaria.txt
@@ -1947,7 +1947,9 @@ Extra - asset_id: collision_camera_022
   > Door to Magma Flow Vista
       Trivial
   > Dock to David Jaffe Room
-      Morph Ball and Speed Booster
+      All of the following:
+          Morph Ball and Speed Booster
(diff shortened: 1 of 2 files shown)


aki-cat:
Hold on I just found you can do infinite bomb jump to get there too, let me add that real quick.

randovania/games/dread/json_data/Artaria.txt
```diff
@@ -1947,7 +1947,13 @@ Extra - asset_id: collision_camera_022
...
       Trivial
   > Dock to David Jaffe Room
-      Morph Ball and Speed Booster
+      All of the following:
+          Morph Ball
+          Any of the following:
+              Infinite Bomb Jump (Intermediate) and Lay Bomb
+              All of the following:
+                  Speed Booster
+                  Charge Beam or After Artaria - Thermal Device
```
henriquegemignani, line 1956: Not entirely sure why you need charge or thermal device to get to the tunnel on top

aki-cat: You can't speed boost your way up there without opening either the thermal door to the left or the charge beam door to the right.

The other door is a Phantom Cloak Door, and using Phantom Cloak cancels out your Speed Boost.

henriquegemignani: Ah. This is assuming these doors are charge/thermal doors. These go against how the database is organized, because it'd break if we change these door types to something else which isn't a feature right now but has always been planned (and can easily be implemented for dread/prime 1 rn).

We can add this, but it's worth a comment explaining things and a TODO to revisit in the future.

aki-cat: Is there some way to take those doors into consideration then? Is this such a corner case that we'd need to add an event for each of those doors?

henriquegemignani: Proper speed booster charging considerations are going to be a nightmare and require some extra work elsewhere I'd say.

aki-cat: Added a comment. Hopefully this is good enough.

PR #2673, opened 2022-01-30
Game Session simplify

henriquegemignani:


duncathan (commented):
don't see anything obviously wrong. which line changes the default tab to observers?

henriquegemignani:
> don't see anything obviously wrong. which line changes the default tab to observers?

``` python
self.splitDockWidget(self.game_dock, self.observers_dock, Qt.Horizontal)
```

It's not being default, but making it appear at the same time as the other.

duncathan (approved):

PR #2704, opened 2022-02-02
Fix Dread db to be strongly connected

henriquegemignani:
I've deleted a bunch of unused Start Point nodes, added a few missing connections and changed some missile/super doors to be matching on both sides.

There's a special case in the blast shield logic that if both sides of a door are the same shield, then it handles opening from the back. Pretty much a kludge until I can get the correct way working, though I think I'll just patch the game to be that way.

### DIFF
--- a/randovania/games/dread/json_data/Burenia.json
+++ b/randovania/games/dread/json_data/Burenia.json
@@ -142,7 +142,7 @@
             }
         },
         "Upper Burenia Hub": {
-            "default_node": "Start Point 2",
+            "default_node": "Start Point",
             "valid_starting_location": true,
             "extra": {
                 "total_boundings": {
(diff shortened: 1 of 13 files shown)


randovania/games/dread/json_data/Burenia.json
```diff
@@ -371,29 +371,6 @@
...
-                        "Door to Navigation Room North": {
-                            "type": "and",
-                            "data": {
-                                "comment": null,
-                                "items": []
-                            }
-                        }
-                    }
-                },
                 "Start Point 2": {
```
kirbymastah, line 374: Rename to "Start Point" since Start Point 1 is gone

kirbymastah:
In Teleport to Artaria (Red):
"Dock to Teleport to Ghavoran" to "Door to collision_camera_064" should require the db_dside_mg_003 event instead of blah_002b_000

PR #2717, opened 2022-02-04
Super Metroid DB Fixes and Reorganization

SolventMercury:
- Fixed some doors being missing in the DB
- Added rooms to represent rooms that were not present in the DB
- Made doors and connections accurate to the underlying game
- Fixed errors involving strongly connected components
- Just generally addresses #2650
- Fixed a broken link in the glossary

NOTE: The error involving the region around Pants Room is deliberate, and not a mistake. There are some actual one-way docks in this bit - this is not a mistake on my part, just how the game works.


randovania/games/super_metroid/json_data/Ceres Station.txt
```diff
@@ -6,25 +6,31 @@ Extra - asset_id: 253
...
   * Normal Door to Hallway A/Door to Elevator Shaft
+  * Extra - different_strongly_connected_component: True
   > Escape
       After Fight Ridley At Ceres
 
 > Escape; Heals? False
   * Teleporter to Crateria - Landing Site
   * Extra - scan_asset_id: None
   * Extra - teleporter_instance_id: None
+  * Extra - different_strongly_connected_component: True
```
henriquegemignani, line 17: You only need to add this flag into a single node of each component and its best to minimize the user to better catch errors.

PR #2750, opened 2022-02-08
WIP: Cross Game Models

EthanArmbrust:
This isn't complete yet, but I figured I'd ask for some feedback now before I start going to far in the wrong direction. 

What's done:

- [x] Serialize PickupModel class
- [x] Choose correct model during `patch_game()`
- [x] Cache models in internal_copies directory, and use them if they exist
- [x] Cache versioning

What still needs to be done:

- [x] Have GameInputDialog let the user provide an iso for models
- [ ] Allow the user to not use external models (possibly with a checkbox in customize cosmetic options) 
- [x] Don't use external models for single world seeds
- [ ] Allow the user to delete the cached model assets

### DIFF
--- a/randovania/game_description/resources/pickup_entry.py
+++ b/randovania/game_description/resources/pickup_entry.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
 from typing import Optional, Iterator, Tuple
 
+from randovania.bitpacking.json_dataclass import JsonDataclass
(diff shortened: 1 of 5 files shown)


toasterparty:
I highly recommend you integrate the converter version number into your cache system so you can delete the cache anytime the conversion code version doesn't match the cached version.

PR #2791, opened 2022-02-14
Added support for DELTARUNE

jonloveslegos:
- Added support for DELTARUNE (You are supposed to spell it with all caps)
- Added deltapatcher, a file patcher, to be included and used by the deltarune auto patcher.
- Can randomize items, key items, weapons, armor, starting abilities, and adds doors that require keys for progression.


henriquegemignani:
Before any kind of review can be started on this PR, you'll need to clean it up. Make sure the branch only contains the necessary files for the functionality.
Things like the `.vs` subfolder, a few test files in the root and etc should be gone. Preferably from history as well.
The `dist` folder should never have been commited, especially since it's on the gitignore.

PR #2808, opened 2022-02-16
Logic fixes and additions

MeriKatt:
Fixed dairon big hub item, cataris green emmi introduction logic, added tricks for artaria in thermal door tutorial for pseudo-wave of lower blob, for single wall jump to item / emmi door in Energy recharge station south, in burenia for crossbomb to fan item, for reaching the fan from the bottom by having lowered the water.

### DIFF
--- a/randovania/games/dread/json_data/Artaria.json
+++ b/randovania/games/dread/json_data/Artaria.json
@@ -5376,8 +5376,42 @@
                     "editable": true,
                     "connections": {
                         "Event - Post Chain Reaction Debris": {
-                            "type": "template",
-                            "data": "Shoot Diffusion or Wave"
+                            "type": "or",
+                            "data": {
+                                "comment": null,
+                                "items": [
+                                    {
+                                        "type": "template",
+                                        "data": "Shoot Diffusion or Wave"
+                                    },
+                                    {
+                                        "type": "resource",
+                                        "data": {
+                                            "type": "tricks",
+                                            "name": "Pseudo",
+                                            "amount": 2,
+                                            "negate": false
+                                        }
+                                    }
+                                ]
+                            }
+                        },
+                        "Event - Early SA Shortcut Blob": {
+                            "type": "or",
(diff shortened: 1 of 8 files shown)


randovania/games/dread/json_data/Artaria.json
```diff
@@ -5376,8 +5376,42 @@
                     "editable": true,
                     "connections": {
                         "Event - Post Chain Reaction Debris": {
-                            "type": "template",
-                            "data": "Shoot Diffusion or Wave"
```
kirbymastah, line 5380: We'll want another node with the same event - one for destroying from above/the teleport, and one for destroying from the water.

As it is, logic thinks if you destroy the blob from below, then the blob's only connection is to the teleport, so it thinks you can go from water -> blob -> teleport, without morph ball.

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -295,7 +298,9 @@ Extra - asset_id: collision_camera_002
   > Door to Underwater Horseshoe
       After Burenia - db_reg_aq_001 or After Burenia - db_reg_aq_002
   > Fan Platform - Below
-      Gravity Suit or Water Bomb Jump (Beginner)
+      Any of the following:
+          Gravity Suit or Water Bomb Jump (Beginner)
+          After Burenia - db_reg_aq_001 and After Burenia - db_reg_aq_002
```
kirbymastah, line 303: Remove these blob requirements. The "Raft - Top Water Level" is a node for the same altitude as the raft's highest point - i.e. neither blob is destroyed

The logic for destroying the blobs to get to the higher level are already accounted for with "Door to Underwater Horseshoe -> Raft - Top Water Level" and vice versa.

kirbymastah: EDIT: never mind, I see why you added these.

kirbymastah (approved):

PR #2815, opened 2022-02-17
Improved Database Video Export

toasterparty:
- Minimum trick difficulty above each video
- Sort videos by difficulty on each connection (beginner first)
- Fix for multiple videos on one comment
- Skip duplicate videos (same video start time)

### DIFF
--- a/randovania/cli/commands/export_db_videos.py
+++ b/randovania/cli/commands/export_db_videos.py
@@ -184,19 +184,22 @@ def generate_world_html(name, areas):
     '''
 
     for area in sorted(areas):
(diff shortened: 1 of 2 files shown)


henriquegemignani:
Think you can add a test that just calls `export_videos` and checks if the output is not empty or something?

It'd also be neater if the file would decode the game data to a GameDescription instead of depending on the json data...

toasterparty:
> Think you can add a test that just calls `export_videos` and checks if the output is not empty or something?

Added a test. I may refactor to use `GameDescription` at a later date.

PR #2830, opened 2022-02-20
Game-specific GameExporter, PatchDataFactory and GameExportDialog

henriquegemignani:


duncathan (commented):
This introduces a lot of duplicate code in the game export UI classes. Would it be appropriate to create some mixin classes to handle that code duplication?

PR #2842, opened 2022-02-22
Cross Game Models 

EthanArmbrust:
image

Continuation of #2750 on a new branch due to merge issues.  
This is mostly feature complete, save for maybe giving the user an option to delete the model caches manually


randovania/patching/prime/asset_conversion.py
```diff
@@ -127,52 +150,130 @@ def id_generator(asset_type):
...
+                resource=decoded_from_raw,
+            )
+            converted_assets[asset_id] = converted_asset
+
+        with open(cache_path.joinpath("meta.json"), "r")as data_additions_file:
+            randomizer_data_additions = json.load(data_additions_file)["data"]
+
+
+    else:
+        print("Creating assets....")
```
henriquegemignani, line 182: All prints should be logging calls.

randovania/patching/prime/asset_conversion.py
```diff
@@ -127,52 +150,130 @@ def id_generator(asset_type):
...
+
+        try:
+            asset_provider = prime_asset_provider()
+        except RuntimeError:
```
henriquegemignani, line 186: Now that we can properly prompt the user for the ISO, this should be a fatal error

randovania/games/prime1/exporter/game_exporter.py
```diff
@@ -67,7 +132,7 @@ def export_game(self, patch_data: dict, export_params: GameExportParams,
         try:
             py_randomprime.patch_iso_raw(
                 patch_as_str,
-                py_randomprime.ProgressNotifier(lambda percent, msg: progress_update(msg, percent)),
+                py_randomprime.ProgressNotifier(lambda percent, msg: updaters[2](msg, percent)),
```
henriquegemignani, line 141: It should only use `updaters[2]` when actually converting assets, otherwise this will skip the first 66% of the bar.

randovania/gui/game_session_window.py
```diff
@@ -1118,7 +1118,9 @@ async def save_iso(self):
         patch_data = await self._admin_player_action(membership, SessionAdminUserAction.CREATE_PATCHER_FILE,
                                                      options.options_for_game(game).cosmetic_patches.as_json)
 
-        dialog = game.gui.export_dialog(options, patch_data, self._game_session.game_details.word_hash, False)
+        all_presets = self._game_session.presets
```
henriquegemignani, line 1121: unused line?

randovania/patching/prime/asset_conversion.py
```diff
@@ -20,6 +21,28 @@
 from randovania.lib import status_update_lib
 from randovania.lib.status_update_lib import ProgressUpdateCallable
 
+PRIME_MODELS_VERSION = 1
+ECHOES_MODEL_VERSION = 1
```
henriquegemignani, line 25: MODELS

randovania/patching/prime/asset_conversion.py
```diff
@@ -252,17 +332,23 @@ def id_generator(asset_type):
...
+            try:
+                status_update(f"Converting {data['Name']} from Prime 2", i / len(assets_to_change))
+                result["{}_{}".format(RandovaniaGame.METROID_PRIME_ECHOES.value, data["Name"])] = Asset(
+                    ancs=converter.convert_id(data["AnimSet"], Game.ECHOES, missing_assets_as_invalid=False),
+                    cmdl=converter.convert_id(data["Model"], Game.ECHOES, missing_assets_as_invalid=False),
+                    character=data["Character"],
+                    scale=data["Scale"][0],
+                )
+            except (InvalidAssetId, UnknownAssetId) as e:
+                logging.error("Unable to convert {}: {}".format(data["Name"], e))
```
henriquegemignani, line 351: I think I'd go with fatal as well. This would mean a bad ISO.

henriquegemignani (changes requested):
This needs a lot of more tests. Something close to full coverage on the new part would be good, as there's a lot of moving parts here and we don't want to break it accidentally.

I think it'd be nicer to have the conversion from Echoes to Prime to be able to read the Echoes ISO directly. We'd need some new tooling in py-nod, but should be doable.

randovania/patching/prime/asset_conversion.py
EthanArmbrust: There's a few models that aren't in the vanilla game (translators, dark and light ammo expansions) that can cause this

henriquegemignani: I'd prefer that at first we skip converting these, unless we figure out a way to consistently do it.

EthanArmbrust:
Any chance decode_randomizer_data() could be moved out of the EchoesGameExporter file? I'm trying to pass Options through this class but the server is using this function.

henriquegemignani:
> Any chance decode_randomizer_data() could be moved out of the EchoesGameExporter file? I'm trying to pass Options through this class but the server is using this function.

Absolutely can be moved out, but what exactly do you need the `Options` object there?

EthanArmbrust:
> Absolutely can be moved out, but what exactly do you need the `Options` object there?
I'm trying to pass an options object through to the export game function to remove the with_default_data_dir() uses.

henriquegemignani:
> > Absolutely can be moved out, but what exactly do you need the `Options` object there?
> 
> I'm trying to pass an options object through to the export game function to remove the with_default_data_dir() uses.

I'd see if you can add whatever path is needed directly to `PrimeGameExportParams`, so the PrimeGameExportDialog uses Options to calculate whatever is needed.

EthanArmbrust:
For reference, here is the export game dialog window with the new checkbox
Use_correct_models

randovania/interface_common/persisted_options.py
```diff
@@ -81,6 +81,9 @@ def _convert_v15(options: dict) -> dict:
 
 def _convert_v16(options: dict) -> dict:
     per_game_options = options.pop("per_game_options", {})
+    for game_name in ["prime1", "prime2"]:
```
henriquegemignani, line 84: This function shouldn't be touched.

EthanArmbrust:
Some known issues that may want to be put in the changelog:
-Non-vanilla items are not converted. These include dark/light beam ammo expansions, translators, and Phazon suit
- Slight stutter in Echoes while loading a room with a Prime item in it

henriquegemignani (changes requested):

henriquegemignani:
> Some known issues that may want to be put in the changelog:
> 
>     * Non-vanilla items are not converted. These include dark/light beam ammo expansions, translators, and Phazon suit
> 
>     * Slight stutter in Echoes while loading a room with a Prime item in it

These should be readable inside Randovania's help somewhere.

PR #2855, opened 2022-02-23
fixes a couple weird issues with the resolver/map tracker

duncathan:


randovania/resolver/bootstrap.py
```diff
@@ -78,7 +78,7 @@ def calculate_starting_state(self, game: GameDescription, patches: GamePatches,
 
         starting_energy, energy_per_tank = self.energy_config(configuration)
 
-        if isinstance(starting_node, PlayerShipNode):
+        if starting_node.is_resource_node:
```
henriquegemignani, line 81: could use an `assert isinstance(starting_node, ResourceNode)` to prevent type-based warnings

henriquegemignani (approved):

PR #2856, opened 2022-02-24
Reworked logic for Ghavoran

XenoWars:
99% of this is the Ghavoran rework. The other 1% is a few more things in Artaria, and removing the speed booster block group in Gravity Suit Room.

Also, not quite sure why it's saying almost all of Artaria is changed. It should only be adding flash shift wall jumps to where there were already single-wall jumps


henriquegemignani:
I'm not really sure what's going on with your branch. Can you try rebasing? Though I tested it here and got a bunch of conflicts.

henriquegemignani:
Ah, it seems to be duplicated with your last PR. Yeah, lots of weirdness. It'd be best if you re-create the branch or something.

XenoWars:
Should i recreate? will that fix it?

XenoWars:
It doesn't let me resolve the conflicts, the option is grayed out. It should only be the speedbooster option in East Lava Middle Room in Artaria

henriquegemignani:
Please take your modified files and re-create/reapply things manually then re-open.

PR #2872, opened 2022-02-28
"Add manual action" in tracker with Area Picker dialog

dansalvato:
## Add Manual Action

- An "Add manual action" button is added next to "Undo last action". This allows the player to manually add an action that is not exposed in the Map Tracker due to it being "illogical" in the current settings—for example, if out-of-logic tricks are performed.
- The text instructions were moved to tooltips that appear when hovering over each button, which explains each of their functions.

## Area Picker

The Area Picker dialog widget is a general-purpose dialog that prompts the user to select a node or an area (via `get_node()` or `get_area()`), which gets returned as a `NodeIdentifier` or `AreaIdentifier`. Similarly to `QInputDialog`, there is no boilerplate to create/connect the dialog. It also accepts a `valid_areas` list which will show only those areas in the dialog.

### DIFF
--- a/randovania/gui/lib/area_picker.py
+++ b/randovania/gui/lib/area_picker.py
@@ -1,0 +1,81 @@
+from typing import Callable, Optional
+from randovania.game_description import default_database
+from randovania.game_description.world.area import Area
+from randovania.game_description.world.area_identifier import AreaIdentifier
+from randovania.game_description.world.world import World
+from randovania.games.game import RandovaniaGame
+from randovania.game_description.world.node import Node, NodeIdentifier
+from randovania.gui.generated.area_picker_dialog_ui import Ui_AreaPickerDialog
+from PySide6 import QtWidgets
+from PySide6.QtCore import Qt
+
+def get_node(parent: QtWidgets.QWidget, game: RandovaniaGame, filter: Optional[Callable[[Node], bool]]=None) -> Optional[NodeIdentifier]:
+    dialog = AreaPickerDialog(parent, game, filter)
+    result = dialog.exec()
+    if result == QtWidgets.QDialog.DialogCode.Accepted:
+        return NodeIdentifier(AreaIdentifier(dialog.current_world.name, dialog.current_area.name), dialog.current_node.name)
+    else:
+        return None
+
+class AreaPickerDialog(QtWidgets.QDialog, Ui_AreaPickerDialog):
+    _world_list: list[World]
+    _form: QtWidgets.QFormLayout
+    world_combo_box: QtWidgets.QComboBox
+    area_combo_box: QtWidgets.QComboBox
+    node_combo_box: QtWidgets.QComboBox
+    confirm_buttons: QtWidgets.QDialogButtonBox
+    current_world: World
+    current_area: Area
+    current_node: Node
+    filter: Optional[Callable[[Node], bool]]
+
+    def __init__(self, parent: QtWidgets.QWidget, game: RandovaniaGame, filter: Optional[Callable[[Node], bool]]=None):
+        super().__init__(parent)
+        self.setupUi(self)
+        self._world_list = default_database.game_description_for(game).world_list.worlds
+        for world in self._world_list:
+            item = QtWidgets.QListWidgetItem(world.name)
+            item.setData(Qt.UserRole, world)
+            self.worldList.addItem(item)
+        self.worldList.currentItemChanged.connect(self._update_areas)
+        self.areaList.currentItemChanged.connect(self._update_nodes)
+        self.nodeList.currentItemChanged.connect(self._node_selected)
+        self.confirmButtonBox.button(QtWidgets.QDialogButtonBox.Ok).setEnabled(False)
+
+    def _update_areas(self, world: QtWidgets.QListWidgetItem, previous: QtWidgets.QListWidgetItem):
+        self.areaList.clear()
+        self.nodeList.clear()
+        self.current_world = None if not world else world.data(Qt.UserRole)
+        if not world:
+            self.areaList.setDisabled(True)
+            self.nodeList.setDisabled(True)
+            return
+        area_list = world.data(Qt.UserRole).areas
+        area_list.sort(key=lambda area: area.name)
+        self.areaList.clear()
+        self.areaList.setDisabled(not area_list)
+        for area in area_list:
+            item = QtWidgets.QListWidgetItem(area.name)
+            item.setData(Qt.UserRole, area)
+            self.areaList.addItem(item)
+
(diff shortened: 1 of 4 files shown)


randovania/gui/lib/area_picker.py
```diff
@@ -0,0 +1,262 @@
...
+        world_list = default_database.game_description_for(game).world_list
+        self._pick_node = pick_node
+        self._model = AreaPickerModel(self, world_list, valid_areas)
+        self._setup_lists()
+        self.worldList.selectionModel().currentChanged.connect(self._update_areas)
+        self.searchLineEdit.textEdited.connect(self._update_search)
+        self.searchLineEdit.returnPressed.connect(self._confirm_selection)
+        self.confirmButtonBox.button(QtWidgets.QDialogButtonBox.Ok).setEnabled(False)
+        self.searchLineEdit.setFocus()
+        if window_title:
```
henriquegemignani, line 181: if `window_title` is supposed to be optional, the check should be `window_title is not None`. (and updating the type hint to be `Optional[str]`.

randovania/gui/lib/area_picker.py
```diff
@@ -0,0 +1,262 @@
...
+        """Checks if the object is any type in a list of types."""
+        return any ([isinstance(obj, t) for t in types])
+
+
+class AreaPickerDialog(QtWidgets.QDialog, Ui_AreaPickerDialog):
+    _model: AreaPickerModel
+    _world_list: WorldList
+    _empty_index: QtCore.QModelIndex
+    _pick_node: bool
+    worldList: QtWidgets.QListView
```
henriquegemignani, line 158: `world_list`

randovania/gui/lib/area_picker.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def __init__(self, header_text: str, parent: QtWidgets.QWidget, game: RandovaniaGame, valid_areas: list[AreaIdentifier] = None, pick_node: bool = True, window_title: str = None):
+        super().__init__(parent)
+        self.setupUi(self)
+        world_list = default_database.game_description_for(game).world_list
+        self._pick_node = pick_node
+        self._model = AreaPickerModel(self, world_list, valid_areas)
+        self._setup_lists()
+        self.worldList.selectionModel().currentChanged.connect(self._update_areas)
+        self.searchLineEdit.textEdited.connect(self._update_search)
```
henriquegemignani, line 177: I like to name the fields in the .ui files in snake case as well.

randovania/gui/tracker_window.py
```diff
@@ -17,6 +17,7 @@
 from matplotlib.backends.backend_qt5agg import FigureCanvasQTAgg as FigureCanvas
 from matplotlib.backends.backend_qt5agg import NavigationToolbar2QT as NavigationToolbar
 from matplotlib.figure import Figure
+from randovania import game_description
```
henriquegemignani, line 20: weird import

randovania/gui/tracker_window.py
```diff
@@ -35,9 +36,11 @@
 from randovania.generator import generator
 from randovania.gui.dialog.scroll_label_dialog import ScrollLabelDialog
 from randovania.gui.generated.tracker_window_ui import Ui_TrackerWindow
+from randovania.gui.lib import area_picker
 from randovania.gui.lib.common_qt_lib import set_default_window_icon
 from randovania.gui.lib.scroll_protected import ScrollProtectedSpinBox
 from randovania.layout.base.base_configuration import BaseConfiguration
+from randovania.layout.lib import location_list
```
henriquegemignani, line 43: weird import

randovania/gui/tracker_window.py
```diff
@@ -324,6 +328,13 @@ def _undo_last_action(self):
         self.actions_list.takeItem(len(self._actions))
         self._refresh_for_new_action()
 
+    def _add_manual_action(self):
```
henriquegemignani, line 331: if `get_node` becomes `async`, this would also need to be async. You can then use `@asyncSlot()` to solve things

randovania/gui/lib/area_picker.py
```diff
@@ -0,0 +1,262 @@
...
+        return (False, None, None, None)
+
+
+class AreaPickerModel(QtCore.QSortFilterProxyModel):
+    _world_list: WorldList
+    _show_locations: bool = True
+    _show_pickups: bool = True
+    _show_events: bool = True
+    _valid_areas: list[AreaIdentifier]
+    _LOCATION_NODES: list[type] = [DockNode, GenericNode, TeleporterNode]
```
henriquegemignani, line 54: Those could be module fields and not class members. And tuples to be immutable too.

henriquegemignani (changes requested):

henriquegemignani (changes requested):
The `area_picker` module itself definitely needs some unit testing. I'm fine with skipping one in the usage in TrackerWindow.

dansalvato:
Thanks for the extensive feedback, I'll start getting these changes in over the next few days.

randovania/gui/lib/area_picker.py
dansalvato: I chose camel case to indicate that I was referencing fields generated from a non-Python source, but if it's a project-wide convention to use snake case in .ui files then I'm happy to change it.

randovania/gui/lib/area_picker.py
```diff
@@ -0,0 +1,262 @@
...
+    :param game: the game whose areas will be used
+    :param valid_areas: limit the areas shown in the dialog to the areas in this list
+    :return: an AreaIdentifier of the selected area, or None if the dialog was canceled
+    """
+    accepted, world, area, _ = _get_item(header_text, parent, game, valid_areas, False, title_text if title_text else "Select Area")
+    return None if not accepted else AreaIdentifier(world.name, area.name)
+
+def _get_item(header_text: str, parent: QtWidgets.QWidget, game: RandovaniaGame, valid_areas: list[AreaIdentifier] = None, pick_node = True, window_title: str = None) -> tuple[bool, Optional[World], Optional[Area], Optional[Node]]:
+    dialog = AreaPickerDialog(header_text, parent, game, valid_areas, pick_node, window_title)
+    result = dialog.exec()
```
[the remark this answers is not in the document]
dansalvato: Maybe there could be `async` variants of the functions, eg. `get_area_async()`? I was hoping that this dialog would be as easy to use as `QInputDialog` where the result can be easily retrieved inline without having to connect to an async function. Do you have any thoughts on this?

duncathan:
I'm closing this but I *will* keep bugging you about it in twitch chat

ThanatosGit:
> I'm closing this but I _will_ keep bugging you about it in twitch chat

I already prepared the three year celebration :(

PR #2875, opened 2022-02-28
Fixed Golzuna Arena logic

XenoWars:
- Added "Dock to Golzuna Arena" in "Above Golzuna" for the SSC
- Added "Dock from Above Golzuna" in "Golzuna Arena" for the SSC
- Added connections to "Door to Cross Bomb Tutorial" and "Before Golzuna" from "Dock from Above Golzuna"
- Removed "Tile (WEIGHT)" connection to "Door to Cross bomb Tutorial"


dyceron:
PR has old commits

PR #2893, opened 2022-03-02
Refactoring Nodes

henriquegemignani:
Some refactoring for how Nodes are defined and a few of the methods. No functional changes.

- Each node class has their own file.
- NodeContext now has access to `NodeProvider`, an abstraction on top of `WorldList`.
- ResourceNode.resource now has access to NodeContext.
- NodeIdentifier is now a valid ResourceInfo.
- Hints are now assigned to a NodeIdentifier, instead of to a LogbookAsset.
- LogbookAsset has been removed.


randovania/game_description/resources/resource_info.py
```diff
@@ -28,7 +29,7 @@ def add_resource_gain_to_current_resources(resource_gain: ResourceGain,
 
 
 def add_resources_into_another(target: CurrentResources, source: CurrentResources) -> None:
-    resource_gain: ResourceGain = source.items()
+    resource_gain = typing.cast(ResourceGain, source.items())
```
duncathan, line 32: what's the difference here?

randovania/layout/description_migration.py
```diff
@@ -192,6 +192,69 @@ def _migrate_v9(json_dict: dict) -> dict:
...
+            '9': "Sand Zone/Jenka's House/Hint - Jenka 1"
+        }
+    }
+
+    for game in json_dict["game_modifications"]:
+        game["hints"] = {
+            asset_id_conversion[game["game"]][asset_id]: hint
+            for asset_id, hint in game["hints"].items()
+        }
+        pass
```
duncathan, line 253: ?

duncathan (commented):
Don't see any obvious errors. While you're in here, though, it might be a good idea to rename `LogbookNode` to the more generic `HintNode` since 2/3 of the games that use hints do not have a logbook lol

randovania/game_description/resources/resource_info.py
henriquegemignani: The first I'm declaring the variable is of type `ResourceGain`, but I can still get a warning if the argument isn't.
The new one I'm explicitly casting the variable to the new type. Kinda like asserting, but it could be wrong.

From what I understand, `source.items()` really should match the `ResourceGain` type, but at least PyCharm gets lost somewhere in the types and gives me a warning.

randovania/layout/description_migration.py
henriquegemignani: Guess how the for loop began :)

henriquegemignani:
> Don't see any obvious errors. While you're in here, though, it might be a good idea to rename `LogbookNode` to the more generic `HintNode` since 2/3 of the games that use hints do not have a logbook lol

That makes sense to touch when also cleaning up a bunch of other weirdness with it, like the `scan_visor` and `required_translator` fields, moving `string_asset_id` to the extra. It's just some work I don't really want to do right now :X

duncathan:
> > Don't see any obvious errors. While you're in here, though, it might be a good idea to rename `LogbookNode` to the more generic `HintNode` since 2/3 of the games that use hints do not have a logbook lol
> 
> That makes sense to touch when also cleaning up a bunch of other weirdness with it, like the `scan_visor` and `required_translator` fields, moving `string_asset_id` to the extra. It's just some work I don't really want to do right now :X

yeah fair

duncathan (approved):

PR #2897, opened 2022-03-03
Ensure Hanubia is reachable only by Golzuna or SSC

XenoWars:
- Added "Dock to Golzuna Arena" in "Above Golzuna" for the SSC
- Added "Dock from Above Golzuna" in "Golzuna Arena" for the SSC
- Removed "Tile (WEIGHT)" connection to "Door to Cross Bomb Tutorial"

### DIFF
--- a/randovania/games/dread/json_data/Ghavoran.txt
+++ b/randovania/games/dread/json_data/Ghavoran.txt
@@ -2065,6 +2065,9 @@ Extra - asset_id: collision_camera_026
 
 > Dock from Above Golzuna; Heals? False
   * Blocked Passage to Above Golzuna/Dock to Golzuna Arena
+  * This dock is only used with the SSC.
+
+
   > Before Golzuna
       After Elun - Release X Parasites
   > Start Point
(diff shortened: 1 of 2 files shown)


randovania/games/dread/json_data/Ghavoran.txt
```diff
@@ -2052,15 +2056,20 @@ Extra - asset_id: collision_camera_026
...
-      Before Elun - Release X Parasites
   > Before Golzuna
       After Elun - Release X Parasites
 
 > Start Point; Heals? False; Spawn Point
   > Door to Cross Bomb Tutorial
       Trivial
 
+> Dock from Above Golzuna; Heals? False
+  * Blocked Passage to Above Golzuna/Dock to Golzuna Arena
```
henriquegemignani, line 2067: Can you add a description to this node and the other dock, so it's clear it's for an SSC?

henriquegemignani (changes requested):

randovania/games/dread/json_data/Ghavoran.txt
XenoWars: Added some descriptions

PR #2900, opened 2022-03-04
Move Randomprime Cache Dir to Appdata

toasterparty:


randovania/games/prime1/exporter/patch_data_factory.py
```diff
@@ -326,6 +326,10 @@ def create_data(self) -> dict:
 
         seed = self.description.get_seed_for_player(self.players_config.player_index)
 
+        from randovania.interface_common.options import Options
+        from os import path
+        cache_dir = str(path.join(Options.with_default_data_dir().data_dir, "cache"))
```
henriquegemignani, line 331: This is a system dependent path, so it should be done in the game exporter and not in the patch data factory since the later runs in the server.

Additionally, the cache path should be provided via the GameExportParams, so the game export dialog can use the existing Options object and not recreate one ad-hoc.

The cache for likely needs a "prime1" in the name somewhere.

When using Path objects, they have a joinpath argument to avoid using os.join

henriquegemignani (changes requested):

randovania/games/prime1/exporter/patch_data_factory.py
toasterparty: I don't know how to do your first two points

henriquegemignani: 1. Check `PrimeGameExporter`. You should set the `cacheDir` field in the `new_config` variable there directly.
2. In the same file as 1, there's a `PrimeGameExportParams`. Add a new field there like `cache_path` or something. Then in `PrimeGameExportDialog.get_game_export_params`, you'll define the value of cache_path and there you also have an `self._options` so you don't need to create a new one.

EthanArmbrust: Take a look at my cross game models PR, I have to do this for the external assets

toasterparty:
looks good, I'll get those error messages in a bit

henriquegemignani (approved):

PR #2966, opened 2022-03-16
Implement correct tracking of door locks

### LINKED ISSUE #2933: Properly handle unlocking dock locks
Opening a dock node's lock (aka destroying a blast shield) should be a resource collection, with different criteria based on the lock type.

henriquegemignani:
Fixes #2933.


randovania/game_description/world/dock_lock_node.py
```diff
@@ -0,0 +1,87 @@
...
+            if not context.has_resource(self.dock_identifier):
+                return False
+
+        dock_target = dock.get_target_identifier(context)
+        if front_weak.can_unlock_from_back(dock.get_back_weakness(context)):
+            if not context.has_resource(dock_target):
+                return False
+
+        return True
+
```
duncathan, line 66: couldn't this whole function be `return not self.can_collect(context)`?

randovania/game_description/world/dock_node.py
```diff
@@ -45,50 +55,100 @@ def __hash__(self):
...
+        target_identifier = self.get_target_identifier(context)
+        if target_identifier is None:
+            # Explicitly connected to nothing.
+            return
 
-        yield target_node, RequirementAnd(reqs).simplify() if len(reqs) != 1 else reqs[0]
+        yield self._open_dock_connection(context, target_identifier)
+        lock_connection = self._lock_connection(context)
+        if lock_connection is not None:
+            yield lock_connection
```
duncathan, line 154: I think this could be
```py
if (lock_connection := self._lock_connection(context)) is not None:
    yield lock_connection
```

duncathan (commented):
there's also a lot of duplicate/very similar code in a few places but it looks like it's been done for performance reasons for the most part so might be better as is

randovania/game_description/world/dock_lock_node.py
henriquegemignani: I'll have to draw a table of results for each state and compare, because I'm not sure.

PR #2972, opened 2022-03-17
Use echoes missile expansion for unlimited missiles

EthanArmbrust:
When converting assets to Prime 1, use the echoes missile expansion model instead of the launcher. The echoes launcher model causes the Prime to crash.


toasterparty (approved):
I tested this. Looks good.

PR #3001, opened 2022-03-29
Prime1 - Fix Starting Memo

toasterparty:


toasterparty:
@henriquegemignani don't merge until the multi is over. I already interrupted it once

PR #3029, opened 2022-04-02
Add AppImage Build Support (Closes #2927)

### LINKED ISSUE #2927: Improve Packaging for Linux
The current methods for running RDV on Linux involve some quite intimate meddling with system packages and the occasional hacky workaround. In no particular order:

* For multiworld, the system-wide Python has to have a particular `CAP_SYS_PTRACE` to allow it to access Dolphin's memory, but only if Dolphin is launched outside of RDV. If it's a child process of RDV, then multiworld Just Works.
* `.config` overrides need to be shipped with bundled Mono apps to allow them to access `liblzo2.so.2`
* Different distros ship different versions of Python with their own idiosyncrasies

**EDIT**: I had previously attempted to package this myself, but worked in the wrong direction (Flatpak is the way to go) and lost motivation. [See here; all of my work should be pretty self-documented and I'm always open to answer any questions](https://github.com/randovania/randovania/issues/2927#issuecomment-1385550420).

My previous work had solved basically all of the problems mentioned above via `mkbundle`. The outstanding work now would be to:

* Port that work to Flatpak and build whatever ancillary build scripts need to be built
* Set up CI to build and push Flatpaks
* Submit PRs to flathub and flathub-dev

rehashedsalt:
These commits add a series of tools to the main branch to enable building Randovania into Linux x86_64 AppImages. A quickstart build guide has been added to the root README.md, and a more in-depth one can be found at tools/appimage/README.md. The short of it is this:

* Clone the repo
* `docker-compose -f docker-compose-appimage-builder.yml up --build`
* Upload both the `.AppImage` and `.AppImage.zsync` files found in `out` to the GitHub release

The dockerfile `Dockerfile.appimage-builder` has been included and is used by the aforementioned docker-compose to build an image. The distro chosen is intimately tied to various build scripts because of our good friend Mono, so while a different image can be used, it would require a bit of effort to change and would warrant testing.

---

This changeset comes with a number of advantages over distributing and executing Randovania by source or by Python package:

* Enormously simplified installation -- download and run a file and you're done
* Both of the .NET executables -- Randomizer.exe and EchoesMenu.exe -- have been bundled with Mono and statically linked, allowing them to run on practically any system and kicking our dependency on a system-wide Mono
* Bundled liblzo2, avoiding another system-wide dependency
* Automatic updates; supported desktop environments (Plasma at least) will allow the user to update Randovania by simply right-clicking the AppImage and choosing "Update". The AppImage uses the GitHub API to access the `zsync` files created by the build scripts; if the application differs from latest, it fetches the latest version. This PR only includes support for updating to the latest stable release of Randovania, but adding a nightly channel would be easy to implement due to how well the toolchain integrates updates and how easy it is to just serve a couple files over HTTP.
* Unsandboxed, and so should Just Work when migrating from a different form of install
* In general, we get a known-good environment with a known-good Python version

And, of course, a few caveats and issues:

* If a user does not have Mono installed but does have configuration files in `/etc/mono`, the .NET executables will crash as they will read those configs and try to load libraries that don't exist. Solution is to move or remove `/etc/mono`. If a properly-configured system-wide Mono *is* installed, the bundles will use that instead and run without issue. As far as I am aware, there is no way to have Mono behave differently. Thanks Mono.
* Users on versions of glibc much older than that of the system used to build the AppImage will be unable to patch games; I've chosen a 2-year-old LTS distribution as the image for this reason
* Since the Python interpreter used for Randovania is sealed in a squashfs archive and mounted read-only, there is no way to give it `CAP_SYS_PTRACE`. The only way for Randovania to observe Dolphin's memory is for Dolphin to be a subprocess of Randovania. A function will have to be added to Randovania -- if one does not already exist -- to launch Dolphin as a subprocess.
* The bundle is like 375M (1.1G uncompressed!) because PySide6 is friggin enormous. This is a problem any Linux distribution of Randovania will have, but I'm not familiar enough with the codebase to try to prune the bloat. I did pitch QtWebEngine though, which is 100M uncompressed on its own.
* A handful of build scripts had to be copy-pasted so they execute outside of the context of other actions like sourcing a venv and launching the application
* We'll probably want to ship a copy of the PSF license with the application since we're now bundling Python
* I had to duplicate the Randovania icon; if it ever needs changed, it needs changed in two places. Imagemagick doesn't like ICNS files.

---

My internet access is currently less than good, so I don't have the bandwidth to download and test several different distributions right now. I can give it the stamp of approval for Ubuntu 21.10 and nothing else at the moment. I'm 95% sure that everything Python will work just fine, it's Mono that I'm worried about.

If anybody has the time and download speed to perform some more thorough testing, it'd be appreciated, otherwise I'll circle back and try a few larger-named distros later.


tools/appimage/overlay/usr/share/metainfo/org.randovania.randovania.appdata.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
henriquegemignani, line 1: I'd prefer if this file would be generated based on the rest. Particularly since it duplicates part of the README and is in a path that will be absolutely hidden.

henriquegemignani (changes requested):
Thank you for looking into this!

Is this step of running a docker compose file the standard of creating an app image?

My impression overall is that we could be adding to the app image the executable created with PyInstaller: it already features a logic for following imports/etc to figure out the minimum needed, so it helps minimize the number of different ways to run randovania.
Also, `create_release.py` does some extra preparing for release in the form of creating the config json and downloading nintendont.

Lastly, I'm less concerned with how easy it is to prepare the final executable in many different distros, but a lot more that these images should be built and provided via CI.

rehashedsalt:
>Is this step of running a docker compose file the standard of creating an app image?

So, there isn't too much of a "standard" for creating AppImages besides needing to execute AppImageTool or some variant thereof to build the final product. I bundled everything up into a Docker image for a few reasons:

* I have no experience with GH Actions and assumed containerization was a first-class citizen, which would make building easier for you
* It makes reproducibility easier
* I needed access to another distro's Mono quickly and easily from my workstation, and this was the leanest and quickest way

We're more or less just constructing a filesystem, then AppImageTool zips it up and sticks an ELF on top; the toolchain can be adapted to whatever platform we need.

>My impression overall is that we could be adding to the app image the executable created with PyInstaller: it already features a logic for following imports/etc to figure out the minimum needed, so it helps minimize the number of different ways to run randovania.

That should be easily doable. I'll see about calling out to create_release.py and installing that package into the AppImage.

>Lastly, I'm less concerned with how easy it is to prepare the final executable in many different distros, but a lot more that these images should be built and provided via CI.

Agreed. AppImage just happens to neatly solve some of the problems we'd encounter when packaging for several distros, like hosting repositories and providing a version of Python that RDV likes (Buster still ships 3.7, for example).

tools/appimage/overlay/usr/share/metainfo/org.randovania.randovania.appdata.xml
rehashedsalt: I'm onboard. Should I parse out a section of the README or include a chunk of data in the build script?

There's quite a bit of boilerplate here; realistically we only need to worry about generating the summary and description.

henriquegemignani:
> Agreed. AppImage just happens to neatly solve some of the problems we'd encounter when packaging for several distros, like hosting repositories and providing a version of Python that RDV likes (Buster still ships 3.7, for example).

There's a difference between being able to create the published artifact/running from source/development environment, and just being an end user of the published artifact.

I have no concerns of raising minimum requirements of the former. The later is a bit more important, but as long as it runs on the latest version of the major distros, it's ok for me. Debian Buster, for example, is out because there's a new stable.
But AppImage should run mostly everywhere so it's not a big deal anyway.

tools/appimage/overlay/usr/share/metainfo/org.randovania.randovania.appdata.xml
duncathan: conveniently, the README is already formatted in such a way to facilitate pulling individual sections out of it. there are comments marking various sections: the description tag in this file is in the WELCOME section. worth noting that you'd need to convert the markdown to XML, ofc

tools/appimage/build.sh
```diff
@@ -0,0 +1,207 @@
...
+	popd
+}
+build-install-randovania() {
+	# Enter the Randovania directory inside squashfs-root and install its
+	# dependencies. We avoid activating the venv here because the Python
+	# interpreter in the AppImage we extracted already has a pseudo-venv going on
+	# with respect to the AppImage root.
+	pushd squashfs-root/"$randovania_location"
+	# Install dependencies
+	python -m pip install --upgrade -r requirements-setuptools.txt
```
henriquegemignani, line 146: This entire function would be just using the executable from `create_release.py`.

henriquegemignani (changes requested):

Dockerfile.appimage-builder
```diff
@@ -0,0 +1,15 @@
...
+		liblzo2-2 \
+		mono-complete \
+		rsync \
+		wget \
+		zsync \
+		&& \
+	# The homedir for this user is required for nuget
+	groupadd builder --gid 1000 && \
+	useradd builder --uid 1000 -g builder -d /home/builder -m
+CMD /bin/bash -c "cd /opt && ./tools/appimage/build.sh"
```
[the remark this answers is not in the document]
rehashedsalt: Yep, dockerfile fail on my part. Resolved in dd1593bc

tools/appimage/build.sh
```diff
@@ -0,0 +1,207 @@
...
+
+set -e
+
+#
+# Configuration vars
+# These variables can be freely modified
+#
+
+# The URL to the Python AppImage we'll bundle RDV in
+python_appimage="https://github.com/niess/python-appimage/releases/download/python3.9/python3.9.10-cp39-cp39-manylinux1_x86_64.AppImage"
```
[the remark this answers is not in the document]
rehashedsalt: Just pushed a few commits that fix this. We also remove our dependency on that Python appimage since we're shipping a bundled python with the release image

tools/appimage/overlay/AppRun
```diff
@@ -0,0 +1,34 @@
...
+    executable=$(which "${ARGV0}")
+fi
+
+# This is a reimplementation of tools/start_client.sh from the base repository
+# The UI build was performed during AppImage build, so we skip that step
+(
+	set -e -x
+	export PATH="${APPDIR}/usr/bin:${PATH}"
+	cd "$APPDIR/opt/randovania"
+	"$APPDIR/usr/bin/python3" -m randovania gui --preview main
```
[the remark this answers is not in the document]
rehashedsalt: They absolutely can. Since we're using the binary release from create_release.py, this line has been changed to just pass the rest of the cmdline args along. It's more or less just `/path/to/randovania "$@"` now.

rehashedsalt:
After refactoring to leverage create_release.py instead of running Randovania from source, I'm realizing that calling out to that script is duplicated work. We already build the application once; can we configure CI to just pass a built release to this script to repack it?

We'll still need sources available during this stage of the build so we can read the README and pack it into AppImage metainfo. Can we pass this container the repository in a state where `./dist/randovania/` exists and is populated with the built Linux release?

tools/appimage/overlay/usr/share/metainfo/org.randovania.randovania.appdata.xml
rehashedsalt: Added in 0d4af72a

randovania/games/prime2/patcher/claris_randomizer.py
```diff
@@ -36,7 +36,11 @@ def _get_randomizer_folder() -> Path:
 
 
 def _get_randomizer_path() -> Path:
-    return _get_randomizer_folder().joinpath("Randomizer.exe")
+    nativepath = _get_randomizer_folder().joinpath("randomizer")
```
henriquegemignani, line 39: This different path is weird, we'd need to have at least a description of what it is.

randovania/games/prime2/patcher/claris_randomizer.py
```diff
@@ -36,7 +36,11 @@ def _get_randomizer_folder() -> Path:
...
+    if nativepath.is_file():
```
henriquegemignani, line 40: I prefer if we'd be able to check for the situation where the different path is needed, and then use the different path.
So if when running from app image uses this path, the if should check for that.

The reasoning means we don't hide errors. If the file isn't where we expect, I want the error message to yell the correct file path that is wrong.

henriquegemignani (changes requested):

duncathan:
@rehashedsalt @henriquegemignani what's the status on this?

rehashedsalt:
Outstanding changes right now are:

* [Address these undocumented file paths](https://github.com/randovania/randovania/pull/3029/files/0d4af72a1f8bbc46eb51fd6d6038fbbe5406df2f) in `claris_randomizer.py`. The reason why they're there is because that's what I name the mono-bundled binaries. A rename to something more immediately-apparent and some light logging is probably in order, maybe something like "randomizer-monobundle" and similar.
* Incorporate the build script into `create_release.py`. At this point, the actual bash-written build scripts are all small enough that most of them could be reasonably rewritten in Python straight-out to simplify things.

The biggest blocker right now is that I work an IT job as my nine-to-five and am suffering from burnout, so I have no motivation to work on any of this at all right now. You know how it goes.

rehashedsalt:
It's become clear that not only do I not have the motivation to continue this work, but that (with the advent of the Steam Deck) AppImage is the wrong package format to target. I'll be closing this PR accordingly.

The work done in my branch is still valuable to any future attempts at packaging for Flatpak, though, and I've documented accordingly in #2927. I'd be more than happy to work with someone on this, but it's not work I can follow through with myself as of now.

PR #3031, opened 2022-04-03
Fix Ghavoran Golzuna tower logic

ArcanoxDragon:
A friend and I got softlocked on two different seeds because the logic expected us to be able to ballspark out of the Morph Ball maze under Pulse Radar out into the Golzuna Tower before being able to destroy the Enky in that hallway. My friend was able to do some insane cross bomb + bomb + jump combos using a turbo controller, but I didn't have regular bombs, and couldn't get out.

The connection from the missile tank in the diagonal tunnel out into the Golzuna Tower should only be marked as accessible via Ballspark if that Enky has been destroyed already (unless there's some INSANE speed strat we're both entirely unaware of)

### DIFF
--- a/randovania/games/dread/json_data/Ghavoran.json
+++ b/randovania/games/dread/json_data/Ghavoran.json
@@ -7544,8 +7544,25 @@
                     "major_location": false,
                     "connections": {
                         "Door to Save Station East": {
-                            "type": "template",
-                            "data": "Ballspark"
+                            "type": "and",
+                            "data": {
+                                "comment": null,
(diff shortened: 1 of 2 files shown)


henriquegemignani:
@XenoWars is this PR is redundant now that #2984 is merged?

ArcanoxDragon:
> @XenoWars is this PR is redundant now that #2984 is merged?

I looked at that PR and it does make this one obsolete. It includes the logic changes from this one and then some; I'll close this.

PR #3074, opened 2022-04-18
Pulse Radar Room logic does not work as a starting location when it contains required progression to exit

ArcanoxDragon:
The Pulse Radar room had a connection configured from "Start Point" to the item pickup. This led to a seed where my starting location was the Pulse Radar room, and the item that was acquired from that room was Morph Ball. I was unable to leave the room, because spawning in the room doesn't trigger the item acquisition.

Logic has been adjusted so that the only connection to the Pickup is from the Pitfall Blocks tile group.

### DIFF
--- a/randovania/games/dread/json_data/Ghavoran.json
+++ b/randovania/games/dread/json_data/Ghavoran.json
@@ -12231,13 +12231,6 @@
                         "start_point_actor_def": "actordef:actors/logic/startpoint/charclasses/startpoint.bmsad"
                     },
                     "connections": {
-                        "Pickup (Pulse Radar)": {
-                            "type": "and",
-                            "data": {
-                                "comment": null,
-                                "items": []
-                            }
-                        },
                         "Tile Group (POWERBEAM) 1": {
                             "type": "and",
                             "data": {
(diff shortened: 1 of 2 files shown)


duncathan:
I'm not sure the issue this is fixing is actually real. I don't know for certain for Pulse Radar, but I know that in the Flash Shift room if you start there the item is still present, but invisible, and can be obtained by jumping up and touching it in the middle of the statue. That being said, I'd prefer we force the cutscene when starting in these rooms than to remove the connection anyway.

ArcanoxDragon:
> I'm not sure the issue this is fixing is actually real. I don't know for certain for Pulse Radar, but I know that in the Flash Shift room if you start there the item is still present, but invisible, and can be obtained by jumping up and touching it in the middle of the statue. That being said, I'd prefer we force the cutscene when starting in these rooms than to remove the connection anyway.

I thought about that when I spawned, but I tried jumping around every square inch of the room and never found the trigger. It's possible it's at the top-left corner of the room, but I wouldn't have been able to reach it without flash shift or spin boost or something.

I'm not sure if that item is handled via Lua or native code...I'll play around later to see if it's even feasible to trigger that cutscene manually.

XenoWars:
Just tested and can confirm that this is indeed an issue. I spawned with spin boost, jumped all around the room. no trigger for the item or cutscene

duncathan:
this issue has been fixed properly for the next release of open-dread-rando, so this PR is not needed. thanks regardless!

PR #3079, opened 2022-04-20
Add Preset fields for controlling when a logical hint should be placed.

henriquegemignani:


randovania/layout/preset_migration.py
```diff
@@ -584,7 +584,10 @@ def _migrate_v25(preset: dict) -> dict:
 
     return preset
 
+
 def _migrate_v26(preset: dict) -> dict:
+    preset["configuration"]["minimum_available_locations_for_hint_placement"] = 0
+    preset["configuration"]["minimum_location_weight_for_hint_placement"] = 0.0
```
duncathan, line 590: this is bad form and you know it :P

randovania/generator/filler/player_state.py
```diff
@@ -219,6 +219,26 @@ def filter_usable_locations(self, locations_weighted: WeightedLocations) -> Weig
 
         return locations_weighted
 
+    def should_have_hint(self, pickup: PickupEntry, current_uncollected: UncollectedState,
+                         all_locations_weighted: WeightedLocations) -> bool:
+
+        if not pickup.item_category.is_major:
```
duncathan, line 225: shouldn't this be checking the interesting pickup thing?

duncathan (changes requested):

randovania/layout/preset_migration.py
henriquegemignani: The reusing a preset version? I don't care, that was just merged lol

The bunch of similar fields? I could add a class but also bleh...

randovania/generator/filler/player_state.py
henriquegemignani: If it was picked at all it's an interesting pickup.

randovania/layout/preset_migration.py
duncathan: my presets are ruined ;~;

PR #3090, opened 2022-04-23
Properly implement David Jaffe charge beam door.

henriquegemignani:


XenoWars (commented):
The database looks fine, and based on what you posted in Discord about the dock locks and looking at it a bit myself, it looks good to go. There will need to be uses of "front-blast-back-impossible" in some cases though

PR #3148, opened 2022-05-06
Phazon Suit Hint Fix

toasterparty:


henriquegemignani:
What happens when there's multiple? I feel it also throws an exception

toasterparty:
> What happens when there's multiple? I feel it also throws an exception

It just hints the first one it finds.

henriquegemignani (approved):

PR #3211, opened 2022-05-22
Dread: Remove Speed Blocks in Burenia

XenoWars:
- Removes all SPEEDBOOSTER tile groups in Burenia and reapplies logic
- Adds event for the Early Grav Epart to account for door rando when breaking the blocks just outside


Isachu:
Looks good 👍

PR #3212, opened 2022-05-22
Dread: Remove Speed Blocks in Artaria

XenoWars:
- Removes all SPEEDBOOSTER tile groups in Artaria and reapplies the logic


Isachu (approved):
Looks good

PR #3213, opened 2022-05-22
Dread: Remove Speed Blocks in Cataris

XenoWars:
- Removes all SPEEDBOOSTER tile blocks in Cataris and reapplies the logic


Isachu (approved):
Looks good

PR #3215, opened 2022-05-22
Dread: Remove Speed Blocks in Ferenia

XenoWars:
- Removes all SPEEDBOOSTER tile groups in Ferenia and reapplies the logic.


Isachu (approved):
Looks good

PR #3229, opened 2022-05-24
Expansions for multi pickup

henriquegemignani:
I'm not happy with how locked resources work, but it works for now. If I don't make that change for Prime 1 then generator will place expansions first compared to mains and that's just boring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [4.4.0] - 2022-06-01
 
+- Added: When using multi-pickup placement, expansions are now considered for logic.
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -7,10 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [4.4.0] - 2022-06-01
 
+- Added: When using multi-pickup placement, expansions are now considered for logic.
```
duncathan, line 10: this could honestly be a major change

duncathan (approved):

CHANGELOG.md
henriquegemignani: Remind me of marking relevant changes as major before release.

PR #3245, opened 2022-06-01
Update Randomprime and Prime Database

toasterparty:


randovania/games/prime1/json_data/a.py
```diff
@@ -0,0 +1,299 @@
+import os
```
henriquegemignani, line 1: Woops

henriquegemignani (changes requested):

PR #3249, opened 2022-06-03
[WIP] Added hints to Prime 1

ThomasJRyan:
The goal of this PR is to add a hint system to Prime 1. Overriding the STRGs of the Chozo Lore and Space Pirate Logs, it will ideally give hints to the player of where their and other's items are located.

It currently needs to be tweaked. The way it's implemented has the hints being way too direct and doesn't leave much room for mystery. It also uses up far too many slots at once (a total of 24 at the moment, one for each major collectible). It doesn't hint towards expansions which it absolutely should to cut down on the major spoilers.

For what it is, it's completely functional, if not basically cheating
image

There currently isn't a GUI toggle for it. Need to implement that 

TODO:
- [ ] Add GUI
- [ ] Add minor items
- [ ] Add joke hints
- [ ] Possibly limit hints to one per room
- [ ] Enable/Disable Space Pirate Frigate hint
- [ ] Improve code
- [ ] Shuffle hints thoroughly before sending to patcher (Or shuffle them in the patcher)

Randomprime PR that accompanies it: https://github.com/toasterparty/randomprime/pull/46

### DIFF
--- a/randovania/games/prime1/exporter/patch_data_factory.py
+++ b/randovania/games/prime1/exporter/patch_data_factory.py
@@ -17,7 +17,7 @@ from randovania.game_description.world.pickup_node import PickupNode
 from randovania.game_description.world.world_list import WorldList
 from randovania.games.game import RandovaniaGame
 from randovania.games.prime1.exporter.hint_namer import PrimeHintNamer
-from randovania.games.prime1.layout.hint_configuration import ArtifactHintMode, PhazonSuitHintMode
+from randovania.games.prime1.layout.hint_configuration import ArtifactHintMode, PhazonSuitHintMode, LoreHintMode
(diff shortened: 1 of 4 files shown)


henriquegemignani:
`guaranteed_item_hint.create_guaranteed_hints_for_resources` is supposed to be used by things like the artifacts. Hinting the major pickups like this is obviously broken like you mention and not a good solution at all.

Even just the idea of using the Chozo Lore and Space Pirate Logs for hints in Prime 1 doesn't work that well: there's just too many and mostly in Chozo Ruins. I do have an old idea for balacing these: using just one hint per world, with the rest being hints to where the actual hint is. Dunno how useful though.

If actually going with these scans, the proper solution is to create one LogbookNode for each of these and let the existing hint system in Randovania fill hints.

ThomasJRyan:
Unable to complete this PR

PR #3279, opened 2022-06-07
Minor optimizations in Requirements

henriquegemignani:
With an incredible sample size of 1, it sped up a dock rando seed gen by 10%.

### DIFF
--- a/randovania/game_description/requirements.py
+++ b/randovania/game_description/requirements.py
@@ -63,13 +63,15 @@ class Requirement:
 
 
 class RequirementArrayBase(Requirement):
+    __slots__ = ("items", "comment", "_cached_hash")
     items: Tuple[Requirement, ...]
     comment: Optional[str]
-    _cached_hash = None
+    _cached_hash: Optional[int]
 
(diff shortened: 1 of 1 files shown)


randovania/game_description/requirements.py
```diff
@@ -177,10 +179,10 @@ def damage(self, current_resources: ResourceCollection, database: ResourceDataba
...
 
     def satisfied(self, current_resources: ResourceCollection, current_energy: int, database: ResourceDatabase) -> bool:
-        return any(
-            item.satisfied(current_resources, current_energy, database)
-            for item in self.items
-        )
+        for item in self.items:
+            if item.satisfied(current_resources, current_energy, database):
+                return True
+        return False
```
duncathan, line 185: I'm surprised this is faster tbh

duncathan (approved):

randovania/game_description/requirements.py
henriquegemignani: I didn't test and don't expect it to be, but it's also simpler and in one of the hottest functions.

PR #3311, opened 2022-06-09
Export preset with suggested name

EthanArmbrust:
When exporting a preset, default to `{preset-slug-name}.rdvpreset` instead of nothing


henriquegemignani (changes requested):
Can you add an entry to the changelog?

henriquegemignani (approved):

PR #3352, opened 2022-06-14
Dread: Revise logic for the entrance rooms of Ghavoran

XenoWars:
- Added various methods for getting around Left Entrance and Right Entrance
- Added event for fighting the Gold Robot
- Removed templates for Water Bomb Jump and Water Cross Jump and combined them into one
- Added template for fighting the Gold Robot

### DIFF
--- a/randovania/games/dread/json_data/Ghavoran.txt
+++ b/randovania/games/dread/json_data/Ghavoran.txt
@@ -88,7 +88,9 @@ Extra - asset_id: collision_camera_001
   > Grapple Block Alcove
       All of the following:
           Morph Ball and After Ghavoran - Burenia Transport Grapple Box
-          Grapple Beam or Spider Magnet or Simple IBJ or Use Spin Boost
+          Any of the following:
+              Spider Magnet or Speed Booster or Simple IBJ or Use Spin Boost
+              Grapple Beam and Movement (Beginner)
 
 > Pickup (Missile Tank); Heals? False
   * Layers: default
@@ -110,13 +112,18 @@ Extra - asset_id: collision_camera_001
   > Tile Group (MISSILE)
       All of the following:
           Morph Ball
-          Gravity Suit or Space Jump or Speed Booster
+          Gravity Suit or Speed Booster or Perform WBJ or Use Spin Boost
   > Grapple Block Alcove
       All of the following:
           Morph Ball and Before Ghavoran - Burenia Transport Grapple Box
           Any of the following:
-              Spider Magnet or Space Jump
-              Flash Shift and Use Spin Boost
+              Spider Magnet or Space Jump or Speed Booster
+              Spin Boost and Movement (Beginner)
+              All of the following:
+                  Grapple Beam
+                  Any of the following:
+                      Movement (Advanced)
+                      Movement (Intermediate) and Walljump (Beginner)
 
 > Total Recharge; Heals? True
   * Layers: default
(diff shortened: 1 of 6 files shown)


randovania/games/dread/json_data/Ghavoran.txt
```diff
@@ -88,7 +88,9 @@ Extra - asset_id: collision_camera_001
   > Grapple Block Alcove
       All of the following:
           Morph Ball and After Ghavoran - Burenia Transport Grapple Box
-          Grapple Beam or Spider Magnet or Simple IBJ or Use Spin Boost
+          Any of the following:
+              Spider Magnet or Speed Booster or Simple IBJ or Use Spin Boost
+              Grapple Beam and Movement (Beginner)
```
kirbymastah, line 93: I'm not sure if this grapple beam requirement should require beginner movement - there's a similar requirement done in these two spots with no movement at all:

- Artaria - Energy Recharge Station South, going from Energy Recharge -> Missile Tank
- Dairon - Hub Access, going from Above Crumble Blocks -> Dock to Save Station East

At the very least, they should be consistent with each other?

randovania/games/dread/json_data/Ghavoran.txt
```diff
@@ -179,7 +186,19 @@ Extra - asset_id: collision_camera_002
...
+      Any of the following:
+          After Elun - Release X Parasites
+          All of the following:
+              Before Elun - Release X Parasites
+              Any of the following:
+                  Fight Gold Robot
+                  All of the following:
+                      Speed Booster and Combat (Beginner)
+                      Any of the following:
+                          Missiles ≥ 29 and Shoot Missile
```
kirbymastah, line 200: I'm a little confused by the extra missile requirements when that should be covered by "Fight Gold Robot". I notice they're lower than the reqirements for "Fight Gold Robot", so is this a different robot then, like one of the silver ones? If so, we might want a different template or that

randovania/games/dread/json_data/Ghavoran.txt
```diff
@@ -319,14 +353,16 @@ Extra - asset_id: collision_camera_003
...
   > Tile Group (POWERBEAM) 3
       Morph Ball
 
 > Floating Platform; Heals? False
   * Layers: default
   > Dock to Blue EMMI Introduction
-      Grapple Beam or Spider Magnet or Simple IBJ or Use Spin Boost
+      Any of the following:
+          Spider Magnet or Walljump (Beginner) or Simple IBJ or Use Spin Boost
+          Grapple Beam and Movement (Beginner)
```
kirbymastah, line 365: Same comment here about if we want Movement (Beginner) or not - though that grapple point looks higher and more annoying to reach, so maybe it doesn't need to be consistent with the artaria and dairon examples I listed?

kirbymastah (changes requested):
One other thing: remove the Screw Attack requirement from the golzuna fight while we're making updates to this area

randovania/games/dread/json_data/Ghavoran.txt
XenoWars: Yeah, I've been trying to make ones that you have to use the grapple momentum or jump out of it as a movement trick. forgot to do the one in Artaria i guess.

randovania/games/dread/json_data/Ghavoran.txt
XenoWars: This is coming in from the right entrance door, where you are able to get a shinespark onto the gold robot here, which knocks off almost half of its health. The shinespark can only be used once, and only on this robot

XenoWars: This is the speed method i am referring to in the commit

kirbymastah (approved):

PR #3366, opened 2022-06-18
Dread: Added a Linear Damage Runs option.

ScorelessPine:
This allows the user to change the damage over time dealt by environmental rooms to a linear damage over time, rather than increasing exponentially.

Requires randovania/open-dread-rando#106


duncathan (approved):
will need to update the requirements for open-dread-rando once the PR goes through and a new release is pushed

randovania/layout/preset_migration.py
```diff
@@ -6,7 +6,7 @@
 from randovania.layout.base.dock_rando_configuration import DockRandoMode, DockTypeState
 from randovania.lib import migration_lib
 
-CURRENT_VERSION = 33
+CURRENT_VERSION = 34
```
duncathan, line 9: you'll need to adjust the new migration to prevent merge conflicts, since I bumped the preset version

duncathan (changes requested):

randovania/layout/preset_migration.py
ScorelessPine: Merged upstream into my main while using the github tool to fix merge conflicts in preset describer, preset migration, and dread configuration. Also fixed an oopsie in function ordering that I made. I would recommend testing a game export before accepting the PR however, as I am unable to do so on my current computer setup.

duncathan (approved):

duncathan:
looks like tests are failing - you'll need to fix those before it's possible to merge

PR #3401, opened 2022-06-27
Dread: Revise logic for Freezer and those related

XenoWars:
Completely revamped all the logic for Freezer
- Room now accounts for coming in before activating Power Switch 2
- Removed the Speed Blocks and redid logic accordingly
- Added the Power Switch 2 event for getting into Freezer from Energy Recharge Station West
- Map Station platform now properly accounts for the power switch
- Door to Bomb Room now properly accounts for the power switch
- Various other methods of traveling through Freezer


randovania/games/dread/json_data/Dairon.txt
```diff
@@ -2294,105 +2304,58 @@ Extra - asset_id: collision_camera_025
...
-          Gravity Suit
-          Heat/Cold Runs (Intermediate) and Cold Damage ≥ 20
-  > Door to Energy Recharge Station West (doorpowerpower_030)
-      All of the following:
-          Morph Ball and After Dairon - db_dif_b1_001
-          Any of the following:
-              Gravity Suit
-              Heat/Cold Runs (Intermediate) and Cold Damage ≥ 100
-
-> Tile Group (SCREWATTACK); Heals? False
```
kirbymastah, line 2359: "Tile Group (SCREWATTACK)" -> "Bottom-Left Corner" should be an available connection that requires speedbooster.

As it is, there is only one path to bottom-left corner, which is "Tile Group (SCREWATTACK)" -> "Tile Group (WEIGHT)" -> "Bottom-Left Corner", which requires morph ball, when that isn't necessarily required to get to the bottom left corner.

randovania/games/dread/json_data/Dairon.txt
```diff
@@ -2405,38 +2368,69 @@ Extra - asset_id: collision_camera_025
...
+              Before Dairon - powergenerator_002
+              Any of the following:
+                  Grapple Beam or Spider Magnet or Space Jump or Speed Booster
+                  All of the following:
+                      Gravity Suit
+                      Spin Boost or Walljump (Beginner) or Simple IBJ
+                  Flash Shift and Walljump (Beginner)
+  > Door to Energy Recharge Station West (Upper)
+      All of the following:
+          Speed Booster
```
kirbymastah, line 2392: Currently, "Door to Energy Recharge Station West (Upper)" -> "Freezer Center" doesn't have a speedboost requirement, and if a room shuffle were to occur, then the speedboost blocks would block the player. However, I don't think it's possible to charge a speedboost from the top here.

It's finefor now, but I'm not sure what the best long-term solution is - maybe an event for destroying those speedboost blocks? At the very minimum, I would add a "TODO" comment saying that this eventually needs to be addressed.

kirbymastah (changes requested):
few quick comments

randovania/games/dread/json_data/Dairon.txt
XenoWars: it is required to get the item though, which is the only reason to go down there, but i probably should do that

randovania/games/dread/json_data/Dairon.txt
XenoWars: I considered adding an event, but then remembered you can't even get up there without destoyring the speed blocks anyway. And IIRC, room rando is never going to be a thing. That would be a thing to ask Dunc probably.

XenoWars:
What should I do about these tests? I correctly resolved the conflict. And it's weird it didn't occur with the Artaria one, just this one.

henriquegemignani:
Make sure you open the data editor and just press save, after you resolve conflicts.

randovania/games/dread/json_data/Dairon.txt
kirbymastah: ah you do have a point. but yeah I guess we can keep the extra path

kirbymastah (approved):

PR #3402, opened 2022-06-27
Dread: Remove Speed Blocks in Hanubia

XenoWars:
Removed the SPEEDBOOSTER tile groups in Hanubia and revised logic accordingly
- Added new event in Tank Room to track that the player has access to that side of the door
- Added Wave Beam or new event in Tank Room as logic for the Speedboost Puzzle Room item
- Added "Inside Breakable Tube" node to account for breaking it from both above and below


Isachu:
Looks good

Isachu (approved):
Looks good

PR #3441, opened 2022-07-08
Updated Ferenia/Energy recharge station (gate)

### LINKED ISSUE #3483: Grapple beam got placed behind a grapple block.
Found by Juice on 07/06/2022. 
Ferenia / Energy recharge station secret pickup.
Logic showed that grapple beam is not required for the grapple block.

https://discordapp.com/channels/914291389293027329/914508458739511358/994159884478202006

SFJucelino:
Grapple block now needs grapple beam. 
Added two more ways to bypass grapple event.
1) https://www.youtube.com/watch?v=NdVesJ6F_6A
2) https://twitter.com/i/status/1545111396760690692

fixes #3483


randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -1713,18 +1713,28 @@ Extra - asset_id: collision_camera_019
...
-          Morph Ball
+          Grapple Beam and Morph Ball
           Any of the following:
               Cross Bomb
               All of the following:
                   # Bomb & Power Bombs, to desync the tunnel bomb blocks from the ledge bomb blocks
                   Bomb and Power Bombs
+              All of the following:
+                  # Power bombs (video: https://twitter.com/i/status/1545111396760690692 )
+                  Power Bombs ≥ 2
```
XenoWars, line 1724: This should use some sort of movement trick for taking up the spot of the bomb block so it doesnt break.

randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -1713,18 +1713,28 @@ Extra - asset_id: collision_camera_019
...
+              All of the following:
+                  # Bombs + flash shift (video: https://www.youtube.com/watch?v=NdVesJ6F_6A )
+                  Bomb and Flash Shift
```
XenoWars, line 1727: This again needs some sort of movement trick to be able to do things in time before all the blocks reform.

randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -1713,18 +1713,28 @@ Extra - asset_id: collision_camera_019
...
       Morph Ball
   > Tunnel to Energy Recharge Station Secret
       All of the following:
           Morph Ball and After Ferenia - grapplepulloff1x2_000
-          Cross Bomb or Power Bombs
+          Any of the following:
+              Cross Bomb or Power Bombs ≥ 3
+              All of the following:
+                  # bomb + flash ( video: https://www.youtube.com/watch?v=NdVesJ6F_6A )
+                  Bomb and Flash Shift
```
XenoWars, line 1737: This can also probably have "Use Spin Boost" as an option rather than just Flash Shift. You'll have enough time to get back to the tunnel after the first block gets destroyed (and therefore does not need this video as no trick is involved)

randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -1713,18 +1713,28 @@ Extra - asset_id: collision_camera_019
...
+                  # Bombs + flash shift (video: https://www.youtube.com/watch?v=NdVesJ6F_6A )
+                  Bomb and Flash Shift
   > Tile Group (WEIGHT)
       Morph Ball
   > Tunnel to Energy Recharge Station Secret
       All of the following:
           Morph Ball and After Ferenia - grapplepulloff1x2_000
-          Cross Bomb or Power Bombs
+          Any of the following:
+              Cross Bomb or Power Bombs ≥ 3
```
XenoWars, line 1734: Change this back to 1 Power Bomb, or just "Lay Power Bomb". You don't need to cover for getting back out of Energy Recharge Station Secret. I recommend putting "Lay Power Bomb" for all cases where a PB can be required as it covers both the Main item and at least one in inventory. If more are needed, you can list Power Bombs with it.

randovania/games/dread/json_data/Ferenia.txt
SFJucelino: I agree. Either movement or knowledge should be added. How would you rate the skill level needed for this movement? I do not want to rate skill level since I just got here.

randovania/games/dread/json_data/Ferenia.txt
SFJucelino: By using walljump it is also possible to destroy this block, see video https://twitter.com/i/status/1545679535546372098.
What skill level for walljump should be added for this?
I should also check if the grapple block can be destroyed with spin boost (it is unlikely to do so with wall jumps).

randovania/games/dread/json_data/Ferenia.txt
SFJucelino: Does the randomizer know how many powerbombs are needed if I change it back into 1 PB needed? If so, the room Energy Recharge Station Secret should be changed a little bit because one PB is sufficient to grab the pickup.

SFJucelino:
Since I am adding skills required for tricks. Should I add the trick "knowledge" to beginner reach the bomb ledge with screw attack? The map doesn't show until area unless you destroy the SA blocks.

randovania/games/dread/json_data/Ferenia.txt
XenoWars: Yes, the randomizer will know how many you need. If you are going to add the Secret room to the PR, I suggest removing the "Tile Group (BOMB WEIGHT)" node as that one is essentially useless, then you'll need to mark going to "Corner" to include Power Bombs. And then in the Gate room you need to add Power Bombs on "Tunnel to Energy Recharge Station Secret" to "Tile Group (WEIGHT)"

randovania/games/dread/json_data/Ferenia.txt
XenoWars: Knowledge is not needed. It's vanilla game knowledge to know this area is here. 1 and 2 can be trickless. 3 I would personally mark as beginner, and then 4 maybe intermediate or advanced. I'm not sure how to mark that one without doing it myself. But it doesnt look as hard as like, bombing up a slope which can be either advanced or expert depending on the surroundings.

XenoWars:
Because coming up here is knowledge required in the vanilla game, a knowledge trick is not needed. Plus Pulse Radar exists.

SFJucelino:
> Because coming up here is knowledge required in the vanilla game, a knowledge trick is not needed. Plus Pulse Radar exists.

I disagree with this statement. Existence of Pulse Radar does not mean you have acquired Pulse Radar in a randomized game.
Also when you are in Energy Recharge Station (Gate), the map does not highlight that there is a secret in that room. So a player does not have a incentive to use Pulse Radar there. Lastly if a person has obtained all items in the game once, does not necessarily mean that the person knows about the existence of Energy Recharge Station Secret in his next playthrough. Map knowledge is a skill.

Besides this disagreement, I reworked the two rooms.

XenoWars:
Keep in mind all items need to have a trickless method to obtain them.  And map knowledge is always an assumed skill in any randomizer.

randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -1705,26 +1705,42 @@ Extra - asset_id: collision_camera_019
...
   > Event - Grapple Block (grapplepulloff1x2_000)
       All of the following:
-          Morph Ball
+          Grapple Beam and Morph Ball
           Any of the following:
               Cross Bomb
               All of the following:
                   # Bomb & Power Bombs, to desync the tunnel bomb blocks from the ledge bomb blocks
-                  Bomb and Power Bombs
+                  Bomb and Main Power Bomb and Power Bombs
```
XenoWars, line 1723: As was said in the discord, Main Power Bomb apparently does not need to be listed

randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -1705,26 +1705,42 @@ Extra - asset_id: collision_camera_019
...
   > Tunnel to Energy Recharge Station Secret
       All of the following:
-          Morph Ball and After Ferenia - grapplepulloff1x2_000
-          Cross Bomb or Power Bombs
+          After Ferenia - grapplepulloff1x2_000
+          Any of the following:
+              Any of the following:
+                  # Without flash shift or bombs
+                  Lay Cross Comb or Lay Power Bomb
+                  Bomb and Morph Ball and Movement (Beginner)
```
XenoWars, line 1739: Is it possible to grab the ledge without spin or flash?

randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -1749,6 +1765,10 @@ Extra - asset_id: collision_camera_020
   * Morph Ball Tunnel to Energy Recharge Station (Gate)/Tunnel to Energy Recharge Station Secret
   > Tile Group (BOMB WEIGHT)
       Lay Bomb
+  > Corner
```
XenoWars, line 1768: Remove the node for Tile Group (BOMB WEIGHT) too

Migs3:
I agree with Xeno on ERSS.  All randomizers (that I know of) sort of "expect" the player to have played through the vanilla game and 100%'d it.  Basically, "tricks" should be defined as anything outside of the "casual" playthrough or as anything outside of the intended methods of acquiring items.  While it's not always easy to know the true "developer intended" method, we can make generalizations about them, and in this case it's quite obvious that it's a case of "check every room with X-ray" (or Pulse Radar, in this case).  The "X-Ray every room" strat was a Super Metroid thing, for those unaware, and was fully intended to acquire all items.

randovania/games/dread/json_data/Ferenia.txt
SFJucelino: Added it because the template "Lay Power Bomb" includes main power bomb (link: https://github.com/randovania/randovania/blob/main/randovania/games/dread/json_data/header.json?plain=1#L1413-L1421). I'll remove it for this instance. It could also be removed for the "Lay Power Bomb" template.

randovania/games/dread/json_data/Ferenia.txt
SFJucelino: Yeah, by simply unmorphing as you can see in my bombs only video  https://twitter.com/i/status/1545703077138665472. Timing is easier than destroying that block with a bomb.

randovania/games/dread/json_data/Ferenia.txt
SFJucelino: I somehow did not save my last edit...

SFJucelino:
First time working on a randomizer and I will keep disagreeing with this. But if you guys say that having full map knowledge is expected of a player, the logic I'll update will be according to that assumption.

randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -1705,26 +1705,46 @@ Extra - asset_id: collision_camera_019
...
+      Trivial
   > Tunnel to Energy Recharge Station Secret
       All of the following:
-          Morph Ball and After Ferenia - grapplepulloff1x2_000
-          Cross Bomb or Power Bombs
+          After Ferenia - grapplepulloff1x2_000
+          Any of the following:
+              Any of the following:
+                  # Without flash shift or bombs
+                  Lay Cross Comb
```
XenoWars, line 1738: Wouldnt this also require movement, spin, or flash? wouldnt you just fall down to the weight block if you try to use cross to reach the tunnel?

XenoWars: This is the last potential issue as far as i can see

XenoWars: You could also move Morph Ball back to be with the event. Don't really know why you took out of always being needed. Can leave it there and just list Cross Bomb instead of Lay Cross Comb

SFJucelino: > You could also move Morph Ball back to be with the event. Don't really know why you took out of always being needed. Can leave it there and just list Cross Bomb instead of Lay Cross Comb

This is a nitpicky comment/review which does not hold any merit. "Lay Cross Comb" is currently not even used. For both the grapple event and tunnel morph ball is currently at the highest level. It is grouped with grapple beam for grapple event.
image
And morphball is also grouped with after grapple event after 
image

Did you forget to fetch the latest commit?

XenoWars (approved):
It looks good now

randovania/games/dread/json_data/Ferenia.txt
XenoWars: this commit was not in yet when i commented

XenoWars:
fixes #3483

PR #3476, opened 2022-07-18
Added Dread trick descriptions

Hugoshido:
Take 2


XenoWars (changes requested):
This should be fine after these two minor tweaks

PR #3496, opened 2022-07-24
Dread: Add Short boost as a Trick

SFJucelino:
1) Added Short Boost as a Trick.
~~2) Added a path that requires SSC and Short Boost in Screw Attack room.~~
2) Updated trick description and added link to tutorial/showcase where needed.

### DIFF
--- a/randovania/games/dread/json_data/Artaria.json
+++ b/randovania/games/dread/json_data/Artaria.json
@@ -23443,36 +23443,6 @@
                     "keep_name_when_vanilla": true,
                     "editable": true,
                     "connections": {
-                        "Door to Shortcut to Screw Attack": {
-                            "type": "and",
-                            "data": {
-                                "comment": null,
-                                "items": [
-                                    {
-                                        "type": "template",
-                                        "data": "Can SSC"
-                                    },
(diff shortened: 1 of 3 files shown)


XenoWars (approved):
When it comes time to add implementations for it, I would suggest putting in all the ones you can think of, region by region.

XenoWars:
Would like another opinion on the extensions of the trick descriptions. I personally do not feel they are necessary and could just have the link to a vod added to what was already there.

henriquegemignani:
> Would like another opinion on the extensions of the trick descriptions. I personally do not feel they are necessary and could just have the link to a vod added to what was already there.

I don't trust links to external resources to stay up long term. (This is also a problem I have with link to videos in the db, but for _usages_ of a trick it's more necessary).
The only link I'd like with trick descriptions would be to a community wiki.

duncathan (changes requested):
Please remove the video links from the trick descriptions for the time being. We can discuss them in a separate PR if desired

duncathan:
if we want to add short boost as a trick we can do so in a PR that does only that

PR #3504, opened 2022-07-26
Dread: Remove Speed Blocks in Dairon

XenoWars:
The final step in removing all the speed block nodes in the database. This removes the rest of the speed blocks in Dairon and applies logic accordingly.

- Removed Speed Blocks in Central Unit Access
- Removed Speed Blocks in EMMI Zone Exit South
- Removed Speed Blocks in Ammo Recharge Station
- Removed Speed Blocks in Storm Missile Gate Room
- Added event for destroying the tunnel block in the above room for future proof of random starting area
- Removed Speed Blocks in Shinespark Tutorial
- Removed Speed Blocks in EMMI Zone Exit West
- Removed Speed Blocks in EMMI Zone Exit North
- Removed Speed Blocks in EMMI Zone Exit East
- Added event for destroying the Speed Blocks in the above room as it is impossible to do so from the transport
- Removed Speed Blocks in Early Grapple Room


randovania/games/dread/json_data/Dairon.txt
```diff
@@ -405,16 +405,16 @@ Extra - asset_id: collision_camera_005
   * Extra - actor_def: actordef:actors/props/dooremmy/charclasses/dooremmy.bmsad
   > Door from Yellow EMMI Introduction
       Trivial
-  > Tile Group (SPEEDBOOST)
-      Trivial
+  > Event - Ferenia Transport Speed Blocks Destroyed
+      Speed Booster
```
kirbymastah, line 409: "Dock to Hub Access" is missing a direct connection to "Door to East Transport to Ferenia". Yes destroying the speedboost blocks will take you to the door, but if logic enters this room again with the speedboost blocks destroyed, then it will think "Dock to Hub Access" has no way to access "Door to East Transport to Ferenia".

randovania/games/dread/json_data/Dairon.txt
```diff
@@ -1202,6 +1172,8 @@ Extra - asset_id: collision_camera_012
 
 > Lava Floor with Gravity Suit; Heals? False
   * Layers: default
+  > Pickup (Missile Tank Speedboost)
+      Gravity Suit and Ballspark
```
kirbymastah, line 1176: This logically doesn't make a difference but IMO probably best to set this to "Gravity suit and morph ball and speedbooster" , for readability sake since this doesn't require an actual ballspark.

Also I don't think ballsparking would ever become an actual trick, but if it ever does, then this spot shouldn't have it.

randovania/games/dread/json_data/Dairon.txt
```diff
@@ -1670,10 +1536,10 @@ Extra - asset_id: collision_camera_018
   * Extra - right_shield_def: None
   > Dock to Big Hub
       Morph Ball and After Dairon - Central Unit
-  > Tile Group (SPEEDBOOST)
```
kirbymastah, line 1673: Remove this connection: "Dock to EMMI Zone Exit North" -> "Lower-Left Area"
Add this connection: "Dock to EMMI Zone Exit North" -> "Door to EMMI Zone Exit East", with a speedbooster requirement

kirbymastah (changes requested):
Few comments

randovania/games/dread/json_data/Dairon.txt
XenoWars: I did this because speed can get you down there, while you are not guaranteed to have charge (I have been in this situation). If you were to stop your speed before going through the flash gate, you'd have to go back and get the speed again. And then it's trivial to go from Lower-left to the charge door, so it's essentially the same. But i can still add the connection from Exit North to Exit East if you feel it's necessary.

randovania/games/dread/json_data/Dairon.txt
XenoWars: Does this actually matter? I thought there was something a bit ago that allows for revisiting events

kirbymastah (approved):

PR #3508, opened 2022-07-29
Dread: Revise logic for Spin Boost Tower

XenoWars:
- Added some speedbooster methods for climbing to various points in the room
- Added a speedbooster method of getting the power bomb tank
- Properly accounted for the ability to get the energy part item at the grapple block
- Removed the "Energy Part Platform" node as I thought it was pretty redundant
- Added advanced HBJ for Left Entrance item

### DIFF
--- a/randovania/games/dread/json_data/Ghavoran.txt
+++ b/randovania/games/dread/json_data/Ghavoran.txt
@@ -347,7 +347,9 @@ Extra - asset_id: collision_camera_003
   > Pickup (Missile Tank)
       All of the following:
           Morph Ball
-          Flash Shift or Grapple Beam or Space Jump or Speed Booster
+          Any of the following:
+              Flash Shift or Grapple Beam or Space Jump or Speed Booster
+              Bomb and Horizontal Bomb Jump (Advanced)
   > Tile Group (POWERBEAM) 3
       Morph Ball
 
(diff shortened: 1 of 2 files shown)


randovania/games/dread/json_data/Ghavoran.txt
```diff
@@ -1583,25 +1625,16 @@ Extra - asset_id: collision_camera_021
 > Ledge Below PB Tank; Heals? False
   * Layers: default
   > Door to Above Pulse Radar
-      Simple IBJ or Use Spin Boost
+      Movement (Beginner) or Simple IBJ or Use Spin Boost
```
kirbymastah, line 1628: Not sure what the "Movement (Beginner)" method is here - is it a slide jump? Could you add a comment clarifying this?

kirbymastah (changes requested):
Just one quick comment

randovania/games/dread/json_data/Ghavoran.txt
XenoWars: you can jump to the edge, but it's not a jump that is very simple. the jump doesnt even look doable, but it is

XenoWars: i also tested this a while ago back when i first did the revision. i could try again and see if it should just be a trivial connection

kirbymastah (approved):

PR #3509, opened 2022-07-29
Dread: Finalize Ghavoran logic revision

XenoWars:
- A few more changes to Spin Boost Tower
- Total Recharge Station North removed the crumble block nodes as they are redundant
- Made it impossible to reach the Screw Attack nodes in the above room and Teleport to Burenia if coming from above since you cannot use Screw Attack while in morph ball
- Added a connection to the Screw Attack node in Total Recharge Station North as there was not one previously
- Added various tricks to connections in Energy Recharge Station
- Added Flash Shift and Spin Boost methods for crossing the crumble blocks in Cross Bomb Tutorial and Teleport to Burenia
- Changed Ballspark to go from Orange to Golzuna Tower proper to beginner movement
- Changed Movement to Stand on Frozen Enemy now that this trick exists for climbing Golzuna Tower with Ice Missiles
- Removed the connection from Start Point to Pulse Radar pickup as this can cause issues if starting here in the future
- Added connection to the powerbeam block below orange as there was not one previously

### DIFF
--- a/randovania/games/dread/json_data/Ghavoran.txt
+++ b/randovania/games/dread/json_data/Ghavoran.txt
@@ -1523,7 +1523,7 @@ Extra - asset_id: collision_camera_021
   * Extra - actor_name: dooremmy
   * Extra - actor_def: actordef:actors/props/dooremmy/charclasses/dooremmy.bmsad
   > Upper Center Platform
-      Flash Shift or Simple IBJ or Use Spin Boost
+      Flash Shift or Movement (Beginner) or Simple IBJ or Use Spin Boost
 
 > Dock to EMMI Zone Exit Northeast (dooremmy_007); Heals? False
   * Layers: default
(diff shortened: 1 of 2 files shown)


randovania/games/dread/json_data/Ghavoran.txt
```diff
@@ -1941,7 +1947,7 @@ Extra - asset_id: collision_camera_024
   * Layers: default
   * Slide Tunnel to Above Pulse Radar/Tunnel to Golzuna Tower
   > Tile Group (POWERBEAM) 1
-      Morph Ball
+      Can Slide
```
kirbymastah, line 1959: This distance is too far to reach with sliding alone if i remember correctly - I would keep this as morph ball

randovania/games/dread/json_data/Ghavoran.txt
```diff
@@ -2116,11 +2130,9 @@ Extra - asset_id: collision_camera_025_B
   * Extra - right_shield_entity: Root:pScenario:rEntitiesLayer:dctSublayers:default:dctActors:doorshieldmissile_001
   * Extra - right_shield_def: actordef:actors/props/doorshieldmissile/charclasses/doorshieldmissile.bmsad
   > Door from Chozo Warrior Arena
-      Morph Ball and Before Ghavoran - Ferenia Transport Grapple Box
+      Before Ghavoran - Ferenia Transport Grapple Box and Can Slide
```
kirbymastah, line 2142: Same here, this distance is too far to slide alone - keep this as morph ball

randovania/games/dread/json_data/Ghavoran.txt
XenoWars: Without morph ball you just keep sliding until you fall or hit a wall

randovania/games/dread/json_data/Ghavoran.txt
XenoWars: And as before, you'll just keep sliding.

kirbymastah (changes requested):
two small comments about sliding

kirbymastah (approved):

PR #3511, opened 2022-07-30
Dread: Begin Dairon logic revision

XenoWars:
This marks the beginning of Dairon's logic revision. This covers the rooms you encounter when you first arrive to get Wide Beam.

- Correctly accounted for the power switch for certain doors (does not yet account for the one in the heated room)
- Added Wave Beam methods for the first two wide beam blocks, the first of which being trickless and the second requiring movement
- Corrected the door to Yellow EMMI Introduction in EMMI Zone Exit East to be Access Locked rather than Access Closed
- Added reverse grapple to the East Ferenia Transport block
- Added various methods of climbing Big Hub and Big Hub Access


randovania/games/dread/json_data/Dairon.txt
```diff
@@ -177,7 +194,9 @@ Extra - asset_id: collision_camera_002
   > Dock to EMMI Zone Exit East
       After Dairon - Wide Beam Block by Save Room
   > Above Crumble Blocks
-      Spider Magnet or Walljump (Beginner) or Simple IBJ or Use Spin Boost
+      Trivial
```
XenoWars, line 197: what even was the thought process here the first time around

randovania/games/dread/json_data/Dairon.txt
```diff
@@ -143,7 +154,13 @@ Extra - asset_id: collision_camera_001
   > Top Floor
       After Dairon - Teleport To Artaria Speed Blocks Destroyed
   > Event - Speed Booster blocks Destroyed
-      Speed Booster
+      All of the following:
+          Speed Booster
+          Any of the following:
+              Grapple Beam or Spider Magnet or Space Jump or Simple IBJ
```
kirbymastah, line 160: If you climb with spider magnet without walljumping, can you still make this spark in time?

randovania/games/dread/json_data/Dairon.txt
kirbymastah: Looks like this logic was intended to go from down to up, not the other way lmao. Whoops.

randovania/games/dread/json_data/Dairon.txt
```diff
@@ -288,7 +311,9 @@ Extra - asset_id: collision_camera_004
   > Pickup (Missile Tank)
       All of the following:
           Morph Ball
-          Grapple Beam or Space Jump or Speed Booster or Simple IBJ
+          Any of the following:
+              Grapple Beam or Space Jump or Speed Booster or Simple IBJ
+              Flash Shift and Use Spin Boost
```
kirbymastah, line 316: I feel like you could also add a cross bomb alternative:

Flash Shift AND (Use Spin Boost OR Lay Cross Comb)

randovania/games/dread/json_data/Dairon.txt
```diff
@@ -566,7 +589,7 @@ Extra - asset_id: collision_camera_007
   * Extra - start_point_actor_name: powergeneratorengine_001
```
kirbymastah, line 589: For visualization purposes, could you adjust the coordinates of the start point so it doesn't overlap with the event? At the moment, it looks like they're the same node, and I thought the event had no connections coming out because of it. Maybe set X=700 or something like that.

randovania/games/dread/json_data/Dairon.txt
```diff
@@ -1586,7 +1621,9 @@ Extra - asset_id: collision_camera_018
   * Layers: default
   * Open Passage to EMMI Zone Exit North/Dock to Yellow EMMI Introduction
   > Pickup (Energy Part)
-      Missiles and Lay Bomb
+      All of the following:
```
kirbymastah, line 1624: I would just remove the "Pickup (Energy Part)" -> "Door to EMMI Zone Exit North"  connection entirely - not sure why that was there in the first place. Just have the energy part require morph ball to get out via the weight tile.

kirbymastah (changes requested):
Few comments

randovania/games/dread/json_data/Dairon.txt
XenoWars: there's a slope you can spark into from the wall, so yes

randovania/games/dread/json_data/Dairon.txt
XenoWars: i would need to test to see if that is doable. you wouldnt get any horizontal movement from the cross bomb

randovania/games/dread/json_data/Dairon.txt
XenoWars: also this was the wrong area for it. this comment was placed on the opposite connection

kirbymastah (approved):

PR #3535, opened 2022-08-03
Make First Thermal Device event less dangerous

henriquegemignani:


randovania/games/dread/json_data/Cataris.txt
```diff
@@ -257,6 +259,13 @@ Extra - asset_id: collision_camera_004
   > Door to Navigation Station Southeast
       Morph Ball
```
XenoWars, line 260: If you want to be consistent with your changes, I'd suggest making the Dock and Tunnel to Above Z-57 Fight point to Start Point instead of Navigation Station Southeast

XenoWars (changes requested):
Just one small thing

PR #3544, opened 2022-08-05
Optimize customize preset

henriquegemignani:
By keeping only the current tab active, we massively decrease the time it takes to create the window at the cost of some small freeze when opening the slow tabs.
As the tabs are now re-created each time it's opened, they now use the filtered database for the current configuration.

Remove the hidden elements from the location pool tab, as they cause significant slowdown to creation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed: The multi-pickup placement, using the new weighting, is now the default mode. The old behavior has been removed.
 - Changed: Error messages when a permalink is incompatible have been improved with more details.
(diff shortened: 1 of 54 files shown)


toasterparty (approved):

duncathan:
do you think it may be valuable to cache the item pool tab in particular? that tab is probably one of the more commonly used, and doesn't benefit from the db filtering as far as I'm aware

henriquegemignani:
> do you think it may be valuable to cache the item pool tab in particular? that tab is probably one of the more commonly used, and doesn't benefit from the db filtering as far as I'm aware

It's also annoyingly one of the few that actually takes a perceptible time.
I'll leave further optimisation to another PR, so it can easily start from a working state.

duncathan:
yeah, that was why I focused on that tab. most of the others are more or less instant so caching them would serve no purpose

PR #3552, opened 2022-08-07
Dread: Revise logic for the middle section of Dairon

XenoWars:
This revises the logic of the middle section of Dairon. It adds various methods of getting around these rooms, and also renames the events to make them easier to read. This also begins to remove the POWERBOMB block nodes since those were deemed unnecessary. The rooms affected are as follows:

- Transport to Artaria
- Early Grapple Room
- Early Grapple Access
- Heated Room West
- Energy Recharge Station Middle
- Test Chamber Access
- Total Recharge Station West
- Power Switch 2
- Cross Bomb Puzzle Room
- EMMI Zone Exit North
- West Transport to Ferenia


XenoWars:
I apologize ahead of time for the 1000+ lines changed.

randovania/games/dread/json_data/Dairon.txt
```diff
@@ -1085,10 +1120,12 @@ Extra - asset_id: collision_camera_012
           Any of the following:
```
kirbymastah, line 1120: How about adding this path: "Ledge before Slide Jump" -> "Door to Early Grapple Access"?

This would provide a non-lava method of going left without grapple beam. Should be doable with flash shift / spin boost at least.

kirbymastah (changes requested):
One suggestion

randovania/games/dread/json_data/Dairon.txt
XenoWars: done, don't know why I didn't do this myself. I even thought about it, just dont know why i didnt do it

kirbymastah (approved):

PR #3559, opened 2022-08-08
Automatically calculate the current version for migrations

henriquegemignani:


randovania/lib/migration_lib.py
```diff
@@ -6,22 +6,20 @@ class UnsupportedVersion(ValueError):
...
-def unsupported_migration(data: dict) -> dict:
-    raise UnsupportedVersion("Support for migrating from this version has been removed")
+Migrations = list[Callable[[dict], dict] | None]
 
 
-def migrate_to_version(data: dict, version: int, migrations: dict[int, Callable[[dict], dict]],
-                       *, copy_before_migrating: bool = False) -> dict:
+def apply_migrations(data: dict, migrations: Migrations, *, copy_before_migrating: bool = False) -> dict:
     schema_version = data.get("schema_version", 1)
+    version = len(migrations) + 1
```
duncathan, line 14: this should use `get_version` too

duncathan (changes requested):

PR #3561, opened 2022-08-08
Support optional values in bitpacking

henriquegemignani:
This allows us to have values in configuration classes that can be None.

### DIFF
--- a/randovania/bitpacking/bitpacking.py
+++ b/randovania/bitpacking/bitpacking.py
@@ -6,6 +6,8 @@ from typing import Iterator, TypeVar, Callable
 
 import bitstruct
 
+from randovania.lib import type_lib
+
 T = TypeVar("T")
 
 ARBITRARY_INT_SIZE_LIMITS = (4, 16, 128, 1024)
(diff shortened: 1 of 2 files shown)


duncathan (approved):

randovania/bitpacking/bitpacking.py
```diff
@@ -196,7 +198,14 @@ def bit_pack_encode(self, metadata) -> Iterator[tuple[int, int]]:
                 continue
 
             item = getattr(self, field.name)
-            bit_pack_value = _get_bit_pack_value_for(item, field.type)
+
+            resolved_type, optional = type_lib.resolve_optional(field.type)
+            if optional:
+                yield from encode_bool(item is not None)
```
toasterparty, line 204: `item is not None` this is the only confusing thing to me in this PR.

Wouldn't it be:

```py
            if optional:
                if item is None:
                    continue
                yield from encode_bool(item)
```

Or do I not know how yielding works

henriquegemignani: The `bit_pack_encode` functions are generators of `(value, max_value)`. Think of yielding values as returning an array of all values, but you can do so one at a time.

These functions effectively define a binary data format, using the yield values for it.
What I'm yielding there is a boolean that indicates if `item is not None`. Only when that value is true is more data encoded.

PR #3565, opened 2022-08-08
Dread: Add connection to get Main Hub Tower Top Energy Tank

Tyranisaur:
Before:
image

After:
image

Using the method presented in this video https://youtu.be/KX6N59HDpNM?t=115

### DIFF
--- a/randovania/games/dread/json_data/Burenia.txt
+++ b/randovania/games/dread/json_data/Burenia.txt
@@ -1110,8 +1110,36 @@ Extra - polygon: [[-900.0, 3700.0], [-5000.0, 3700.0], [-5000.0, -2115.0], [-900
       Trivial
   > Pickup (Energy Tank)
       All of the following:
-          Flash Shift and Morph Ball
-          Grapple Beam or Spider Magnet or Walljump (Beginner)
+          Morph Ball
+          Any of the following:
+              All of the following:
+                  Use Spin Boost
+                  Movement (Intermediate) or Slide Jump (Beginner)
+              All of the following:
+                  Speed Booster and Disabled Door Lock Randomizer
+                  Spider Magnet or Movement (Beginner)
+              All of the following:
+                  Flash Shift
+                  Any of the following:
+                      Spider Magnet
+                      All of the following:
+                          Movement (Beginner)
+                          Walljump (Beginner) or Use Spin Boost
+              All of the following:
+                  Grapple Beam
+                  Any of the following:
+                      Spider Magnet
+                      All of the following:
+                          Movement (Beginner)
+                          Walljump (Beginner) or Use Spin Boost
+              All of the following:
+                  Lay Cross Comb
+                  Any of the following:
+                      Spider Magnet
+                      All of the following:
+                          Movement (Intermediate)
+                          Walljump (Beginner) or Use Spin Boost
+                      Damage Boost (Beginner) and Stand on Frozen Enemy (Intermediate) and Shoot Ice Missile
 
 ----------------
 Main Hub Tower Middle
--- a/randovania/games/dread/json_data/Burenia.json
+++ b/randovania/games/dread/json_data/Burenia.json
@@ -5324,15 +5324,6 @@
                                         "type": "resource",
                                         "data": {
                                             "type": "items",
-                                            "name": "Flash",
-                                            "amount": 1,
-                                            "negate": false
-                                        }
-                                    },
-                                    {
-                                        "type": "resource",
-                                        "data": {
-                                            "type": "items",
                                             "name": "Morph",
                                             "amount": 1,
                                             "negate": false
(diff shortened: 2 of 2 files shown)


XenoWars:
is this the tank directly below the path to the bottom part of dairon hub?

Tyranisaur:
Yes

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1110,8 +1110,12 @@ Extra - polygon: [[-900.0, 3700.0], [-5000.0, 3700.0], [-5000.0, -2115.0], [-900
...
   > Pickup (Energy Tank)
       All of the following:
-          Flash Shift and Morph Ball
-          Grapple Beam or Spider Magnet or Walljump (Beginner)
+          Morph Ball
+          Any of the following:
+              Grapple Beam and Spider Magnet
+              All of the following:
+                  Flash Shift
+                  Grapple Beam or Spider Magnet or Walljump (Beginner)
```
XenoWars, line 1118: Beginner walljumps like this are treated as being trivial now, so this can just be flash shift. You can also walljump after using grapple. Space Jump is a viable option as well. So it should be Grapple or Flash Shift or Space Jump.

XenoWars (changes requested):

randovania/games/dread/json_data/Burenia.txt
Tyranisaur: I've unsuccessfully tried to pull it off with just space jump before, but after you said it here I went back and was able to do it without much trouble.

I also figured out that you can get this by using the same cross bomb jump that can be used to pass through shutters (flash shift doors), to launch you into the spider surface. And you can charge speed in the save room. I can't immediately tell if wall jumping off from either of those actions is viable, so at least there's some usefulness to the spider surface.

f-raZ0R:
For the record, Spider Magnet is not actually needed for the shinespark method, as after the shinespark, you have enough time to perform a wall jump off of the right wall. This, however, is a tight enough timing that I'd propose making it actually require Walljump (Beginner), as opposed to being trivial.


https://user-images.githubusercontent.com/30442287/184248748-abf9e1a2-7ac2-4b2b-8561-e9e19454f95f.mp4

f-raZ0R:
this is also possible but is very stupid

https://user-images.githubusercontent.com/30442287/184258754-6c6bc504-aac4-49b2-99c6-6c136cc47db4.mp4

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1110,8 +1110,21 @@ Extra - polygon: [[-900.0, 3700.0], [-5000.0, 3700.0], [-5000.0, -2115.0], [-900
...
+              All of the following:
+                  # Speed boost into spider magnet surface or beginner walljump
+                  Speed Booster
+                  Spider Magnet or Walljump (Beginner)
+              All of the following:
+                  # Midair sideways cross bomb into spider magnet surface or onto frozen enemy
+                  Lay Cross Comb
+                  Any of the following:
+                      Spider Magnet
+                      Stand on Frozen Enemy (Intermediate) and Shoot Ice Missile
```
XenoWars, line 1127: i still want to say you can probably walljump after the cross bomb without using the enemy but would need tested. this should also probably include Damage Boost (Beginner).

Also, Spin Boost alone is also enough to get into the hole but it's rather difficult. I'd place it at Movement (Intermediate)

XenoWars (changes requested):

randovania/games/dread/json_data/Burenia.txt
Tyranisaur: I did this. It's unclear if the damage boost or spin boost should then also be paired with a walljump.

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1110,8 +1110,24 @@ Extra - polygon: [[-900.0, 3700.0], [-5000.0, 3700.0], [-5000.0, -2115.0], [-900
...
+              All of the following:
+                  # Speed boost into spider magnet surface or beginner walljump
+                  Speed Booster
+                  Spider Magnet or Walljump (Beginner)
+              All of the following:
+                  # Midair sideways cross bomb into spider magnet surface,  beginner damage boost or onto frozen enemy
+                  Lay Cross Comb
+                  Any of the following:
+                      Spider Magnet or Damage Boost (Beginner)
+                      Stand on Frozen Enemy (Intermediate) and Shoot Ice Missile
```
XenoWars, line 1127: So you can get a wall jump off with just cross? Probably should be wall jump beginner then if the timing is any bit tight. The damage boost would also need paired with the Stand on Frozen Enemies

So,
Spider Magnet or Wall Jump (Beginner)
Damage Boost (Beginner) and Stand on Frozen Enemies (Intermediate) and Shoot Ice Missile

XenoWars (changes requested):

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1110,8 +1110,24 @@ Extra - polygon: [[-900.0, 3700.0], [-5000.0, 3700.0], [-5000.0, -2115.0], [-900
...
-          Flash Shift and Morph Ball
-          Grapple Beam or Spider Magnet or Walljump (Beginner)
+          # You always need morph
+          Morph Ball
+          Any of the following:
+              # Either flash or grapple with a trivial wall jump, or space jump directly
+              Flash Shift or Grapple Beam or Space Jump
+              All of the following:
+                  # Speed boost into spider magnet surface or beginner walljump
+                  Speed Booster
```
XenoWars, line 1120: Oh, I'm pretty sure you have to use the save station to get this, so this also needs "Disabled Door Lock Randomizer"

XenoWars (changes requested):

randovania/games/dread/json_data/Burenia.txt
f-raZ0R: Not true, although this one also requires disabled door lock randomizer.
i think that the area near the flash gate is large enough that you could charge a short charge there though.

https://user-images.githubusercontent.com/30442287/184507361-37374695-1606-423a-861f-06b2faa894ee.mp4

.

XenoWars: yeah this still needs the door lock thing

f-raZ0R: im also currently labbing whether you can use spark conservation to bring a spark in all the way from the top left door, and it is... very dumb. if the slope near the missile was one block lower and aligned with the door, this would be so much easier...

f-raZ0R: the spark requires flash shift anyways to actually bring it far enough, but i'm still curious at least...

Tyranisaur: I swear I had the door lock thing on an earlier version. I must have accidentally dropped it because it's a hassle to add new rows above existing nested ones. 😅 The spark from the flash shift room would be a more complicated version of the other spark, so until there's sufficient door lock logic in place to account for different doors then it's redundant.

randovania/games/dread/json_data/Burenia.txt
Tyranisaur: I'll try to figure out if the simple cross bomb trick can be done with a wall jump. But the damage boosts and standing on frozen I'm not super interested in checking personally, but I'll add it in as per requested.

XenoWars: No need to check for damage boost. the damage is done together with the frozen enemy, so they must be combined, not separated.

Tyranisaur:
Besides the spin boost option, which I figured you can make simpler by doing a coyote jump, there are four different options in regards to how to hit the wall.

For all these, having spider magnet makes the pickup trivial.

1. Flash shift: The main challenge of this one is that you need to shift as high up as possible, but if you do the trick is easy. You can either walljump or spin boost. As far as I can tell, the window for spin boosting is slightly more lenient compared to the walljump.
2. Grapple beam: Largely the same situation as the flash shift. It took me slightly longer to figure out where to aim, but since you stand still while doing it, I figure the trick is overall at beginner level.
3. Cross bomb. I placed this at intermediate movement, because I think it is notably harder to align this, since you have to lay the cross bomb at the correct height and then also time the ball jump to hit the blastoff. I suspect that if you go via the frozen enemy, the window isn't as tight, so I placed that outside the movement requirement.
4. Speed Booster. While this can be done similarly to the other options, through a midair spark, I figured that if you do it from the ground at the platform level, it seems to give you a significantly easier time hitting the jump. Also, interestingly, it seems like you don't even need to walljump for this, and instead effectively jump off the pitfall block. You still have to hold right when you jump, but by changing direction midair, you don't need to do a walljump.

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1110,8 +1110,36 @@ Extra - polygon: [[-900.0, 3700.0], [-5000.0, 3700.0], [-5000.0, -2115.0], [-900
       Trivial
   > Pickup (Energy Tank)
       All of the following:
-          Flash Shift and Morph Ball
-          Grapple Beam or Spider Magnet or Walljump (Beginner)
+          Morph Ball
+          Any of the following:
+              All of the following:
+                  Use Spin Boost
+                  Movement (Intermediate) or Slide Jump (Beginner)
```
XenoWars, line 1117: Have you even tried this with a slide jump? it doesnt make it easier at all cause of the crumble blocks. Your slide just gets eaten up by them

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1110,8 +1110,36 @@ Extra - polygon: [[-900.0, 3700.0], [-5000.0, 3700.0], [-5000.0, -2115.0], [-900
...
+              All of the following:
+                  Speed Booster and Disabled Door Lock Randomizer
+                  Spider Magnet or Movement (Beginner)
+              All of the following:
+                  Flash Shift
+                  Any of the following:
+                      Spider Magnet
+                      All of the following:
+                          Movement (Beginner)
```
XenoWars, line 1126: not sure what the movement here is for. I understand for Speed Booster, but i personally dont feel it's needed for flash

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1110,8 +1110,36 @@ Extra - polygon: [[-900.0, 3700.0], [-5000.0, 3700.0], [-5000.0, -2115.0], [-900
...
+                          Walljump (Beginner) or Use Spin Boost
+              All of the following:
+                  Grapple Beam
+                  Any of the following:
+                      Spider Magnet
+                      All of the following:
+                          Movement (Beginner)
+                          Walljump (Beginner) or Use Spin Boost
```
XenoWars, line 1134: Grapple can definitely just be a trivial option. It's definitely not hard at all to just stand in place and grapple the top of the spider wall, and wall jump (simple wall jumps like this are being treated as trivial)

XenoWars (changes requested):

randovania/games/dread/json_data/Burenia.txt
Tyranisaur: The pitfall blocks slow down your slide, making the timing very easy.

randovania/games/dread/json_data/Burenia.txt
Tyranisaur: I was only able to get the walljump or spin boost off correctly if I did the flash from a height that has precision to it.

randovania/games/dread/json_data/Burenia.txt
Tyranisaur: Nothing is hard when you know how to do it.  You need to stand close enough to the edge, aim high enough at the wall , let go of the grapple at the correct time and then wall jump during a small window. All in all, I think this is a complicated operation that I don't think we can expect less experienced dread players to understand that they can do.

XenoWars (approved):

PR #3568, opened 2022-08-09
Use py-cord instead of discord.py and discord-py-interactions

henriquegemignani:
The former is no longer developed, the later changed it's api entirely, so we'd have to rewrite everything anyway.


duncathan (approved):
I'm gonna be honest I don't know nearly enough to review this properly so I'm gonna say if it works it's good lol

PR #3581, opened 2022-08-12
Adding Advanced Installer

henriquegemignani:


.github/workflows/tests.yml
```diff
@@ -168,6 +168,97 @@ jobs:
...
+          $productDetails.Version = "${{ steps.version.outputs.version }}";
+          Write-Host "Version: $productDetails.Version";
+          
+          # Build the project
+          $project.Build();
+
+      - name: Upload installer
+        uses: actions/upload-artifact@v3
+        with:
+          name: Windows Installer
```
duncathan, line 257: should be `Randovania Windows Installer` for consistency prob

duncathan (changes requested):
everything else is over my head lol so if it works then great. i tried the installer and it seems to work so besides the naming thing it looks good

PR #3603, opened 2022-08-17
update blank dev game cover art

duncathan:
my friend made this and we're keeping it. shoutouts to steph


toasterparty:
thanks steph

PR #3611, opened 2022-08-18
Dread: Begin Burenia Logic Revision

XenoWars:
This PR begins the process of revising the logic for Burenia. This covers the upper portion of Burenia stretching from the Ghavoran Transport to the Ferenia Teleport.

- Added event for Drogyga to account for some logic in Underneath Drogyga
- Corrected the logic for fighting Drogyga
- Renamed various event nodes to make them easier to understand
- Removed some redundant WEIGHT tile nodes

Added various methods of traversing the following rooms:
- Transport to Ghavoran
- Teleport to Ferenia
- Drogyga Eyedoor

### DIFF
--- a/randovania/games/dread/json_data/Burenia.json
+++ b/randovania/games/dread/json_data/Burenia.json
@@ -2327,12 +2327,29 @@
                                         "data": "Simple IBJ"
                                     },
                                     {
-                                        "type": "resource",
+                                        "type": "and",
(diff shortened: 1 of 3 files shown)


XenoWars:
I will update this with Grapple Movement once it is accepted.

kirbymastah (approved):

PR #3632, opened 2022-08-23
Improve usability enemy attribute

henriquegemignani:


duncathan:
something broke some prime1 UI tests, and I'm guessing the status update change broke the SM UI tests

PR #3649, opened 2022-08-25
Dread: Adjust lava damage and boss energy requirements

XenoWars:
- Add Heat/Cold Runs trick to all instances of lava damage that did not already have it
- Added energy requirements for all 7 main bosses

### DIFF
--- a/randovania/games/dread/json_data/Cataris.txt
+++ b/randovania/games/dread/json_data/Cataris.txt
@@ -523,6 +523,10 @@ Extra - asset_id: collision_camera_009
           Any of the following:
               Shoot Charge Beam and Shoot Wide Beam
               Missiles ≥ 30 and Shoot Super Missile
+          Any of the following:
+              # Energy Requirements
+              Combat (Intermediate) or Normal Damage ≥ 499
+              Combat (Beginner) and Normal Damage ≥ 249
 
 > Z-57 Room; Heals? False
   * Layers: default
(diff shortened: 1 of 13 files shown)


duncathan (changes requested):
please add normal damage reductions and update the damage values accordingly

XenoWars:
The damage values for the bosses are good? or some need increased?

randovania/games/dread/json_data/Cataris.txt
```diff
@@ -1185,7 +1193,9 @@ Extra - asset_id: collision_camera_015
   > Door to Artaria Transport Access
       All of the following:
           Morph Ball
-          Gravity Suit or Lava Damage ≥ 600
+          Any of the following:
+              Gravity Suit
+              Heat/Cold Runs (Intermediate) and Lava Damage ≥ 1000
```
henriquegemignani, line 1198: Changed to 1000 here?

henriquegemignani (changes requested):

randovania/games/dread/json_data/Cataris.txt
XenoWars: Yes, because the opposite connection is 1000. This direction would also be slower with having to unmorph and regrab into every tunnel

XenoWars:
Is there anything I need to address as to why this wasnt approved yet?

duncathan (approved):

PR #3660, opened 2022-08-28
Combine single requirement with nested array

henriquegemignani:
Few samples of how it looks in the visualizer:

image
image
image
image


duncathan (commented):
I really don't like it tbh. it's a bit more concise but this is way harder for me to read. I prefer the old look

henriquegemignani:
I did another change: if an and/or block would have just a single row, it skips the `all/any of the following:` text.

image
image
image
image

PR #3661, opened 2022-08-28
Artaria waterfall Allow rotatable to be activated with wave beam

Tyranisaur:
If you have wave beam here, you can shoot the blob to rotate the flipper
and walk between the shortcut to screw attack and waterfall rooms.

image


XenoWars:
Since the event is trivial to the door to freezer, a new event node is required to be made to be trivial to the door to screw attack shortcut or right of rotatable. However, I'll be doing an Artaria PR later today that contains this room, so it can just be left til when I do it.

PR #3667, opened 2022-08-29
Improvements for logging related to generator

henriquegemignani:


duncathan (commented):
is this making changes to the behavior of the generator in addition to adding the logging?

henriquegemignani:
> is this making changes to the behavior of the generator in addition to adding the logging?

The closest is `if reduction.inventory_item is not None and not collection.has_resource(reduction.inventory_item):`. But I'm not sure it makes any difference in usage (adding a requirement that is known to be satisfied is.. lol), but it's also just used for the generation failed exception.
But it does mean Varia/Gravity Suit does not show up as possible resources for progression despite already having them.

The `possible_sets = ...` change filters out empty sets, which do nothing but complicate a debugger. Maybe makes things faster?

duncathan (approved):

PR #3677, opened 2022-08-30
Dread: Add templates for fighting the Chozo mini-bosses

XenoWars:
Added templates for the following:

- Silver Robot
- Twin Robots
- Chozo-X
- Red Chozo-X
- Gold Chozo-X

Modified template for Gold Robot. Twin Robots template can be used for both instances.


randovania/games/dread/json_data/header.txt
```diff
@@ -111,31 +111,222 @@ Templates
...
               Cross Bomb and Water Bomb Jump (Intermediate)
 
+* Open Charge Door:
+      Any of the following:
+          Shoot Charge Beam
+          Knowledge (Beginner) and Lay Power Bomb
+
+* Fight Silver Robot:
+      Any of the following:
+          Knowledge (Beginner) and Lay Power Bomb
```
duncathan, line 121: I believe that Knowledge checks such as this, in the context of fights, are better off as Combat tricks. @henriquegemignani can verify but iirc it's used that way in the other games

randovania/games/dread/json_data/header.txt
```diff
@@ -111,31 +111,222 @@ Templates
...
+              Any of the following:
+                  # Methods of fighting
+                  Shoot Plasma Beam
+                  Combat (Intermediate) and Shoot Charge Beam
+                  All of the following:
+                      # 47 Missiles are required, 1.5x leeway for trickless
+                      Shoot Missile
+                      Any of the following:
+                          Missiles ≥ 61
+                          Missiles ≥ 47 and Combat (Beginner)
```
duncathan, line 135: strict missile requirements should probably be higher than Beginner - maybe Intermediate? I'm also not seeing counter to refill missiles being accounted for here. in theory I'm pretty sure 1 missile is sufficient to beat a robot (and would obviously be hypermode) but something like 15ish + a combat level for the counter would be appropriate as well

randovania/games/dread/json_data/header.txt
```diff
@@ -111,31 +111,222 @@ Templates
...
+                  All of the following:
+                      # 28 Super Missiles are required, 1.5x leeway for trickless
+                      Shoot Super Missile
+                      Any of the following:
+                          Missiles ≥ 42
+                          Missiles ≥ 28 and Combat (Beginner)
+                  All of the following:
+                      # 15 Storm Missiles are required, 1.5x leeway for trickless
```
duncathan, line 143: might be better to indicate the number of lock-ons necessary in the comment rather than the raw ammo counts

randovania/games/dread/json_data/header.txt
```diff
@@ -111,31 +111,222 @@ Templates
...
+          All of the following:
+              Any of the following:
+                  # Methods of dodging
+                  Flash Shift or Combat (Beginner)
+              Any of the following:
+                  # Methods of fighting
+                  Shoot Plasma Beam
+                  Combat (Intermediate) and Shoot Charge Beam
+                  All of the following:
+                      # 101 Missiles are required, 1.5x leeway for trickless
```
duncathan, line 209: ammo counts for twins could also account for drops from killing one of them, with a combat trick to accompany ofc

randovania/games/dread/json_data/header.txt
```diff
@@ -111,31 +111,222 @@ Templates
...
+                      Storm Missile and Shoot Missile
+                      Any of the following:
+                          Missiles ≥ 75
+                          Missiles ≥ 50 and Combat (Beginner)
+                  Combat (Intermediate) and Shoot Wide Beam
+                  Combat (Advanced) and Shoot Beam
+
+* Fight Gold Chozo-X:
+      Any of the following:
+          Power Bombs ≥ 2 and Knowledge (Beginner) and Lay Power Bomb
```
duncathan, line 300: Gold chozo-x is NOT vulnerable to power bombs

duncathan (changes requested):

duncathan:
also, if any of these require more ice missiles than super missiles to kill, use the ice missile ammo count for both ice and supers and mention that in a comment

randovania/games/dread/json_data/header.txt
XenoWars: I checked with exo, and flamethrower instead of power beam is listed as both knowledge and combat

randovania/games/dread/json_data/header.txt
XenoWars: yeah i forgot to include the countering.

randovania/games/dread/json_data/header.txt
XenoWars: Oh, oops. i had copied this from the red chozo-x template initially

randovania/games/dread/json_data/header.txt
duncathan: in that case, use both :+1:

randovania/games/dread/json_data/header.txt
duncathan: not sure. this one isn't a huge deal

duncathan:
> > also, if any of these require more ice missiles than super missiles to kill, use the ice missile ammo count for both ice and supers and mention that in a comment
> 
> The only case that is a thing is Chozo-X. Why not list them individually?

for it to be accurate, the supers requirement would need to be `Shoot Super Missile AND N Super Missiles AND No Ice Missile`, which would make collecting Ice Missiles a potentially dangerous action. I'd prefer being slightly generous with the ammo requirements rather than introducing a totally unnecessary dangerous action

duncathan (approved):

PR #3678, opened 2022-08-30
Fixed incorrect actors being randomized, released GIL

Fantaselion:
I fixed 4 actors being scaled that shouldn't of been scaled due to them having the same instance ID as another actor in another PAK (Ex: Two of the pistons in Research Entrance were being randomized instead of some Elite Pirate Actors in the tubes) I also released the GIL.

I was wanting to wait on this pull request until I got the callback thing done, but I ended up getting sick so its quite unlikely that I'll get that done before next month rolls around.

Idk why the 8 commits from my last pull request are showing up but since they don't change anything, I suppose it doesn't matter.

Also this won't make any previous test of this null or anything, cause if my changes broke anything, it would cause an exception.

### DIFF
--- a/randovania/games/dread/json_data/Ferenia.json
+++ b/randovania/games/dread/json_data/Ferenia.json
@@ -231,31 +231,14 @@
                                         "type": "resource",
                                         "data": {
                                             "type": "items",
-                                            "name": "Space",
+                                            "name": "Spin",
                                             "amount": 1,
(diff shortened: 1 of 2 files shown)


toasterparty:
image

You need to clean your branch of these kinds of commits before starting a new proposed change. See my guide pinned in `#dev-talk-general` for future PRs

PR #3705, opened 2022-09-06
Dread: Final Burenia Revision

### LINKED ISSUE #3558: Dread: Dairon Logic caused unresolavble softlock when coupled with Burenia Hub access
I've been playing the latest build of the Dread randomizer (4.6.0 dev52) and have run into an interesting conundrum that caused a permanent softlock that I didn't realize for a very long time. I spent over an hour after causing this softlock trying to find items, but was unsuccessful until I finally realized it.

Permalink to seed: UCFqDedv3R1N7b9RIvSu2MY2Y_VRYv4zOUqVeUlP9Ai (CQQEEIEL). 
Not sure if I need to provide further details about the rules but I effectively did the following: 12 Metroid DNA, Progressive everything except Bomb (so Bomb and Cross Bomb are separate), and the Metroid DNA can be obtained from anything, it doesn't prefer EMMIs nor bosses. I removed 3 Missile Tanks for the 3 extra DNA. X start released.

The problem comes from the fact that, according to the spoiler log, Bomb is accessible via the Yellow EMMI Introduction room which requires Bomb to access. Cross Bomb is in Hanubia which is inaccessible without Bomb or another collection of items I did not get (I checked every possible location). Main Power Bomb is in the Dairon Central Unit Access, which is the speed boost location that is only accessible after using the secondary entrance to Dairon from Burenia.

In Burenia, I started to go towards the access to Dairon, but destroyed one of the two water releases in the Burenia Hub. The other one cannot be destroyed without Bomb. I did this because a Metroid DNA was in the Missile Tank spot on the left water release, and I wanted to grab it while I was there. I had Gravity Suit, but I did not consider using it to get to the Dairon access without releasing the water until it was too late.

The problem comes from the fact that destroying one out of two water releases causes the bridge to lower just far enough to cause the access to Dairon to be inaccessible. With no way to destroy the second water release, I am softlocked and cannot proceed. I have Diffusion Beam but it cannot reach the release.

image
image

I checked the spoiler log for other ways to do this. The only other thing that might help is somehow going from Wide Beam -> Wave Beam, however 2/3 Progressive Beam pickups are in Ferenia and Ghavoran, in areas that are also inaccessible (cannot access that part of Ferenia and Ghavoran's spot requires Bomb).

I know there are other simple ways to softlock yourself currently, like entering a room where you get a major item without that major item, but almost all of them are resolvable with the Reset to Checkpoint. This one is a situation you can cause and not even notice for prolonged periods of time, and I think it's something that is worth looking in to, that's why I'm reporting it here. I believe the Dairon logic was just recently revamped and I'm not sure if this is an inadvertent result of those changes.

XenoWars:
This should finish off the logic revision for Burenia. Fixed a few things from the previous PR as well.

- Added various methods of traversing "Main Hub Tower Bottom"
- Added various methods of traversing "Gravity Suit Tower"
- Corrected some logic in "Gravity Suit Tower"
- Corrected some logic in "Gravity Suit Room"
- Corrected some logic in "Transport to Artaria"
- Fixed the connection between the bomb blocks in "Energy Recharge South"
- Accounted for the magnet wall event when choosing to grapple it to reach the item in "Main Hub Tower Middle"
- Added event for the Twin Robot Fight
- Added event for a speedboost in "Main Hub Tower Bottom"
- Corrected minor typo in the event long name for the Ferenia Twin Robot Fight
- Renamed "Allow Disabled Freezer" to "Highly Dangerous Logic", allowing for many forms of dangerous logic that is not entirely obvious to be covered
- Applied "Highly Dangerous Logic" to the connection from "Water Bottom" to "Door to Lower Transport to Dairon" in "Burenia Hub to Dairon"

fixes #3558

### DIFF
--- a/randovania/games/dread/json_data/Burenia.txt
+++ b/randovania/games/dread/json_data/Burenia.txt
@@ -1426,14 +1426,14 @@ Extra - polygon: [[-900.0, -7000.0], [-5000.0, -7000.0], [-5000.0, -10700.0], [-
   * Extra - right_shield_def: None
   > Alcove Across Grapple Block
       All of the following:
-          Morph Ball and After Burenia - Main Hub Bottom Blob
+          Morph Ball and After Burenia - Main Hub Bottom Blob and After Burenia - Main Hub Bottom Grapple Block
           Any of the following:
               Use Spin Boost
               Flash Shift and Gravity Suit
   > Ledge above Grapple Block
-      Gravity Suit and Morph Ball and After Burenia - Destroy Gravity Suit Floor
+      Gravity Suit and Morph Ball and After Burenia - Main Hub Bottom Grapple Block
   > Water Space Jump Platform
-      Morph Ball and After Burenia - Destroy Gravity Suit Floor
+      Morph Ball and After Burenia - Main Hub Bottom Grapple Block
 
 > Event - Main Hub Bottom Blob, Left; Heals? False
   * Layers: default
@@ -1445,7 +1445,7 @@ Extra - polygon: [[-900.0, -7000.0], [-5000.0, -7000.0], [-5000.0, -10700.0], [-
 
 > Event - Main Hub Bottom Grapple Block; Heals? False
   * Layers: default
-  * Event Burenia - Destroy Gravity Suit Floor
+  * Event Burenia - Main Hub Bottom Grapple Block
   * Extra - actor_name: grapplepulloff1x2_003
   * Extra - actor_def: actordef:actors/props/grapplepulloff1x2/charclasses/grapplepulloff1x2.bmsad
   > Alcove Across Grapple Block
@@ -1495,9 +1495,12 @@ Extra - polygon: [[-900.0, -7000.0], [-5000.0, -7000.0], [-5000.0, -10700.0], [-
       Gravity Suit
   > Door to Storm Missile Gate Room
       All of the following:
-          Morph Ball and After Burenia - Destroy Gravity Suit Floor
+          Morph Ball and After Burenia - Main Hub Bottom Grapple Block
           Any of the following:
-              After Burenia - Prepare Speedboost in Save Station or Use Spin Boost
+              Use Spin Boost
+              All of the following:
+                  Speed Booster and After Burenia - Prepare Speedboost in Save Station and Disabled Door Lock Randomizer
+                  Gravity Suit or Speedbooster Conservation (Intermediate)
               Flash Shift and Gravity Suit
   > Event - Main Hub Bottom Blob, Left
       Lay Bomb or Shoot Beam
@@ -1515,7 +1518,7 @@ Extra - polygon: [[-900.0, -7000.0], [-5000.0, -7000.0], [-5000.0, -10700.0], [-
           Gravity Suit
           Flash Shift or Simple IBJ or Use Spin Boost
   > Door to Storm Missile Gate Room
-      Morph Ball and After Burenia - Destroy Gravity Suit Floor
+      Morph Ball and After Burenia - Main Hub Bottom Grapple Block
   > Alcove Across Grapple Block
       All of the following:
           After Burenia - Main Hub Bottom Blob
@@ -1543,7 +1546,7 @@ Extra - polygon: [[-900.0, -7000.0], [-5000.0, -7000.0], [-5000.0, -10700.0], [-
       Horizontal Water Movement
   > Door to Storm Missile Gate Room
       All of the following:
-          Gravity Suit and Morph Ball and After Burenia - Destroy Gravity Suit Floor
+          Gravity Suit and Morph Ball and After Burenia - Main Hub Bottom Grapple Block
           Speed Booster or Single-wall Wall Jump (Beginner) or Simple IBJ or Use Spin Boost
   > Tile Group (SCREWATTACK)
       Trivial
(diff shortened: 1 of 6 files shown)


randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1422,27 +1424,31 @@ Extra - polygon: [[-900.0, -7000.0], [-5000.0, -7000.0], [-5000.0, -10700.0], [-
...
   * Extra - right_shield_entity: {EMPTY}
   * Extra - right_shield_def: None
-  > Alcove across grapple block
-      Morph Ball and After Burenia - grapplepulloff1x2_003 and Horizontal Water Movement
+  > Alcove Across Grapple Block
+      All of the following:
+          Morph Ball and After Burenia - Main Hub Bottom Blob
+          Any of the following:
+              Use Spin Boost
+              Flash Shift and Gravity Suit
```
Hugoshido, line 1432: Isn´t the grapple pulloff missing?

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1483,63 +1489,88 @@ Extra - polygon: [[-900.0, -7000.0], [-5000.0, -7000.0], [-5000.0, -10700.0], [-
...
 
 > Water Space Jump Platform; Heals? False
   * Layers: default
-  > Door to Save Station South Access (doorpowerpower_023)
+  > Door to Save Station South Access (Lower)
       Horizontal Water Movement
+  > Door to Storm Missile Gate Room
+      All of the following:
+          Gravity Suit and Morph Ball and After Burenia - Destroy Gravity Suit Floor
+          Speed Booster or Single-wall Wall Jump (Beginner) or Simple IBJ or Use Spin Boost
```
Hugoshido, line 1547: Flash shift also possible here. Don´t know if it would be redundant though

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1483,63 +1489,88 @@ Extra - polygon: [[-900.0, -7000.0], [-5000.0, -7000.0], [-5000.0, -10700.0], [-
...
   > Tile Group (SCREWATTACK)
       Trivial
+  > Alcove Across Grapple Block
+      All of the following:
+          Gravity Suit and After Burenia - Main Hub Bottom Blob
+          Flash Shift or Simple IBJ or Use Spin Boost
```
Hugoshido, line 1553: It´s possible to jump onto the ledge trivial. It´s pretty tight though. Would say Movement (Beginner)?

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -2422,11 +2431,11 @@ Extra - asset_id: collision_camera_021
   > Pickup (Missile+ Tank)
       Trivial
   > Tile Group (POWERBEAM) 1
-      Simple IBJ or Use Spin Boost
+      Trivial
```
Hugoshido, line 2434: I would argue that this is not really trivial. You have to stop the screw attack manually so that you don´t destroy the floor you want to stand on. Maybe beginner movement if you don´t have any Jump items.

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -2422,11 +2431,11 @@ Extra - asset_id: collision_camera_021
...
   > Tile Group (POWERBEAM) 2
-      Simple IBJ or Use Spin Boost
+      Trivial
```
Hugoshido, line 2436: Same as above

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -2422,11 +2431,11 @@ Extra - asset_id: collision_camera_021
...
   > Tile Group (SCREWATTACK) 4
-      Simple IBJ or Use Spin Boost
+      Trivial
```
Hugoshido, line 2438: Same as above

Hugoshido (commented):
Just some additions.

randovania/games/dread/json_data/Burenia.txt
XenoWars: oops

randovania/games/dread/json_data/Burenia.txt
XenoWars: it would be, cause it would have required being at the alcove to pull the grapple block, which you can just flash to there then to storm. also it seems events are a bit messed up.

randovania/games/dread/json_data/Burenia.txt
XenoWars: Irrelevant. That would lock going up this way behind a trick. Everything has to be doable trickless. You have SA no matter what here, so you would have to cancel regardless of what you do unless you ibj up which is also a trick. Speed would be locked behind conservation beginner. Even so, the only trickless way to get all the way up is with Space Jump anyway.

Hugoshido: Fair enough

randovania/games/dread/json_data/Burenia.txt
Hugoshido: I can record one

Hugoshido (approved):

randovania/games/dread/json_data/Burenia.txt
XenoWars: This will be added later if it is deemed possible as a movement trick

PR #3722, opened 2022-09-08
Dread: Fix up the logic around getting the Thermal Device item without Morph

XenoWars:
- Removed the event for opening the upper door
- Added a direct connection to the door if Door Rando is disabled


henriquegemignani:
Does this have a practical change for accessing this, or is it just cleanup?

XenoWars:
It's mostly just a cleanup, but does allow the item in logic without Morph Ball

henriquegemignani (approved):

PR #3723, opened 2022-09-09
Dread: Final Ferenia revision

Hugoshido:
Final review of Ferenia
- Looked over logic and searched for mistakes. 
- Changed all Event and Door names which still contained db names.


henriquegemignani (approved):

XenoWars:
Gonna want to remove the Weight nodes in Purple EMMI Introduction instead. Can just add a direct connection from the Center Platform to the Dock to Fan Room Access

PR #3727, opened 2022-09-09
Dread: Final Itorash revision

Hugoshido:
- Changed PB-Block event name

### DIFF
--- a/randovania/games/dread/json_data/Itorash.txt
+++ b/randovania/games/dread/json_data/Itorash.txt
@@ -109,19 +109,19 @@ Extra - asset_id: collision_camera_002
   * Extra - left_shield_def: None
   * Extra - right_shield_entity: {EMPTY}
   * Extra - right_shield_def: None
-  > Event - Breakable (block_pbcolumn_000)
+  > Event - Elevator to Raven Beak PB Block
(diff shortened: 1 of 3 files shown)


randovania/games/dread/json_data/Itorash.txt
```diff
@@ -109,19 +109,19 @@ Extra - asset_id: collision_camera_002
   * Extra - left_shield_def: None
   * Extra - right_shield_entity: {EMPTY}
   * Extra - right_shield_def: None
-  > Event - Breakable (block_pbcolumn_000)
+  > Event - Elevator to Raven Beak PB Block
```
henriquegemignani, line 112: "PB Block" is pretty meh, could just say machine or whatever it is we're breaking?
And don't even need room name, it's fairly obvious given that Itorash is a few rooms long.

henriquegemignani (changes requested):

randovania/games/dread/json_data/Itorash.txt
XenoWars: I always saw it as the ship's energy source, so Destroy Generator perhaps

henriquegemignani (approved):

PR #3729, opened 2022-09-09
Dread: Cataris Revision Part 2

XenoWars:
This revision goes over the entire EMMI Zone and up to the Artaria Transport, renaming events and other nodes, and adding other methods to trickless connection and some tricks. Pretty much every room is affected within the EMMI Zone and between Green EMMI Introduction Access and Transport to Artaria.

### DIFF
--- a/randovania/games/dread/json_data/Cataris.json
+++ b/randovania/games/dread/json_data/Cataris.json
@@ -147,7 +147,7 @@
                                 ]
                             }
                         },
-                        "Event - Missile Tank Blob through wall (breakablemag002)": {
+                        "Event - Artaria Transport Blob through Wall": {
                             "type": "template",
                             "data": "Shoot Diffusion or Wave"
                         },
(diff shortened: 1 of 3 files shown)


henriquegemignani:
You don't have to put the full event name as the node name, especially if part of the event name is the area name.

XenoWars:
Is it really a problem though? I did that for most of Burenia and Dairon already, and I personally feel it makes it easier to tell exactly what the event is at a glance, especially for those who dont know the rooms very well.

henriquegemignani:
> Is it really a problem though? I did that for most of Burenia and Dairon already, and I personally feel it makes it easier to tell exactly what the event is at a glance, especially for those who dont know the rooms very well.

The resource itself will have the room name, so requirements show it.
The node is always displayed with the area name in context. In particular, most people only ever event nodes when checking the area itself in the data visualizer. So having the node include the area name is just harmful by being redundant.

XenoWars:
I did all of them (I think) except the ones in Z-57 Heat Room West (Left) cause those blobs do need extra details to tell them apart

henriquegemignani (approved):

PR #3734, opened 2022-09-10
Dread: Final Trickless Cataris Revision

XenoWars:
Finishes the primarily trickless-focused revision for Cataris. A few rooms in the EMMI Zone were affected, and most rooms ranging from Underlava Puzzle Room 2 all the way to Teleport to Artaria (Blue), and all the way to the upper edge of the EMMI Zone from there.

- Added event in Underlava Puzzle Room 2 for breaking the speed blocks so that going between the two parts can be accounted for
- Added event for the trigger that reopens the door to Central Unit Access, allowing it logical to go back through
- Separated the First Tunnel Blob event into two to account for Diffusion/Wave not needing to be in the tunnel
- Deleted some unnecessary tile nodes
- Accounted for whether the player could have Varia or not when trudging through lava
- Accounted for the upper parts of Thermal Device Room North being heated without pressing the lava button
- Added other various methods of going through rooms

### DIFF
--- a/randovania/games/dread/json_data/Cataris.txt
+++ b/randovania/games/dread/json_data/Cataris.txt
@@ -800,7 +800,7 @@ Extra - asset_id: collision_camera_014
   > Door to Tall Magnet Walls Access
       Trivial
   > Tile Group (POWERBEAM)
-      Speed Booster
+      Speed Booster or Simple IBJ
 
 > Door to Energy Recharge Station (Charge); Heals? False
   * Layers: default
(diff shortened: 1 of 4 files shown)


randovania/games/dread/json_data/Cataris.txt
```diff
@@ -791,6 +799,8 @@ Extra - asset_id: collision_camera_014
   * Extra - right_shield_def: None
   > Door to Tall Magnet Walls Access
       Trivial
+  > Tile Group (POWERBEAM)
+      Speed Booster
```
Hugoshido, line 803: I know trickless.. But IBJ would work or not?

Hugoshido (changes requested):
Just 1 addition

randovania/games/dread/json_data/Cataris.txt
XenoWars: Yeah, it would. forgot

XenoWars: But that can be added once we go through adding the tricks

Hugoshido (approved):

randovania/games/dread/json_data/Cataris.txt
XenoWars: Added this now

PR #3774, opened 2022-09-16
Dread: Remove requirement to have grapple to obtain Elun access pickup

Tyranisaur:


XenoWars (changes requested):
Inaccurate. Grapple is required if you have not used the flipper. If it's simple enough to be trickless, you should change it to an Or listing either Grapple or (After Super Missile Rotatable and Shoot Super Missile) (something along that for the event name)

Tyranisaur:
Right I suppose that accuracy is required for things like elevator rando, different starting locations etc. Otherwise you normally need to flip the rotatable to have access to the super locked door anyway.

Also, if you open the super locked door from behind, does that break the super lock even without supers?

XenoWars:
> Also, if you open the super locked door from behind, does that break the super lock even without supers?

Well, the super lock is on both sides in rando. All door covers break on both sides when you break one of them.

XenoWars:
Might be possible to store a speed if the flipper isnt used?

randovania/games/dread/json_data/Ghavoran.txt
```diff
@@ -554,6 +554,8 @@ Extra - asset_id: collision_camera_007
   * Extra - right_shield_def: None
   > Door to Navigation Station
       After Ghavoran - Super Missile Rotatable
+  > Dock from Spider Magnet Elevator
+      Before Ghavoran - Super Missile Rotatable
```
XenoWars, line 558: Remove this. You can only access it after the flipper has been used. There's also no point in even having it as you can't go through it anyway

CHANGELOG.md
```diff
@@ -32,6 +32,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Some Grapple options to include Grapple Movement
 - Changed: Some Movement tricks to Climb Sloped Tunnels
 - Changed: Some Movement tricks to Skip Cross Bomb
+- Changed: Allow pickup in Ghavoran Elun Transport Access by charging speed via navigation room
+- Changed: Allow traversal through the bottom of Ghavoran Flipper Room without flipping the spinner
```
XenoWars, line 36: This is already in the logic, so take this out.

XenoWars (changes requested):

XenoWars:
Probably just capitalize morph ball in the changelog and it's good.

XenoWars (approved):

PR #3775, opened 2022-09-17
fixed tab order in ui for enemy attribute rando

Fantaselion:
when pressing tab in the 'enemy attribute randomizer' tab, it would be completely out of order, so now that's fixed.

### DIFF
--- a/randovania/gui/ui_files/preset_prime_enemy_stat_randomizer.ui
+++ b/randovania/gui/ui_files/preset_prime_enemy_stat_randomizer.ui
@@ -43,8 +43,8 @@
         <rect>
          <x>0</x>
          <y>0</y>
-         <width>670</width>
-         <height>423</height>
+         <width>653</width>
+         <height>554</height>
         </rect>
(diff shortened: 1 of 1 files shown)


henriquegemignani:
Did it actually fix it? I remember trying to fix this in the past and the only thing that mattered was the order of elements in the ui file

Fantaselion:
> Did it actually fix it? I remember trying to fix this in the past and the only thing that mattered was the order of elements in the ui file

Yea, after looking it up I found that if you go to 'edit - edit tab order' in QT Creator, then you can change the order that it will 'tab' to.

henriquegemignani:
> > Did it actually fix it? I remember trying to fix this in the past and the only thing that mattered was the order of elements in the ui file
> 
> Yea, after looking it up I found that if you go to 'edit - edit tab order' in QT Creator, then you can change the order that it will 'tab' to.

Yeah, I remember seeing that but I also remember it just not working 😅

PR #3784, opened 2022-09-20
Dread: Add Diffusion Abuse trick

XenoWars:
- Add the new Diffusion Abuse trick
- Apply it to pushing Wide Beam Blocks while next to them as intermediate
- Added the trick to activating the Cataris lava buttons from outside as beginner
- Added hypermode trick for the Artaria Wide Beam Block

### DIFF
--- a/randovania/games/dread/json_data/header.json
+++ b/randovania/games/dread/json_data/header.json
@@ -1148,7 +1148,7 @@
             },
             "DiffusionAbuse": {
                 "long_name": "Diffusion Abuse",
-                "description": "Using Diffusion Beam in certain situations can bypass the usual requirements for some objects.",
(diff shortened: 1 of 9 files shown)


randovania/games/dread/json_data/header.json
```diff
@@ -1406,17 +1411,46 @@
...
+                                        "type": "template",
+                                        "data": "Shoot Wide Beam"
+                                    }
+                                ]
+                            }
                         },
                         {
-                            "type": "template",
-                            "data": "Shoot Wide Beam"
+                            "type": "and",
```
Hugoshido, line 1435: Wouldn´t this mean that if you activate diffusion abuse, the resolver thinks you can use it on every Wide Beam Block? But we have blocks where it is impossible to use the trick

Hugoshido (commented):
Just one question / addition

PR #3786, opened 2022-09-20
Add trick for ADC and adjust SSC trick description

Isachu:
Adds a trick for Aim Down Clip. Description is as follows:
> A type of shine sink clip that allows you to clip through 2 tile thick floors.

Also adjusts the description for Shinesink Clip. It is now as follows:
> Shinesparking and firing the grapple beam in certain locations allows you to clip through 1 tile thick floors.

Includes no implementation of the new trick. This will be handled in future PRs.

### DIFF
--- a/randovania/games/dread/json_data/header.json
+++ b/randovania/games/dread/json_data/header.json
@@ -1145,6 +1145,11 @@
                 "long_name": "Short Boost",
                 "description": "Flash Shift can be manipulated to allow you to charge Speed Booster in a smaller area than intended.",
                 "extra": {}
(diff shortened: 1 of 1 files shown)


XenoWars (changes requested):
Minor punctuation nit-pick. 1 tile and 2 tile should have a hyphen between. 1-tile and 2-tile

XenoWars (approved):
Minor punctuation nit-pick. 1 tile and 2 tile should have a hyphen between. 1-tile and 2-tile

PR #3799, opened 2022-09-21
Fixed Space Jump Room Access lower block logic

henriquegemignani:


XenoWars:
For what it's worth, I vaguely recall being able to shoot the blocks before rolling into the tunnel

XenoWars (approved):

henriquegemignani:
While making the changes I thought it might be possible. But it's a pretty long tunnel, so maybe with wide, and I'd put behind some knowledge trick.

PR #3818, opened 2022-09-28
a couple quick dread db changes

duncathan:
see commit messages


XenoWars (approved):
If the speed boost to get the item is possible, i'll add it in later

PR #3820, opened 2022-09-28
Dread: Changed connections in Cataris / Thermal Device Room North

Hugoshido:
To fix an issue with the validator

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Allow pickup in Ghavoran Elun Transport Access by charging speed via navigation room
 - Changed: Help solver by adding Morph Ball requirment on connections to event to flip the spinner in Ghavoran Flipper Room
 - Changed: Shooting occluded objects requires at least Intermediate Knowledge
-- Changed: Connections in Cataris - Thermal Device Room North
+- Changed: Fixed the node connection from Cataris / Door to Dairon Transport Access - Past Magnet Floor to Cataris / Door to Dairon Transport Access - Pickup (Energy Tank)
 - Fixed: Accounted for whether the player could have Varia or not when trudging through lava
 - Fixed: Accounted for the upper parts of Thermal Device Room North being heated without pressing the lava button
 - Fixed: Ghavoran Orange backdoor properly connects to Above Pulse Radar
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Allow pickup in Ghavoran Elun Transport Access by charging speed via navigation room
 - Changed: Help solver by adding Morph Ball requirment on connections to event to flip the spinner in Ghavoran Flipper Room
 - Changed: Shooting occluded objects requires at least Intermediate Knowledge
+- Changed: Connections in Cataris - Thermal Device Room North
```
henriquegemignani, line 45: Please be more detailed on what changed

henriquegemignani (changes requested):

CHANGELOG.md
Hugoshido: Like this?

henriquegemignani: Not really. You have a "Changed" entry that starts with "Fixed".

But other than that, try to read the entry from someone who comes entirely from outside, with no clue of what's going on. Could even be someone who's not familiar with the database.
You don't need to be detailed on how the change was made, but it should be understandable without checking the db.

In this case, something like "Fixed: Reorganised connections in ... to improve the solver".

Hugoshido: Gotcha will change it accordingly

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Allow pickup in Ghavoran Elun Transport Access by charging speed via navigation room
 - Changed: Help solver by adding Morph Ball requirment on connections to event to flip the spinner in Ghavoran Flipper Room
 - Changed: Shooting occluded objects requires at least Intermediate Knowledge
+- Changed: Fixed the node connection from Cataris / Door to Dairon Transport Access - Past Magnet Floor to Cataris / Door to Dairon Transport Access - Pickup (Energy Tank)
```
XenoWars, line 45: I'm not sure how to word it, but for one should probably remove the fixed part. This isn't really "fixing" anything, just changing a connection. Can probably word it something like:

"Changed: Door to Total Recharge Station in Cataris - Thermal Device Room North connects directly to the pickup."

After all, the connection you changed is from Total Recharge, not Dairon Transport.

XenoWars (changes requested):
I'd see if you can add the changes I suggested in the discord to make collecting the event and pickup not dangerous.

PR #3824, opened 2022-09-30
Dread: Make navigating Thermal Device Room North non-dangerous

XenoWars:
- Forces collecting the Energy Tank pickup and Magnet Wall Thermal Device event before traveling to the Door to Dairon Transport Access, and thus the Final Thermal Device
- Adds the option of bypassing the above via Highly Dangerous Logic
- Renamed "Thermal Magnet Platform Lowered" to "Magnet Wall Lowered"
- Added connection from "Path to Kraid Entryway" to "Teleporter to Artaria" in "Teleport to Artaria (Red)"


XenoWars:
This will need the updated the CHANGELOG from the Highly Dangerous Logic PR

Hugoshido (approved):

PR #3870, opened 2022-10-22
Dread: Add Cross Bomb alternative for Flash Gates

XenoWars:


henriquegemignani:
Does this work for every single flash shift gate in the game?

XenoWars:
Every flash gate except the one at the top of Burenia, but one is not a door so it would not apply to it. There's also only 5 flash doors in the game surprisingly enough.

henriquegemignani:
Oh! It definitely should be Knowledge Beginner to use cross bombs.

XenoWars:
Is there a method in the editor to edit weaknesses?

dyceron:
You can just say Flash Shift or (Lay Cross Bomb and Knowledge (Beginner)).

Also, why are they called "Phase Shift" doors in the header? lol

henriquegemignani:
> You can just say Flash Shift or (Lay Cross Bomb and Knowledge (Beginner)).
> 
> Also, why are they called "Phase Shift" doors in the header? lol

Internal name for these doors.

PR #3884, opened 2022-10-31
Integrate the new Echoes patcher in Randovania

henriquegemignani:
image


henriquegemignani:
> Does inverted Aether require the new patcher? If so, if the new patcher is deselected and the inverted aether feature enabled, is that handled correctly?

It requires, yes. In the UI the checkbox is disabled, but in the export the inverted is just ignored.
Hm.. I should make it disable the inverted when unchecking new patcher in the UI.

randovania/games/prime2/exporter/patch_data_factory.py
```diff
@@ -565,8 +565,22 @@ def create_data(self) -> dict:
...
         _apply_translator_gate_patches(result["specific_patches"], self.configuration.elevators.mode)
 
+        if self.configuration.use_new_patcher:
+            result["new_patcher"] = self.new_patcher_configuration()
+
         return result
 
+    def new_patcher_configuration(self):
+        return {
+            "small_randomizations": {
```
toasterparty, line 575: As much as we'd like this new feature tested, I think it's weird that it's forcefully bundled with the new patcher.

The way you have it now, it is impossible for testers to do true "apples-to-apples" testing to unequivocally prove that the new patcher meets functional equivalence with the old. Right now it's like, green apples to red apples, and that introduces ambiguity as to the root cause of potential issues.

I think that the faster and the more confidently we can replace the old patcher with the new, the better.

henriquegemignani: The new patcher for now, and for a non-trivial time, includes only new features. The only code duplication is in being able to read/edit the game itself.

toasterparty: If it's only new features, then why have a separate checkbox for including the new patcher?

henriquegemignani: I wanted to be explicit when the new patcher is enabled, with an easy to disable in case it has issues. Could dynamically detect if any needed feature is enabled, but that's meh.

toasterparty (approved):

PR #3885, opened 2022-10-31
Zoom in and zoom out feature for data visualizer (data editor canvas)

ThanatosGit:
I was so annoyed by the fact that you need to resize the window / `QDockWidget`s to scale, that I just wrote a quick zoom in and zoom out via mouse wheel.


henriquegemignani (approved):

ThanatosGit:
Added another commit with a slider feature. It is a little bit tricky because of the different image sizes but I think this settings works for dread, super metroid and cave story.

randovania/gui/data_editor.py
```diff
@@ -145,6 +147,20 @@ def __init__(self, data: dict, data_path: Path | None, is_internal: bool, edit_m
                                           RandovaniaGame.METROID_PRIME_CORRUPTION}:
             self.area_view_dock.hide()
 
+        # values for dread and super metroid
+        min_zoom = 7
```
henriquegemignani, line 151: I'd prefer that these values were to come from the game itself, not don't random constants here.
Maybe they're actually per-world?

randovania/gui/ui_files/ui_data_editor.py
```diff
@@ -0,0 +1,261 @@
+# -*- coding: utf-8 -*-
```
henriquegemignani, line 1: This file is generated, shouldn't have been committed.

It's also been generated in the wrong path??

henriquegemignani (changes requested):
Please also make sure to ensure all tests pass.

randovania/gui/data_editor.py
ThanatosGit: Yeah, that's pretty dirty. The range for the scale variable really depends on the size / resolution of the images.
Where is a good place for these constants? The `extra` section in the JSONs? Like `Artaria.json` etc?
Than I could change it to be set by the `on_select_world` method.

randovania/gui/ui_files/ui_data_editor.py
ThanatosGit: Yikes! Sorry! Maybe was due to misconfiguration at the beginning. I should have thought again why this generated file contained changes and was under version control

randovania/gui/data_editor.py
henriquegemignani: Yeah, the extra section of each world is perfect.

henriquegemignani: There's already some logic to get the image per world or per area depending on some magical config. You could hook the same events to decide if these values would be per-world or per-area.

ThanatosGit: I now simply add the zoom from the slider to the already existing scale value. Not the best solution but still a little bit more comfort.

henriquegemignani (approved):

PR #3892, opened 2022-11-06
Feature/dread tilegroup patcher

### LINKED ISSUE #3697: Dread: Prevent softlock in Artaria - EMMI Zone Exit Northwest
By changing the blocks by the entrance of the tunnel to Speedbooster blocks, they'll no longer respawn. This has no logical consequences has the pickup is already blocked by speedbooster.

steven11sjf:
Implements Dread's new `tile_group_patches` API implemented in randovania/open-dread-rando#159. Allows the exporter to change breakable tile groups from one block type into another. It currently patches the beam blocks in `EMMI Zone Hub - Tunnel to EMMI Zone Exit Northwest` to use speedbooster blocks which prevents a possible softlock. 

This will remain a draft PR until the odr pull request is merged, since it will break the json schema without the patcher being updated.

Fixes #3697.


henriquegemignani:
New values to the result of `create_data()` needs to be added to the json used for comparison in the tests.

We also need to release a new odr version and update it here, not just merge the PR.

henriquegemignani (changes requested):
Could you please add a change log entry?

henriquegemignani (approved):

PR #3929, opened 2022-11-17
Dread door lock rando

duncathan:


steven11sjf:
Small thing but maybe change the name of sensor lock doors so it's more clear that they're phantom cloak doors? flash shift also uses "sensors" and I was confused for a second until i read the rest of the doors and figured it out.

steven11sjf:
Also, access open doors can be changed to as well. I'd suggest defaulting to off though since it doesn't really add anything other than not needing to shoot them. 

As a side note, the order of doors in the UI seems off. I'd either organize it based on type (ie Power, Wide, Plasma, Wave, Missile, Super, ...) or alphabetically.

duncathan:
> Small thing but maybe change the name of sensor lock doors so it's more clear that they're phantom cloak doors? flash shift also uses "sensors" and I was confused for a second until i read the rest of the doors and figured it out.

it's the official name

duncathan:
> Also, access open doors can be changed to as well. I'd suggest defaulting to off though since it doesn't really add anything other than not needing to shoot them.
> 
> As a side note, the order of doors in the UI seems off. I'd either organize it based on type (ie Power, Wide, Plasma, Wave, Missile, Super, ...) or alphabetically.

the order isn't actually related to any of these changes. that's something for another pr if there's concerns

PR #3937, opened 2022-11-20
Rename of callback functions to use same name pattern

ThanatosGit:
This should only be merged if the pull request in https://github.com/randovania/open-dread-rando/pull/174 is accepted.
Otherwise ODR and Randovania wouldn't be compatible.


ThanatosGit:
Added a commit which bumps ODR (and also MEDS as dependency).
Changed Changelog.

PR #3939, opened 2022-11-21
Better handle solver failures

henriquegemignani:


randovania/gui/lib/generation_failure_handling.py
```diff
@@ -38,25 +38,25 @@ async def handle_exception(self, exception: Exception, progress_update: Callable
...
+            QtWidgets.QMessageBox.Icon.Critical,
             "An error occurred while generating game",
-            str(exception), QtWidgets.QMessageBox.Ok, self.parent)
+            str(exception), QtWidgets.QMessageBox.StandardButton.Ok, self.parent)
         common_qt_lib.set_default_window_icon(box)
 
-        if isinstance(exception.source, UnableToGenerate):
-            box.label.setTextInteractionFlags(QtCore.Qt.TextSelectableByMouse)
+        if isinstance(exception, ImpossibleForSolver):
+            box.setText(box.text() + "\n\nRandovania sometimes generates games with insufficient Energy Tanks. "
```
duncathan, line 47: we should really find a different message for this

duncathan (approved):

randovania/gui/lib/generation_failure_handling.py
henriquegemignani: I don't think this if case happens anymore now. Two places uses `GenerationFailureHandling`: one handles this case especially, the other is multiworld.

PR #4016, opened 2022-12-28
Multiple start points per area

ThanatosGit:
Changed starting_location from Area(Identifier) to Node(Identifier).
Added an integrity_check to allow only one starting_location per area for all games because there is currently no patcher support. (I will add that to dread/odr later).
Added a second checkbox to allow setting the default_node for an area independent of the starting points. default_node can be removed if TeleporterNode is refactored.

To keep your sanity: The first commit are changes to the source, the second to the tests and last one only an update of the db.


ThanatosGit:
As always: The moment you hit enter, you question if something is not correct...
Is the integrity check actually needed? For ODR, multiple spawn points would already work if there is a start_point_actor_name and start_point_actor_def defined for the node. Can't speak for the other games but I guess that's one part of what a code review is for ;)

randovania/layout/preset_migration.py
```diff
@@ -718,6 +718,16 @@ def _migrate_v42(preset: dict) -> dict:
         preset = _update_default_dock_rando(preset)
     return preset
 
+def _migrate_v43(preset: dict) -> dict:
+    game = default_database.game_description_for(RandovaniaGame(preset["game"]))
```
henriquegemignani, line 722: It's best to not use the DB here, as it can change in the future and break the migration of old presets.

henriquegemignani (changes requested):

randovania/layout/preset_migration.py
ThanatosGit: Am I allowed to use `migration_data.json` here, too? I already bloated the json to find the start node name for the description migration.

henriquegemignani: Yeah, the purpose of that json is for schema migrations like this

ThanatosGit: Done. Did a small refactor so that both migrations can use the same function

ThanatosGit:
If there is anything else which should be done, please let me know. Otherwise I will resolve the conflicting file and prepare the pr to allow usage of this new feature in dread

test/layout/test_game_patches_serializer.py
```diff
@@ -109,7 +110,8 @@ def _patches_with_data(request, echoes_game_description, echoes_game_patches, ec
         teleporter: NodeIdentifier = request.param.get("elevator")
         patches = patches.assign_elevators([
             (game.world_list.get_teleporter_node(teleporter),
-             game.starting_location),
+             # TODO: Starting Locations are now NodeIdentifier not AreaIdentifier => was modified but can be refactored back if teleporter also uses Node instead of Area a.k.a default_node is removed
+             AreaIdentifier(game.starting_location.world_name, game.starting_location.area_name)),
```
henriquegemignani, line 114: just `game.starting_location.area_identifier`

henriquegemignani (changes requested):

ThanatosGit:
- Fixed the requested change
- Reverted DB for merge
- Merged Main
- Merged Main again because I forget to fetch upstream 😄
- Used main branch to export (in the meantime modfied) starting locations for SR and copied it to the migration json
- Updated DB
- Renamed "my" migration 43 to 44 because the name was taken in the meantime and updated the new test for "Fake Quad Arena"

At least locally all tests passed again

henriquegemignani (approved):

PR #4020, opened 2023-01-01
Dread: Remove WBJ from Ghavoran Map Station Access Secret

Tyranisaur:
This Water Bomb Jump appears to be impossible.


dyceron:
Yeah I'm not having any luck with Bombs. I bet it's possible with Cross Bombs however.

dyceron:
I will accept this PR, as it doesn't seem possible. It can be readded later.

PR #4034, opened 2023-01-09
fix generation bugs for major/minor split

duncathan:
seems to work? i feel like the code could be cleaner


duncathan:
remind me to changelog

henriquegemignani (changes requested):

randovania/games/dread/item_database/item-database.json
```diff
@@ -675,7 +675,9 @@
             "must_be_starting": false,
             "probability_offset": 0,
             "probability_multiplier": 1,
-            "extra": {}
+            "extra": {
+                "is_major_override": false
```
henriquegemignani, line 679: I see you abusing extra.
Please don't :p

randovania/generator/generator.py
```diff
@@ -1,7 +1,7 @@
 import asyncio
 import dataclasses
 from random import Random
-from typing import Callable
+from typing import Callable, Tuple
```
henriquegemignani, line 4: we can just use `tuple` in randovania, since we use 3.11 (feature is from 3.10 I believe?)

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #4052, opened 2023-01-20
Add Linux binary for liblzo2

flibitijibibo:
This binary was built for CentOS 7, and therefore supports all systems that support glibc 2.17 or newer. The additions follow the same layout as the macOS dylib, which uses a symlink to avoid duplication.

Only downside to bundling this is that the filename is shared with other systems (i.e. *BSD) so if someone is running this on FreeBSD or something, it should still look for your system copy, but my bad if not!


henriquegemignani:
Appreciated!

henriquegemignani (approved):

PR #4054, opened 2023-01-21
Add Flatpak builds

henriquegemignani:


tools/flatpak/io.github.randovania.Randovania.yml
```diff
@@ -0,0 +1,46 @@
...
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
+sdk: org.freedesktop.Sdk
+sdk-extensions: [ org.freedesktop.Sdk.Extension.mono6 ]
+finish-args:
+    - --device=dri          # UI will want hw acceleration
+    - --share=network       # Multiworld support
+    - --share=ipc           # Needed for X11 support
+    - --socket=wayland
+    - --socket=fallback-x11
```
flibitijibibo, line 11: Tested the latest artifact on Steam Deck and can confirm it works, with one added line:

```suggestion
    - --socket=fallback-x11
    - --filesystem=xdg-run/app/com.discordapp.Discord:create
```

This makes it so that the login system can detect Discord even when both are in Flatpak containers; we should probably document this somewhere in the README (maybe the UI too, when `/.flatpak` is detected?):

https://github.com/flathub/com.discordapp.Discord/wiki/Rich-Precense-(discord-rpc)

After this was done, was able to log in and generate possibly the first seed made on Deck hardware!

image

flibitijibibo (approved):
Still works with the latest tag, so I'll throw in a +1. I'd have to look over the [guidelines](https://github.com/flathub/flathub/wiki/App-Submission) again but I think the resulting files is enough to submit to flathub for at least a `stable` branch, not sure about `dev` branches though.

henriquegemignani:
Yeah, I just need to calmly read again precisely what needs to be done to submit to flathub. I did it the other day and got a bit lost of what exactly.

PR #4056, opened 2023-01-21
RequirementSet: simplify when amount differs, but item matches

henriquegemignani:


randovania/game_description/requirements/requirement_list.py
```diff
@@ -112,9 +112,17 @@ def union(self, other: RequirementList) -> RequirementList:
         return RequirementList(itertools.chain(self.values(), other.values()))
 
     def is_subset_of(self, requirement: RequirementList) -> bool:
-        if len(self._items) >= len(requirement._items):
+        if len(self._items) > len(requirement._items):
             return False
-        return all(key in requirement._items for key in self._items.keys())
+
+        if self._items == requirement._items:
+            return False
```
duncathan, line 119: which definition of "subset" are we using here? I would've thought this should be `True` but I may be misunderstanding

henriquegemignani: It was already strictly smaller for this function, as it was `False` when equal length.

Checking wikipedia, I'd say the better naming is "proper subset"

duncathan: 👍

duncathan (approved):

PR #4068, opened 2023-01-23
Dread - Remove Tilegroup Nodes from Elun

dyceron:


XenoWars (changes requested):
I would think the Dock to Gyroscope Room in the Spider Magnet Room is actually a door, so that should probably be changed.

To be consistent with the Ferenia one by QR and the one in Burenia, the Cross Bomb method for the fan should probably be changed to IBJ or the other two to Movement (which is how they were originally).

For all the power bomb connection in the maze rooms, i think i only saw max 2 for going through the whole thing. Is that enough to get through it all?

dyceron:
> I would think the Dock to Gyroscope Room in the Spider Magnet Room is actually a door, so that should probably be changed.
> 
> To be consistent with the Ferenia one by QR and the one in Burenia, the Cross Bomb method for the fan should probably be changed to IBJ or the other two to Movement (which is how they were originally).
> 
> For all the power bomb connection in the maze rooms, i think i only saw max 2 for going through the whole thing. Is that enough to get through it all?

1. Yeah I noticed it but wasn't sure. I can change it.
2. Seems like Movement might be a better choice for these fan skips? Not sure
3. 2 is definitely enough because of all the respawning enemies.

dyceron:
Actually ill bump the next room to 3 just to be safe

XenoWars (approved):

PR #4095, opened 2023-02-03
Added new 8x3 layout for auto tracker. Useful for marathon runs where…

Uncle-Reggie:
Added new 8x3 layout for auto tracker. Useful for marathon runs where horizontal screen real estate is scarce. Also added _alt versions of the Echoes Translator icons with the Luminoth symbol for the translator's color first letter (V, A, E, C) to the Pixel Icon pack and set as default. Old icons still exist in mp2 directory.

randovania_y3Mz57K0HA
randovania_KewB6gomgD


henriquegemignani (changes requested):
I'd prefer if you did not json-format the existing tracker files.

Also it seems you reverted the image change to Darkburst?

henriquegemignani (approved):

PR #4126, opened 2023-02-10
Bundle XDG assets into the Linux package

flibitijibibo:
Addressing feedback from https://github.com/flathub/flathub/pull/3896

Things I didn't check but should have:
- Confirm the paths in the tarball for the desktop/png files
- A nicer way to make the wrapper script


flibitijibibo:
Forgot to update build_flatpak.py, this has now been pushed.

henriquegemignani (approved):
Ideally we could add these xdg_assets only when building the binary for Linux, but this works for now.

PR #4127, opened 2023-02-10
Add xdg_assets to spec

flibitijibibo:
Put this in the root for now, so we don't have to fuss with the flatpak yml more. TODO: Make this Linux-only!

### DIFF
--- a/randovania.spec
+++ b/randovania.spec
@@ -1,4 +1,5 @@
 # -*- mode: python -*-
+import platform
 from PyInstaller.utils.hooks import copy_metadata
 
 import randovania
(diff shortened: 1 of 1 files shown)


henriquegemignani:
`.spec` files are python files, you can just do an `if is_linux`:

flibitijibibo:
Probably should have actually checked how Python does this... can you tell I'm not a scripts guy :P

Moved the xdg_assets line to a platform.system check.

henriquegemignani:
It's missing an `import platform` at the top.

flibitijibibo:
PyInstaller was giving me a bit of sass about the data type, this should be fixed now. (11 years and 80 games and it's today that I finally learn Python, woops)

henriquegemignani (approved):

PR #4154, opened 2023-02-22
Fix error when selecting / deselecting Credits in Elevator tab

ThanatosGit:
Fix for the regression from discord channel.
Summary:
Elevators can be shuffled to use any starting location. There is one corner case: You can activate a warp to the credits for MP1 and MP2. These areas in MP1 and MP2 have a `default_node` which is not a `valid_starting_location`. This means the logic doesn't find this node when builidng the internal nodes list and crashes when try to mark them as selected.
Setting `valid_starting_location` to true for these nodes is wrong because it will appear selectable in `Starting area` tab.

In the call to `nodes_by_areas_by_world_from_locations` the area is not present for `Starting area`.
In the call to `nodes_by_areas_by_world_from_locations` the area is present for `Elevators`.

I just added the check for the special case. I don't like the fact that it needs to be checked at 3 places. Once for building the list, twice when checking if all start nodes are selected (basically for doing the tristate checkbox).

Maybe it's  better to mark them as `valid_starting_location` and just hide them on the `Starting area` tab?

While debugging I just realised that the types were wrong on some parameters and it was only working because `NodeIdentifier` can be used as `AreaIdentifier`  due to the same names for certain methods.

### DIFF
--- a/randovania/gui/lib/node_list_helper.py
+++ b/randovania/gui/lib/node_list_helper.py
@@ -6,6 +6,7 @@ from PySide6 import QtWidgets, QtCore
 from randovania.game_description.game_description import GameDescription
 from randovania.game_description.world.area import Area
 from randovania.game_description.world.area_identifier import AreaIdentifier
+from randovania.game_description.world.node import Node
 from randovania.game_description.world.node_identifier import NodeIdentifier
 from randovania.game_description.world.world import World
 from randovania.patching.prime import elevators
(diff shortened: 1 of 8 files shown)


randovania/gui/lib/node_list_helper.py
```diff
@@ -80,8 +80,8 @@ def _on_check_world(c: QtWidgets.QCheckBox, _):
                 new_node_list = [
                     identifier
                     for area in areas if c.is_dark_world == area.in_dark_aether
-                    for node in nodes_by_area[area.name]
-                    if (identifier := NodeIdentifier(world_list.identifier_for_area(area), node)) in checks_for_node
+                    for node in nodes_by_area[self.game_description.world_list.identifier_for_area(area)]
```
ThanatosGit, line 83: Just use the temp variable `world_list `:
`for node in nodes_by_area[world_list.identifier_for_area(area)]`

randovania/layout/lib/teleporters.py
```diff
@@ -82,8 +82,9 @@ def _valid_teleporter_target(area: Area, game: RandovaniaGame):
 
 class TeleporterTargetList(location_list.LocationList):
     @classmethod
-    def areas_list(cls, game: RandovaniaGame):
-        return location_list.area_locations_with_filter(game, lambda area: _valid_teleporter_target(area, game))
+    def nodes_list(cls, game: RandovaniaGame):
+        temp_area_list = location_list.area_with_filter(game, lambda area: _valid_teleporter_target(area, game))
+        return location_list.node_locations_with_filter(game, lambda node: node.identifier.area_identifier in temp_area_list and temp_area_list[node.identifier.area_identifier].default_node == node.name)
```
ThanatosGit, line 87: This looks too complicated.
Check if there is another method to apply the `_valid_teleporter_target` logic to an area and only yield the `default_node` from it.

henriquegemignani (approved):

PR #4164, opened 2023-02-26
Add game dependent flag in game data to allow multiple start nodes per area

ThanatosGit:


test/game_description/test_integrity_check.py
```diff
@@ -27,7 +27,7 @@ def test_find_database_errors(game_enum: RandovaniaGame):
 
 def test_invalid_db():
     sample_data = {
-        "schema_version": 6,
+        "schema_version": 14,
```
ThanatosGit, line 30: If this test is also supposed to test migration than it should not be modified.
For the integrity check of multiple start nodes per area, we need an updated schema to set `valid_starting_location` per node and not per area.

henriquegemignani (approved):

PR #4165, opened 2023-02-26
Multiple start nodes for dread

ThanatosGit:
Based on https://github.com/randovania/randovania/pull/4164


randovania/games/dread/exporter/patch_data_factory.py
```diff
@@ -117,18 +120,52 @@ def _node_for(self, identifier: AreaIdentifier | NodeIdentifier) -> Node:
...
-        except KeyError as e:
-            raise self._key_error_for_node(node, e)
+            if "start_point_actor_name" in node.extra:
+                return {
+                    "scenario": level_name,
+                    "actor": node.extra["start_point_actor_name"],
+                }
+            else:
+                collision_camera_name = area.extra["asset_id"]
+                spawnpoint_key = f"{level_name}_{collision_camera_name}_{node.location.x}" + \
```
ThanatosGit, line 141: @ArcanoxDragon Was there a specific reason we use such a complicated key and not just the NodeIdentifier? Afaik something containing the world, area and node name should already be unique.

henriquegemignani: The actor we create can literally be a static string, `SP_Randovania`. We'll never need to create more than one starting point for one given seed.

ThanatosGit: This comes from a plando where we had one-way teleporter/transporter/elevator to custom points.
I could remove it and make it more simple or leave it for the opportunity to implement such a feature in the very far future.

ArcanoxDragon: I'd say keep it in; the option for arbitrary teleport points is very beneficial for plando reasons. I don't recall why I chose the particular key format but if it can be simplified, I'm all for it.

henriquegemignani: Then they're named based on the source teleporter at that point. Though could just make it SP_Randovania{i} with a helper that creates these.

ThanatosGit: Sorry, I really can't follow your last message. 
`_start_point_ref_for` is called for the target and not for the source. The purpose is to check and prevent creating multiple times the same start point for the same node. e.g. two teleporters have the same target or one teleporter and the start location referencing the same node etc.
https://github.com/randovania/randovania/blob/d400cc33aef9686f5bdd699c703d41d09679bfd2/randovania/games/dread/exporter/patch_data_factory.py#L425-L431

Currently they are named `SP_RDV_i` where `i` is replaced by three digits number. You want to change `spawnpoint_name_prefix` to be a format string and move the logic in `_start_point_ref_for` to a helper function? Otherwise I don't see a difference from your suggestion compared to the current state.
https://github.com/randovania/randovania/blob/d400cc33aef9686f5bdd699c703d41d09679bfd2/randovania/games/dread/exporter/patch_data_factory.py#L145

It somehow sounds like you misread something like that the spawn points are named by teleporter name or similiar.

henriquegemignani: One: you should have a helper method that takes a node, ensures a spawn point was created for it and returns the name.
Two: Why not just the node itself as a key for this?

henriquegemignani (changes requested):
Please add tests for the new cases added,particularly the failure cases.

randovania/games/dread/exporter/patch_data_factory.py
ThanatosGit: One: Alright.
Two: See above. That was my first message here to Arcanox because I didn't remember if there was a reason for such a complicated key.
Tests: Will be added for the gui and for the case you mentioned.

henriquegemignani: Yeah, if this key is used purely in python there's no reason it can't just use a python object as key. Preferable to use, even.

test/gui/preset_settings/test_starting_area.py
```diff
@@ -107,3 +106,66 @@ def test_check_credits(skip_qtbot, preset_manager):
 
     checkbox_list = window._starting_location_for_node
     assert checkbox_list.get(not_expected, None) == None
+
+@pytest.fixture
+def setup_and_teardown_for_multiple_nodes(preset_manager):
+    # TODO: Currently all areas have only one sp that's why the test is artificially constructed.
```
henriquegemignani, line 112: Add one such case to the Blank game and use that for the test.

henriquegemignani (changes requested):

randovania/games/dread/exporter/patch_data_factory.py
```diff
@@ -351,6 +388,10 @@ def _tilegroup_patches(self):
         ]
 
     def create_data(self) -> dict:
+        # reset spawnpoint variables
+        self.spawnpoint_id = 0
```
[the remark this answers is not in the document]
ThanatosGit: Oh, lol. Because what you wrote above: The variable is shared between all instances and exporting multiple seeds create(d) the wrong dict without resetting it. But this will be removed and is not needed when also moving `spawnpoint_id` and `new_spawn_points` into the `__init__` as you suggested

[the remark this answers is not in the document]
henriquegemignani: Suspected such, but wasn't sure :)

randovania/games/dread/exporter/patch_data_factory.py
```diff
@@ -117,18 +120,52 @@ def _node_for(self, identifier: AreaIdentifier | NodeIdentifier) -> Node:
...
+                        "scenario": level_name
+                    },
+                    "location": {
+                        "x": node.location.x,
+                        "y": node.location.y,
+                        "z": node.location.z
+                    },
+                    "collision_camera_name": collision_camera_name
+                }
+                self.spawnpoint_id += 1
```
henriquegemignani, line 148: Could just use `len(self.new_spawn_points)` instead.

henriquegemignani (changes requested):

test/games/dread/exporter/test_dread_patch_data_factory.py
```diff
@@ -167,3 +169,70 @@ def test_pickup_data_for_a_major(dread_game_description, preset_manager):
...
+
+@pytest.fixture
+def setup_and_teardown_for_custom_spawn():
+    # modify a node to be a valid start point without a vanilla spawn
+    game_desc = default_database.game_description_for(RandovaniaGame.METROID_DREAD)
+    world = game_desc.world_list.world_with_name("Artaria")
+    area = world.area_by_name("Charge Tutorial")
+    node = area.node_with_name("Start Point")
+    modified_node = dataclasses.replace(node, valid_starting_location=True)
+    area.nodes.remove(node)
```
henriquegemignani, line 214: Editing the shared GameDescription is such a scary thing, but I'm not sure the best way of handling this.

test/gui/preset_settings/test_starting_area.py
ThanatosGit: Rewrote the test. Added a node to blank game. Also added test/checks for correct behaviour of world box (therefore marked another node in another area as start point).
After the push it just came to my mind that it maybe was not necessary to add a node and instead use an already existing one.

test/games/dread/exporter/test_dread_patch_data_factory.py
ThanatosGit: I would suggest to just change/rewrite the test as soon as we have this in the db of dread.

randovania/games/blank/gui/preset_settings/__init__.py
```diff
@@ -16,4 +17,5 @@ def preset_tabs(editor: PresetEditor, window_manager: WindowManager):
         PresetItemPool,
         PresetBlankPatches,
         PresetDockRando,
+        PresetStartingArea
```
henriquegemignani, line 20: I would not add this tab unconditionally until it's stable :)

henriquegemignani (changes requested):

randovania/games/blank/gui/preset_settings/__init__.py
ThanatosGit: Is the blank game used for anything in stable?

henriquegemignani: I'm blind. Thought I read Dread 🤣

henriquegemignani (approved):

PR #4198, opened 2023-03-05
Resolver: Optimize solve by using a heuristic when choosing between actions

Tyranisaur:


randovania/resolver/resolver.py
```diff
@@ -64,24 +83,14 @@ def _should_check_if_action_is_safe(state: State, action: ResourceNode, dangerou
...
-    else:
-        pickup_node = action
-
-    if isinstance(pickup_node, PickupNode):
-        target = state.patches.pickup_assignment.get(pickup_node.pickup_index)
-        if target is not None and (target.pickup.item_category.is_major or target.pickup.item_category.is_key):
-            return True
-
-    return False
+    return _is_major_or_key_pickup_node(action, state)
```
duncathan, line 93: this whole function could be written like this:
```py
def _should_check_if_action_is_safe(state: State, action: ResourceNode, dangerous_resources: frozenset[ResourceInfo]) -> bool:
    """
    Determines if we should _check_ if the given action is safe that state
    :param state:
    :param action:
    :param dangerous_resources:
    :return:
    """
    return (not _is_action_dangerous(state, action, dangerous_resources) or isinstance(action, EventNode) or _is_major_or_key_pickup_node(action, state)
```

randovania/resolver/resolver.py
```diff
@@ -160,8 +171,19 @@ async def _inner_advance_depth(state: State,
...
                 # If a safe node was a dead end, we're certainly a dead end as well
                 return new_result
+        if _is_action_dangerous(state, action, logic.game.dangerous_resources):
+            dangerous_actions.append((action, energy))
+        elif _is_major_or_key_pickup_node(action, state):
+            major_pickup_actions.append((action, energy))
+        elif isinstance(action, DockLockNode) or isinstance(action, EventNode) or isinstance(action, EventPickupNode):
+            lock_actions.append((action, energy))
+        else:
+            rest_of_actions.append((action, energy))
```
duncathan, line 181:
```py
action_tuple = (action, energy)
if _is_action_dangerous(state, action, logic.game.dangerous_resources):
    dangerous_actions.append(action_tuple)
elif _is_major_or_key_pickup_node(action, state):
    major_pickup_actions.append(action_tuple)
elif isinstance(action, (DockLockNode, EventNode, EventPickupNode)):
    lock_actions.append(action_tuple)
else:
    rest_of_actions.append(action_tuple)
```

duncathan (dismissed):

randovania/resolver/resolver.py
```diff
@@ -56,34 +58,39 @@ def _simplify_additional_requirement_set(requirements: RequirementSet,
...
+                                    ) -> bool:
+    """
+    Determines if we should _check_ if the given action is safe that state
+    :param state:
+    :param action:
+    :param dangerous_resources:
+    :return:
+    """
+    return (not _is_action_dangerous(state, action, dangerous_resources)
+            or isinstance(action, EventNode)
```
henriquegemignani, line 90: This raises the question. Why are events always safe? There are definitely dangerous events. It's the most common in particular.

henriquegemignani: Oh I see. The old did this. But this new version is logically different!

Tyranisaur: I suppose there are two definitions of safe; one being that the actions not considered dangerous as per the `_is_action_dangerous`  function, and the other includes all the checks that are being done, including the part that involves a `potential_reach`. So previously, due to the bug I'm fixing in 9b416aeea4f7ab29359138d7c3318508200a5ce6, all event actions were considered safe according to the first definition, so the only ones that were considered unsafe are the ones that would be covered by the second definition, which has nothing to do with the event itself, and instead has everything to do with the event being the first action inside a softlock area.

Now that I'm fixing said bug, any actions that are dangerous are correctly defined as unsafe as per the first definition, so by all accounts, not all events are safe.

Side note, I figured that cf5bed2ce45d395e604e10b44afc70d232c7b4b2 provides a more idiomatic way to resolve the bug.

henriquegemignani: Oh, so the old if was completely broken. And checking back, since always that check was added. Neato.
Also I did some testing and unpacking the tuple is faster than indexing it. That was very surprising.

There's probably a better name for `_is_action_dangerous` but that's more picky-ness.

henriquegemignani (approved):

Tyranisaur:
Sorry about stopping the auto merge, but I just realized that there's something broken somewhere within the two last commits that makes it no longer solve a seed that it solved on the third to last commit.

Tyranisaur:
I figured it out, turns out it was the part you commented on, where you correctly pointed out that all events were considered safe in that version, because at that point you need to check for `not dangerous` *and*  the other stuff, not or. If an action is dangerous, then it is by definition not safe, and there's no point in doing any further checks.

PR #4234, opened 2023-03-29
Dread April Fool's 2023

ArcanoxDragon:
This is ridiculously goofy, but if it's sane enough (and _soon_ enough) to be accepted, I'd be honored 😅

It simply replaces the item location template used for Dread's hint messages with a reference to the ["it goes in the square hole" meme](https://www.youtube.com/watch?v=Nz8ssH7LiB0) on April Fool's Day. The hints read something along the lines of `Can you guess where [the item] goes? That's right! It goes in the [area] hole!`

I didn't bother adding a cosmetic toggle, but I could probably get that done pretty quickly if desired.


henriquegemignani:
I'd prefer this via a preset setting, with a different April fools preset and bobs of the date checking.

In particular the date checking might cause tests to fail that day and be unable to release :p

ArcanoxDragon:
This now requires randovania/open-dread-rando#201 so the negative missiles don't cause a Lua crash

henriquegemignani:
Since we sadly missed the boat for having this out for the actual April fools, what do you want to do? I'm not opposed to merging all of this (even the april fools hints), but the negative ammo thing will need an adjustment: add a new field to `Ammo` that says if it supports being negative. I know prime 1/2 works, but I don't think it'd work for Cave Story and maybe not other games.

duncathan:
> I don't think it'd work for Cave Story

would need to test, depending how its set up I might actually already have it configured in such a way it'd work. negative numbers in CS are cursed tho

ArcanoxDragon:
> Since we sadly missed the boat for having this out for the actual April fools, what do you want to do? I'm not opposed to merging all of this (even the april fools hints), but the negative ammo thing will need an adjustment: add a new field to `Ammo` that says if it supports being negative. I know prime 1/2 works, but I don't think it'd work for Cave Story and maybe not other games.

Ah, that was a pretty big oversight I didn't think of 😅 (this started as a very temporary inside joke, and then myself and those helping me test it decided it was hilarious enough to keep)

I can certainly add a flag that can be specified in the game headers as to whether or not each ammo type supports negative quantities.

randovania/layout/base/ammo_state.py
```diff
@@ -21,7 +21,8 @@ def check_consistency(self, ammo: Ammo):
 
         for count, ammo_index in zip(self.ammo_count, ammo.items):
             ammo_item = db.get_item(ammo_index)
-            if not (0 <= count <= ammo_item.max_capacity):
+            minimum_count = -ammo_item.max_capacity if ammo.allows_negative else 0
+            if not (minimum_count <= count <= ammo_item.max_capacity):
                 raise ValueError(f"Ammo count for item {ammo_index} of value {count} is not "
                                  f"in range [0, {ammo_item.max_capacity}].")
```
henriquegemignani, line 27: incorrect range :)

randovania/layout/base/ammo_state.py
```diff
@@ -35,9 +36,10 @@ def bit_pack_encode(self, metadata) -> Iterator[tuple[int, int]]:
         for count, ammo_index in zip(self.ammo_count, ammo.items):
             ammo_item = db.get_item(ammo_index)
             yield from bitpacking.encode_int_with_limits(
-                count,
+                abs(count),
                 (ammo_item.max_capacity // 2, ammo_item.max_capacity + 1),
             )
+            yield from bitpacking.encode_bool(count < 0) # Negative?
```
henriquegemignani, line 42: this is only needed if `ammo.allows_negative`

randovania/layout/base/ammo_state.py
```diff
@@ -52,12 +54,13 @@ def bit_pack_unpack(cls, decoder: BitPackDecoder, metadata) -> "AmmoState":
...
-            ammo_count.append(
-                bitpacking.decode_int_with_limits(
-                    decoder,
-                    (ammo_item.max_capacity // 2, ammo_item.max_capacity + 1),
-                )
+            count = bitpacking.decode_int_with_limits(
+                decoder,
+                (ammo_item.max_capacity // 2, ammo_item.max_capacity + 1),
             )
+            if bitpacking.decode_bool(decoder): # Negative?
```
henriquegemignani, line 61: same, add a `ammo.allows_negative and ` here

henriquegemignani (changes requested):
Please add the changelog entries!

henriquegemignani (changes requested):
This is missing changelog entries!

There should be:
- One for Dread, regarding support for negative expansions.
- One for Dread, regarding the April Fools preset.
- One for Prime 1, regarding the negative expansions.
- One for Prime 2, regarding the negative expansions.

henriquegemignani:
You do get the reward of changelog conflicts! :)

henriquegemignani (approved):

PR #4239, opened 2023-04-04
Added ice missile and diffusion beam shields

steven11sjf:
Configured the database to add ice missile and diffusion beam shields as doors to be changed to. 

Sensor lock doors could also be added but since they are larger sizes I would suggest verifying which doors can be replaced with them in the database.


henriquegemignani (approved):

henriquegemignani:
conflict :)

PR #4256, opened 2023-04-10
Prime 1/2: Check if ISO is of the correct game

henriquegemignani:


toasterparty (approved):

randovania/games/common/prime_family/gui/export_validator.py
```diff
@@ -0,0 +1,47 @@
...
+
+    return header.game_id.decode("UTF-8"), header.game_title.split(b"\x00")[0].decode("UTF-8")
+
+
+def is_prime1_iso_validator(file: Path | None) -> bool:
+    """Returns False when the given path exists and is either not an ISO or a Prime 1 ISO, True otherwise"""
+    if is_file_validator(file):
+        return True
+
+    if file.suffix.lower() == ".iso":
```
duncathan, line 29: is there a reason this line isn't in the echoes verifier?

randovania/games/common/prime_family/gui/export_validator.py
```diff
@@ -0,0 +1,47 @@
...
+        header: nod.DolHeader = data_partition.get_header()
+    except RuntimeError as e:
+        logging.warning("Unable to parse %s: %s", str(iso_path), str(e))
+        return None
+
+    return header.game_id.decode("UTF-8"), header.game_title.split(b"\x00")[0].decode("UTF-8")
+
+
+def is_prime1_iso_validator(file: Path | None) -> bool:
+    """Returns False when the given path exists and is either not an ISO or a Prime 1 ISO, True otherwise"""
```
duncathan, line 25: is this understanding correct? `Returns True if the given path either does not exist, or is a Prime 1 ISO; False otherwise`? if that's correct it would certainly prove easier (at least for mii) to read if it were written that way

randovania/games/common/prime_family/gui/export_validator.py
henriquegemignani: Echoes only supports iso, but randomprime supports a few extra formats that aren't supported by nod.

randovania/games/common/prime_family/gui/export_validator.py
henriquegemignani: I really should change these to be `True` when it passes validation, other than not >.>

The validator succeds (aka False result) if the file is:
- exists, but not an iso
- exists, is ISO and Prime 1

else, fails (aka True)

randovania/games/common/prime_family/gui/export_validator.py
duncathan: might wanna add a comment mentioning that limitation then

duncathan (approved):

PR #4258, opened 2023-04-11
Update pyside6 to 6.5.2

henriquegemignani:


henriquegemignani:
Seems related to https://bugreports.qt.io/browse/QTBUG-114635

PR #4270, opened 2023-04-12
Dread/burenia shinesparks

Tyranisaur:


randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1391,6 +1393,8 @@ Extra - asset_id: collision_camera_010
 
 > Alcove Across Grapple Block; Heals? False
   * Layers: default
+  > Door to Energy Recharge South
+      Speed Booster and After Burenia - Prepare Speedboost in Save Station and After Burenia - Main Hub Bottom Blob and Speedbooster Conservation (Beginner)
```
XenoWars, line 1397: So the reason i put conservation as an alternative is that if you dont have gravity, you cannot retry. I would make this Gravity or conservation, but i do agree this one can be beginner. The other one you have to pull the block then spark within the timer while the enemy is there, so that's why that one is intermediate

XenoWars (changes requested):

randovania/games/dread/json_data/Burenia.txt
Tyranisaur: I added using shinespark with only gravity from above the screw attack blocks, so I don't think there's any reason for this one to have an alternative with gravity.

CHANGELOG.md
```diff
@@ -38,6 +38,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Added: Use Flash Shift to go right after getting the pickup in Artaria EMMI Zone Spinner.
 - Added: Use Flash Shift and Slide Jump to go from Artaria White EMMMI Arena to the top door to EMMI Zone Spinner.
+- Added: Use a Shinespark to climb up from Above Screw Attack Blocks in Burenia Main Hub Tower Bottom with only Gravity Suit.
+- Added: Use a Shinespark to climb up from Alcove Across Grapple Block in Burenia Main Hub Tower Bottom with only Speedbooster using Speedbooster Conservation Beginner.
+- Added: Use a Shinespark with Gravity Suit to reach Ammo Recharge South at the bottom of Burenia Gravity Suit Tower before the Destroy Gravity Suit Floor event.
+- Added: Use Spin Boost And Gravity Suit with different trick strategies to cross the big gap in Burenia Main Hub Tower Middle.
+- Added: Use a Shinespark with Gravity Suit to reach the Spider Magnet wall in Burenia Main Hub Tower Middle from the bottom of the room.
+- Added: Climb up to the Charge Beam Door in Burenia Main Hub Tower Middle using Gravity Suit and either Spin Boost or Flash Shift.
```
XenoWars, line 46: This is already a thing, connected from the lower tunnel to that room. I'm not sure how Flash can get to it though.

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1215,7 +1225,20 @@ Extra - asset_id: collision_camera_010
...
+              Any of the following:
+                  All of the following:
+                      # https://youtu.be/EKucxlQ6XtE
+                      Slide and Slide Jump (Intermediate)
+                  All of the following:
+                      # https://youtu.be/rb01nTDYOXk
+                      Movement (Intermediate)
+                  All of the following:
+                      # https://youtu.be/wlCOhr95pAY
+                      Movement (Beginner) and Walljump (Beginner)
```
XenoWars, line 1241: i personally dont think the wall jump trick here is necessary. or since the difficulty is from performing the wall jump, maybe the movement shouldnt be here. I just dont think it should require both.

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -1224,7 +1247,20 @@ Extra - asset_id: collision_camera_010
...
```
XenoWars, line 1263: Same reasoning here.

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -2153,7 +2193,9 @@ Extra - asset_id: collision_camera_021
               Use Spin Boost
               All of the following:
                   Gravity Suit
-                  Flash Shift or Simple IBJ
+                  Any of the following:
+                      Flash Shift or Simple IBJ
+                      Speed Booster and Disabled Door Lock Randomizer
```
XenoWars, line 2198: As i stated in the Discord, placing an event in the upper part of Gravity Suit Room would also be perfectly safe to require here as well. "Disabled Door Lock Randomizer or Gravity Suit Shinespark"

XenoWars (changes requested):

CHANGELOG.md
Tyranisaur: Ok, I'll move the flash shift to that connection then. The way it works is that you can simply walljump and then flash onto the ledge 
image

randovania/games/dread/json_data/Burenia.txt
Tyranisaur: What does placing an event in the upper part of gravity suit room do for your ability to get across the bottom, into the recharge station?

XenoWars: it lets you charge a speed from gravity suit room even with door rando. I've already explained this in the discord. You break the pb blocks, go into the room, speed, spark. Whether you go in or start in it, it will be safe. Should also add it to this connection.
image

randovania/games/dread/json_data/Burenia.txt
Tyranisaur: I removed movement from this and increased the walljump to intermediate.

randovania/games/dread/json_data/Burenia.txt
```diff
@@ -2310,7 +2314,9 @@ Extra - asset_id: collision_camera_021
           Gravity Suit and After Burenia - Destroy Gravity Suit Floor
           Any of the following:
               Space Jump or Simple IBJ
-              Speed Booster and Disabled Door Lock Randomizer
+              All of the following:
+                  Speed Booster
+                  After Burenia - Gravity Suit Room Store Shinespark or Disabled Door Lock Randomizer
```
XenoWars, line 2327: Thinking about it, this one doesnt really need the door lock disabled on it.

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added: Use a Shinespark with Gravity Suit to reach the Spider Magnet wall in Burenia Main Hub Tower Middle from the bottom of the room.
 - Added: Climb up to the Charge Beam Door in Burenia Main Hub Tower Middle using Gravity Suit and Flash Shift.
 - Added: Climb up from the Charge Beam Door in Burenia Main Hub Tower Middle using Gravity Suit, a Slide Jump, Spin Boost and a Wall Jump.
+- Added: Allow using Shinesparks in Gravity Suit Room by storing speed in the upper part of Gravity Suit Room.
```
XenoWars, line 48: This entry I don't think is needed, since it was already in logic to do so, just not in door rando.

XenoWars (changes requested):

randovania/games/dread/json_data/Burenia.txt
Tyranisaur: Right, because it's either storing it in the gravity suit room using the event, or if you can't access the top of the gravity suit room then you would do it from the lower part of gravity suit room, by breaking the power bomb blocks. But since we don't have an event there, we would need the "disable door lock rando" on that. And if we have that, the only thing that prevents us from storing speed in the upper part of the room again, is a missile cover.

So I guess it should be `[event] or disabled door locks rando and (missiles or lay power bomb)`

CHANGELOG.md
Tyranisaur: I guess entry should mention that it works also in door lock rando.

XenoWars (approved):

PR #4274, opened 2023-04-13
Add configuration option for enabling Raven Beak consistent damage table

ArcanoxDragon:
Myself and a number of other people I play with dislike the fact that Raven Beak's damage table is normalized so that the debuff on Wave/Ice is applied to other weapons as well. Most of us are speedrunners, and some of the speedrunning categories actually utilize the fact that Plasma/Supers and below deal more damage to RB and skip Wave/Ice, which makes the run _up to_ that point more challenging. When running rando seeds, it's not as enticing to "speedrun" a seed because the RB fight is agonizing with nerfed damage unless you go out of your way to find Wave/Ice.

I have added a preset option under Game Modifications > Other to turn off this normalization so that the RB fight behaves as it does in Vanilla. The preset option defaults to enabled, so new presets will continue to use the normalized damage tables unless the player decides to turn it off.


duncathan:
I won't be merging this pr, personally. If dark wants to he's welcome to though. I am strongly opposed to discouraging the player from picking up a major item such as wave or ice. It's not healthy rando design. I'd maybe be willing to merge a PR that buffs ice and wave to match their weaker counterparts, but I have no interest in reintroducing vanilla behavior

ArcanoxDragon:
It just seems too highly opinionated to not even give the player a choice. Some seeds will need those items for other reasons (doors or other locks), some might need one but not the other.

Having rando change vanilla boss damage logic that isn't related to randomizing the game has always seemed out-of-scope and I know several players who have repeatedly expressed frustration towards it...I don't see why giving the player a _choice_ is a problem.

henriquegemignani:
I'll all for being opionated in what the rando does. I also don't want this as an option because I want less options overall.

duncathan:
I have very limited interest in supporting options that serve only to introduce bad gameplay. the vanilla behavior makes ice missiles a strict downgrade from supers in 99% of seeds, and makes wave a substantial downgrade in many seeds. there's nothing interesting about that - you just never pick up ice, and avoid picking up wave unless you know for certain it's required for something. finding an item and choosing not to pick it up because doing so will always actively hurt you is the antithesis of randomizer gameplay.

compare to the patched behavior: I've won a race because I chose to collect a couple combat upgrades before fighting Raven Beak, while my opponent didn't and died to the boss. That's an *interesting* decision. There's legitimate reason to go either way. If my opponent had made a couple fewer mistakes during the fight, I'd have lost. On the other hand: I've won a race because I was willing to take the risk and fight Raven Beak with wide beam and regular missiles, while my opponent wasted ages looking for plasma or supers. In that case, I was rewarded heavily for choosing to "speedrun" the seed as you mentioned. 

Do you have any thoughts on my suggestion of buffing the strongest items to match the weaker ones? What about an alternative nerf for the weaker ones? Making them only *as* strong?

duncathan:
also, to address the comment on scope: the final stage of a progressive item being a downgrade from the penultimate stage is (as far as I'm concerned) a bug, and fixing that bug with progressive items is entirely within scope of the randomizer

ArcanoxDragon:
I think the only solution _I'd_ be satisfied with if the option to restore vanilla is off the table is for Wave and Ice to deal _at least_ as much damage as Plasma and Supers do in vanilla. If you want to incentivize people to collect all the items, make items better (not worse), especially since players who are familiar with speedrunning but not rando are going to be extremely disappointed to find out that Plasma and Supers deal barely any damage at all in rando.

We have a pretty large group of people in the Dread Speedrunning server who play rando, and we've had _multiple_ players expression confusion as to why Plasma/Supers aren't dealing enough damage to RB.

henriquegemignani:
And then quadruple the boss health so the fight isn't invalidated?

If confusion is your argument, then a setting won't help because they'd still not know to enable it.

duncathan:
> I think the only solution I'd be satisfied with if the option to restore vanilla is off the table is for Wave and Ice to deal at least as much damage as Plasma and Supers do in vanilla. If you want to incentivize people to collect all the items, make items better (not worse)

I would be fine with offering this as an option

henriquegemignani (changes requested):
Summing up the discussion we had on discord:

This PR is acceptable if you have the setting not visible by default in the customize preset window.
Ideally this would be via a toggle similar to the Simple/Medium/Advanced from the [VARIA randomizer](https://randommetroidsolver.pythonanywhere.com/randomizer).

ArcanoxDragon:
What was the final decision on adding a third option to buff Wave/Ice to do the same damage as Plasma/Supers? My ideal implementation of this would be the following:
- Presets default to the same thing they do today, which is nerfing all weapons the same as Wave/Ice
- An option is displayed in the UI to alter this such that Wave/Ice instead get buffed to _match_ the damage values of Plasma/Supers, respectively
- A third option is available only by editing the preset JSON that completely disables modifications to RB's damage table and restores Vanilla behavior

duncathan:
> What was the final decision on adding a third option to buff Wave/Ice to do the same damage as Plasma/Supers? My ideal implementation of this would be the following:
> 
> * Presets default to the same thing they do today, which is nerfing all weapons the same as Wave/Ice
> * An option is displayed in the UI to alter this such that Wave/Ice instead get buffed to _match_ the damage values of Plasma/Supers, respectively
> * A third option is available only by editing the preset JSON that completely disables modifications to RB's damage table and restores Vanilla behavior

I'm fine with this suggested implementation

randovania/games/dread/layout/dread_configuration.py
```diff
@@ -30,6 +50,7 @@ class DreadConfiguration(BaseConfiguration):
     hanubia_shortcut_no_grapple: bool
     hanubia_easier_path_to_itorash: bool
     x_starts_released: bool
+    consistent_raven_beak_damage_table: DreadRavenBeakDamageMode
```
henriquegemignani, line 53: Remove the `consistent_` from the variable name

randovania/games/dread/layout/dread_configuration.py
```diff
@@ -22,6 +23,25 @@ def unsupported_features(self):
         return []
 
 
+class DreadRavenBeakDamageMode(BitPackEnum, Enum):
+    DISABLED = "disabled"
```
henriquegemignani, line 27: `UNMODIFIED`

randovania/games/dread/layout/dread_configuration.py
```diff
@@ -22,6 +23,25 @@ def unsupported_features(self):
...
+    NERF_ALL = "nerf_all"
```
henriquegemignani, line 28: `CONSISTENT_LOW`

randovania/games/dread/layout/dread_configuration.py
```diff
@@ -22,6 +23,25 @@ def unsupported_features(self):
...
+    BUFF_FINAL = "buff_final"
```
henriquegemignani, line 29: `CONSISTENT_HIGH`

randovania/games/dread/layout/dread_configuration.py
```diff
@@ -22,6 +23,25 @@ def unsupported_features(self):
...
+
+    @property
+    def long_name(self) -> str:
+        if self == DreadRavenBeakDamageMode.DISABLED:
+            return "Vanilla damage values"
```
henriquegemignani, line 34: `Unmodified`

randovania/games/dread/layout/dread_configuration.py
```diff
@@ -22,6 +23,25 @@ def unsupported_features(self):
...
+        elif self == DreadRavenBeakDamageMode.NERF_ALL:
+            return "Nerf all weapons uniformly"
```
henriquegemignani, line 36: `Consistent, with reduction`

randovania/games/dread/layout/dread_configuration.py
```diff
@@ -22,6 +23,25 @@ def unsupported_features(self):
...
+        elif self == DreadRavenBeakDamageMode.BUFF_FINAL:
+            return "Buff Wave/Ice to match Plasma/Supers"
```
henriquegemignani, line 38: `Consistent, without reduction`

randovania/gui/ui_files/preset_dread_patches.ui
```diff
@@ -116,6 +116,32 @@
           </layout>
          </widget>
         </item>
+        <item>
+         <widget class="QGroupBox" name="bosses_group">
+          <property name="title">
+           <string>Bosses</string>
+          </property>
+          <layout class="QVBoxLayout" name="verticalLayout_3">
```
henriquegemignani, line 124: Name this layout too

randovania/gui/ui_files/preset_dread_patches.ui
```diff
@@ -116,6 +116,32 @@
...
+            <widget class="QCheckBox" name="consistent_raven_beak_damage_table_check">
+             <property name="text">
+              <string>Use alternate Raven Beak damage table</string>
+             </property>
+            </widget>
+           </item>
+           <item>
+            <widget class="QLabel" name="consistent_raven_beak_damage_table_label">
+             <property name="text">
+              <string>Instead of nerfing Plasma Beam, Super Missiles, and all prior weapons using the same "debuff" factor as Wave Beam/Ice Missiles, Wave Beam and Ice Missiles will have their debuff factors altered such that they do the same damage as Plasma Beam and Super Missiles do against Raven Beak in the vanilla game.</string>
```
henriquegemignani, line 135: `Normally, Raven Beak has a high resistance to only Wave Beam and Ice Missiles. This setting decides if that resistance should be removed, added to the previous weapons or left untouched.`

CHANGELOG.md
```diff
@@ -14,11 +14,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Major** - Added: Random Starting Locations is now supported. This enables all Save Stations, Navigation Stations, and Map Stations as possible starting options.
 - Added: New cosmetic option to display Randovania's area names on the HUD, either always or after room transitions
 - Added: Ice Missile Cover and Diffusion Beam Cover can now be added to door lock randomizer
+- Added: New option under "Game Modifications" to choose whether to use the vanilla Raven Beak damage table or the normalized damage table Randovania provides.
```
henriquegemignani, line 17: to choose how the inconsistency with Raven Beak's damage resistance is handled.

henriquegemignani (changes requested):
I very much dislike using buff/nerf anywhere. It's a jargon term and sentences in general tend to be better if you use simpler terms.

ArcanoxDragon:
randovania/open-dread-rando#210 will be needed for this now

duncathan (commented):
would you mind posting a screenshot of the UI for this setting?

ArcanoxDragon:
For reference, I just tested the exact damage values for each phase and each mode:
image

ArcanoxDragon:
> would you mind posting a screenshot of the UI for this setting?

Certainly :)

image

duncathan:
> For reference, I just tested the exact damage values for each phase and each mode: image

the numbers in consistent high seem strange to mii, but if it's what the people requesting this would like then that's fine

duncathan:
> For reference, I just tested the exact damage values for each phase and each mode: image

huh, is consistent high supposed to be so... inconsistent, lol

henriquegemignani (changes requested):
Can you add an additional rdvgame for testing purposes that uses a different value for this setting?

Try to add a bunch of weird settings there if you could, so we use it for that purpose in the future.

randovania/layout/preset_migration.py
```diff
@@ -802,6 +803,12 @@ def _migrate_v49(preset: dict) -> dict:
 
     return preset
 
+def _migrate_v50(preset: dict) -> dict:
+    if preset["game"] == "dread":
+        preset["configuration"]["raven_beak_damage_table_handling"] = DreadRavenBeakDamageMode.CONSISTENT_LOW
```
henriquegemignani, line 808: use pure string here instead of depending on any rdv API.
Also the enum is wrong here, curious why any test didn't fail.

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #4279, opened 2023-04-13
Rewrite to RemoteConnectorV2-API

ThanatosGit:
- removes the "heavily tied to prime games" logic from `ConnectionBackend`
- rewrite of the existing connector to RemoteConnectorV2

Nintendont still needs to be tested!

Quite unsure about the rewrite of the tests. They are now testing the nested objects and are more like integrationt tests of the connector within the whole connection backend.


henriquegemignani (approved):

henriquegemignani:
For a first step it looks fine-ish, and if it still works I'm ok.

duncathan:
oh right I was gonna test nintendont. it should be sufficient to just use the autotracker to test, yea?

ThanatosGit:
> it should be sufficient to just use the autotracker to test, yea?

afaik: yes. this should test all functions of the connector builder and executor. That should be the only difference.

We could also start another multiworld session.

EthanArmbrust:
Just tested out a Dolphin + Nintendont multiworld with this branch and everything was working correctly

PR #4292, opened 2023-04-15
Revamp how locations and pickups are major/minor

henriquegemignani:


duncathan (commented):
couple questions:
1) does the major_override serve any purpose any more, after this pr? I don't see any reference to it in the changes but I may be missing it. if it's no longer needed, we might as well remove it outright

2) is the TODO next to nothing items related to their weird state of being neither major nor minor? does it affect anything for them to be one way or the other atm?

henriquegemignani:
>     1. does the major_override serve any purpose any more, after this pr? I don't see any reference to it in the changes but I may be missing it. if it's no longer needed, we might as well remove it outright

I thought I removed all references to the major override. Did you find any left?

> 
>     2. is the TODO next to nothing items related to their weird state of being neither major nor minor? does it affect anything for them to be one way or the other atm?

Right now these are only created in places we only care about exporting the pickup, but I do plan to make the nothings a pickup. Though after a day thinking, nothings should be minor, which would then spill over to major since you'd have more minors than majors.

duncathan:
> Did you find any left?

nah, just wanted to make sure it was gone

> nothings should be minor, which would then spill over to major since you'd have more minors than majors.

I actually dislike the behavior of that implementation. it would mean that, with a complete pool of minors and an incomplete pool of majors, you can end up with a minor in a major location. imo majors and minors should only end up in the opposite location if there's an excess in their respective pool, whereas nothings are acceptable in either kind of location. perhaps they warrant a third category altogether with special handling?

henriquegemignani:
> > nothings should be minor, which would then spill over to major since you'd have more minors than majors.
> 
> I actually dislike the behavior of that implementation. it would mean that, with a complete pool of minors and an incomplete pool of majors, you can end up with a minor in a major location. imo majors and minors should only end up in the opposite location if there's an excess in their respective pool, whereas nothings are acceptable in either kind of location. perhaps they warrant a third category altogether with special handling?

Probably a better idea yeah. I explicitly designed it as an enum to allow for a third value :)

PR #4298, opened 2023-04-18
Added is in cinematic check when interacting with the game

UltiNaruto:
It now checks when the player is in a cutscene when handling multiworld interactions.

On Prime 1 and 2 it checks for the number of active instances of cameras to tell if you're in a cutscene. It's also what is used in metaforce to check if the player is in a cutscene.

https://git.axiodl.com/AxioDL/metaforce/src/branch/main/Runtime/Camera/CCameraManager.hpp#L79 :
`bool IsInCinematicCamera() const { return x4_cineCameras.size() != 0; }`


UltiNaruto:
Moved the check to game side rather than rdv side

randovania/patching/prime/all_prime_dol_patches.py
```diff
@@ -58,8 +58,29 @@ class BasePrimeDolVersion(DolVersion):
...
+
+    if game == RandovaniaGame.METROID_PRIME:
+        cutscene_check = [
+            lwz(r3, 0x870, r31),
+            cmpwi(r3, 0),
+            beq(0x10, True),
+            lwz(r0, 0x8, r3),
+            cmpwi(r0, 0),
+            ble(return_code_byte_count + 4, True),
+            *return_code,
```
henriquegemignani, line 73: The cutscene check is being added right after `intro`, which ends in `return_code`. We can reuse that same return block instead of duplicating it here.
In particular, the cutscene check should just be added to `intro`.

henriquegemignani (changes requested):

randovania/patching/prime/all_prime_dol_patches.py
UltiNaruto: so I put *return_code after *cutscene_check?

randovania/patching/prime/all_prime_dol_patches.py
```diff
@@ -78,13 +93,21 @@ def remote_execution_patch_start() -> list[BaseInstruction]:
...
 
     return [
         *intro,
 
+        # if camera count > 0 then we're in a cutscene so we return as we don't want
+        # to handle receiving items during a cutscene
+        *cutscene_check,
+        cmpwi(r0, 0),
+        ble(return_code_byte_count + 4, True),
+        *return_code,
```
henriquegemignani, line 109: This return code is still duplicated. Instead of having all the return block again, you could just jump to the previous return block.
(The if condition would need to be inverted though).

henriquegemignani (changes requested):

randovania/patching/prime/all_prime_dol_patches.py
UltiNaruto: I had to implement bgt

randovania/patching/prime/all_prime_dol_patches.py
```diff
@@ -78,13 +93,21 @@ def remote_execution_patch_start() -> list[BaseInstruction]:
         *return_code,
     ]
 
-    num_bytes_to_invalidate = _remote_execution_max_byte_count - assembler.byte_count(intro)
+    num_bytes_to_invalidate = _remote_execution_max_byte_count - (assembler.byte_count(cutscene_check) + 8 +
```
henriquegemignani, line 96: If you add the cutscene_check directly into `intro`, this line won't need to be changed

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #4301, opened 2023-04-20
Rename "lounge" to "Lounge" in Cave Story database

TomChapple:
I have been wanting to generate a Cave Story randomiser, but found that I was unable to do so on my Linux machine due to issue described in #4177: game exports failing due to different capitalisation between "lounge" in the database and "Lounge" in the file system of the game itself. I aim to resolve the bug described in this Issue without impacting other platforms.

The change is essentially a find and replace of "lounge" with "Lounge" in both the database and the test cases that verify its functionality. Once this change was completed, I was able to export Cave Story randomisers perfectly. In addition, I no longer experience a crash in Randovania when attempting to edit the database for that specific area (i.e. "Plantation" - "Rest Area") as I assume it was unable to find a resource due to a similar naming issue.

I have tested this by running the unit tests and running an exported randomiser. In particular, I made sure to arrive in "Rest Area" and verify that aspects like the music and interactions with other characters worked as intended. I found no issues during this time.

The unit tests are a little peculiar as they had an odd setup to begin with. The expected values contained two entries for this area in `test/games/cave_story/exporter/test_caver_patcher.py`: "lounge" and "Lounge". The former had pickup entries whereas the latter had the music definitions. My assumption in this change is that this is an erroneous setup and that the two areas should be merged. As such, there is now a single area "Lounge" that has the corresponding music and pickups combined for each test.

## Outstanding Tasks
- [ ] This has only been tested on my Linux machine using WINE 8.6 and has not been verified on other platforms just yet. As such, I may need some help verifying that this continues to work on Windows and Mac.
- [ ] I am also aware that there may be issues with how the executable interacts with the file being referred to as "Lounge" at all times. This is mentioned in #4177.
- [x] I have not updated the changelog just yet. This is mostly as a precaution in case that is a particular formatting or style I need to adhere to before this is merged. I aim to have something in place soon.
- [ ] I'm not sure how this change impacts existing permalinks or methods to replicate a seed. Any feedback on what impact this has would be appreciated and I can update the changelog appropriately.


henriquegemignani:
Appreciate the pull request and all the comments!
This strictly needs a review from @duncathan as it involves some game specific knowledge, but as he's rather busy right now it might take a while.

duncathan:
Yes, my apologies for the delay. I need to double check a few things on my PC to ensure this will work as expected with the game itself

duncathan (approved):
Great news - apparently I (or another contributor, I suppose) fixed this in the executable ages ago, so `Lounge` is indeed the correct case for rando. Thanks very much for the PR!

PR #4303, opened 2023-04-20
[Prime1+] Preset Configuration Cleanup

toasterparty:
Non-game specific hidden by default:
- generator dangerous action behavior
- The two in-dev generator actions (should we just absorb these as default?)
- local first progression

Prime hidden by default:
- Minimum artifact actions
- attribute rando
- chaos options

Removed from GUI:
- No Doors mod

*I couldn't figure out how to pipe a message from `_on_menu_action_experimental_settings()` to `PresetTabRoot.showEvent()`, which means that if a user changes the box with the preset editor open, they have to change tabs before the experimental settings update*


randovania/gui/preset_settings/customize_preset_dialog.py
```diff
@@ -61,7 +63,8 @@ def __init__(self, window_manager: WindowManager, editor: PresetEditor):
         self._tabs = {}
         self._current_tab = None
 
-        for extra_tab in self._tab_types:
+        for i in range(len(self._tab_types)):
```
henriquegemignani, line 66: ?


(and if you ever need the item and index, use `enumerate`)

randovania/interface_common/options.py
```diff
@@ -112,6 +112,7 @@ def from_json(cls, value: dict) -> "PerGameOptions":
     "advanced_timeout_during_generation": Serializer(identity, bool),
     "auto_save_spoiler": Serializer(identity, bool),
     "dark_mode": Serializer(identity, bool),
+    "experimental_settings": Serializer(identity, bool),
```
henriquegemignani, line 115: Add a new version/migration in `persisted_options`, just so newer rdv releases uses a different config file to older. (and then using an old version won't revert this setting randomly).

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #4310, opened 2023-04-23
Add new "upgrade" items for Metroid Dread

ArcanoxDragon:
This PR adds some new items to Dread to expand gameplay options. Corresponds to randovania/open-dread-rando#211.

Currently, the following new items are added:

A "Flash Shift Upgrade" item that increases the number of times you can "chain" a Flash Shift after the first time without needing to wait for Aeion to recharge
A "Speed Boost Upgrade" item that decreases the amount of time it takes for Speed Booster to charge up by 0.25s (for each item collected)


randovania/games/dread/item_database/item-database.json
```diff
@@ -709,6 +709,42 @@
...
+            "default_shuffled_count": 0,
+            "default_starting_count": 0,
+            "ammo": [],
+            "unlocks_ammo": false,
+            "hide_from_gui": false,
+            "must_be_starting": false,
+            "probability_offset": 0,
+            "probability_multiplier": 1,
+            "preferred_location_category": "major",
+            "extra": {}
```
duncathan, line 747: please add a `warning` field to each of these (which is admittedly somewhat poorly named) with the description of each one. for speed upgrade, please mention the minimum possible charge time of 0.25s as well

duncathan (changes requested):

ArcanoxDragon:
Will definitely require a new ODR build; randovania/open-dread-rando#214 is necessary for the "ammo" changes.

henriquegemignani:
Mapping as additional ammo given by Flash Shift means collecting that also gives you 2 upgrades, so if you place 2 Flash Shifts in the pool, the second one you find will also give two additional shifts. Feels like a very unintuitive behavior for those who add extra copies of upgrades, especially because that'd always happen.

I guess you tried it this way to allow the flash upgrades to give you a nerfed flash shift and also out of logic?

If having additional Flash Shifts have consequences, I'd go for merging it with the expansions and just having the first you find give full flash shift and the rest is just +1 shift.

duncathan:
there were three behaviors they had mentioned they wanted to be possible:
1) collecting main flash shift gives you flash shift and at least 2 chains, while collecting upgrades increases the number of chains without giving you flash shift
2) collecting main flash shift gives you flash shift and 2 chains, while collecting upgrades gives you flash shift and 1 chain
3) instead of having two separate pickups, there's a progressive flash shift pickup that provides 2 chains and flash shift when first collected, and adds 1 chain for subsequent pickups

using ammo was the best way I could come up with to model all of these behaviors, though I don't know if they've all actually been implemented in this pr

ArcanoxDragon:
All of that is implemented as describe, except for two differences:
- Main Flash Shift gives you _any_ configurable number of chains, including "0" (i.e. you must also collect upgrades to chain)
- Each subsequent Main Flash Shift gives you as many _more_ chains as the first one gave you to begin with. This is due to the way Ammo works in general; PBs are the same way.

randovania/games/dread/exporter/patch_data_factory.py
```diff
@@ -45,31 +45,34 @@ def get_item_id_for_item(item: ItemResourceInfo) -> str:
...
-        item_id = "ITEM_WEAPON_POWER_BOMB"
 
-    return {"item_id": item_id, "quantity": quantity}
+def get_resources_for_details(detail: ExportedPickupDetails) -> list[list[dict]]:
+    resources = [
+        list(convert_conditional_resource(conditional_resource))
+        for conditional_resource in detail.conditional_resources
+    ]
 
+    if (detail.original_pickup.resource_lock is not None and not detail.original_pickup.respects_lock
```
henriquegemignani, line 66: Doing `pickup = detail.original_pickup` probably makes this more readable.

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #4321, opened 2023-04-26
Prepare GameConnection and family to support multiple ConnectorBuilders

henriquegemignani:
Massive preparation for #4207.

This makes GameConnection a module that uses multiple ConnectionBuilder to get RemoteConnector instances. All logic for interacting with the games are now responsability of the game-specific ConnectionBuilder/RemoteConnector, including doing any kind of polling as needed.

To properly use multiple RemoteConnector at once it'll be necessary that these provide an UUID, which is patched into the game and is decided by the server. This is not implemented and everything uses the same dummy UUID and shall be implemented later. As a result, this explicitly prevents multiple ConnectionBuilder at once.

Regressions:
- Server isn't updated on your connection state
- Not sure if it's updated with your inventory
- Doesn't properly detect and/or prevent a player from playing the wrong game

image
image


duncathan:
We should probably make it clear to users that multiworld on the dev builds is going to be much more unstable than usual for the near future.

randovania/gui/game_session_window.py
```diff
@@ -897,7 +892,7 @@ async def update_multiworld_client_status(self):
             if not self.multiworld_client.is_active:
                 you = self._game_session.players[self.network_client.current_user.id]
                 session_id = f"{self._game_session.id}_{self._game_session.game_details.seed_hash}_{you.row}"
-                persist_path = self.network_client.server_data_path.joinpath(f"game_session_{session_id}.json")
+                persist_path = self.network_client.server_data_path.joinpath(f"game_session_{session_id}")
```
duncathan, line 895: any particular reason for dropping the file extension?

randovania/gui/multiworld_client.py
```diff
@@ -104,73 +94,93 @@ async def stop(self):
...
     def start_notify_collect_locations_task(self):
         if self._notify_task is not None and not self._notify_task.done():
             return
 
         self._notify_task = asyncio.create_task(self._notify_collect_locations())
 
-    async def on_location_collected(self, game: RandovaniaGame, location: PickupIndex):
-        if game != self._expected_game:
+    async def _on_new_state(self):
+        if self._all_data is None:
```
duncathan, line 143: could this be `if not self.is_active`?

randovania/gui/widgets/game_connection_window.py
```diff
@@ -0,0 +1,223 @@
...
+            action.triggered.connect(functools.partial(self.open_debug_connector_window, builder))
+            ui.menu.addAction(action)
+
+        ui.description.setText(builder.pretty_text)
+        self.ui_for_builder[builder] = ui
+
+        self.builders_layout.addWidget(ui.group)
+
+    def open_auto_tracker(self, builder: ConnectorBuilder):
+        pass
```
duncathan, line 194: TODO?

duncathan (changes requested):

randovania/game_connection/connector/debug_remote_connector.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self, game: RandovaniaGame):
+        super().__init__()
+        self._game = game
+
+    @property
+    def game_enum(self) -> RandovaniaGame:
+        return self._game
+
+    def description(self) -> str:
+        return f"{self.game_enum.long_name}"
```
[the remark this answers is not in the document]
henriquegemignani: Copy paste. Bit surprised ruff didn't complain

randovania/gui/game_session_window.py
henriquegemignani: This path is now a directory. Though this is changing again later, likely

duncathan (approved):

PR #4354, opened 2023-05-03
Add "Shuffled (avoid majors)" option to Location Pool settings

ArcanoxDragon:
This new option for locations in the Location Pool will exclude the location from containing progression items, but it will also try to avoid placing any item marked as "major" in the location unless there are no other options left.

### DIFF
--- a/randovania/generator/generator.py
+++ b/randovania/generator/generator.py
@@ -144,9 +144,19 @@ def _distribute_remaining_items(rng: Random,
     def assign_pickup(node_player: int, node: PickupNode, pickup_target: PickupTarget):
         assignments[node_player].append((node.pickup_index, pickup_target))
 
-    def assign_while_both_non_empty(nodes: list[tuple[int, PickupNode]], pickups: list[PickupTarget]):
+    def shuffle_assignment_lists(nodes: list[tuple[int, PickupNode]], pickups: list[PickupTarget]):
         rng.shuffle(nodes)
         rng.shuffle(pickups)
+
+        # after shuffling, sort both lists such that locations marked as "no majors" come first and major items come last
+        # this will attempt to prevent majors from being assigned to locations marked as "no majors", but allows for such
+        # assignments to happen if there aren't enough remaining locations
+        nodes.sort(key=lambda n: 0 if n in avoid_majors_pickup_nodes else 1)
+        pickups.sort(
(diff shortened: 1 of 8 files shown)


henriquegemignani:
Please do try to get coverage of the PR up to the target :)

randovania/generator/generator.py
```diff
@@ -140,10 +144,20 @@ def _distribute_remaining_items(rng: Random,
...
         assignments[node_player].append((node.pickup_index, pickup_target))
 
-    def assign_while_both_non_empty(nodes: list[tuple[int, PickupNode]], pickups: list[PickupTarget]):
+    def shuffle_assignment_lists(nodes: list[tuple[int, PickupNode]], pickups: list[PickupTarget]):
         rng.shuffle(nodes)
         rng.shuffle(pickups)
 
+        # after shuffling, sort both lists such that locations marked as "no majors" come first and major items come
+        # last this will attempt to prevent majors from being assigned to locations marked as "no majors", but allows
+        # for such assignments to happen if there aren't enough remaining locations
```
duncathan, line 153: probably better to make this a docstring for the function rather than a series of comments

randovania/layout/base/available_locations.py
```diff
@@ -37,38 +37,69 @@ def description(self) -> str:
...
+            "excluded_indices": self._sorted_excluded_indices,
+            "minor_only_indices": self._sorted_minor_only_indices,
         }
 
     @classmethod
     def from_json(cls, value: dict, game: RandovaniaGame) -> "AvailableLocationsConfiguration":
         return cls(
             randomization_mode=RandomizationMode(value["randomization_mode"]),
             excluded_indices=frozenset(PickupIndex(item) for item in value["excluded_indices"]),
+            minor_only_indices=frozenset(PickupIndex(item) for item in value.get("minor_only_indices", [])),
```
duncathan, line 85: this is slightly irresponsible. we shouldn't be using `get()` in this context - should instead write a preset migration and guarantee the key will be present

duncathan (changes requested):
I don't totally love that we're using two separate lists of nodes, now, but I'm not certain how best to address that

duncathan:
stale and conflicting. please open a new PR if you'd like to address the review comments!

PR #4356, opened 2023-05-03
Improve the base Preset describer

henriquegemignani:
Left is before, right is after
image
image
image

### DIFF
--- a/randovania/layout/preset_describer.py
+++ b/randovania/layout/preset_describer.py
@@ -1,5 +1,5 @@
 import collections
-from typing import Iterable, Sequence
+from collections.abc import Iterable, Sequence
 
 from randovania.game_description import default_database
 from randovania.game_description.pickup.standard_pickup import StandardPickupDefinition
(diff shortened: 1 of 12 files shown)


randovania/layout/preset_describer.py
```diff
@@ -131,16 +131,15 @@ def format_params(self, configuration: BaseConfiguration) -> dict[str, list[str]
         starting_locations = configuration.starting_location.locations
         if len(starting_locations) == 1:
             area = game_description.world_list.area_by_area_location(starting_locations[0])
-            starting_location = game_description.world_list.area_name(area)
+            starting_location = f"Starts at {game_description.world_list.area_name(area)}"
```
duncathan, line 134: we should probably have this support starting nodes in the future

randovania/layout/preset_describer.py
```diff
@@ -1,70 +1,70 @@
...
-
-
 PresetDescription = tuple[str, list[str]]
 
 
 class GamePresetDescriber:
-    def _calculate_starting_items(self, configuration: BaseConfiguration) -> list[str]:
-        expected_count = self.expected_starting_item_count(configuration)
-        starting_items = []
+    def _calculate_pickup_pool(self, configuration: BaseConfiguration) -> list[str]:
```
duncathan, line 21: it might be worthwhile to split the pool size into major and minor for the preset describer as well, when it's active

duncathan (approved):
my comments are probably out of scope for this PR but I wanted to get them written down at least

randovania/layout/preset_describer.py
henriquegemignani: oh I plan to do that. Thinking of changing `pool_creator.calculate_pool_pickup_count` to always return the two values for each `LocationCategory`.

randovania/layout/preset_describer.py
henriquegemignani: Oh. Didn't even notice. hm

PR #4357, opened 2023-05-03
Change calculate_pool_pickup_count to return the count per category

henriquegemignani:
image
image

### DIFF
--- a/randovania/generator/pickup_pool/pool_creator.py
+++ b/randovania/generator/pickup_pool/pool_creator.py
@@ -1,4 +1,5 @@
 from random import Random
+from typing import NamedTuple
 
 from randovania.game_description.game_description import GameDescription
 from randovania.game_description.game_patches import GamePatches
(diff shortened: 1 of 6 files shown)


randovania/generator/pickup_pool/pool_creator.py
```diff
@@ -55,43 +55,43 @@ def calculate_pool_results(layout_configuration: BaseConfiguration,
     return base_results
 
 
-def calculate_pool_pickup_count(layout: BaseConfiguration) -> tuple[int, int]:
+def calculate_pool_pickup_count(layout: BaseConfiguration) -> dict[LocationCategory | str, tuple[int, int]]:
```
duncathan, line 58: I think maybe this should be a simple `NamedTuple`. it's not obvious anywhere that index 0 is pickups and index 1 is locations

randovania/generator/pickup_pool/pool_creator.py
```diff
@@ -55,43 +55,43 @@ def calculate_pool_results(layout_configuration: BaseConfiguration,
...
+    }
+
+    for node in game_description.world_list.iterate_nodes():
+        if isinstance(node, PickupNode):
+            result[node.location_category][1] += 1
+
     pool_results = calculate_pool_results(layout, game_description,
                                           rng_required=False)
-    min_starting_pickups = layout.standard_pickup_configuration.minimum_random_starting_pickups
+    result["Starting"] = [0, layout.standard_pickup_configuration.minimum_random_starting_pickups]
```
duncathan, line 77: curious why we're assigning the number of nodes here, rather than number of pickups

duncathan (changes requested):

randovania/generator/pickup_pool/pool_creator.py
henriquegemignani: Minimum random starting pickups means there are more "places" that pickups can be placed at, so the pool can have more pickups in total. That's the same as having extra nodes.

duncathan (approved):

PR #4366, opened 2023-05-05
Allow presets to be placed at root, save order

henriquegemignani:


toasterparty:
Does this mean you can use CLI to generate from a non-default preset now?

henriquegemignani:
No, that's just because the CLI doesn't have a decent way of choosing which presets. Names aren't unique if you consider user presets. 
But mostly because it wasn't implemented :X

randovania/interface_common/options.py
```diff
@@ -459,6 +478,12 @@ def hidden_preset_uuids(self):
     def hidden_preset_uuids(self, value):
         self._edit_field("hidden_preset_uuids", value)
 
+    def get_preset_order_for(self, game: RandovaniaGame) -> list[uuid.UUID]:
+        return list(_return_with_default(getattr(self, f"preset_order_{game.value}"), list))
```
toasterparty, line 482: Isn't `list` uninitialized here?

henriquegemignani: You meant the second argument to `_return_with_default`? That function receives a callable that creates a default value. By passing `list` it can create a list when `preset_order_...` is None.

toasterparty (approved):

PR #4372, opened 2023-05-05
Add arguments for configuring where data is stored

henriquegemignani:


toasterparty (approved):
Yessssss! This will be a useful troubleshooting step for user issues that are resolved by "deleting your appdata folder"

PR #4373, opened 2023-05-05
Resover speedups

Tyranisaur:


randovania/resolver/resolver.py
```diff
@@ -24,15 +24,13 @@ def _simplify_requirement_list(self: RequirementList, state: State,
...
-        if item.negate:
-            return None
-
         if item.satisfied(state.resources, state.energy, state.resource_database):
             continue
 
         # We don't want to mark collecting a pickup/event node as a requirement to collecting that node.
         # This could be interesting for DockLock, as indicating it needs to be unlocked from the other side.
-        if item.resource.resource_type == ResourceType.NODE_IDENTIFIER:
+        if item.resource.resource_type == ResourceType.NODE_IDENTIFIER\
```
henriquegemignani, line 32: could either use `item.resource.resource_type in {...}`, or wrap the condition in ()`s so we avoid the `\` there.
In general you can wrap a line in ()s to avoid using the "backslash as last char" thing

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #4377, opened 2023-05-08
Fix user-facing messages mentioning docks

henriquegemignani:


randovania/generator/dock_weakness_distributor.py
```diff
@@ -231,7 +231,7 @@ async def distribute_post_fill_weaknesses(rng: Random,
         if patches.configuration.dock_rando.mode == DockRandoMode.VANILLA:
             continue
 
-        status_update(f"Preparing dock randomizer for player {player + 1}.")
+        status_update(f"Preparing door locks randomizer for player {player + 1}.")
```
duncathan, line 234: this should probably be "door lock", rather than "door locks"

duncathan (changes requested):

duncathan (approved):

PR #4378, opened 2023-05-08
Remove One-way door lock rando

henriquegemignani:


randovania/layout/base/dock_rando_configuration.py
```diff
@@ -13,27 +13,22 @@
 
 class DockRandoMode(BitPackEnum, Enum):
     VANILLA = "vanilla"
-    TWO_WAY = "two-way"
-    ONE_WAY = "one-way"
+    FULL = "two-way"
```
duncathan, line 16: I don't like the use of "full" here. I'd prefer we keep "two-way" unless we can come up with a better name

duncathan (changes requested):

duncathan (changes requested):
oh also we SHOULD changelog this even if one way has been disabled accidentally for a while

duncathan (approved):

PR #4381, opened 2023-05-08
Echoes: automatically decide noise removal

henriquegemignani:
Also dynamically generate the tab description.


duncathan (commented):
We should probably indicate in the UI when the SFX is being disabled and why. Otherwise I can imagine us getting bug reports about missing SFX in the future

henriquegemignani:
Could it be ambiguous when? The one-way anywhere is simply to explain, but wording "when stg, sky temple ec or aerie are shuffled" is less so.

duncathan (approved):

PR #4388, opened 2023-05-09
Add description editor to preset tabs

JeffGainsNGames:
- New tab in customize preset window, text edit accepts Markup formatting. Wasn't able to incorporate new lines with Plain Text, I can look into it more if having the markup option hidden from users is proffered.
Desc1
Desc2


duncathan:
are you able to make it use markdown, rather than markup?

henriquegemignani:
> are you able to make it use markdown, rather than markup?

[GenerateGameWidget](https://github.com/randovania/randovania/blob/main/randovania/gui/widgets/generate_game_widget.py#L446) does put the description inside a markup tag, but it's just for size so it could be changed.

duncathan:
for user-facing formatting markdown is vastly preferable imo so if we can do that we should

JeffGainsNGames:
I now have it supporting both. I could change the top label to not mention Markup to avoid users trying it but I don't believe I can remove the functionality
Desc1

henriquegemignani:
Might be better to just not mention it. We can also force markdown only for the label in the description window too.

I'm a more review side, please add a changelog.

henriquegemignani (approved):

PR #4389, opened 2023-05-09
DockLock fixes and db improvement

henriquegemignani:
image
image


duncathan (commented):
don't have time for a full review rn, but if we're adding the door lock db features to the editor we should make sure to add the field that excludes a given door from being randomized

duncathan (approved):

PR #4394, opened 2023-05-10
Multiworld: prefer starting pickup when there's no own locations left

henriquegemignani:


randovania/generator/filler/player_state.py
```diff
@@ -248,6 +248,13 @@ def should_have_hint(self, pickup: PickupEntry, current_uncollected: Uncollected
 
         return can_hint
 
+    def count_self_locations(self, locations: WeightedLocations) -> int:
+        count = 0
+        for player, index in locations.keys():
+            if player == self:
+                count += 1
+        return count
```
duncathan, line 256: any reason this shouldn't be `return sum(1 for player, _ in locations.keys() if player == self)`?

randovania/generator/filler/retcon.py
```diff
@@ -278,6 +278,18 @@ def debug_print_weighted_locations(all_locations_weighted: WeightedLocations):
         ))
 
 
+def should_be_starting_pickup(player: PlayerState, locations: WeightedLocations) -> bool:
+    cur_starting_pickups = player.num_starting_pickups_placed
+    minimum_starting_pickups = player.configuration.minimum_random_starting_pickups
+    maximum_starting_pickups = player.configuration.maximum_random_starting_pickups
+
+    result = cur_starting_pickups < minimum_starting_pickups or not locations
+    if player.count_self_locations(locations) == 0 and cur_starting_pickups < maximum_starting_pickups:
```
duncathan, line 287: would appreciate a comment along the lines of `prioritize placing starting pickups over off-world pickups` to clarify the intent here for anyone reading later

duncathan (changes requested):

randovania/generator/filler/player_state.py
henriquegemignani: tired brain couldn't confirm they were equivalent

duncathan (approved):

PR #4400, opened 2023-05-14
Don't export locked resources from other games in dread

ThanatosGit:
I failed to export dread multiworld seed because of locked resources from prime 1 were added to the patcher data.
E.g.:
image
where 1007 is the id of the Missile Launcher in prime 1 and shouldn't be added to dread's patcher data / file.

Patch will simply ignore /not add locked resources if it is for ITEM_NONE


randovania/games/dread/exporter/patch_data_factory.py
```diff
@@ -63,13 +63,14 @@ def get_resources_for_details(detail: ExportedPickupDetails) -> list[list[dict]]
...
     ]
 
-    if pickup.resource_lock is not None and not pickup.respects_lock and not pickup.unlocks_resource:
-        # Add the lock resource into the pickup in addition to the expansion's resources
-        assert len(resources) == 1
-        resources[0].append({
-            "item_id": get_item_id_for_item(pickup.resource_lock.locked_by),
-            "quantity": 1,
-        })
+    if resources[0][0]["item_id"] != "ITEM_NONE":
```
ArcanoxDragon, line 66: I think personally I'd turn this into an early return instead of putting all the following code into another nesting level. I'm not sure if @duncathan or @henriquegemignani have a convention/preference regarding early returns or not, but it prevents lots of deep nesting.

henriquegemignani: I prefer whatever makes code more readable. If an early return makes it more readable, then yeah.

In this case, I'd say what's most readable is detect if it's a pickup for another game. Maybe check for `not detail.other_player` in this if, and that's it.

ThanatosGit: Oh, that's a nice flag to use. Thanks!
Updated the pr

henriquegemignani:
Can you add a test file with multiworld for Dread?

ThanatosGit:
> Can you add a test file with multiworld for Dread?

Uh, I'm willing to do so but how we conceptually do this? We don't have rdvgame files for multiworld, do we? Permalinks would break and include all steps until a layout is generated.
So, I kinda create a multiworld layout and test the exporter or do you just want a small test for this specific line with e.g. prime1 missile?

henriquegemignani:
> We don't have rdvgame files for multiworld, do we?

Every rdvgame is a multiworld file, sometimes they just have 1 player :)
There's even two multiworld rdvgame test files already! `multiworld.rdvgame` and `prime1_and_2_multi.rdvgame`
(these files really need some renaming and organizing...)

Could just be one with 2 dread (can easily be made via gui if you run it with `--preview`: it adds a "how many players to generate for" to the solo UI).

ThanatosGit:
Wow, I'm sorry. I looked for other multiworld tests yesterday and somehow I managed to search for it in a way such that I haven't found any of the mentioned files.

henriquegemignani:
Oh there's absolutely no organisation at all for these things.

ThanatosGit:
Updated the pr with a dread-dread and dread-prime1 export.
Shamelessly copied the test functions from the existing tests and modified it

PR #4409, opened 2023-05-21
Rename World to Region

henriquegemignani:


randovania/exporter/hints/hint_formatters.py
```diff
@@ -18,16 +18,16 @@ def format(self, game: RandovaniaGame, pick_hint: PickupHint, hint: Hint, with_c
...
         self.namer = namer
-        self.with_world = with_world
+        self.with_region = with_region
         self.upper_pickup = upper_pickup
 
     def format(self, game: RandovaniaGame, pick_hint: PickupHint, hint: Hint, with_color: bool) -> str:
         node_name = self.namer.format_location(
             location=PickupLocation(game, hint.target),
-            with_world=self.with_world or hint.precision.location == HintLocationPrecision.WORLD_ONLY,
+            with_region=self.with_region or hint.precision.location == HintLocationPrecision.WORLD_ONLY,
```
duncathan, line 30: should rename this enum

randovania/game_description/db/region.py
```diff
@@ -23,14 +23,10 @@ def __repr__(self):
...
 
-    @property
-    def world_asset_id(self) -> int:
-        return self.extra["asset_id"]
-
     @property
     def all_nodes(self) -> Iterator[Node]:
         """
-        Iterates over all nodes in all areas of this world.
+        Iterates over all nodes in all areas of this db.
```
duncathan, line 29: I think this one ought to be world, not db

randovania/game_description/db/region.py
```diff
@@ -50,16 +46,16 @@ def area_by_name(self, area_name: str, is_dark_aether: bool | None = None) -> Ar
...
     def area_by_identifier(self, location: AreaIdentifier) -> Area:
-        if self.name != location.world_name:
-            raise ValueError(f"Attempting to use AreaIdentifier for {location.world_name} with world {self.name}")
+        if self.name != location.region_name:
+            raise ValueError(f"Attempting to use AreaIdentifier for {location.region_name} with region {self.name}")
         return self.area_by_name(location.area_name)
 
-    def correct_name(self, in_dark_world: bool) -> str:
-        if in_dark_world and self.dark_name is not None:
+    def correct_name(self, use_dark_name: bool) -> str:
```
duncathan, line 53: while we're in here renaming things, what do you think of renaming `dark_name` etc to something else? cave story kinda abuses this system although maybe I should just make it not abuse it instead lol

duncathan (changes requested):
I wanna say there are documentation files that need updating as well. also, we should consider putting all these recent renames in a changelog entry as they are at least somewhat user-facing

randovania/game_description/db/region.py
henriquegemignani: I'm slowly collecting my courage to refactor Prime 2 to have different ~~World~~Regions for Light/Dark. I'm afraid it's gonna do _shenanigans_ with pickup distribution because the weighting is based on how filled up each region is.

Could be renamed "alternative_name" I guess?

randovania/game_description/db/region.py
henriquegemignani: pycharm renamed a bunch of comments and strings when I renamed the folder, it's weird

randovania/game_description/db/region.py
duncathan: if we're both thinking about refactoring our games to do that, then maybe we should just leave it be? but yeah probably `alternative_name` (and should update `Area`'s `is_in_dark_aether` too while we're at it lol) if we do change it

duncathan (approved):

PR #4418, opened 2023-05-22
Add automatic construct encoding for JsonDataclass

henriquegemignani:


test/bitpacking/test_construct_dataclasss.py
```diff
@@ -45,7 +49,30 @@ def test_encode_dataclass():
...
     assert decoded == data
+
+
+@dataclasses.dataclass()
+class J(JsonDataclass):
+    a: int
+    b: str | None
+    c: list[uuid.UUID]
+    d: datetime.datetime
+
```
duncathan, line 64:
```py
@dataclasses.dataclass()
class K(JsonDataclass):
    a: bool

@dataclasses.dataclass()
class J(JsonDataclass):
    a: int
    b: str | None
    c: list[uuid.UUID]
    d: datetime.datetime
    e: K
```
maybe add something like that just to make sure we test nested dataclasses too

duncathan (changes requested):

randovania/bitpacking/construct_dataclass.py
```diff
@@ -34,12 +45,14 @@ def construct_for_dataclass(cls) -> construct.Construct:
 
 _direct_mapping = {
     bool: construct.Flag,
-    int: construct.VarInt,
+    int: construct.ZigZag,
```
duncathan, line 48: out of curiosity, was being unsigned causing any problems anywhere or are you just future proofing?

duncathan (approved):

randovania/bitpacking/construct_dataclass.py
henriquegemignani: there's certainly values that could be thrown around that are negative, though I guess PickupEntry doesn't use this rn

PR #4431, opened 2023-05-25
support for NamedTuple in json_dataclass and construct_dataclass

duncathan:


randovania/bitpacking/construct_dataclass.py
```diff
@@ -42,6 +42,20 @@ def _construct_for_dataclass(cls) -> construct.Construct:
 
     return construct.Struct(*fields)
 
+def _construct_for_named_tuple(cls) -> construct.Construct:
+    resolved_types = typing.get_type_hints(cls)
+
+    fields = [
+        construct.Renamed(
+            construct_for_type(resolved_types[field_name]),
```
henriquegemignani, line 50: does this work for files that _don't_ use `from __future__ import annotations`?

henriquegemignani (changes requested):

randovania/bitpacking/construct_dataclass.py
duncathan: seems so

henriquegemignani (approved):

PR #4445, opened 2023-06-01
Add AM2R logic database

Miepee:
This PR adds an initial logic database for AM2R, as well as add all basic functionality so that seeds can be generated.

<details> 
  <summary>Original Description</summary>
   This is still a work in progress. I'm opening this PR already to better publically track how far progress is, and for reviewers to have a chance at looking at potential problems I have in the database.
All nodes/transitions can be looked at.

Stuff that still needs to be done before I'll put this out of draft (ca. sorted by effort):
- [x] Fix typos in template names as well as rename them all, to be clearer of what they actually do
- [x] Fix typo in entrance rando short name
- [x] Remove unnecessary events/templates
- [x]  Add proper item capacaity
- [x]  Sort events somewhat properly
- [x] Add the rest of the Metroid events, and place them for each Metroid. Currently every Metroid event is just assigned to "Alpha1"
- [x] Fill out "defeat zeta/omega" templates
- [x] Double check area spawns. Ideally one per area.
- [x] Add better logic for bosses
- [x] Logic for water wall in A2
- [ ] Run through each "damage" trick, and add proper health logic
- [ ] Run through each Morph Glide and mid air morph trick and make morph a requirement there
- [ ] Run through A3/A4/A5 Spider ball checks and see if they need spring/bombs/etc.
- [x] ack -i for "TODO" and fix those
- [x] Add the internal name as the map name for every room from every region except for `Golden Temple`
- [ ] Figure out better on how to deal with the teleporter pipes 
- [x] Double check logic for the Tester event locked doors in `The Tower`
- [x] Figure out how to deal with the EMP Events in Distribution Center
- [x] Add the missing logic for `The Nest`, `Genetics Laboratory` and the exploding factory part in the `The Tower`
- [x] Redo all doors in header.json to be proper
- [x] Come up with human readable room names for all rooms from every region except for `Golden Temple` and rename accordingly

And probably a few others I missed.  
Patcher and wishlist of potential rando features will come in future PRs.
</details>


henriquegemignani:
Try to run [pngquant](https://pngquant.org/) over all the map images to minimize their file size as much as possible. Both to decrease the release size, but to minimize repository size.

Miepee:
Apart from doing a second more extensive logic pass and renaming almost every room, all nodes/transitions should be reviewable now!

Miepee:
It was decided to move the rest of logic stuff into future PRs. Thus, the PR is now out of draft

test/generator/test_generator_reach.py
```diff
@@ -100,6 +100,9 @@ def _compare_actions(first_reach: GeneratorReach,
     # These 3 indices are in Olympus and are unreachable given the default preset
     RandovaniaGame.METROID_PRIME_CORRUPTION: {0, 1, 2},
 
+    # These items require shinesparking, which due to varying difficulty has been made into a trick
```
henriquegemignani, line 103: I'd say it'd be nice to run the check with a few tricks then, maybe? Hm

henriquegemignani (changes requested):

test/generator/test_generator_reach.py
Miepee: Would be nice, yeah.
Not quite sure how to implement though. I wanted to avoid modifying the starter preset, as there are quite a few who don't have that level of shinesparking and I didn't want to exclude them so I wanted to opt for a "include tricks when checking for all pickups" approach. However, I'm a bit lost on what actually excludes the tricks in the first place, as far as I can see they're already included [here](https://github.com/randovania/randovania/blob/main/test/generator/test_generator_reach.py#L129). Probably something in reach_lib?

henriquegemignani: You can change the preset dynamically during the preset. This example for a Dread test:
```python
    gd = default_database.game_description_for(preset.game)
    suitless = gd.resource_database.get_by_type_and_index(ResourceType.TRICK, "Suitless")

    configuration = dataclasses.replace(
        configuration,
        trick_level=configuration.trick_level.set_level_for_trick(suitless, LayoutTrickLevel.HYPERMODE),
    )
```

Could be a new set of overrides at the top or something.

henriquegemignani (approved):

PR #4477, opened 2023-06-10
Construct Dataclass improvements

henriquegemignani:
- Fix Enum encoding
- UUIDs are fixed length

### DIFF
--- a/randovania/bitpacking/construct_dataclass.py
+++ b/randovania/bitpacking/construct_dataclass.py
@@ -61,7 +61,7 @@ _direct_mapping = {
     bool: construct.Flag,
     int: construct.ZigZag,
     str: BinStr,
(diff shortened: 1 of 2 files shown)


duncathan (approved):

henriquegemignani:
I want to make this encode UUID's via bytes so it's 16 bytes instead of 36, but I'd either do a roundtrip or not abuse the json encoding, preferably the later.

But that's a lot of effort.

duncathan:
UUID objects do convert pretty easily to and from bytes objects, but I imagine the annoying part is encoding bytes in JSON?

henriquegemignani:
oh it's trivial to do the encoding. But I find stupid how we'd go from UUID object to UUID string in dict, to then convert back to UUID object to go bytes.

PR #4478, opened 2023-06-10
Multiple Improvements

henriquegemignani:
many different things I've changed while working on multiworld server

### DIFF
--- a/randovania/bitpacking/construct_dataclass.py
+++ b/randovania/bitpacking/construct_dataclass.py
@@ -1,6 +1,7 @@
 import dataclasses
 import datetime
 import functools
+import json
 import typing
 import uuid
 from enum import Enum
@@ -128,6 +129,13 @@ def construct_for_type(type_: type) -> construct.Construct:
 
     elif dataclasses.is_dataclass(type_):
         return _construct_for_dataclass(type_)
+
+    elif hasattr(type_, "as_json") and hasattr(type_, "from_json"):
+        return construct.ExprAdapter(
+            BinStr,
+            encoder=lambda obj, ctx: json.dumps(obj),
+            decoder=lambda obj, ctx: json.loads(obj),
+        )
 
     raise TypeError(f"Unsupported type: {type_}.")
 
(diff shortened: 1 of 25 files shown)


randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -34,7 +34,7 @@ class PrimeRemoteConnector(RemoteConnector):
     version: all_prime_dol_patches.BasePrimeDolVersion
     game: GameDescription
     _last_message_size: int = 0
-    _world: Region | None = None
+    _last_emitted_region: Region | None = "PLACEHOLDER"
```
duncathan, line 37: when is `_last_emitted_region` set to `None`, now? what does it mean for it to be `"PLACEHOLDER"`?

randovania/bitpacking/construct_dataclass.py
```diff
@@ -129,6 +130,13 @@ def construct_for_type(type_: type) -> construct.Construct:
     elif dataclasses.is_dataclass(type_):
         return _construct_for_dataclass(type_)
 
+    elif hasattr(type_, "as_json") and hasattr(type_, "from_json"):
+        return construct.ExprAdapter(
+            BinStr,
+            encoder=lambda obj, ctx: json.dumps(obj),
+            decoder=lambda obj, ctx: json.loads(obj),
+        )
```
duncathan, line 138: is it a problem that building -> parsing the same object will transform it into JSON, rather than returning it to the class it previously was? I can't recall whether we enforce that here

randovania/game_connection/connector/prime_remote_connector.py
henriquegemignani: During `update`:
```python
            has_pending_op, region = await self.current_game_status()
            if region != self._last_emitted_region:
                self.PlayerLocationChanged.emit(PlayerLocationEvent(region, None))
            self._last_emitted_region = region
```
The idea of the change is to always emit a location changed the first time it happens, even if it's None. `GameConnection` uses that signal being emitted as a sign the RemoteConnector is not disconnected.

Though thinking about it I could make it default to `GameConnectionStatus.TitleScreen` whenever a RemoteConnector is created? hm...

duncathan: kk. in that case I just don't love seeing a magic string here - that enum entry you mentioned would be fine by mii (or whatever else you settle on)

duncathan (approved):

PR #4488, opened 2023-06-13
Remove TeleporterNode

ThanatosGit:
If it makes any sense: The state is very draft.
~~Since it needs manually editing at the curernt state, I probably trash the current partial automated migration (which is very dirty anyways).
At least it will require migrations for presets and rdvgame files.~~

PR #4494, opened 2023-06-14
Fix some easy logic errors for AM2R

Miepee:
This PR
- fixes launchers not unlocking their ammo
- adds a test to ensure that very common logic errors aren't missed (i.e. complain if a logic requires bombs, but no morph)
- fixes the errors that the above test showed


randovania/games/am2r/assets/migration_data.json
```diff
@@ -0,0 +1,96 @@
+{
+    "remove_teleporternode_data": {
```
henriquegemignani, line 2: Skip this file. Let's avoid having to worry about migrating preset/rdvgame for a game that is quite new

test/games/am2r/test_am2r.py
```diff
@@ -0,0 +1,75 @@
...
+        if requirement.template_name == resource:
+            return True
+        template_req = requirement.template_requirement(db)
+        return does_requirement_contain_resource(template_req, resource, db)
+    if isinstance(requirement, ResourceRequirement):
+        if requirement.pretty_text == resource:
+            return True
+        return False
+    for subreq in requirement.items:
+        temp_result = does_requirement_contain_resource(subreq, resource, db)
```
henriquegemignani, line 39: Temp variable here is not needed. Just "if func(): return true`

henriquegemignani (changes requested):
It's best to not combine a pr with big structural changes with logical changes, makes ir harder to read. No need to split this one though

Miepee:
Went through the logic changes I did with @DruidVorse .
Resolve tests fail, because these have been taken from #4488 as I'll be needing a few more DB changes before theirs gets merged and didn't want them to deal with rebasing.

ThanatosGit:
> Resolve tests fail, because these have been taken from #4488 as I'll be needing a few more DB changes before theirs gets merged and didn't want them to deal with rebasing.

Hmmm expected to get a little bit awkward if we both changing things in the db.
Is it really the resolver? The integrity check failed tests could be "skipped" by adding AM2R to `_acceptable_database_errors`. Though I would mark it with a TODO and we can meet again in a week or so and glue everything together.

henriquegemignani:
Changing TeleporterNodes to DockNodes should never have logical consequences (and any that do is very suspicious), but integrity errors could happen.
I'd just add to _acceptable_database_errors for now.

henriquegemignani (approved):

PR #4499, opened 2023-06-15
New integrity checks for docks

ThanatosGit:
Kinda cursed for me...

`base_dock_names_raw` is returning a list of names, where the first one is always the preffered one.
Since now every `base_dock_names_raw` will contain something like `to Region`, I have to get the real used base name from the node and just can't compare the lists. So, I wrote `docks_used_base_name` but since some of the games are using`(...)` in the area name, I wasn't able to only use the regex but had to also add the first if case to remove false postivies.

Not many changes but this still gave me a headache. I'm also fine if this pr get's trashed.


henriquegemignani:
I changed things around:
- `<dock type> to <region>` is only allowed if the dock goes to a different region
- `<dock type> to <region>` and `<dock type> to <area>` in the same area that points to the same area allows for both nodes to not have a suffix
- The node name suggestion skips the suffix if there are nodes incorrectly named

I also removed the command line for renaming docks because it's annoying :p

PR #4507, opened 2023-06-17
AM2R:Add new DNA goal

Miepee:
This goes over the AM2R database to prepare it for the new dna goal.
Specifically it:
- adds a new trivial pickup that can be reached from/to every metroid. Alphas and Gammas are minor, Zetas and Omegas are Major
- Removes the vanilla metroid gate locks
- Adds a dna gate lock in the lab
- Adds the rest of the save stations as possible candidates for area rando now.


Miepee:
Just realized i need extra data for all the new pickup nodes..

Miepee:
PR came out a little bigger than expected, tell me if you would like this to be split.

randovania/games/am2r/generator/pool_creator.py
```diff
@@ -2,11 +2,111 @@
...
+        }
+        item_pool = [key for key in keys_to_shuffle if key not in new_assignment.values()]
+
+    return PoolResults(item_pool, new_assignment, starting_keys)
+
+_c = NodeIdentifier.create
+_METROID_INDICES = [
+    # Alphas
+    _c("Main Caves", "Surface Alpha Nest", "Pickup (Alpha)"),
+    _c("Golden Temple", "Exterior Alpha Nest", "Pickup (Alpha)"),
```
ThanatosGit, line 54: Disclaimer: I'm not a core dev, which means it's not up to me to decide. I also know that most other games already doing stuff like that and especially this part is basically a copy of dread. Eventually, we better ask someone additional.

I really dislike hard coding stuff like names from the db. This breaks easily as soon as any part is renamed. My suggestion would be to use the extra section and simply filter the nodes according to these optional flags. That would make renaming much easier and less faulty. But as I said, that's more a general problem in randovania IMO and someone else should be asked.

randovania/games/am2r/gui/preset_settings/am2r_goal_tab.py
```diff
@@ -0,0 +1,89 @@
...
+    @property
+    def num_preferred_locations(self) -> int:
+        preferred = 0
+        if self.prefer_metroids_check.isChecked():
+            preferred += 46
+        if self.prefer_bosses_check.isChecked():
+            preferred += 6
+        if preferred > 46:
+            preferred = 46
+        return preferred
```
ThanatosGit, line 58: That's some weird logic. You always start with 0. If the first condition is true, you will always have 46 and second condition will always make it 6 or greater than 46, which will than be reset to 46.
Maybe the following (???):
```
    @property
    def num_preferred_locations(self) -> int:
        if self.prefer_metroids_check.isChecked():
           return 46
        if self.prefer_bosses_check.isChecked():
            return 6
        return 0
```

ThanatosGit (changes requested):

randovania/games/am2r/generator/pool_creator.py
Miepee: I copied dreads implementation for now as a way to get this implemented and playable faster.
I do agree though that the hardcoded nodes should go away for Dread, AM2R and whatever uses them as well. 
Due to tests, if something would change in the DB, it would complain here so a "something broken and generation is impossible now" shouldn't happen, but at minimum for when the "place DNA at normal pickups" option gets implemented this basically *has* to get refactored because I wouldn't want have all 88 pickup nodes in here.

randovania/games/am2r/gui/preset_settings/am2r_goal_tab.py
Miepee: Indeed is much cleaner, thanks

randovania/games/am2r/generator/pool_creator.py
henriquegemignani: Great refactor idea, did it for Dread https://github.com/randovania/randovania/pull/4538

Miepee: Addressed now as well.

henriquegemignani (changes requested):
Can you add explicit tests for the artifact config settings and the goal tab?

Miepee:
> Can you add explicit tests for the artifact config settings and the goal tab?

Do you have any references for those? Am unsure what I should test for, and how I should do them.

henriquegemignani:
> > Can you add explicit tests for the artifact config settings and the goal tab?
> 
> Do you have any references for those? Am unsure what I should test for, and how I should do them.

A lot of the dialog tests works well. The cosmetic patches are also good to look at.

test/games/am2r/gui/preset_settings/test_preset_am2r_goal.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Run
+    tab.prefer_metroids_check.setChecked(prefer_metroids)
+    tab.prefer_bosses_check.setChecked(prefer_bosses)
+    tab._update_slider_max()
+
+    # Assert
+    assert tab.num_preferred_locations == expected
+    assert tab.dna_slider.maximum() == expected
+    assert tab.dna_slider.isEnabled() == (expected > 0)
```
henriquegemignani, line 37: do also assert if the `editor.configuration` has the correct values

test/games/am2r/generator/test_am2r_pool_creator.py
```diff
@@ -0,0 +1,139 @@
...
+
+def test_am2r_artifact_pool_metroids(am2r_game_description, preset_manager):
+    # Setup
+    configuration = preset_manager.default_preset_for_game(am2r_game_description.game).get_preset().configuration
+
+    configuration = dataclasses.replace(configuration,
+                                        artifacts=dataclasses.replace(configuration.artifacts,
+                                                                      prefer_metroids=True, prefer_bosses=False, required_artifacts=46))
+
+    print(configuration.artifacts.prefer_metroids)
```
henriquegemignani, line 63: prints

test/games/am2r/generator/test_am2r_pool_creator.py
```diff
@@ -0,0 +1,139 @@
...
+
+def test_am2r_artifact_pool_bosses(am2r_game_description, preset_manager):
+    # Setup
+    configuration = preset_manager.default_preset_for_game(am2r_game_description.game).get_preset().configuration
+
+    configuration = dataclasses.replace(configuration,
+                                        artifacts=dataclasses.replace(configuration.artifacts,
+                                                                      prefer_metroids=False, prefer_bosses=True, required_artifacts=6))
+
+    print(configuration.artifacts.prefer_metroids)
```
henriquegemignani, line 84: more prints

test/games/am2r/generator/test_am2r_pool_creator.py
```diff
@@ -0,0 +1,139 @@
...
+from random import Random
+
+from randovania.game_description.game_patches import GamePatches
+from randovania.game_description.db.node_identifier import NodeIdentifier
+from randovania.games.am2r.generator.pool_creator import artifact_pool
+from randovania.generator.pickup_pool import pool_creator
+
+
+def test_am2r_pool_creator(am2r_game_description, preset_manager):
+    return
```
henriquegemignani, line 12: !!!!

test/games/am2r/gui/preset_settings/test_preset_am2r_goal.py
```diff
@@ -0,0 +1,40 @@
...
+    assert editor.configuration.artifacts.prefer_metroids == prefer_metroids
+    assert editor.configuration.artifacts.prefer_bosses == prefer_bosses
+    assert 0 <= editor.configuration.artifacts.required_artifacts <= expected
```
henriquegemignani, line 40: Why can't you know how many it'd be in the test?

henriquegemignani (changes requested):

test/games/am2r/gui/preset_settings/test_preset_am2r_goal.py
Miepee: Because `required_artifacts` will not necessarily adjust itself after the other values have been set.
i.e. on `prefer_metroids=True, prefer_bosses=false`, the `required_artifacts` will have value of 15, because that's what's in the default configuration.

henriquegemignani: Then assert for that? The good of being precise with the assertion is that we can catch is things change unexpectedly.

PR #4509, opened 2023-06-17
AM2R: Fix BG3 softlock

Miepee:
Fixes generator getting stuck in BG3

### DIFF
--- a/randovania/games/am2r/json_data/Distribution Center.json
+++ b/randovania/games/am2r/json_data/Distribution Center.json
@@ -9152,7 +9152,7 @@
                                 ]
                             }
                         },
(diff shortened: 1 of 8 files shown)


Miepee:
Generator still fails on other places, i'll collect the other ones first and then mark out of draft

Miepee:
Ready for review
Worth noting that i removed a "Before Alpha 6 or after First omega" connection, as the generator had issues evaluating that out correctly
The DB is technically wrong now, but it's a temporary really minor thing and will fixed by the new dna goal #4507 anyway. Just wanted to have it generate now so i can have more test sessions.

henriquegemignani (approved):

PR #4510, opened 2023-06-17
Make dataclass construct encoding not depend on jsondataclass

henriquegemignani:
Please disregarding the jank with server stuff, that's all changed in the other PR.


randovania/bitpacking/construct_pack.py
```diff
@@ -0,0 +1,200 @@
...
+    if type_ in _direct_mapping:
+        return _direct_mapping[type_]
+
+    if issubclass(type_, Enum):
+        enum_arr = list(type_)
+        return construct.ExprAdapter(
+            construct.VarInt,
+            decoder=lambda obj, ctx: enum_arr[obj],
+            encoder=lambda obj, ctx: enum_arr.index(obj),
+        )
```
duncathan, line 140: how safe is it to depend on the exact order of values in an enum? I don't think we change them that frequently but I figured I'd ask

henriquegemignani: For anything that uses construct dataclass, my assumption is that both need to use the exact same version.
Adding am enum value can break, or reordering, but the same applies to changing the dataclass

duncathan (approved):

PR #4519, opened 2023-06-20
Very weird way of doing random default items

henriquegemignani:


duncathan:
is there a particular reason this PR is a draft?

henriquegemignani:
I hated how hacky the solution was. And I think it uses rng during pool generation which is known to be broken now?

PR #4523, opened 2023-06-20
Don't rename elevator rooms when using new patcher

henriquegemignani:


duncathan (approved):

duncathan (changes requested):
probably best to just go with #4526

PR #4546, opened 2023-06-23
MultiworldClient: ignore connected game if it has invalid uuid

henriquegemignani:


randovania/gui/multiworld_client.py
```diff
@@ -81,9 +82,10 @@ def __init__(self, network_client: QtNetworkClient, game_connection: GameConnect
             try:
                 uid = uuid.UUID(f.name)
```
ThanatosGit, line 83: It's not a change from this PR but I have already seen this and it reminded me to ask: How can a filename which ends with ".json" ever be a valid UUID? Shouldn't this be something like `basename` which cuts off the file ending?

ThanatosGit (approved):

randovania/gui/multiworld_client.py
henriquegemignani: Because I never tested it. #4547

PR #4555, opened 2023-06-25
Swap to session tab when exporting, so progress bar is visible

henriquegemignani:
Idea from ulti


ThanatosGit (approved):
I'm all for it.
I was confused by "where is my progress bar" the very first time I exported something.
@MasierAzulite was also asking me the same question when he used it the first time,

henriquegemignani:
I don't really like just swapping to other tab, but I don't like placing the bar at the bottom like before, so doing this one

duncathan:
solution: add a separate floating window for the progress bar alone :)

duncathan (approved):

PR #4564, opened 2023-06-25
Detect unmodified build strings

henriquegemignani:


randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -75,7 +76,11 @@ async def check_for_world_uid(self) -> bool:
         expected = bytearray(self.version.build_string)
         expected[6:6 + 16] = world_uid
         if build_string == expected:
-            self._layout_uuid = uuid.UUID(bytes=world_uid)
+            if build_string == self.version.build_string:
+                # Game exported with old version, act as if it's invalid uuid
+                self._layout_uuid = INVALID_UUID
+            else:
+                self._layout_uuid = uuid.UUID(bytes=world_uid)
```
ThanatosGit, line 83: Hmmm I don't understand that part.
If `build_string == expected` and `expected = bytearray(self.version.build_string)` how can `if build_string == self.version.build_string` not be `true`?

henriquegemignani: The exact line between the two you quoted: `expected[6:6 + 16] = world_uid`

ThanatosGit (approved):

PR #4565, opened 2023-06-25
Remove the Back button, always reply with new message

henriquegemignani:
There's a bug in the bot currently that it seems to randomly fail to work. I can reproduce it easily by:
1. Starting the bot
2. Run `/database-inspect` with a game
3. Select one section of regions
4. Press back
5. Repeat 3 and 4 until it fails

The error messages comes as if we're sending too much data, but adding prints to the libraries tells we don't. But since that request is a PATCH one, I'm wondering if we're hiting a _lifetime_ total message or something?

This PR dodges the issue by just never editing the message anymore.

### DIFF
--- a/randovania/server/discord/database_command.py
+++ b/randovania/server/discord/database_command.py
@@ -14,10 +14,10 @@ from discord import Embed
 from discord.ext.commands import Converter, Context
 
 from randovania.game_description import default_database, pretty_print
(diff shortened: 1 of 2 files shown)


henriquegemignani:
Added a description to the PR.

duncathan (approved):

PR #4574, opened 2023-06-26
Add "Open Session Window" action / Allow joining without password 

ThanatosGit:
1. Adds "Open Session Window" action to game connection window
2. Allow joining without password if user already belongs to the session

Got somehow distracted by 2. because that made the refactor for 1. a little bit easier.

Not really happy with some parts: Moving the whole "which game session windows are open" part to `WindowManager` introduced some cyclic dependencies. Atm I fixed that by importing when used in the functions.

Let's see what codecov says.

### DIFF
--- a/randovania/gui/lib/qt_network_client.py
+++ b/randovania/gui/lib/qt_network_client.py
@@ -2,21 +2,73 @@ import datetime
 import functools
 import json
 import webbrowser
+import asyncio
 from pathlib import Path
 
 from PySide6.QtCore import Signal
 from PySide6.QtWidgets import QWidget
+from qasync import asyncSlot
 
 import randovania
-from randovania.gui.lib import async_dialog
+from randovania.gui.lib import async_dialog, wait_dialog
 from randovania.network_client.network_client import NetworkClient, ConnectionState, UnableToConnect
 from randovania.network_common.error import (InvalidAction, NotAuthorizedForAction, ServerError, RequestTimeout,
-                                             NotLoggedIn, UserNotAuthorized, UnsupportedClient)
+                                             NotLoggedIn, UserNotAuthorized, UnsupportedClient, WrongPassword)
 from randovania.network_common.multiplayer_session import (
-    MultiplayerSessionEntry, User, MultiplayerSessionActions,
+    MultiplayerSessionEntry, MultiplayerSessionListEntry, User, MultiplayerSessionActions,
     MultiplayerWorldPickups, MultiplayerSessionAuditLog,
     WorldUserInventory)
+from PySide6.QtWidgets import QDialog, QInputDialog, QLineEdit
+from PySide6.QtCore import Qt
 
+
+def handle_network_errors(fn):
+    @functools.wraps(fn)
+    async def wrapper(self, *args, **kwargs):
+        try:
+            return await fn(self, *args, **kwargs)
+
+        except InvalidAction as e:
+            await async_dialog.warning(self, "Invalid action", f"{e}")
+
+        except ServerError:
+            await async_dialog.warning(self, "Server error",
+                                       "An error occurred on the server while processing your request.")
+
+        except NotLoggedIn:
+            await async_dialog.warning(self, "Unauthenticated",
+                                       "You must be logged in.")
+
+        except NotAuthorizedForAction:
+            await async_dialog.warning(self, "Unauthorized",
+                                       "You're not authorized to perform that action.")
+
+        except UserNotAuthorized:
+            await async_dialog.warning(
+                self, "Unauthorized",
+                "You're not authorized to use this build.\nPlease check #dev-builds for more details.",
+            )
+
+        except UnsupportedClient as e:
+            s = e.detail.replace('\n', '<br />')
+            await async_dialog.warning(
+                self, "Unsupported client",
+                s,
+            )
+
+        except UnableToConnect as e:
+            s = e.reason.replace('\n', '<br />')
+            await async_dialog.warning(self, "Connection Error",
+                                       f"<b>Unable to connect to the server:</b><br /><br />{s}")
+
+        except RequestTimeout as e:
+            await async_dialog.warning(self, "Connection Error",
+                                       f"<b>Timeout while communicating with the server:</b><br /><br />{e}"
+                                       f"<br />Further attempts will wait for longer.")
+
+        return None
+
+    return wrapper
 
 class QtNetworkClient(QWidget, NetworkClient):
     Connect = Signal()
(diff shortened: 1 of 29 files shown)


randovania/gui/lib/window_manager.py
```diff
@@ -55,6 +61,22 @@ def remove_window():
         window.CloseEvent.connect(remove_window)
         self.tracked_windows.append(window)
 
+    async def activate_or_create_game_session_window(self, network_client: QtNetworkClient,
+                                                session_entry: MultiplayerSessionEntry, options: Options
+                                                ):
+        session_window = self.opened_session_windows.get(session_entry.id, None)
```
henriquegemignani, line 67: I would move this entire implementation to MainWindow

randovania/gui/lib/qt_network_client.py
```diff
@@ -117,51 +169,48 @@ def available_login_methods(self) -> set[str]:
...
+        except WrongPassword:
+            await async_dialog.warning(self, "Incorrect Password", "The password entered was incorrect.")
 
-        except UnsupportedClient as e:
-            s = e.detail.replace('\n', '<br />')
-            await async_dialog.warning(
-                self, "Unsupported client",
-                s,
-            )
+    async def ensure_logged_in(self, parent: QWidget | None):
```
henriquegemignani, line 196: I don't really like this being here.
In particular I'm planning to have "weak" and "strong" auth. 

When you restore a session, if your discord re-auth fails you'll be "weak" auth. You can still do world sync and play multi, but can't open and session gui.
Explicitly doing a login flow always grants "strong" auth.

henriquegemignani (changes requested):

randovania/gui/widgets/game_connection_window.py
```diff
@@ -165,12 +197,57 @@ def setup_builder_ui(self):
...
+            return False
+        
+        # check the uuid
+        layout_uuid = associated_rc._layout_uuid
+        if layout_uuid == INVALID_UUID:
+            return False
+
+        if not await self.network_client.ensure_logged_in(self):
+            return False
+        session_list_entry = await self.network_client.get_session_for_uuid(layout_uuid)
```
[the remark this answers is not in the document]
ThanatosGit: Hmmm you mean both calls? From the other comment, I get that we could remove L225-226 since you can just sync it without being logged in.
L227 is just to make sure that the UUID really belongs to a session. It could be also removed and just be like "not invalid = show button" and do the login when the user really pressed the button to open the session window.

randovania/gui/lib/qt_network_client.py
ThanatosGit: If we remove the call to that function in the section of the first comment, it get's only called by `_host_game_session`, `_browse_for_session` and `_attempt_join` (pressing the new action in the game connection window).
That sounds fine to me because that is where you want the strong auth.

randovania/gui/widgets/game_connection_window.py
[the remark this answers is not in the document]
ThanatosGit: After working on it locally: The list entry, which is holding the fields for "has password" and "user in session", is not accessible by `WorldDatabase` (or I just don't find it). 
That means I just delay the calls until the user presses the join action? Do a login, ask for session entry, optional password prompt and eventually open the session window.
I must admit asking for a session list entry is mostly needed because of rare corner cases like:
- Having a valid UUID but there is no session for it
- valid UUID + session exists but you are not in the session (and it has a password). That's like trying to steal someone's session by getting access to the world's iso / romfs.

Otherwise we could just login and allow to join / open the session window by just sending a uuid and the server does the rest and just declines it if something is wrong.

randovania/gui/lib/qt_network_client.py
henriquegemignani: I'd just remove the CLI command for joining a session directly from `qt.py`

randovania/gui/widgets/game_connection_window.py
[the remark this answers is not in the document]
henriquegemignani: My idea is that if the user knows about a World and even has session data about it, then 99% of the time [1] that World belongs to a Session the user also belongs to.
So if you're connected to a Game and you know session details about that Game's World, we should assume "user in session" is valid, and if that's valid we should be able to not need a password. This follows with:
- Call `multiplayer_listen_to_session` to make the server start sending you session details.
   - This request fails if you're not a member already.
- Call `multiplayer_request_session_update` to get a full entry, though looking at `MultiplayerSessionWindow.create_and_update` we could get away with just passing a session id here, especially since it already requests the update.

[1]: I can only think of "logged out and then logged in as another user" and "kicked out of session" as reasons why that wouldn't happen. Both are rare enough that just handling as failure case seems good enough.

> Having a valid UUID but there is no session for it

If WorldDatabase has session details for a world, then a session absolutely exists, and you have it's id!
And if doesn't (session got deleted? then the world itself also got deleted and other mess), you `multiplayer_listen_to_session` will fail because you're not a member :)

> valid UUID + session exists but you are not in the session

I would just reject completely if you're not a member. Put the session name in the error message if needed, but let the user join via the normal flow. This is a rare case, let's not overcomplicate it.

randovania/gui/lib/qt_network_client.py
ThanatosGit: That works but dev builds need to call the login method. I only broke every future dev build that's why I thought I maybe should fix it :p

randovania/gui/widgets/game_connection_window.py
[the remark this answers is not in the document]
ThanatosGit: Changed the logic again. Re-requested review but take your time. Earliest I will fix the tests is somewhere on the weekend.

randovania/server/database.py
```diff
@@ -178,8 +178,11 @@ def layout_description(self, description: LayoutDescription | None):
     @property
     def creation_datetime(self) -> datetime.datetime:
         return datetime.datetime.fromisoformat(self.creation_date)
+    
+    def is_user_in_session(self, user: User):
+        return any(member for member in self.members if member.user == user)
```
henriquegemignani, line 183: Not obvious, but this is terrible code!
This involves an additional query (or many? not sure) to get each member, so this is very slow.

If keeping this, best to do a direct query for the membership and handle found/not found.

henriquegemignani (changes requested):

randovania/gui/widgets/game_connection_window.py
```diff
@@ -164,10 +201,61 @@ def setup_builder_ui(self):
...
 
+    @asyncSlot()
+    async def _attempt_join(self, layout_uuid: uuid.UUID):
+        if not await self.network_client.ensure_logged_in(self):
+            return
+        session_id = self._check_session_data(layout_uuid)
+        joined_session = await self.network_client.attempt_join(session_id, None)
+        if joined_session is not None:
+            await self.window_manager.activate_or_create_game_session_window(
+                self.network_client, joined_session,
```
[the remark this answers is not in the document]
ThanatosGit: You mean `activate_or_create_game_session_window`  by `that`? Sounds like a logical consequence of using `listen_to_session` which just returns the session ids?
No idea how to do that. It needs a `MultiplayerSessionEntry` for `MultiplayerSessionWindow.create_and_update`, which uses it for `MultiplayerSessionApi` and `_session`. `_session` is used to acess users, calling `get_world` etc.
Providing only the id will just move requesting the session list entry into `activate_or_create_game_session_window`.

randovania/gui/widgets/game_connection_window.py
```diff
@@ -164,10 +201,61 @@ def setup_builder_ui(self):
 
         self.update_builder_ui()
 
+    @asyncSlot()
+    async def _attempt_join(self, layout_uuid: uuid.UUID):
+        if not await self.network_client.ensure_logged_in(self):
+            return
+        session_id = self._check_session_data(layout_uuid)
+        joined_session = await self.network_client.attempt_join(session_id, None)
```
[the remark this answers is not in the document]
ThanatosGit: So it must be a new implementation for this window cause the current `handle_network_errors` are catching it already.

randovania/server/database.py
ThanatosGit: Was scrolling through documentation of peewee. Direct query would be(?!):
```
    def is_user_in_session(self, user: User):
        result = MultiplayerMembership.get_or_none((MultiplayerMembership.user == user) & (MultiplayerMembership.session == self.id))
        return result is not None
```

randovania/gui/widgets/game_connection_window.py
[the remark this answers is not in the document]
henriquegemignani: `handle_network_errors` wraps around your function. So if you don't catch something, it catches for you.
So if you explicitly handle one case inside, then it doesn't get to `handle_network_errors`.

randovania/server/database.py
henriquegemignani: Can just do `MultiplayerMembership.get_by_ids` and use a try except to catch `DoesNotExist`

henriquegemignani (approved):

PR #4575, opened 2023-06-26
[Echoes] Various Tweaks to Bioenergy Production Logic

f-raZ0R:
This pull request does several changes to the logic database for the room Bioenergy Production:

-Readds the vanilla boost+spider method to get to the Storage C door, since that seems to have gotten messed up at some point.
-Lowers the trick difficulty of a couple of ways to reach the door to Security Station B NSJ
-Adds a new Movement (Advanced) trick to reach the pickup in the room NSJ, with Screw Attack.


henriquegemignani (dismissed):
Could you add videos for the new path, and the one moved to begginer?

f-raZ0R:
Added a single video containing both. Yes, it does contain the use of savestates near the end, but I am not quite consistent enough at the trick, and have already spent upwards of hours attempting to get a better recording without the use of states. If this makes the video not acceptable, then I could either split it into two, and crop the state out, or attempt further to get a better recording.

dyceron (approved):

PR #4581, opened 2023-06-26
Move Data out of MultiworldClient into WorldDatabase

henriquegemignani:


randovania/interface_common/world_database.py
```diff
@@ -0,0 +1,123 @@
...
+    _persist_path: Path
+
+    def __init__(self, persist_path: Path):
+        super().__init__()
+        self.logger = logging.getLogger(__name__)
+
+        persist_path.mkdir(parents=True, exist_ok=True)
+        self._persist_path = persist_path
+
+        self._remote_games = {}
```
ThanatosGit, line 70: Is this field of `WorldDatabase` used somewhere? Can only find usages of the field with the same name of `MultiworldClient`

henriquegemignani: oops

PR #4589, opened 2023-06-27
Optimize resolver, focused on Door Lock rando

henriquegemignani:
Before:
```
python -m randovania layout generate-from-permalink --permalink DXK8dMMXI5StDw00PJ6wAvjukIKSxjASlIAMf_FmOOeWEPnvnFDM9SMqAHvZ door.rdvgame
Took 159.693 seconds. Hash: AUQKVSKI
```

After:
```
python -m randovania layout generate-from-permalink --permalink DXK8dMMXI5StDw00PJ6wAvjukIKSxjASlIAMf_FmOOeWEPnvnFDM9SMqAHvZ door.rdvgame
Took 93.556 seconds. Hash: AUQKVSKI
```
Though there's some variance in runtime.

## What could be done to improve further?
99% of the execution time is in `ResolverReach.calculate_reach`.
image

`_build_satisfiable_requirements` is the biggest single contributor, though `fast_as_alternatives` helped a lot to reduce that. Only way I can see of improving this is to have less data. Maybe adding the additional requirements there is not needed?
image

The other frequent problem is calling `potential_nodes_from`. There's just way too many calls of that and the `connections_from` of Docks likes to show up a lot.
But there's also a bunch of damage/satisfied and general time spent on the function itself.

I can see value in trying to make `ResolverReach.calculate_reach` be able to leech out of a previous `ResolverReach` instance, to avoid having to navigate a bulk of the world.
Something else to do is suspend the reach calculation when an action is found.


henriquegemignani:
Writing the description made me realise I could cache the entire result of `DockNode.connections_from`. Made it ~16s faster.

```
python -m randovania layout generate-from-permalink --permalink DXK8dMMXI5StDw00PJ6wAvjukIKSxjASlIAMf_FmOOeWEPnvnFDM9SMqAHvZ door.rdvgame
Took 93.556 seconds. Hash: AUQKVSKI
```

duncathan (approved):
some of the innermost functions are a little bit ugly at points but that's just the price of speed

PR #4591, opened 2023-06-27
Improve Auto Tracker and no connected games

### LINKED ISSUE #4595: Auto Tracker theme selection is a checklist, not radio buttons
This means you can select multiple themes for the same game, which is very confusing behavior as it's unclear which one takes precedence and you may think you cannot change your theme somehow

henriquegemignani:
- Option to which game to use when you've never been connected to a game
- Don't lose connected game and inventory when game disconnects

Fixes #4595

### DIFF
--- a/randovania/gui/auto_tracker_window.py
+++ b/randovania/gui/auto_tracker_window.py
@@ -42,6 +42,7 @@ class AutoTrackerWindow(QtWidgets.QMainWindow, Ui_AutoTrackerWindow):
     item_tracker: ItemTrackerWidget | None = None
     _dummy_tracker: QtWidgets.QLabel | None = None
     _last_source: RemoteConnector | None = None
+    _last_selected_builder: ConnectorBuilder | None = None
 
     def __init__(self, game_connection: GameConnection, window_manager: WindowManager | None, options: Options):
         super().__init__()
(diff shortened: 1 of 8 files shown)


randovania/gui/auto_tracker_window.py
```diff
@@ -153,5 +196,7 @@ def selected_builder(self) -> ConnectorBuilder | None:
     def on_game_state_updated(self, state: ConnectedGameState):
         self.create_tracker()
         expected_connector = self.game_connection.get_connector_for_builder(self.selected_builder())
-        if expected_connector == state.source:
+        if expected_connector == state.source or self._last_source == state.source:
+            self._last_source = state.source
             self.item_tracker.update_state(state.current_inventory)
+            self.connected_game_state_label.setText(f"{state.source.game_enum.long_name}: {state.status.pretty_text}")
```
ThanatosGit, line 202: That's missig the update if you open the auto tracker after the connector updated the status and will say only `Disconnected`.
Maybe read it once on open + changing the selected connector? Cause I was easily a able to get these wrong results:
Disconnected but I'm on title screen:
image
Echoes showing status of Dread:
image

ThanatosGit (changes requested):
Run out of time. Will check the rest in the evening

randovania/gui/auto_tracker_window.py
henriquegemignani: Addressed, with a bunch other weird cases too

ThanatosGit: Has still some unexpected behaviour for me. Like switching from a connected dread to an unconnected dolphin, stays on the dread tracker and an empty connection state label. Don't know if someone will every do this but at least I want to show it:
image

henriquegemignani: Ah it's intentional to keep the tracker: if you go from non-disconnected to disconnected, it keeps the tracker in case it was a temporary connection loss.
The lack of any sort of label below is weird, but that's the message I put when the builder has no connector associated. Maybe it should say "Not Connected"

ThanatosGit: Keeping the tracker makes only sense to me if the game connection was not changed or better to say: Discard the old state if some is fiddling around in the selector box.

henriquegemignani: Made the tracker reset if changing builders

ThanatosGit (approved):

PR #4592, opened 2023-06-27
Delay importing the actual RemoteConnector classes at startup

henriquegemignani:


ThanatosGit (approved):
Afaik not a best practice but that heavily speeds up starting time for me.

duncathan:
Nah, it's appropriate to delay imports when they're known to be significantly slow

henriquegemignani:
In the quest for speed, everything goes.
But just make sure something is actually slow before doing it.

PR #4597, opened 2023-06-28
AM2R: add new beam doors and make tower doors shuffable

Miepee:
Adds doors for all available beams (charge, wave, spazer, plasma, ice)
The plasma beam chamber door was accidentally assigned the wrong type
And the tester arena logic has been revised to better reflect what actually happens in-game. (need to review this with @DruidVorse)


Miepee:
Reviewed by @DruidVorse

PR #4604, opened 2023-06-29
Add button to edit cosmetic options in MultiplayerSession

henriquegemignani:


ThanatosGit (approved):
The implementation in the multiworld part is not blocking other UIs while the one in the game details window does.
Afaik multiworld one also allows opening multiple times the same cosmetic window dialog.
I assume it is intentional because it was explicitly rewritten to use `asyncSlot`....so I approve.

henriquegemignani:
> The implementation in the multiworld part is not blocking other UIs while the one in the game details window does.
> Afaik multiworld one also allows opening multiple times the same cosmetic window dialog.
> I assume it is intentional because it was explicitly rewritten to use `asyncSlot`....so I approve.

That's why async_dialog exists! The main reason is that doing exec causes the asyncio loop to stop, só ser avoid that.

Now, there being theme differences is not expected and I dunno how to fix it :(

PR #4607, opened 2023-06-29
Improve logging for the server

henriquegemignani:


randovania/server/app.py
```diff
@@ -8,20 +8,56 @@
 
 import randovania
 import randovania.server.multiplayer.world_api
+from randovania.network_common import error
 from randovania.server import user_session, database, client_check, multiplayer
 from randovania.server.multiplayer import world_api
 from randovania.server.server_app import ServerApp
 
 
+class CustomFormatter(logging.Formatter):
```
duncathan, line 17: surely we can come up with a more descriptive name than this

duncathan (changes requested):

randovania/server/app.py
henriquegemignani: I accept suggestions :p

duncathan: `ServerLoggingFormatter`?

duncathan (approved):

PR #4619, opened 2023-06-30
Fix foldable being cut off

### LINKED ISSUE #4617: "Expandable widgets" sometimes chop off text
IDK if this is potentially linux related or something else. I went through the whole GUI and this was the only place where i could find them
grafik
These should say `Charge Combos` and `Energy Tanks` but look like `Charoe Combos` and `Eneroy Tanks` due to the chopped of `g`

Miepee:
Dunno if increasing the height is really the right fix for it, but it does work.
Fixes #4617 
grafik

### DIFF
--- a/randovania/gui/lib/foldable.py
+++ b/randovania/gui/lib/foldable.py
@@ -15,12 +15,12 @@ class Foldable(QtWidgets.QWidget):
         self._folded = initially_folded
 
         self._toggle_button = QtWidgets.QToolButton(self)
-        self._toggle_button.setStyleSheet("QToolButton { height: 20px; }")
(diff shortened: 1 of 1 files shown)


henriquegemignani:
Does less height work, like 25?

Miepee:
Barely, depending on the installed font and/or other characters, this might clip
grafik

Miepee:
This is with 28, doesn't seem like it'd cause any issues. If wanted I can change it to this
grafik

henriquegemignani:
If things look ugly we can revisit.

henriquegemignani (approved):

PR #4625, opened 2023-06-30
Place restrictions on names for Sessions and Worlds

henriquegemignani:


duncathan (approved):
looks good. I like the new text prompt class. maybe in the future we can use it for things like renaming regions/areas in the db

henriquegemignani:
> looks good. I like the new text prompt class. maybe in the future we can use it for things like renaming regions/areas in the db

I changed everything that was using QInputDialog and async_dialog. I bet the editor stuff is using `QInputDialog.getText` which is a blocking call.

duncathan:
> > looks good. I like the new text prompt class. maybe in the future we can use it for things like renaming regions/areas in the db
> 
> I changed everything that was using QInputDialog and async_dialog. I bet the editor stuff is using `QInputDialog.getText` which is a blocking call.

I'm not even sure if we have a way to rename areas and regions tbh, I only know for sure we can rename nodes

Miepee:
> I'm not even sure if we have a way to rename areas and regions tbh

Areas should be renamable, if a bit buggy, AFAIK no way to rename regions

PR #4632, opened 2023-07-01
Better error and logging messages

henriquegemignani:


randovania/server/database.py
```diff
@@ -84,10 +84,6 @@ class User(BaseModel):
     name: str = peewee.CharField()
     admin: bool = peewee.BooleanField(default=False)
 
-    @classmethod
-    def get_by_id(cls, pk) -> Self:
-        return cls.get(cls._meta.primary_key == pk)
-
```
ThanatosGit, line 90: If that's the default implementation of peewee it might could also be removed from `MultiplayerSession`.

henriquegemignani: It's actually a reimplementation in BaseModel, but yeah

ThanatosGit (approved):

PR #4636, opened 2023-07-01
Dread: Check for RomFS contents to validate input

henriquegemignani:


ThanatosGit (approved):
Why not check all approx. 9000 files :p

henriquegemignani:
> Why not check all approx. 9000 files :p

I'm lazy to make the list. Also might be a bit slow haha

duncathan (approved):

PR #4659, opened 2023-07-03
AM2R: Add option for bomb block before A3

Miepee:
Still needs to be reviewed by @DruidVorse


Miepee:
Reviewed by @DruidVorse

PR #4662, opened 2023-07-03
Revise MW description in Main window

Miepee:
I think this now raises the minimum height of the main RDV window, but I don't think shortening this text would be a good idea to for UX.

preview:
grafik


Miepee:
Alternative, where the main description is now tucked away in the help tab:
grafik

henriquegemignani:
> Alternative, where the main description is now tucked away in the help tab: grafik

Prefer this one, it's a good summary and they can click to read more.

PR #4675, opened 2023-07-04
Improve server logging

henriquegemignani:


duncathan (approved):
is there a particular reason for not using fstrings here?

henriquegemignani:
> is there a particular reason for not using fstrings here?

Using fstrings is not best practice for logging, as you're eagerly doing the formatting even if logging for that level is disabled.

PR #4680, opened 2023-07-04
Dread: Add Wide Beam to Burenia Hub to Dairon Pseudo Beam

Tyranisaur:


CHANGELOG.md
```diff
@@ -14,6 +14,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Tweaked the error reporting for generating and exporting games.
 - Fixed: Pickups providing negative resources can now be sent in multiworld games.
 
+### Metroid Dread
+
+#### Logic Database
+
+- Fixed: The Pseudo Wave Beam in Burenia Burenia Hub to Dairon now correctly requires Wide Beam.
```
XenoWars, line 21: Maybe only have one Burenia here. PR is otherwise fine

XenoWars (changes requested):

CHANGELOG.md
Tyranisaur: I get this. The convention I've been using is simply "AreaName RoomName", where this jank happens because in this case the room name is "Burenia Hub to Dairon". I think using a separator is a better idea than to drop the duplicate.

XenoWars: I would do that then.

ThanatosGit (approved):

PR #4694, opened 2023-07-05
AM2R: Fixes model for launchers

Miepee:
See title


henriquegemignani:
Didn't you do this already before?

PR #4702, opened 2023-07-05
More ruff rules

henriquegemignani:


ThanatosGit:
Is this done via autofix from ruff? Don't know if I want to scroll through 411 files.

henriquegemignani:
With a few here and there where I had to do a manual fix, yes

ThanatosGit (approved):
> With a few here and there where I had to do a manual fix, yes

Then I shamelessly approve without watching every file. I assume this will introduce a lot of merge conflicts in the open PRs, just like the last merged ones did here

PR #4706, opened 2023-07-06
Add Dread Multiworld support

ThanatosGit:


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [6.1.0] - 2023-??-??
+
+- Added: Multiworld support for Dread.
```
duncathan, line 10: this should probably be marked as major, and in the dread category

randovania/game_connection/connector/dread_remote_connector.py
```diff
@@ -66,6 +107,21 @@ def new_player_location_received(self, state_or_region: str):
                                         None)
         self.PlayerLocationChanged.emit(PlayerLocationEvent(self.current_region, None))
 
+    def new_collected_locations_received(self, new_indices: bytes):
+        locations = set()
+        if new_indices.startswith(b"locations:"):
+            index = 0
+            for c in new_indices[len(b"locations:"):]:
```
duncathan, line 114: might be best to split `b'locations:'` into a variable to prevent using the same magic string by hand twice

randovania/game_connection/connector/dread_remote_connector.py
```diff
@@ -84,3 +140,56 @@ def new_inventory_received(self, json_string: str):
         self.last_inventory = inventory
         self.InventoryUpdated.emit(inventory)
 
+    @asyncSlot()
+    async def new_received_pickups_received(self, new_received_pickups: str):
+        new_recv_as_int = int(new_received_pickups)
+        self.logger.warning(f"Received Pickups: {new_received_pickups}")
```
duncathan, line 146: why warning?

randovania/game_connection/connector/dread_remote_connector.py
```diff
@@ -84,3 +140,56 @@ def new_inventory_received(self, json_string: str):
...
+    async def set_remote_pickups(self, remote_pickups: tuple[PickupEntryWithOwner, ...]):
+        self.remote_pickups = remote_pickups
+        await self.receive_remote_pickups()
+
+    async def receive_remote_pickups(self) -> None:
+        remote_pickups = self.remote_pickups
+        inventory = self.last_inventory
+
+        # in that case we never received the numbers (at least 0) from the game
+        if self.received_pickups == -1 or self.inventory_index == -1:
```
duncathan, line 160: would it be possible/appropriate to use `None` for these instead of `-1`?

randovania/game_connection/executor/dread_executor.py
```diff
@@ -50,9 +51,21 @@ class DreadExecutorToConnectorSignals(QObject):
...
+            if not Scenario.CurrentScenarioID then return "not-in-game" end
+            local r,v,i,p = {},0,1,Game.GetPlayerBlackboardSectionName()
+            for _,t in ipairs(RL.Pickups) do
+                if Blackboard.GetProp(p,t) then v=v+i end
+                i=i*2;if i>=256 then table.insert(r,string.char(v));v=0;i=1 end
+            end
+            if i>1 then table.insert(r,string.char(v)) end
+            RL.SendIndices("locations:"..table.concat(r))
+        end""").strip(),
+        f"for i=1,{game.region_list.num_pickup_nodes} do RL.Pickups[i]='' end",
```
duncathan, line 68: we should move any substantial amounts of Lua like this into `.lua` files and load from there imo

randovania/game_connection/connector/dread_remote_connector.py
```diff
@@ -66,6 +107,21 @@ def new_player_location_received(self, state_or_region: str):
...
+        locations = set()
+        if new_indices.startswith(b"locations:"):
+            index = 0
+            for c in new_indices[len(b"locations:"):]:
+                for i in range(8):
+                    if c & (1 << i):
+                        locations.add(PickupIndex(index))
+                    index += 1
+        else:
+            self.logger.warning(f"Unknown response: {new_indices}")
```
duncathan, line 120: dark has taught mii that it's best practice to avoid fstrings in logging, since it'll always evaluate the fstring even if the log is ignored at the current log level. another form of string formatting would be more appropriate for this and other logging calls throughout the pr

duncathan (changes requested):

randovania/game_connection/connector/dread_remote_connector.py
ThanatosGit: Probably some leftover from my tests way back.
Do we need this at all or should I just remove the line? Or more general: Do we want to keep any of the logging from this pr?

henriquegemignani: Keep at first at debug level, so it shows in the log files.

randovania/game_connection/connector/dread_remote_connector.py
ThanatosGit: Hmm what would be the proper way for doing that? The code is send in chunks because of the buffer limit of 4096. It get blocks of `all_code` and concats them as long as they not exceed the limit.
If I just move this to one lua file, I have to make sure that it is still in some sort of blocks.
And also the dynamcially created parts from the region_list or resource_database of the game needs to be filled.

randovania/game_connection/executor/dread_executor.py
ThanatosGit: Hmm what would be the proper way for doing that? The code is send in chunks because of the buffer limit of 4096. It get blocks of `all_code` and concats them as long as they not exceed the limit.
If I just move this to one lua file, I have to make sure that it is still in some sort of blocks.
And also the dynamcially created parts from the region_list or resource_database of the game needs to be filled.

duncathan: 1) just split it into multiple files that's nbd
2) you could use ODR's lua editing library to handle the codegen parts

ThanatosGit: Tried to use `replace_lua_template` of odr's `lua_util`. Just realised the first parameter is always a file within odr's `Path(__file__).parent.joinpath("templates")` 😥

duncathan: ah, dang, just slap a `# FIXME` on it for now then

ThanatosGit: Adressed the requested changes. Should be good now?

henriquegemignani (approved):

randovania/games/dread/assets/lua/bootstrap_locations.lua
```diff
@@ -0,0 +1 @@
+for n,i in pairsTEMPLATE("pairs")do RL.Pickups[i]=RandomizerPowerup.PropertyForLocation(TEMPLATE("location")..n) end
```
duncathan, line 1: should this not be `pairs(TEMPLATE("pairs"))`? or am I missing something

henriquegemignani: In Lua `func({...})` is the same as `func{...}`, which I think is the case here?

ThanatosGit: Ah, I could move on of those `{}` to the lua if you like.
Currently it's done via 
`replacements["pairs"] = "{{{}}}".format(",".join(entries))`
so it creates something like
`for n,i in pairs{Item_EnergyTank001=4, SHORTENED item_missiletankplus_000=16,item_energyfragment_000=19}do RL.Pickups[i]=RandomizerPowerup.PropertyForLocation('s010_cave_'..n) end`

duncathan: ah okay so it's done this way to save a couple bytes? :+1:

duncathan: moving the outermost pair of braces into the Lua file would be a bit more readable yeah

ThanatosGit: Done

duncathan (approved):

PR #4711, opened 2023-07-06
Add add-new-game command

henriquegemignani:
Depends on #4710.


randovania/cli/commands/new_game.py
```diff
@@ -0,0 +1,379 @@
...
+        )
+
+        VersionedPreset.with_preset(new_preset).save_to_file(
+            _GAMES_PATH.joinpath(new_game.value, "presets", path)
+        )
+
+
+def new_game_command_logic(args):
+    enum_name: str = args.enum_name
+    enum_value: str = args.enum_value
```
duncathan, line 293: this should prob be regex checked too. existing games match `[a-z0-9_]+`

duncathan (changes requested):
please update the new game guide in docs as well!

duncathan (approved):

PR #4715, opened 2023-07-07
Teleporters: Remove extra flags, split echoes into two types

henriquegemignani:
Echoes: Create Elevator dock type, for the actual editable elevators.
Dread/Corruption: Make non-editable teleporters into Other docks


duncathan (approved):

randovania/gui/tracker_window.py
```diff
@@ -492,17 +492,10 @@ def setup_elevators(self):
         region_list = self.game_description.region_list
         nodes_by_region: dict[str, list[DockNode]] = collections.defaultdict(list)
 
-        areas_to_not_change = {
-            "Sky Temple Gateway",
-            "Sky Temple Energy Controller",
-            "Aerie Transport Station",
-            "Aerie",
-        }
```
ThanatosGit, line 500: `tracker_window.py` my favorite file...this can be removed because it was only relevant for the teleporters and not the elevators?

ThanatosGit (approved):

randovania/gui/tracker_window.py
henriquegemignani: this can be removed because a few **years** ago these were made shuffleable and should've been removed back then

PR #4721, opened 2023-07-08
AM2R: hints preperation

Miepee:
This does some preperation for am2r hints. In particular, changeable settings for DNA (which newly placed objects will provide), and changeable settings for Ice Beam (which the last chozo log will provide)


randovania/games/am2r/layout/hint_configuration.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class IceBeamHintMode(BitPackEnum, Enum):
+    DISABLED = "disabled"
+    HIDE_AREA = "hide-area"
+    PRECISE = "precise"
+
+    @classmethod
+    def default(cls) -> "IceBeamHintMode":
+        return cls.PRECISE
```
duncathan, line 26: why have two identical classes?

randovania/games/am2r/exporter/hint_namer.py
```diff
@@ -0,0 +1,124 @@
...
+    area = region_list.nodes_to_area(pickup_node)
+    if hide_region:
+        return area.name
+    else:
+        return region_list.area_name(area)
+
+
+def colorize_text(color: AM2RColor, text: str, with_color: bool):
+    if with_color:
+        return f"{color.value}{text}{AM2RColor.WHITE.value}"
```
duncathan, line 39: huh, does am2r really color its text this way? that's remarkably similar to dread's system lol

randovania/games/am2r/gui/preset_settings/am2r_hints_tab.py
```diff
@@ -0,0 +1,55 @@
...
+        super().__init__(editor, game_description, window_manager)
+        self.setupUi(self)
+
+        self.hint_layout.setAlignment(QtCore.Qt.AlignTop)
+
+        for i, art_hint_mode in enumerate(ArtifactHintMode):
+            self.hint_artifact_combo.setItemData(i, art_hint_mode)
+        self.hint_artifact_combo.currentIndexChanged.connect(self._on_art_combo_changed)
+
+        for i, psuit_hint_mode in enumerate(IceBeamHintMode):
```
duncathan, line 27: ~~damn when did am2r add the phazon suit~~

randovania/games/am2r/layout/hint_configuration.py
Miepee: You mean just using a `ItemHintMode` class, which is the type for both dna and ice beam?
If yes, sound neat, but that should then probably be refactored for Prime1 too, as it was the place where i was copying this all from
https://github.com/randovania/randovania/blob/main/randovania/games/prime1/layout/hint_configuration.py

randovania/games/am2r/exporter/hint_namer.py
Miepee: AM2R doesn't color log text at all. I'll have to write my own drawing text routine, which needs to be backwards compatible with the existing text in the game, and figured this was a good enough solution.

duncathan (approved):

duncathan:
conflicts

PR #4729, opened 2023-07-09
Automatically create new worlds when importing in a session

henriquegemignani:


randovania/gui/multiplayer_session_window.py
```diff
@@ -522,11 +523,11 @@ async def _should_overwrite_presets(self, parameters: GeneratorParameters, perma
...
             source_name = "game file"
 
-        if parameters.player_count != len(self._session.worlds):
+        if parameters.world_count > len(self._session.worlds):
             await async_dialog.warning(
                 self, "Incompatible permalink",
-                f"Given {source_name} is for {parameters.player_count} players, but "
-                f"this session only have {len(self._session.worlds)} rows.")
+                f"Given {source_name} is for {parameters.world_count} worlds, but "
+                f"this session has {len(self._session.worlds)} rows.")
```
ThanatosGit, line 530: Wouldn't it be better to say `worlds` instead of `rows`?

ThanatosGit (changes requested):

randovania/gui/multiplayer_session_window.py
henriquegemignani: I saw the players but not the rows :p

ThanatosGit (approved):

PR #4739, opened 2023-07-10
AM2R Starter preset improvements

Miepee:
Makes launchers required (whoo!), reduces dna and better misc  options

### DIFF
--- a/randovania/games/am2r/presets/starter_preset.rdvpreset
+++ b/randovania/games/am2r/presets/starter_preset.rdvpreset
@@ -143,7 +143,13 @@
                         "Power Bomb Door",
                         "Super Missile Door"
                     ],
-                    "can_change_to": []
+                    "can_change_to":
+                    [
+                        "Normal Door",
+                        "Missile Door",
+                        "Power Bomb Door",
+                        "Super Missile Door"
+                    ]
                 }
             }
         },
--- a/randovania/games/am2r/presets/starter_preset.rdvpreset
+++ b/randovania/games/am2r/presets/starter_preset.rdvpreset
@@ -107,21 +107,21 @@
                         5
                     ],
                     "pickup_count": 44,
-                    "requires_main_item": false
+                    "requires_main_item": true
                 },
                 "Super Missile Expansion": {
                     "ammo_count": [
                         2
                     ],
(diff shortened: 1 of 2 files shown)


randovania/games/am2r/presets/starter_preset.rdvpreset
```diff
@@ -136,7 +136,13 @@
...
                 "door": {
-                    "can_change_from": [],
+                    "can_change_from":
+                    [
+                        "Normal Door",
+                        "Missile Door",
+                        "Power Bomb Door",
+                        "Super Missile Door"
+                    ],
                     "can_change_to": []
```
duncathan, line 146: you should at least include the unlocked weakness in the default change_to settings - otherwise it'll be unchecked in the UI and greyed out, but it'll always still be an option regardless. more generally, it's conventional to enable all the vanilla door types in the default preset, and disable any custom types. so long as the actual mode is set to vanilla, enabling these doors will have no effect on generation

test/games/am2r/generator/test_am2r_pool_creator.py
```diff
@@ -52,14 +52,9 @@ def test_am2r_pool_creator(am2r_game_description, preset_manager):
...
         c("The Nest", "Depths Omega Nest East", "Pickup (Omega)"),
-        c("The Tower", "Exterior Zeta Nest West", "Pickup (Zeta)"),
-        c("Main Caves", "Mines Alpha Nest", "Pickup (Alpha)"),
-        c("The Tower", "Gamma Nest South East", "Pickup (Gamma)"),
-        c("The Nest", "Depths Omega Nest South West", "Pickup (Omega)"),
-        c("Hydro Station", "Exterior Alpha Nest", "Pickup (Alpha)"),
     ]
 
-    assert len(results.to_place) == wl.num_pickup_nodes - 31 - len(results.assignment)
+    assert len(results.to_place) == wl.num_pickup_nodes - 36 - len(results.assignment)
```
duncathan, line 57: you can safely use `len(locations)` here instead of a magic number since you're already asserting that the contents of the list are what you expect

duncathan (changes requested):

duncathan (approved):

PR #4752, opened 2023-07-12
Remove any usage of random numbers in pool creation

henriquegemignani:
Randomness being involved in the creation of a pickup pool is wrong, as we'd need to be able to reproduce the random number generator used when loading an rdvgame.


randovania/games/am2r/generator/pool_creator.py
```diff
@@ -35,42 +52,8 @@ def artifact_pool(game: GameDescription, configuration: AM2RConfiguration, rng:
...
-
-_boss_items = [
-    "oItemM_111",
-    "oItemJumpBall",
-    "oItemSpaceJump",
-    "oItemPBeam",
-    "oItemIBeam",
-    "oItemETank_50"
-]
+    return PoolResults(keys_to_shuffle, {}, starting_keys)
```
Miepee, line 59: For AM2R I need to shuffle the `keys` list, [in order to not have very predictable hints in-game](https://github.com/randovania/randovania/pull/4721/files#diff-beb11d84256c963bed4aa3521fc664616f21ef0d504cf3f04e42af55e07db312). What's the best place to do that now in / should that be done differently in any way?
My current naive line of thought is to just keep this here as-is for the gui (and whatever else needs it) to work, and then in `assign_pool_results` reevaluate again which dna numbers are starting.

henriquegemignani:
> This won't break rdvgame imports in any way that they weren't already broken, right?

I sure hope so, and the fact I didn't have to touch any of the rdvgame used in the tests does help a bit for that.

randovania/games/am2r/generator/pool_creator.py
henriquegemignani: My counter question is why your hints are predictable is that list is not sorted?

Miepee: The way I currently have hints implemented in AM2R, is that each area has a hint location, that gives out hints for DNA which would all normally be in that area. So the hint location in Area 1 will always give out hints for DNA 2, 3, 4, 5. If this list is now sorted, and you have a preset with 5 shuffled dna, that means that dna 1, 2, 3, 4, 5 are shuffled. Which means in all seeds generated from that preset, the only hint locations you ever need to visit is the one in area 1, and the one that spoils the location for dna 1. All other hint locations would just give you the "you already start with this item" hint, thus becoming useless.

If the list is shuffled, then the keys are shuffled too, so on every seed generated from that preset, you'd have to visit different hint locations to get all needed hints, and not just always the same locations.

henriquegemignani: Addressed by shuffling the keys before assigning them.

PR #4766, opened 2023-07-13
Fix creating worlds for new permalink/rdvgame

henriquegemignani:


duncathan:
What exactly was broken here?

henriquegemignani:
The entire thing.
it's supposed to create new worlds if the permalink/layout had more worlds than the session. The check was inverted........

duncathan (approved):
whoops!

PR #4774, opened 2023-07-13
pytest styling

henriquegemignani:


test/generator/test_dock_weakness_distributor.py
```diff
@@ -64,8 +64,8 @@ def test_distribute_pre_fill_weaknesses_swap(empty_patches):
     assert list(result.all_weaknesses_to_shuffle()) == []
 
 
-def test_distribute_pre_fill_weaknesses_swap_force_two_way(empty_patches, force_blank_two_way,
-                                                           monkeypatch):
+@pytest.mark.usefixtures("_force_blank_two_way")
```
duncathan, line 67: what's this change about?

duncathan (approved):

test/generator/test_dock_weakness_distributor.py
henriquegemignani: Two of the style rules are:
- fixtures have names starting with `_` iff they don't have return values
- use `usefixtures` for fixtures without return values

PR #4775, opened 2023-07-13
add Dread progressive tracker layout

duncathan:
Also adds support for custom disabled icons in tracker layouts (useful for progressive items)

image
image


randovania/gui/widgets/item_tracker_widget.py
```diff
@@ -102,9 +114,25 @@ def __init__(self, tracker_config: dict):
...
+                labels.append(disabled_image)
             for label in labels:
-                self._layout.addWidget(label, element["row"], element["column"], row_span, col_span)
+                self._layout.addWidget(
+                    label,
+                    element["row"],
+                    element["column"],
+                    row_span,
+                    col_span,
+                    QtCore.Qt.AlignmentFlag.AlignCenter
```
duncathan, line 134: i didn't end up using it in the dread layout, but this change means that if an image has a `col_span` greater than 1, it'll end up centered horizontally instead of aligned left. makes no real difference for `col_span` of 1

henriquegemignani (approved):

PR #4785, opened 2023-07-14
Refactor offworld models

duncathan:
Offworld models are now integrated into RDV core. The pickup database now defines a default model to use for offworld items in its game, and pickups in that database have a dict mapping games to model names. This dict is used to determine `PickupModel.name` when exporting pickups. `ExportedPickupDetails` has a new field, `original_model`, which can be used by games which have *very* special handling of offworld models (e.g. prime/echoes crossgame models).


henriquegemignani (approved):

randovania/games/prime1/pickup_database/pickup-database.json
```diff
@@ -441,5 +518,6 @@
             "allows_negative": true
         }
     },
-    "default_pickups": {}
+    "default_pickups": {},
+    "default_offworld_model": "Nothing"
```
toasterparty, line 522: This will be `"Gamecube"` in the next randomprime release

PR #4786, opened 2023-07-14
Dread - Ferenia Bomb Tricks

dyceron:
Crossing the water gap in the Ferenia EMMI Zone with Bombs or Cross Bombs.


randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -2143,6 +2143,12 @@ Extra - asset_id: collision_camera_032
           All of the following:
               Gravity Suit
               Simple IBJ or Use Spin Boost
+          All of the following:
+              # Ledge Warp (Advanced Movement) makes this viable https://www.youtube.com/watch?v=ND5zby3zaz8
```
ThanatosGit, line 2147: Do we have a consensus about which form of links we use? Most of the links use the shortened version but not all.

randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -2143,6 +2143,12 @@ Extra - asset_id: collision_camera_032
...
+              Diagonal Bomb Jump (Hypermode) and Infinite Bomb Jump (Hypermode) and Knowledge (Advanced) and Lay Normal Bomb
+          All of the following:
+              # Ledge Warp is optional but makes attempts easier https://www.youtube.com/watch?v=AqlUANN4F_k
+              Cross Bomb and Movement (Advanced) and Can Slide
```
Tyranisaur, line 2151: Surely this has to be just "Slide" instead of the template.

Tyranisaur (changes requested):

randovania/games/dread/json_data/Ferenia.txt
dyceron: Yeah I noticed that when I added this the line was so long. I can shorten it to make it more readable.

randovania/games/dread/json_data/Ferenia.txt
dyceron: Oops I definitely forgot how the template was laid out. Yes this should be slide and lay cross bomb.

PR #4800, opened 2023-07-16
Fix integrity checks

ThanatosGit:
Fix for the prime games.

Open questions:
1. Delay SM and SR as they are not stable and can be freely changed without migration? (Though I did one for Corruption).
2. ~~`test/generator/test_base_patches_factory.py` has a TODO because the connecitons looks kinda weird to me.~~
3. ~~Echoes still has the following error `Torvus Bog/Torvus Temple/Elevator Translator Scan is not a Dock Node, naming suggests it should be.` If we want to rename it, it needs a migration for `translator_requirement` dict. What would be a good new name?~~


henriquegemignani:
3. is a in-room elevator that must be scanned to activate.

Maybe the integrity check should check if a node is named `{type} (to/from` instead of just starting with `type`

ThanatosGit:
> 3. is a in-room elevator that must be scanned to activate.
> 
> Maybe the integrity check should check if a node is named `{type} (to/from` instead of just starting with `type`

I'm fine with that but woudl move that into a different PR then.

Regarding 2.: I just realised that the Elevators are really connected in this way:
```
"Temple Grounds", "Sky Temple Gateway", "Elevator to Great Temple" -> "Great Temple", "Sky Temple Energy Controller", "Save Station"

"Great Temple", "Sky Temple Energy Controller", "Elevator to Temple Grounds" -> "Temple Grounds", "Sky Temple Gateway", "Spawn Point/Front of Teleporter"
```
I change that such that the elevators are connected to each other and add it to the `dock_connection_fixes` migration?

henriquegemignani:
The first probably doesn't matter, but the later is certainly because leaving is a one-way trip.
If you'd connect directly to the dock, logically you'd be able to take it right after you leave. (one-way elevators can make the trip back go somewhere else)

ThanatosGit:
Where do you actually spawn in the game?
From `Temple Grounds` to `Great Temple` logically and currently you start at the `Save Station` and the connection to `Elevator to Temple Grounds` is not trivial (has a `Dark World Damage` constraint). If I change it to connect to the `Elevator to Temple Grounds`, there are lot of more constraints to reach the `Save Station` node.

Well, what's the conclusion on both of these elevators? Leave it as it is?

henriquegemignani:
In both cases, the answer is "depends". The spawn point of an area in Echoes changes depending on the active layers in the room. 

For Sky Temple Energy Controller, the game does some jank to try to spawn you in the elevator or the save station depending if you're coming from the elevator or if you loaded a save. In either case the dark damage there is small enough if that much damage matters, somewhere else will break harder. So where it is doesn't matter much.

For Sky Temple Gateway it also depends on whether the gate has been opened, but it's not that different. But the important thing is that the portal is just not there, even if you take a portal from elsewhere.

ThanatosGit:
That sounds to me that I better leave it untouched. I mean, the elevators were always connected to the default node, so it wont't change anything if I'm not mistaken here.

henriquegemignani:
@toasterparty Do you have opinions on renaming the Elevator nodes in Prime 1?

toasterparty:
No

henriquegemignani:
"No" as in no preferences, or "No" as in "don't do that"?

toasterparty:
No opinion. Change away

henriquegemignani (approved):

PR #4801, opened 2023-07-16
AM2R: add offworld models

Miepee:
Adds offworld models to AM2R

While this is probably a good enough base, some thoughts i had earlier were:
- AM2R's power grip could potentially be represented primes/dreads grapple beam
- AM2R's spring ball could potentially be represented by morph ball models
- AM2R's space jump could potentially be represented in echoes by gravity booster. Currently it's space jump boots for consistency

Adding am2r as offworld models to other games will follow in a different PR

### DIFF
--- a/randovania/games/am2r/pickup_database/pickup-database.json
+++ b/randovania/games/am2r/pickup_database/pickup-database.json
@@ -100,7 +100,11 @@
             "pickup_category": "morph_ball",
             "broad_category": "morph_ball_related",
             "model_name": "Bombs",
-            "offworld_models": {},
+            "offworld_models": {
+                "prime1": "Morph Ball Bomb",
+                "prime2": "MorphBallBomb",
+                "dread": "powerup_bomb"
+            },
             "progression": [
                 "Bombs"
             ],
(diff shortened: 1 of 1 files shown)


randovania/games/am2r/pickup_database/pickup-database.json
```diff
@@ -184,7 +203,11 @@
             "pickup_category": "movement",
             "broad_category": "movement",
             "model_name": "Speed Booster",
-            "offworld_models": {},
+            "offworld_models": {
+                "prime1": "Boost Ball",
+                "prime2": "BoostBall",
+                "dread": "item_speedboostupgrade"
```
duncathan, line 209: this should be `powerup_speedbooster`

randovania/games/am2r/pickup_database/pickup-database.json
```diff
@@ -232,7 +267,10 @@
             "pickup_category": "beam",
             "broad_category": "beam_related",
             "model_name": "Ice Beam",
-            "offworld_models": {},
+            "offworld_models": {
+                "prime1": "Ice Beam",
+                "dread": "powerup_icemissile"
```
duncathan, line 272: I think it's better to have no model than to use ice missiles for ice beam tbh, but that one's up to you

duncathan (changes requested):

randovania/games/am2r/pickup_database/pickup-database.json
Miepee: Every other beam has an equivalent in dread right now, so even though ice missiles might not be a 100% perfect fit, it's close enough for me

duncathan (approved):

PR #4802, opened 2023-07-16
Adds AM2R offworld models to primes/dread

Miepee:
Apart from what title says, also adds dread's charge beam to echoes charge beam as I noticed it was missing

### DIFF
--- a/randovania/games/dread/pickup_database/pickup-database.json
+++ b/randovania/games/dread/pickup_database/pickup-database.json
@@ -96,6 +96,7 @@
             "broad_category": "beam_related",
             "model_name": "powerup_widebeam",
             "offworld_models": {
+                "am2r": "Spazer Beam",
                 "prime1": "Power Beam"
             },
             "progression": [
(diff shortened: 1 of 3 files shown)


randovania/games/dread/pickup_database/pickup-database.json
```diff
@@ -517,6 +538,7 @@
             "broad_category": "misc",
             "model_name": "powerup_doublejump",
             "offworld_models": {
+                "am2r": "Space Jump",
```
duncathan, line 541: am2r high jump uses the spin boost model in dread, so maybe it should be the other way around for spin boost in am2r? not a huge deal since most people use progressive spin

duncathan (changes requested):

randovania/games/dread/pickup_database/pickup-database.json
Miepee: Whoops, good catch

duncathan (approved):

PR #4814, opened 2023-07-17
Adds a "Joined Session?" filter and column for multiworld browse list

ThanatosGit:
Filtering for already joined sessions.
image

### DIFF
--- a/randovania/gui/dialog/multiplayer_session_browser_dialog.py
+++ b/randovania/gui/dialog/multiplayer_session_browser_dialog.py
@@ -29,14 +29,12 @@ class MultiplayerSessionBrowserDialog(QDialog, Ui_MultiplayerSessionBrowserDialo
         common_qt_lib.set_default_window_icon(self)
         self.network_client = network_client
 
-        self.item_model = QtGui.QStandardItemModel(0, 6, self)
-        self.item_model.setHorizontalHeaderLabels(["Name", "Is Member?", "State", "Players", "Worlds", "Password?",
-                                                   "Creator","Creation Date"])
+        self.item_model = QtGui.QStandardItemModel(0, 8, self)
+        self.item_model.setHorizontalHeaderLabels(["Name", "Is Member?", "Join Date", "State", "Players", "Worlds",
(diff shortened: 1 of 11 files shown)


ThanatosGit:
> "Joined Session?" reads a bit weird for me. Maybe "Is member?", but I'd let other people pitch as I suck at text.

Haha, I feel that. I had "In Session?" then switched to "Joined Session?" and then I thought about changing it again.
I'm also fine with "Is member?"

henriquegemignani (changes requested):
Add some worlds to `test_list_sessions` so you can check it works properly.

Uncle-Reggie:
A couple suggestions for the “Joined session” feature:

> Could fold it into the “State” column as a new status “Joined”

If keeping it a separate column is preferred:

> “Joined Date” - put the date/time the player joined the session, bonus feature of being able to sort by joined date just like the room creation date

> “Prior Activity” - either “Yes” or the above date suggestion but for most recent time the player was in the session window

henriquegemignani:
Joined date should be easy to add, but last activity is not data we have right now. Closest is the world activity, but that's a bit tricky and potentially slow to fetch.

ThanatosGit:
> A couple suggestions for the “Joined session” feature:
> 
> > Could fold it into the “State” column as a new status “Joined”

Hmmm the state is currently `Setup`, `In-Progress` or `Finished` I don't think that `Joined` really belongs here. Also it would be much harder to filter for it if you have some weird combination field which says `Setup (Joined)` or `In-Progress (Not joined)`

The joined date can be easily added.

I will add it and write some proper tests for the whole filtering stuff later or tomorrow. 

If we add even more fields after that, I think at some point we need a `Hide/Show column` feature as configuration for future™

ThanatosGit:
Still no tests for filtering but got distracted by `create_list_entry` because it had only one usage.
The list entry is collecting stuff all over the place, I decided to do one query and construct it with peewee and a helper function directly. 
`is_user_in_session` also only has one usage where the attribute does not exists.

image

henriquegemignani (approved):

PR #4817, opened 2023-07-17
Properly disconnect from server if rejected on session restore

henriquegemignani:
Reliable way of reproducing:
- Start local dev server with debug enabled
- Start client, connect to server and open a multiplayer session
- Update SERVER_API_VERSION to a bigger value, which causes the server to reload disconnecting user and no longer allow connection
- Do something in a connected game which causes a world_sync to happen
- World sync -> server_call -> connect_to_server. It connects, but then is rejected due to version. Sync fails.
- Another World sync -> server_call -> connect_to_server. This logs "sio is already connected" in _internal_connect_to_server, and then the server_call does not succeed and times out.

### DIFF
--- a/randovania/network_client/network_client.py
+++ b/randovania/network_client/network_client.py
@@ -142,7 +142,8 @@ class NetworkClient:
         aiohttp.ClientSession.ws_connect = wrap_ws_connect
 
         self._connection_state = ConnectionState.Disconnected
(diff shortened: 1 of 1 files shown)


henriquegemignani:
Still not great, and I had this disable reconnect

henriquegemignani:
Eh, this is fine.

duncathan (approved):

PR #4820, opened 2023-07-18
Add `Skippable` Cutscene Mode

toasterparty:
Hidden unless you turn on experimental

Update Randomprime to v1.14.1

### DIFF
--- a/randovania/games/prime1/gui/preset_settings/prime_patches_qol.py
+++ b/randovania/games/prime1/gui/preset_settings/prime_patches_qol.py
@@ -39,6 +39,30 @@ class PresetPrimeQol(PresetTab, Ui_PresetPrimeQol):
         self.cutscene_combo.setItemData(1, LayoutCutsceneMode.COMPETITIVE)
         self.cutscene_combo.setItemData(2, LayoutCutsceneMode.MINOR)
         self.cutscene_combo.setItemData(3, LayoutCutsceneMode.MAJOR)
+        self.cutscene_combo.setItemData(4, LayoutCutsceneMode.SKIPPABLE)
+
+        if editor._options.experimental_settings:
+            # ruff made me do it this way
+            self.cutscene_label.setText("""
+                <html>
+                    <head/>
+                    <body>
+<p><span style=" font-weight:600;">Original</span>
+: No changes to the cutscenes are made.</p>
+<p><span style=" font-weight:600;">Competitive</span>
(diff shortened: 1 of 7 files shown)


randovania/games/prime1/gui/preset_settings/prime_patches_qol.py
```diff
@@ -39,6 +39,30 @@ def __init__(self, editor: PresetEditor, game_description: GameDescription, wind
...
+: Removes cutscenes that don't affect the game too much when removed.</p>
+<p><span style=" font-weight:600;">Major</span>
+: Allows you to continue playing the game while cutscenes happen.</p>
+<p><span style=" font-weight:700;">Skippable</span>
+: Keeps all of the cutscenes in the game, but makes it so that they can be skipped with the START button.</p>
+                    </body>
+                </html>
+            """)
+        else:
+            self.cutscene_combo.removeItem(4)
```
duncathan, line 64: shouldn't this else block also reset the label to its original text?

randovania/games/prime1/gui/preset_settings/prime_patches_qol.py
```diff
@@ -60,7 +84,11 @@ def persist(value: bool):
 
     def _on_cutscene_changed(self, value: LayoutCutsceneMode):
         with self._editor as editor:
-            editor.set_configuration_field("qol_cutscenes", value)
+            try:
+                editor.set_configuration_field("qol_cutscenes", value)
+            except Exception:
```
duncathan, line 89: this should ideally be catching the specific relevant type of Exception rather than Exception itself

setup.cfg
```diff
@@ -39,7 +39,7 @@ install_requires =
     attrs>=21.4.0
     frozendict
     dulwich>=0.20
-    py_randomprime>=1.13.0
+    py_randomprime>=1.14.1
```
duncathan, line 42: ~~what happened to 1.14.2~~

duncathan (changes requested):

randovania/games/prime1/gui/preset_settings/prime_patches_qol.py
toasterparty: A current limitation of experimental mode is that it's changes do not apply until the preset editor tab is changed.

setup.cfg
toasterparty: The CI uses the oldest tag if there are duplicates I guess. The pypi package is .1

duncathan (approved):

PR #4822, opened 2023-07-18
Fixes for PySide 6.5

henriquegemignani:


randovania/gui/lib/qt_network_client.py
```diff
@@ -97,10 +98,16 @@ class QtNetworkClient(QWidget, NetworkClient):
...
 
     def __init__(self, user_data_dir: Path):
-        super().__init__()
-        NetworkClient.__init__(self, user_data_dir.joinpath("network_client"), randovania.get_configuration())
-        from randovania.gui.lib import common_qt_lib
-        common_qt_lib.set_default_window_icon(self)
+        configuration = randovania.get_configuration()
+        user_data_dir = user_data_dir.joinpath("network_client")
+
+        if PySide6.__version_info__ >= (6, 5):
```
duncathan, line 104: will this eventually be removed?

henriquegemignani: When we can go to PySide 6.5 without breaking Linux

PR #4836, opened 2023-07-19
Don't install server extras when preparing to create an installer

henriquegemignani:


duncathan (dismissed):
it looks like the executables need graphviz?

henriquegemignani:
> it looks like the executables need graphviz?

graphviz is listed only as a GUI dependency, but there's a CLI that uses them. So our code/dependencies are just wrong. Gotta move things around.

PR #4841, opened 2023-07-20
Partially remove default_node

ThanatosGit:
Definitely goes into the "I have no idea what I'm doing", especially with some of the reverts I had to do.
Should also go over the changes myself again because of that.

What this PR should acheive:
- remove `default_node` from ui and all usages
- except for keeping them in the db without any intended way to change
- `default_node`s are not removed from db because of the integrity check skipping strongly connected components for `default_node`s  (only dread has a lot of them)
- as `default_node`s will be logically gone, elevators now needs to be connected directly to a node. It is also a logical consequence of making elevators `DockNode`s and they are already connected to nodes in the db.

### DIFF
--- a/randovania/gui/data_editor.py
+++ b/randovania/gui/data_editor.py
@@ -129,7 +129,6 @@ class DataEditorWindow(QMainWindow, Ui_DataEditorWindow):
         self.node_details_label.linkActivated.connect(self._on_click_link_to_other_node)
         self.node_heals_check.stateChanged.connect(self.on_node_heals_check)
         self.area_spawn_check.stateChanged.connect(self.on_area_spawn_check)
-        self.default_node_check.stateChanged.connect(self.on_default_node_check)
         self.node_edit_button.clicked.connect(self.on_node_edit_button)
         self.other_node_connection_swap_button.clicked.connect(self._swap_selected_connection)
         self.other_node_connection_edit_button.clicked.connect(self._open_edit_connection)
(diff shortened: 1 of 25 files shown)


randovania/gui/data_editor.py
```diff
@@ -355,7 +354,7 @@ def _on_click_link_to_other_node(self, link: str):
             self.focus_on_region_by_name(region_name)
             self.focus_on_area_by_name(area_name)
             if node_name is None:
-                node_name = self.current_area.default_node
+                node_name = self.current_area.nodes[0]
```
ThanatosGit, line 357: What is/was the use case of that?
Clicking a link to another node but the node has no name sounds kinda weird to me. I was about to dropping this case completly.

henriquegemignani: It's for having a link to an Area, not a Node.
For sure the teleporter description had these links, dunno if anything else.

randovania/layout/lib/teleporters.py
```diff
@@ -100,12 +100,11 @@ def ensure_has_locations(self, area_locations: list[NodeIdentifier], enabled: bo
 
 def _valid_teleporter_target(area: Area, node: Node, game: RandovaniaGame):
     if (game in (RandovaniaGame.METROID_PRIME, RandovaniaGame.METROID_PRIME_ECHOES) and
-            area.name == "Credits" and node.name == area.default_node):
+            area.name == "Credits" and node.name == "Event - Credits"):
```
henriquegemignani, line 103: This doesn't work for Echoes, the event is called something different.

randovania/gui/data_editor.py
ThanatosGit: Found only one `href` using node in this file. The other one was gone with the `TeleporterNode` => removed the check

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #4843, opened 2023-07-20
Optimize History tab with lots of actions

henriquegemignani:


randovania/game_description/db/region_list.py
```diff
@@ -119,16 +119,19 @@ def all_regions_areas_nodes(self) -> Iterable[tuple[Region, Area, Node]]:
...
 
         if distinguish_dark_aether:
             return region.correct_name(area.in_dark_aether)
         else:
             return region.name
 
     def region_name_from_node(self, node: Node, distinguish_dark_aether: bool = False) -> str:
-        return self.region_name_from_area(self.nodes_to_area(node), distinguish_dark_aether)
+        region = self.nodes_to_region(node)
+        return self.region_name_from_area(self.nodes_to_area(node), distinguish_dark_aether, region)
```
ThanatosGit, line 134: Could you tell me what is the purpose of that? The new parameter `region` is only used by `region_name_from_node` and now get's the region from the node instead of getting the area for the node and then the region from the area...but this should be equivalent?
I guess `nodes_to_region` is significant faster than `region_with_area` because it is only a lookup in a dict instead of iterating stuff.

ThanatosGit (approved):

randovania/game_description/db/region_list.py
henriquegemignani: It's faster, that's it. It happens to be high in the profiler and the double iteration is dumb.

PR #4852, opened 2023-07-21
AM2R: add differences description

Miepee:
No clue if Ruff will complain due to the very long line


duncathan:
ruff only cares about .py files

duncathan (approved):

PR #4857, opened 2023-07-21
Fix tracker and detect some incompatible configurations

henriquegemignani:


duncathan (approved):

duncathan:
does this fix need a changelog entry, or was it broken after 6.0?

henriquegemignani:
the brokenness came after 6.0 I believe. The dock lock rando thing is just declaring explicitly that it never worked.

PR #4862, opened 2023-07-22
AM2R: Export dialog prep work

Miepee:
This does some initial preperation work for AM2R's patcher.
The `_do_export_game` function is mostly stubbed, please do not pay as much attention to it.

### DIFF
--- a/randovania/games/am2r/exporter/game_exporter.py
+++ b/randovania/games/am2r/exporter/game_exporter.py
@@ -1,13 +1,8 @@
 from __future__ import annotations
 
 import dataclasses
-import json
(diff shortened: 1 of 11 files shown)


Miepee:
While looking through other games again, i noticed that the intended naming scheme for most of them *is* [ID]GameExporter etc. i thought blank's id was BlankGame and thus thought i accidentally forgot to remove that after copypasting. so i'll go revert that change
Will also add tests while i'm at it.

Miepee:
Both should be done now.

ThanatosGit (approved):

PR #4864, opened 2023-07-23
AM2R: Implement cosmetic options

Miepee:
Adds a few cosmetic options for am2r. Wondering whether the color rotation function should maybe be somewhere in some color lib.
grafik

### DIFF
--- a/randovania/games/am2r/gui/dialog/cosmetic_patches_dialog.py
+++ b/randovania/games/am2r/gui/dialog/cosmetic_patches_dialog.py
@@ -14,9 +14,9 @@ if TYPE_CHECKING:
 
     from randovania.layout.base.cosmetic_patches import BaseCosmeticPatches
 
-DEFAULT_HEALTH_COLOR = (255, 228, 2)
-DEFAULT_ETANK_COLOR = (112, 221, 250)
-DEFAULT_DNA_COLOR = (46, 208, 3)
+DEFAULT_HEALTH_COLOR = (255, 225, 0)
+DEFAULT_ETANK_COLOR = (112, 222, 250)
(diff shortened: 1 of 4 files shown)


ThanatosGit (changes requested):
As the cosmetic options now get some nice features, it would also be nice to have some tests for the dialog.

randovania/games/am2r/gui/dialog/cosmetic_patches_dialog.py
```diff
@@ -11,10 +14,28 @@
...
+    h = color.hue() + rotation
+    s = color.saturation()
+    v = color.value()
+    while h >= 360:
+        h -= 360
+    while h < 0:
+        h += 360
+
+    rotated_color = QColor.fromHsv(h, s, v)
+    return (rotated_color.red(), rotated_color.green(), rotated_color.blue())
```
ThanatosGit, line 32: We should maybe move this function somewhere that prime and am2r can access it since both implementations are identical

Miepee: Can I just be lazy for now and mark that as a `TODO`?
Also not sure where the location should be. I thought about `randovania/lib/color_lib.py` but making a whole class for one function seems a bit overkill.

ThanatosGit: I'm also fine with adding a `TODO`.

The helper function could be added later to `base_cosmetic_patches_dialog.py` as both are importing that module anyways or just to a new file in `gui\lib` as there are already other modules with only one helper function like `wait_dialog.py` or `error_message_box.py`

ThanatosGit (approved):

PR #4869, opened 2023-07-23
Prime: Fix tower of light access -> ruined shrine door

Miepee:
The door from ToL acces -> ruined shrine is a normal door.
grafik


dyceron (changes requested):
Add changelog

dyceron (approved):

PR #4871, opened 2023-07-23
AM2R: Add Patcher

Miepee:
Adds final piece of the puzzle to make AM2R patcher work.

Some points
- I don't know if yams should exist in the folder where I currently have it. If there's a better place for it, let me know.
- Regarding yams, while the folder itself is 5.5MB, ~4MB of those are libs, with the rest being the patcher itself. Thus for the future, diffs will hopefully be rather small when the patcher updates.
- yams itself doesnt have any releases yet, I wanted to make one after i got cosmetics implemented. the one in place here is taken from dirty commit
- please let me know on how to do that pythonnet importing madness better

### DIFF
--- a/randovania/games/am2r/exporter/game_exporter.py
+++ b/randovania/games/am2r/exporter/game_exporter.py
@@ -9,6 +9,7 @@ import subprocess
 import sys
 import tempfile
 from typing import TYPE_CHECKING
+from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


henriquegemignani:
Make a Python package for hosting YAMS. Maybe it even is responsible for loading it and running.

PR #4880, opened 2023-07-24
Arbitrary messages for dread

ThanatosGit:
Very simple patch to show a message. Message will be queued if the player is not in a menu and will be shown for 10 seconds.

This can make the message and granting an item out of sync, if there are multiple items in the queue to be received like shown in the video. That's a rather edge case, I don't really care about.
Especially when you consider that the messages will be in sync again as soon as you are long enough in the menu or receiving items stopped for a few seconds (which will happen eventually).

https://github.com/randovania/randovania/assets/117127188/a88a1fa8-8909-47fd-b120-ef6c4fae1389


randovania/game_connection/connector/dread_remote_connector.py
```diff
@@ -196,3 +196,7 @@ async def receive_remote_pickups(self) -> None:
 
         await self.executor.run_lua_code(execute_string)
         return
+
+    async def display_arbitrary_message(self, message: str):
+        execute_string = f"Game.AddSF(0, 'Scenario.QueueAsyncPopup', 'si', '{message}', 10.0)"
```
henriquegemignani, line 201: shouldn't it properly wrap message, so it doesn't break if it contains a `'`?

henriquegemignani (changes requested):

randovania/game_connection/connector/dread_remote_connector.py
ThanatosGit: Oops, yeah, Just `escaped_message = message.replace("'", "\\'")` to escape the single quotes or is there a fancy python way?

henriquegemignani: Can't we grab the lua_util from OPR?

ThanatosGit: OPR? :p
Maybe I have a brain lag but ODR is just doing this:
```
def lua_convert(data, wrap_strings: bool = False) -> str:
    if isinstance(data, list):
        return "{\n" + "\n".join(
            f"{lua_convert(item, wrap_strings)},"
            for item in data
        ) + "\n}"
    if isinstance(data, dict):
        return "{\n" + "\n".join(
            f"{key} = {lua_convert(value, wrap_strings)},"
            for key, value in data.items()
        ) + "\n}"
    if isinstance(data, bool):
        return "true" if data else "false"
    if data is None:
        return "nil"
    if wrap_strings and isinstance(data, str):
        return wrap_string(data)
    return str(data)

def wrap_string(data: str) -> str:
    return f'"{data}"'
```
If I wrap the message with that, a single quote would still break it.

henriquegemignani: Too many letters.

Hm... yeah, it would hm. I guess it'd also break multiworld stuff too, so I'm glad I restricted world names lol

henriquegemignani: Just escaping `'` should work for now

henriquegemignani (approved):

PR #4881, opened 2023-07-24
AM2R: Add Patcher

Miepee:
Attempt number 2!
Tested this on Linux, Jeff has been running 0.0.1 for a bit on Windows, I have absolutely no idea how this behaves on Mac.

### DIFF
--- a/randovania/games/am2r/exporter/game_exporter.py
+++ b/randovania/games/am2r/exporter/game_exporter.py
@@ -1,11 +1,16 @@
 from __future__ import annotations
 
 import dataclasses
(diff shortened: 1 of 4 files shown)


henriquegemignani (changes requested):
Add a test that mocks `patch_game` and asserts it's called.

henriquegemignani (approved):

Miepee:
Currently blocked by the fact that the dotnet runtime from am2r_yams keeps being loaded, and causing the linux runner to crash.

henriquegemignani (approved):

PR #4886, opened 2023-07-25
Better handle invalid presets

henriquegemignani:
- Disable menu actions that need a preset if it's invalid
- Add a clickable link that opens the preset dir when there's an error
- Show full error stack for unknown errors

image
image

### DIFF
--- a/randovania/gui/lib/common_qt_lib.py
+++ b/randovania/gui/lib/common_qt_lib.py
@@ -237,7 +237,13 @@ def set_clipboard(text: str):
     QtWidgets.QApplication.clipboard().setText(text)
 
 
-def open_directory_in_explorer(path: Path, fallback_dialog: tuple[str, str, QtWidgets.QWidget] | None = None):
+class FallbackDialog(typing.NamedTuple):
+    title: str
+    text: str
+    parent: QtWidgets.QWidget
+
+
+def open_directory_in_explorer(path: Path, fallback_dialog: FallbackDialog | None = None):
     try:
         if platform.system() == "Windows":
             os.startfile(path)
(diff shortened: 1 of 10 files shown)


randovania/gui/lib/common_qt_lib.py
```diff
@@ -237,13 +237,26 @@ def set_clipboard(text: str):
...
 
 
-def open_directory_in_explorer(path: Path):
-    if platform.system() == "Windows":
-        os.startfile(path)
-    elif platform.system() == "Darwin":
-        subprocess.run(["open", path])
-    else:
-        subprocess.run(["xdg-open", path])
+def open_directory_in_explorer(path: Path, fallback_dialog: tuple[str, str, QtWidgets.QWidget] | None = None):
```
duncathan, line 240: a NamedTuple for fallback dialog might be nice

randovania/gui/widgets/select_preset_widget.py
```diff
@@ -316,17 +325,40 @@ def on_preset_changed(self, preset: VersionedPreset | None):
...
+                    f"or delete this preset.</p>"
+                    f"{exception_desc}"
                 )
 
         self.create_preset_description.setText(description)
+        self._can_generate = can_generate
         self.CanGenerate.emit(can_generate)
 
+    def _on_click_create_preset_description(self, link: str):
+        info = re.match(r"^open-preset://([0-9a-f\-]{36})$", link)
```
duncathan, line 348: might as well `re.compile` this for reuse imo

duncathan (changes requested):

randovania/gui/widgets/select_preset_widget.py
henriquegemignani: I doubt this search is done more than once per app, and most of the time none. Better performance doing it here.

henriquegemignani:
It is, but the stack trace is pre-formatted so it doesn't line break, which causes the text above to not be either

duncathan (approved):

PR #4888, opened 2023-07-25
Dread: Diagonal Bomb Jump in Ferenia - Speedboost Slopes Maze

Tyranisaur:


ThanatosGit (approved):
Sometimes I'm really surprised what was out of logic the whole time

Tyranisaur:
To be fair, it seems exceptionally niche for this path to be required.

PR #4890, opened 2023-07-26
AM2R: Add missile-less fight to differences

Miepee:
See title, forgot to add it before


Miepee:
Thanks QT Creator for the horrible diff.
FWIW, the only thing i added was this line of text:
grafik

PR #4896, opened 2023-07-26
AM2R: Add new trick for missile less metroid fights

Miepee:
Needs review from other people


Miepee:
Have been going through druid/abyss/jeff for review, no complains

dyceron (approved):

PR #4897, opened 2023-07-26
Dread: Add ledge warps to release X faster

Tyranisaur:


randovania/games/dread/json_data/Ghavoran.txt
```diff
@@ -475,6 +475,8 @@ Extra - asset_id: collision_camera_007
               After Ghavoran - Burenia Transport Grapple Box or After Dairon - Power Switch 2
   > Event - Super Missile Rotatable Enky
       Destroy Enky
+  > Event - Super Missile Rotatable with Ledge Warp
+      Morph Ball and Ledge Warp (Intermediate)
```
XenoWars, line 479: I would personally tack on Highly Dangerous Logic to this, since you can lock out the three checks below permanently, and one of those items could say be screw attack with the seed expecting you to fight golzuna to get into hanubia.

XenoWars (changes requested):

randovania/games/dread/json_data/Ghavoran.txt
Tyranisaur: A better example would be, among these three items, you have both gravity suit and a required grapple beam.

Tyranisaur: It just crossed my mind that we could put an event in the early supers area, which can act as an alternative to highly dangerous logic.

duncathan: I'm thinking about this some more, and I'd actually rather prefer raising the trick difficulty to represent the danger involved with the trick. Ledge warps are an advanced enough trick that anyone enabling them in their preset will understand what they're getting into

duncathan: To be clear: I'd rather raise the trick difficulty than include highly dangerous logic here.

Tyranisaur: The part that makes me not want to use a difficulty rating for the purpose of the dangerous action information is that there are other setups for ledge warps, which seems like a much better usage of difficulty levels than reasoning about dangerous logic.

CHANGELOG.md
```diff
@@ -22,15 +22,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Changed: in Burenia - Main Hub Tower Middle, lowering the Spider Magnet Wall is now "dangerous" only when Highly Dangerous Logic is enabled. The connection from the bottom of the room to the Pickup Platform that uses Grapple Movement requires the Spider Magnet Wall to not be lowered now requires Highly Dangerous Logic. The randomizer currently doesn't have the necessary options to make this connection mandatory in any seeds anyway.
+- Changed: Most instances of pushing Wide Beam Blocks by using Wave Beam through walls now no longer need Wide Beam. Notable exception is Dairon - West Transport to Ferenia, from below.
+- Changed: Boss fight logic using Ice Missile without Super Missile is no longer an option, and effectively requires as many missiles as with normal Missiles.
+- Changed: Exclude Ghavoran door between FLipper Room and Elun Transport Access from being shuffled as a Grapple Beam door in Door Lock rando. This is to enable a Ledge Warp to flip the Spinner from below.
```
ThanatosGit, line 32: Could change that to `Flipper` instead of `FLipper`

ThanatosGit (changes requested):
Dunno if someone else wants to review that, otherwise I'm gonna merge it later today before it hangs here forever.

CHANGELOG.md
Tyranisaur: Fixed this in the merge commit that resolved the conflicts.

randovania/games/dread/json_data/Ghavoran.txt
```diff
@@ -524,14 +528,24 @@ Extra - asset_id: collision_camera_007
   > Door to Elun Transport Access
       Before Ghavoran - Super Missile Rotatable
   > Event - Super Missile Rotatable
-      Morph Ball and Before Ghavoran - Super Missile Rotatable
+      Morph Ball and Before Ghavoran - Super Missile Rotatable and Knowledge (Intermediate)
+  > Event - Super Missile Rotatable with Ledge Warp
+      All of the following:
+          Before Ghavoran - Super Missile Rotatable and Knowledge (Intermediate) and Ledge Warp (Intermediate) and Lay Bomb
+          After Ghavoran - Super Missile Area Access or Enabled Highly Dangerous Logic
```
XenoWars, line 535: This doesn't need "Before Ghavoran - Super Missile Rotatable".  Just makes it dangerous when it doesnt need to be, for the same reason some "Before" events were removed in other prs, like the Dairon - Teleport to Artaria speed blocks.

XenoWars (changes requested):

randovania/games/dread/json_data/Ghavoran.txt
Tyranisaur: Makes, sense, I'll remove it from this one. However, the resource has to be dangerous because you can only rotate it the intended way once, and there's a connection below the spinner that is only available before the rotation. The connections that are relevant when entering through the door at the bottom before rotating the spinner could have misc resource on them, because they are only relevant in transport rando.

PR #4898, opened 2023-07-26
Update the inventory of offline worlds

henriquegemignani:


ThanatosGit:
I'm getting too tired to read it carefully but what would like to ask / comes to my mind: Does this work with non-required mains? Like flash shift is granted because of a flash shift upgrade was sent as item?

henriquegemignani:
It'll respect whatever setting is present in their setting, as that's how `PickupEntry.resource_gain` works.

ThanatosGit:
Hmmm I tried that with two dread worlds and only debug connectors but the flash shift did not show up in the inventory.
Though it's a little bit awkward to test because the receiver first needs to have a non null inventory.

duncathan (approved):

henriquegemignani:
> Hmmm I tried that with two dread worlds and only debug connectors but the flash shift did not show up in the inventory when granting flash shift upgrade. Though it's a little bit awkward to test because the receiver first needs to have a non null inventory.

Hm... I wonder where the issue could be.

This isn't perfect, but as long as it doesn't cause issues I'm fine with a little weirdness. We can fix things later.

PR #4900, opened 2023-07-27
AM2R: Fix shinies being located in metroids

Miepee:
I encountered a shiny missile in a metroid, dont want that again.


Uncle-Reggie:
Alternatively, what if a shiny missile made the Metroid shiny?

Miepee:
Sounds neat on paper, however practically:
A) i'd need to recolor a bunch of metroid sprites which i dont want to
B) I'd need to fiddle with the draw code of metroids, which is already complex enough as-is as they're all split into multiple parts and draw each other a bit differently.

JeffGainsNGames:
I think it's kinda funny to leave in imo

henriquegemignani:
Is all that is happening is getting a "Shiny X" message from a metroid drop?

Miepee:
Yup. Usually one would see the sprite too if it'd be a normal drop.

Miepee:
All righty then.

PR #4901, opened 2023-07-27
Remove need for starting sessions

### LINKED ISSUE #4875: Abolish starting sessions
Clearing a generated game of a MultiplayerSession now resets the UUID of all games.
If we extend to also deleting all `WorldAction` associated with worlds of that session, it will correctly handle the reason why Setup/Start session exists at all.

Clearing the game should have a confirmation dialog, maybe double, if a location has been collected.

henriquegemignani:
Also replaces Finishing sessions with Hiding sessions, and allow unhiding

Fixes #4875.


CHANGELOG.md
```diff
@@ -28,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed: Tracking another user's inventory now properly keeps working after a connection loss.
 - Fixed: Sorting the session history and audit log now works properly.
 - Fixed: In Multiworld session, the Claim world button is now properly disabled when you don't have permissions.
+- Removed: Starting sessions is no longer necessary and has been removed as an option.
```
duncathan, line 32: this is a very significant UX change and ought to be close to the top of the list imo. possibly even marked as major to ensure everyone sees it

randovania/gui/dialog/multiplayer_session_browser_dialog.py
```diff
@@ -164,12 +161,11 @@ def update_list(self):
...
-            self.item_model.setItem(i, 4, num_users)
-            self.item_model.setItem(i, 5, num_worlds)
-            self.item_model.setItem(i, 6, has_password)
-            self.item_model.setItem(i, 7, creator)
-            self.item_model.setItem(i, 8, creation_date)
+            self.item_model.setItem(i, 3, num_users)
+            self.item_model.setItem(i, 4, num_worlds)
+            self.item_model.setItem(i, 5, has_password)
+            self.item_model.setItem(i, 6, creator)
+            self.item_model.setItem(i, 7, creation_date)
```
duncathan, line 168:
```py
for col, item in enumerate((
    name,
    is_user_in_session,
    join_date,
    num_users,
    num_worlds,
    has_password,
    creator,
    creation_date,
):
    self.item_model.setItem(i, col, item)
```
so that we don't need to update magic indices?

duncathan (changes requested):
when a session has been hidden, how can it be found later?

henriquegemignani:
> when a session has been hidden, how can it be found later?

The same one could find finished sessions, there's a checkbox for it in session browser.

henriquegemignani:
Adding a warning:
image

duncathan (approved):

PR #4906, opened 2023-07-27
Better UX for connectors

henriquegemignani:
image
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed: Tracking another user's inventory now properly keeps working after a connection loss.
 - Fixed: Sorting the session history and audit log now works properly.
 - Fixed: In Multiworld session, the Claim world button is now properly disabled when you don't have permissions.
+- Removed: Connecting to Dolphin on Linux executable builds is now hidden, as it doesn't work properly.
 
 ### Metroid Dread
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -30,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed: Tracking another user's inventory now properly keeps working after a connection loss.
 - Fixed: Sorting the session history and audit log now works properly.
 - Fixed: In Multiworld session, the Claim world button is now properly disabled when you don't have permissions.
+- Removed: Connecting to Dolphin on Linux executable builds is now hidden, as it doesn't work properly.
```
ThanatosGit, line 34: I know that it doesn't work properly but it was working somehow with doing it as sudo? Do we know how many are actually using it on Linux and how many players we kick out with that?
If it was long time part of randovania, I don't see a reason to remove the always not so good working solution completly and kick out players after long time.

duncathan: there are reports of it working on Linux executables? my understanding was that it had to be done without the executable

ThanatosGit: Yes, see what Miepee posted in the discord.
I'm really concerned about moving even more players to do dev shenanigans just to use a feature (especially when they were using the feature before in a stable version for a long time).

henriquegemignani: Bleh, forgot about root. You shouldn't be running something as rdv as root...
Oh well, I need to add more complexity for this then

henriquegemignani: Made it check for flatpak (never allow) and root (allow when ran as root)

ThanatosGit (approved):

PR #4915, opened 2023-07-28
Adds elevator/shuttle rando for dread in dev builds

ThanatosGit:
- `ElevatorRandoBasePatchesFactory` sounds weird.
- In the meantime `PresetElevators` moved to `prime_family` (was working on older branch...oops) but I'm using it for Dread as I just need some small modifications of the same. Dunno if we want to make a basic copy of the UI  + File just for dread.
Also moved the differences into subclasses instead of doing all these `if`-conditions since this was just too much if fiddling dread into it.


randovania/games/dread/exporter/patch_data_factory.py
```diff
@@ -482,7 +492,8 @@ def create_data(self) -> dict:
                 for pickup_item in pickup_list
                 if (data := self._pickup_detail_for_target(pickup_item)) is not None
             ],
-            "elevators": [],
+            # FIXME: keep elevators empty in vanilla mode in case something is broken with patching elevators
```
duncathan, line 495: this probably doesn't need a FIXME. this is a pretty typical pattern in patch data factories

randovania/games/dread/gui/preset_settings/__init__.py
```diff
@@ -20,6 +23,9 @@ def dread_preset_tabs(editor: PresetEditor, window_manager: WindowManager):
 
     return [
         PresetTrickLevel,
+        *([
+            PresetElevatorsDread,
+        ] if randovania.is_dev_version() else []),
```
duncathan, line 28: don't even bother tbh. we can add this check if it turns out it's not stable enough in time for 6.2

randovania/layout/lib/teleporters.py
```diff
@@ -204,3 +204,25 @@ def dangerous_settings(self):
         if self.mode == TeleporterShuffleMode.ONE_WAY_ANYTHING:
             return ["One-way anywhere elevators"]
         return []
+
+
+@dataclasses.dataclass(frozen=True)
+class DreadTeleporterConfiguration(TeleporterConfiguration):
+    # Dread has only save stations as start nodes in the db. Elevators are no start nodes but
+    # are valid targets and also the only valid targets. Preset settings makes sure
+    # that nothing else is selected
```
duncathan, line 213: does the UI permit selecting transports correctly despite this?

randovania/games/common/prime_family/gui/elevators_tab.py
```diff
@@ -279,3 +250,83 @@ def on_preset_changed(self, preset: Preset):
...
+class PresetElevatorsPrime2(PresetElevators):
+    compatible_modes = list(enum_lib.iterate_enum(TeleporterShuffleMode))
+
+    custom_weights = {
+            "Great Temple": 0,  # Great Temple
+            "Agon Wastes": 1,  # Agon Wastes
+            "Torvus Bog": 2,  # Torvus Bog
+            "Sanctuary Fortress": 3,  # Sanctuary Fortress
+            "Temple Grounds": 5,  # Temple Grounds
+        }
```
duncathan, line 284: ...we probably don't need these comments any more lol

randovania/layout/preset_migration.py
```diff
@@ -945,6 +945,21 @@ def _migrate_v59(preset: dict) -> dict:
...
+    config = preset["configuration"]
+    game = preset["game"]
+
+    if game in {"dread"}:
+        config["elevators"] = {
+            "mode": "vanilla",
+            "excluded_teleporters": [
+            ],
+            "excluded_targets": [],
+            "skip_final_bosses": False,
```
duncathan, line 958: yuck. I know it's built into the base class but ugh

duncathan (changes requested):

randovania/games/dread/gui/preset_settings/__init__.py
ThanatosGit: We have the choice of missing it when we want to make it stable or missing to deactivate it, if it shouldn't go to stable :D

randovania/layout/lib/teleporters.py
ThanatosGit: With the current modes I added:
image

I removed the "anywhere" options because it looks like the following screenshot. If we want that to be included, it needs a little refactor because all save stations are removed. I think that's some prime specific code as it explicity is removing save stations? (and for dread that means only two are left).
```
...
    has_save_station = any(node.name == "Save Station" for node in area.nodes)
    return (area.has_start_node() and node in area.get_start_nodes() and not has_save_station)
...
```
Though I don't know how stable `anywhere` option is. Iirc remember there was some weird bug that you were not able to move if you never visited a save station before and took a teleporter that was leading to one.
image

randovania/layout/preset_migration.py
ThanatosGit: Is this a comment for future™ or do we want to refactor that now?

duncathan: up to you

randovania/layout/lib/teleporters.py
duncathan: I'm down for just getting rid of anywhere for now. We should remove intro room as a valid target as well then

randovania/layout/lib/teleporters.py
```diff
@@ -204,3 +204,25 @@ def dangerous_settings(self):
         if self.mode == TeleporterShuffleMode.ONE_WAY_ANYTHING:
             return ["One-way anywhere elevators"]
         return []
+
+
+@dataclasses.dataclass(frozen=True)
+class DreadTeleporterConfiguration(TeleporterConfiguration):
```
henriquegemignani, line 210: This should be in `games/dread` folder

randovania/generator/base_patches_factory.py
```diff
@@ -108,7 +108,7 @@ def configurable_node_assignment(self, configuration: BaseConfiguration, game: G
 TrilogyConfiguration = PrimeConfiguration | EchoesConfiguration | CorruptionConfiguration
 
 
-class PrimeTrilogyBasePatchesFactory(BasePatchesFactory):
+class ElevatorRandoBasePatchesFactory(BasePatchesFactory):
```
henriquegemignani, line 111: As a "Base for Prime Trilogy" it's fine, but not for a feature. It'd need to be some sort of composition

randovania/layout/preset_migration.py
henriquegemignani: I'd like for the field to be removed from the base class in this case.

henriquegemignani (changes requested):

randovania/layout/lib/teleporters.py
ThanatosGit: See first screenshot. That's how I want to merge it. No `Valid elevator targets` section at all and no anywhere option

randovania/games/common/prime_family/gui/elevators_tab.py
```diff
@@ -279,3 +250,83 @@ def on_preset_changed(self, preset: Preset):
...
+        TeleporterShuffleMode.TWO_WAY_RANDOMIZED,
+        TeleporterShuffleMode.TWO_WAY_UNCHECKED,
+        TeleporterShuffleMode.ONE_WAY_ELEVATOR,
+        TeleporterShuffleMode.ONE_WAY_ELEVATOR_REPLACEMENT,
+    ]
+
+    def __init__(self, editor: PresetEditor, game_description: GameDescription, window_manager: WindowManager):
+        super().__init__(editor, game_description, window_manager)
+
+        self.elevators_allow_unvisited_names_check.setVisible(False)
```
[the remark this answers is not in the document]
ThanatosGit: That's what I also mentioned in the description of this pr. Though somehow I still need to use the most parts of this file minus the skip final bosses and allow unvisted names stuff.

randovania/generator/base_patches_factory.py
ThanatosGit: No idea what you mean. It's only renamed to accommodate that is now also used by dread instead of only prime games. The usage is still the same.
If that's an issue, then it shouldn't have been there in the first place as it was only adding the elevator rando feature to prime games with a `if echoes` hack.

henriquegemignani: In practice, this PR is changing code intended only for the three prime games into something usable by any game in Randovania.

I'm okay with weird code and `if game ==` in components that are game specific, but it's very no for anything generic.

ThanatosGit: Maybe I just could not imagine how other games do it or will do it but that does not really hold up for me because game specific and generic is flipped between patcher factory and the elevator tab.

`PrimeTrilogyBasePatchesFactory` is not game specific for me as it is generic for configurations which (at least) implemented a `elevators` field. Which btw isn't the case for `CorruptionConfiguration`, so it shouldn't be there. It's "Prime trilogy without corruption" and prime trilogy is not game specific for me. Besides the name and the type for `TrilogyConfiguration`, the code does not look like it was only intended for prime. The only "not so nice" thing is the `elevator_echoes_shuffled` method mixed into it. Hence, I can not understand the reason why this needs a rewrite

`PresetElevators` is used in a generic way, as it is just throwed into the (prime) games which should use it, but it's not generic as it explicitly needs to check for the game and needs to do all the visibility and text changes. Also the `sound_bug_warning` and `custom_weight` needs to check for the instance / game during runtime. The ui files for prime1, echoes and (deactivated) corruption should be seperated but, once again, that should not have been like that in the first place and did not arose with this pr.

Also what you mean by 
> I'm okay with weird code and `if game ==` in components that are game specific

that sounds like a contradiction as game specific components already know which game "they are" by the pure fact that the line is currently executed. If it needs to check for the game, it's not game specfic.

henriquegemignani: "Prime Trilogy" is game specific. It's for these games and these games alone. The widget and elevator makes assumptions it'll always be about a prime game. Critical example is having the skip final bosses setting be related to elevator rando, because in these games these are entirely related.
Even Corruption, where it doesn't even have elevators in the same way the same assumptions work because the game still has elevators. But everything related to Corruption and Elevator rando is dumb and could/should be removed, it provides no value.

`PrimeTrilogyBasePatchesFactory` code as-is might fit other games, but the purpose of that class was for any shared code for the prime games. I think it had more in the past, but that got refactored away?

> The only "not so nice" thing is the elevator_echoes_shuffled method mixed into it. Hence, I can not understand the reason why this needs a rewrite

It's because it's a bad design overall. Composition would go a long way. It could just be `add_elevator_connections_to_patches` as a standalone function. Even simpler, a thing that takes a `GameDescription`, a `Random` and a `TeleporterConfiguration` and returns a list of `DockConnection` things.


> The ui files for prime1, echoes and (deactivated) corruption should be seperated but, once again, that should not have been like that in the first place and did not arose with this pr.

Don't forget that you're touching code written before 99% of the tooling we have for per-game things didn't exist. The only thing one could use to identify which game it was by checking for the `RandovaniaGame`.
It's not acceptable for Prime 1/2 itself, but since it's just in the corner being used by only these two it's kinda fine. But I'd prefer a proper cleanup if it's going to be used by anything else.

ThanatosGit: > "Prime Trilogy" is game specific. It's for these games and these games alone. The widget and elevator makes assumptions it'll always be about a prime game. Critical example is having the skip final bosses setting be related to elevator rando, because in these games these are entirely related.

In the case of the ui, I disagree with that. skip final boss check is one thing they have common but echoes additionally adds custom weights, sound bug warning and allow unvisted names check. For the tab dread and prime1 are closer together than echoes and prime1.

In my current local WIP, I have a hierarchy of `PresetElevatorsCommon` -> `PresetElevatorsPrimeTrilogy` -> `PresetElevatorsPrime1` or `PresetElevatorsPrime2`
Common is what all shares, Trilogy is what is adding skip final boss check, Prime1 just mixin the UI + removes `ECHOES_SHUFFLED` from `compatible_modes`, Prime2 mixin the ui, custom weights, allow unvisted names check, sound bug warning.
Don't really like that as `PresetElevatorsCommon` needs to make assumptions of what is mixed in via the ui to a child's class. E.g. IDE will complain that there is no such thing as `elevator_layout`, `setupUi` etc. as it's only defined in child classes.

Is there a better way to define the differences + using a different ui for each game while sharing a common logic which is heavily tied to the ui?

> It could just be `add_elevator_connections_to_patches` as a standalone function. Even simpler, a thing that takes a `GameDescription`, a `Random` and a `TeleporterConfiguration` and returns a list of `DockConnection` things.

I have something like `def get_dock_connections_for_elevators(game_description: GameDescription, rng: Random, configuration: GamesWithElevatorConfiguration)` but don't know when to call it.
If I call it within `BasePatchesFactory`'s `add_elevator_connections_to_patches`, I need to check if the game supports it but  there is no flag for that?
If I overwrite `add_elevator_connections_to_patches` in a subclass which is inherited by every game which supports it, it's very much the solution which is already in the current state of this pr.

henriquegemignani: The unvisited names was supposed to be in prime 1 too, but lacks patcher support :(

> Don't really like that as PresetElevatorsCommon needs to make assumptions of what is mixed in via the ui to a child's class.
I'd just have the entire UI be per-game. All that is really being shared is code for creating the checkboxes for each elevator, and even that is going to be different for Dread no?

>  If I call it within BasePatchesFactory's add_elevator_connections_to_patches, I need to check if the game supports it but there is no flag for that?

`BasePatchesFactory` shouldn't have a `add_elevator_connections_to_patches`. It should have a `add_dock_connections_to_patches` that is by default empty. Better yet would be a `dock_connection_assignment`, similar to `configurable_node_assignment`.
Then any game that has elevator rando can just implement that method with `return get_dock_connections_for_elevators(...)`

It's fine to have a bit repeated code in each game, as long as it's not the _exactly_ same code in _every_ game.

ThanatosGit: > I'd just have the entire UI be per-game. All that is really being shared is code for creating the checkboxes for each elevator, and even that is going to be different for Dread no?

That part was well hidden. I pushed the WIP of only the hierarchy. Everything in common is shared among prime1, prime2 and dread. And I have some inner struggle in repeating all this stuff for each Ui file.
BTW: ignore the nice typo `PrimeTriologyTeleporterConfiguration` and that it is used there. it is really "just have a `elevators` field in your configuration"

ThanatosGit: Changed it to use a seperate file + ui for each game even if it hurts me to have that many copy & paste code.

test/generator/test_base_patches_factory.py
```diff
@@ -289,7 +293,9 @@ def test_create_base_patches(mocker):
         game, 0, layout_configuration,
     )
 
-    mock_add_elevator_connections_to_patches.assert_called_once_with(factory, layout_configuration, rng, patches[0])
+
+    mock_dock_connections_assignment.assert_called_once_with(factory, layout_configuration, game, rng)
+    # patches[0].assign_node_configuration.assert_called_once_with(mock_gate_assignment_for_configuration.return_value)
```
duncathan, line 298: why the commented line?

ThanatosGit: Oops. Looks like I was distracted. This line should be activated and also changed for the call.
Addressed it

duncathan (approved):

PR #4919, opened 2023-07-28
Dread: Make Escape Sequence safe by creating a separate Hanubia

### LINKED ISSUE #4914: Dread: Making beating Raven Beak not be a dangerous resource
https://discord.com/channels/914291389293027329/914294339235106836/1089628829405950002

In its current form, the database has a bunch of these conditions in Hanubia 
image
When you defeat Raven Beak, you cut off access to the rest of the game, which technically speaking is dangerous, but the generator detects these as dangerous because there are connections with the `Before Escape Sequence` condition on them. It would be sufficient for correctness if there was simply no path back to the rest of the game.

Improving this situation would help both the generator and resolver.

Tyranisaur:
Closes #4914


ThanatosGit:
Somehow I think it's pretty ugly to add a second hanubia, though I know why it was done. So, I pass and leave it to others to decide if we should do that or not.

Miepee:
What are the technical reasons that all the duplicate/unreachable rooms exist?

Tyranisaur:
> What are the technical reasons that all the duplicate/unreachable rooms exist? Also, wouldn't this cause issues with door lock rando?
They still exist here for two reasons. To keep the map aligned, and to make sure the collision camera to room name translations are kept.

There are no randomizable doors in the escape sequence.

henriquegemignani:
I prefer to just have the entire escape sequence be just a single node connection after the RB fight. Then the areas in normal Hanubia can be kept, just maybe without the paths?

XenoWars:
Shouldn't be too hard to add an escape node in RB's Arena. Either Speed and Use Spin Boost, Simple IBJ, or Space Jump. Then do you connect it to Total Recharge Station East and have to ensure you can only take a specific path, or do you connect it to Ship Room and have to account for getting through the Central Unit as well as getting up RB-X Arena?

henriquegemignani:
I'd straight up add the victory condition event in the RB Area room

PR #4942, opened 2023-07-31
Mock all usages of asyncio.run in tests

henriquegemignani:
This might be what leaks sockets?


henriquegemignani:
I made `socket.socket` raise and checked which tests would fail. The default EventLoop on Windows creates one, which is annoying. Some tests also just never respond too, which was **very** annoying.

ThanatosGit (approved):

PR #4946, opened 2023-08-01
Super Metroid: Add DB options for (map) tweaks

Miepee:
- Adds DB options for `Dachora Pit Tweaks`, `Early Supers Bridge Tweaks`, `Pre Hi-Jump Tweaks`, `Moat Tweaks`, `Pre Spazer Tweaks`, `Red Tower Tweaks`, `Nova Boost Tweaks`, `Can't Use Supers On Red Doors` and `Mother Brain Rainbow Beam Nerf`
- Also adds `No Lower Norfair Space Jump Check` as a patch, but doesn't expose it in the GUI yet
- Fix logic condition for missile doors to account for `Can't Use Supers On Red Doors`
- Add damage reduction with suits

### DIFF
--- a/randovania/games/super_metroid/game_data.py
+++ b/randovania/games/super_metroid/game_data.py
@@ -23,14 +23,14 @@ def _gui() -> game.GameGui:
 
 
 def _generator() -> game.GameGenerator:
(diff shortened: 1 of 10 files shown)


SolventMercury (approved):
Gave it a quick test, looks all good to me.

PR #4947, opened 2023-08-01
Slightly improve SM patcher page

Miepee:
Adds a few word wraps, makes pre spazer and pre hijump consistent in naming, clarifies skip intro better to indicate that its not ceres.

### DIFF
--- a/randovania/gui/ui_files/preset_patcher_super_patches.ui
+++ b/randovania/gui/ui_files/preset_patcher_super_patches.ui
@@ -46,8 +46,8 @@
         <rect>
          <x>0</x>
          <y>0</y>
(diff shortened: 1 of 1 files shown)


SolventMercury (approved):
Looks about right. I'm not going to load this into QT to test myself, but this seems like pretty minor changes and I don't foresee any negative outcomes.

dyceron:
For future reference, including pictures of ui changes in the pr comments helps a lot when reviewing.

Miepee:
Am aware, but the only visible change is `Removes Samus' intro monologue.`. The wordwrapping didnt seem to make visible difference in my experience

PR #4953, opened 2023-08-02
Super Metroid: Add enable moonwalk by default + redo cosmetic window

Miepee:
grafik


SolventMercury (approved):
Assuming everything works as expected behind the scenes, then I see no reason to object.

PR #4954, opened 2023-08-02
Super Metroid: Add vanilla locations

Miepee:
Missile launcher, super launcher and PB launcher don't have one assigned. *technically* the first vanilla locations could be assigned but as these are actually expansions and not launchers i feel like it'd be weird to do that.


SolventMercury (approved):
Quickly checked this against my list of hardcoded patcher locations. Everything is correct.

PR #4955, opened 2023-08-02
Super Metroid: Fix Elevators after teleporter refactor

Miepee:
Fixes the teleporters to adhere to new naming scheme, and to not go to previous default nodes.

Would be appreciated if someone can double check the elevator to lower norfair <-> main hall connection in norfair, because based on the names, not sure if it sounds correct.

### DIFF
--- a/randovania/games/super_metroid/json_data/Brinstar.json
+++ b/randovania/games/super_metroid/json_data/Brinstar.json
@@ -5,7 +5,7 @@
     },
     "areas": {
         "Green Brinstar Main Shaft": {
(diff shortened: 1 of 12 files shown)


SolventMercury (approved):
This is really just a node renaming. I get that I'm the SM guy but this one doesn't really need my experties.

Miepee:
Better be safe than sorry. Who knows, maybe something actually does something unexpected with how rdv rounded the coordinates or i accidentally goofed up an elevator connection

PR #4956, opened 2023-08-02
AM2R: Get rid of serris entrance doors

Miepee:
Can't get rid of the BEFORE's sadly, but should make serris logic a bit easier to follow.


JeffGainsNGames (approved):

Miepee:
Approved additionally by druid too

PR #4958, opened 2023-08-02
Fix duplicate prime entry

Miepee:


henriquegemignani:
Thanks. I'll merge it directly in stable to skip the CI

PR #4971, opened 2023-08-04
Add tool for verifying your installation

henriquegemignani:
https://github.com/randovania/randovania/assets/884928/aaaa70cc-7ee1-4fb1-9438-33c471e0643a

### DIFF
--- a/randovania.spec
+++ b/randovania.spec
@@ -1,5 +1,9 @@
 # -*- mode: python -*-
+import hashlib
+import json
+import os
 import platform
+import pathlib
 from PyInstaller.utils.hooks import copy_metadata
 
 import randovania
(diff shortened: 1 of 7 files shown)


randovania.spec
```diff
@@ -57,6 +61,25 @@ a = Analysis(['randovania/__main__.py', 'randovania/cli/__init__.py'],
...
+def calculate_hash(file_name: str):
+    print(f"Hashing {file_name}")
+    with open(file_name, "rb") as f:
+        return hashlib.sha256(f.read()).hexdigest()
+
+
+frozen_file_list = pathlib.Path("randovania").joinpath("data", "frozen_file_list.json")
+frozen_file_list.write_text(json.dumps({
+    entry[0]: calculate_hash(entry[1])
+    for entry in a.datas + a.binaries
```
duncathan, line 75: this will raise failures for certain files for which we support user modification. most notably configuration.json, but also autotracker layouts

henriquegemignani: I would say that it changing these files has it's use cases and work, but I wouldn't say it's a recommended workflow.

And if you changed these and then clicked to see if there's anything wrong in your installation, it certainly should catch these.

henriquegemignani:
bad coverage because coverage reports don't count the tests ran when frozen :(

henriquegemignani:
@duncathan ping

duncathan (approved):

PR #4995, opened 2023-08-06
Resolver: Try actions past point of no return before dangerous actions

Tyranisaur:


duncathan:
what kind of difference does this make?

Tyranisaur:
@duncathan, are you asking about the difference in qualitative or quantitative terms? In terms of what it does qualitative, it's the kind decisions I pointed to here https://discord.com/channels/914291389293027329/914294339235106836/1137831326158958705 With the point being that the EMMI Single Use Slide Event and the Cental Unit Event are both dangerous resources, while the Water Reservoir Blob Event is not a dangerous resource, but is only past a point of no return. 

I added the framework to use heuristics for which actions should be tried before others when the resolver doesn't have any actions that are immediately safe and valuable. This change simply differentiates between dangerous actions, meaning those that give you a resource for which there exists a "before resource" condition somewhere in the database, and actions that would have been considered safe but are behind a point of no return. Previously these were in the same bucket, but with this change the latter of these are tried before the former. Previously, the relative order in which these would be explored was effectively determined by which order they were yielded in, by the for loop.

henriquegemignani:
Can you add a changelog entry?

henriquegemignani (approved):

PR #5005, opened 2023-08-09
AM2R: revise hint faq question.

Miepee:
Since hints moved locations, and I feel like putting names just straight up into an faq is a bad choice, it will now just mention that the in-game map shows the locations.

### DIFF
--- a/randovania/games/am2r/game_data.py
+++ b/randovania/games/am2r/game_data.py
@@ -70,14 +70,16 @@ game_data: game.GameData = game.GameData(
         ("Can I defeat Serris without Ice Beam?",
          "Yes, Serris automatically changes her weakness to not "
          "require Ice Beam if you fight her before acquiring it."),
-        ("Where can I find the Wisdom Septoggs?", "There are seven Wisdom Septoggs. They can can be found in:\n\n"
-            "- Main Caves - Research Site Storage\n"
+        ("Where can I find the Hints?", "There are eight hints in total, and the map will show "
+                                        "an 'H' icon on places where hints exist. They can can be found in:\n\n"
+            "- Main Caves - Research Site Access\n"
             "- Golden Temple - Breeding Grounds Hub\n"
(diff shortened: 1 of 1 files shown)


randovania/games/am2r/game_data.py
```diff
@@ -70,14 +70,9 @@ def _exporter():
...
-            "- Main Caves - Research Site Storage\n"
-            "- Golden Temple - Breeding Grounds Hub\n"
-            "- Hydro Station - Breeding Grounds Lobby\n"
-            "- Industrial Complex - Breeding Grounds Fly Stadium\n"
-            "- The Tower - Tower Exterior North\n"
-            "- Distribution Center - Distribution Facility Tower East\n"
-            "- The Nest - Depths Shinespark Cave East"),
+        ("Where can I find the hints?",
+         "The map will show an 'H' icon on places where hints exist. There is one Hint "
+         "per area. GFS Thoth is the only exception, as it has no hints."),
```
ThanatosGit, line 75: It's per region, isn't it?

ThanatosGit (changes requested):

randovania/games/am2r/game_data.py
Miepee: Yesn't.
Areas in AM2R is what Randovania calls Regions, yes. Most AM2R players are not familiar with the rdv glossrary tho, so I chose to use the am2r one here.

henriquegemignani: I would have be explicit of exactly what you're taking about. This text is also aimed at people who aren't familiar with am2r

Miepee: So should I just go back then to list all room names + mention the H on map thing? That's the best for players I think, i just kinda dislike it for maintainability reasons.

ThanatosGit: Maybe? For unexperienced players, it's probably better. It's not that this list contains 100 area nodes or sth

PR #5016, opened 2023-08-10
Dread/logic fixes for separate beams

Tyranisaur:


duncathan (commented):
note that some of these wide blocks are complicated - particularly the one next to the DNA hints in dairon. is that complexity being accounted for here?

also, we need to make sure that there are no stray beam resource requirements. we should never have just `Wide Beam` as a requirement; it needs to use the `Shoot Wide Beam` template

Tyranisaur:
For the one next to the DNA hints in Dairon, I left doing it from below and from the right as before, and added doing it from above. The effect is that doing it from below or the side requires wide as before, while the new one doesn't.

I did a pass of "Find in Files..." in PycCharm and could not see any usages of Plasma Beam or Wide Beam as a requirements aside from  of `Shoot X Beam` templates.

duncathan:
what about of wave beam? I saw there were a handful such uses in this pr that you changed, but I just wanna make sure that it's comprehensive

Tyranisaur:
I looked for all the Wide Beam Blocks I could think of that matter, and it looks like I've covered all the cases where the `Push Wide Beam Block` template was in use combined with Wave Beam, which I turned into a new template. The outliers are the DNA hint block, which was discussed earlier. I also left the block in Dairon - Big Hub alone, because if you have Wave Beam, you can break the Blob and effectively makes pushing the block redundant.

duncathan (approved):

PR #5021, opened 2023-08-11
add split beams and missiles in dread

duncathan:


ThanatosGit (approved):
RIP overpowered wave beam :)

PR #5023, opened 2023-08-11
AM2R: Change names of torture room + a3 emp reward

Miepee:
Torture room -> Guardian Storage
A3 Storage Cavern -> Sentry Cavern Vault


ThanatosGit (approved):
Better do it now before we need migrations :)

PR #5028, opened 2023-08-11
Base for a further rewrite of Teleporter stuff

ThanatosGit:


ThanatosGit:
Would be nice to get some feedback on how we process it further. Things I'm still thinking about:
- Renaming stuff like "Preset...Prime2" to "Preset...Echoes"?
- moving `_create_source_teleporters` to the common thing?
- moving less stuff to the common thing? On the other hand if there is a game which wants to do it completly different  compared to Prime games and Dread, it should just create it own stuff. Except it's easy feasible by just overwriting one or two methods.
- Stay with calling it `teleporter` internal as it is teleporting you from point x to y and stay with more appropiate names outside / to the user? Like "Transporter" for Dread for a combination of Elevator + Shuttle, and Eleveators for Prime1+2 as they are only using Elevators

randovania/gui/preset_settings/preset_transporter_tab.py
```diff
@@ -0,0 +1,120 @@
...
+    _teleporters_target_for_node: dict[NodeIdentifier, QtWidgets.QCheckBox]
+    compatible_modes = []
+    teleporters_layout: QtWidgets.QVBoxLayout
+    teleporters_combo: ScrollProtectedComboBox
+    teleporters_description_label: QtWidgets.QLabel
+    teleporters_source_group: QtWidgets.QGroupBox
+    teleporters_target_group: QtWidgets.QGroupBox
+    teleporters_target_layout: QtWidgets.QGridLayout
+
+    def __init__(self, editor: PresetEditor, game_description: GameDescription, window_manager: WindowManager):
```
henriquegemignani, line 39: If the class fields are named teleporters, then the class name should also be Teleporter

randovania/generator/elevator_distributor.py
```diff
@@ -164,85 +160,46 @@ def create_elevator_database(region_list: RegionList,
...
+    if not teleporters.is_vanilla:
+        if rng is None:
+            raise MissingRng("Elevator")
+
+        elevator_dock_types = game.dock_weakness_database.all_teleporter_dock_types
+        elevator_db = create_elevator_database(region_list, teleporters.editable_teleporters,
+                                                                    elevator_dock_types)
+        if teleporters.mode == TeleporterShuffleMode.ECHOES_SHUFFLED:
+            from randovania.games.prime2.generator.elevator_distributor import elevator_echoes_shuffled
+            connections = elevator_echoes_shuffled(game, rng)
```
henriquegemignani, line 179: hm... I don't like this wild echoes-specific thing here.
Maybe just error completely if this happens, and Echoes' base patch factory does the check and redirects instead.

randovania/gui/preset_settings/preset_transporter_tab.py
```diff
@@ -0,0 +1,120 @@
...
+        self.setup_ui()
+
+        self.teleporters_layout.setAlignment(QtCore.Qt.AlignmentFlag.AlignTop)
+        self.teleporter_types = game_description.dock_weakness_database.all_teleporter_dock_types
+
+        descriptions = ["<p>Controls where each transporter connects to.</p>"]
+        for value in self.compatible_modes:
+            self.teleporters_combo.addItem(value.long_name, value)
+            descriptions.append(
+                f'<p><span style="font-weight:600;">{value.long_name}</span>: {value.description}</p>'
```
henriquegemignani, line 50: Hmm... these descriptions should be per-game since elevator vs transporter.

randovania/gui/preset_settings/preset_transporter_tab.py
```diff
@@ -0,0 +1,120 @@
...
+    teleporters_target_layout: QtWidgets.QGridLayout
+
+    def __init__(self, editor: PresetEditor, game_description: GameDescription, window_manager: WindowManager):
+        super().__init__(editor, game_description, window_manager)
+        self.setup_ui()
+
+        self.teleporters_layout.setAlignment(QtCore.Qt.AlignmentFlag.AlignTop)
+        self.teleporter_types = game_description.dock_weakness_database.all_teleporter_dock_types
+
+        descriptions = ["<p>Controls where each transporter connects to.</p>"]
```
henriquegemignani, line 46: same with this as next comment

henriquegemignani (changes requested):

randovania/gui/preset_settings/preset_transporter_tab.py
ThanatosGit: Yeah, see the last point of my last message: If no one is opposed, I will call it always teleporter internal for all games and any more fitting name outside for the UI and presentation to the user.

randovania/generator/elevator_distributor.py
ThanatosGit: Is there an exception, I should raise or sth?

ThanatosGit:
Renamed alot to teleporter.
There is `randovania\patching\prime\elevators.py` where I don't know if I should rename all that, too.
Also it is used at places where it maybe shouldn't but that's not really something for this pr.

henriquegemignani:
> Renamed alot to teleporter. There is `randovania\patching\prime\elevators.py` where I don't know if I should rename all that, too. Also it is used at places where it maybe shouldn't but that's not really something for this pr.

That's game specific details and actually should be stuff in the `extra` of the prime 1 and 2 dbs. It's just, you know, from before that was a thing.
Keeping with the trend of all the code to random elevators being old and not flexible.

ThanatosGit:
> That's game specific details and actually should be stuff in the `extra` of the prime 1 and 2 dbs.

Yeah, but it's also used in `create_node_list_selection` which is used by random starting locations. Please, let us do this in another pr

henriquegemignani:
> > That's game specific details and actually should be stuff in the `extra` of the prime 1 and 2 dbs.
> 
> Yeah, but it's also used in `create_node_list_selection` which is used by random starting locations. Please, let us do this in another pr

Oh I was just mentioning what it's used for. Refactoring that can be later for sure.

henriquegemignani (changes requested):

randovania/games/prime1/gui/preset_settings/prime_teleporters_tab.py
```diff
@@ -0,0 +1,182 @@
...
+    from randovania.games.common.prime_family.layout.lib.prime_trilogy_teleporters import (
+        PrimeTrilogyTeleporterConfiguration,
+    )
+    from randovania.games.prime1.layout.prime_configuration import PrimeConfiguration
+    from randovania.gui.lib.window_manager import WindowManager
+    from randovania.interface_common.preset_editor import PresetEditor
+    from randovania.layout.preset import Preset
+
+
+class PresetTeleportersPrime1(PresetTeleporterTab, Ui_PresetTeleportersPrime1, NodeListHelper):
```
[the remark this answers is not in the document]
duncathan: as far as I'm concerned, `Teleporter` *is* a game-specific term for the prime series

randovania/generator/elevator_distributor.py
henriquegemignani: ValueError, or make a custom exception. We don't have good discipline with Exceptions right now.

randovania/games/prime1/gui/preset_settings/prime_teleporters_tab.py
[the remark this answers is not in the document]
ThanatosGit: Hmmm renaming that to elevator again will come with a pretty bad disadvantage: It is using the common part of the ui. It will be very mixed between elevator naming and the common things like `teleporter_mode_to_description` or `teleporters_source_layout` etc. which in my opinion makes it even worse.

randovania/games/game.py
```diff
@@ -40,7 +40,7 @@ def _get_none(h):
 @dataclass(frozen=True)
 class GameLayout:
     configuration: type[BaseConfiguration]
-    """Logic and gameplay settings such as elevator shuffling."""
+    """Logic and gameplay settings such as teleporter shuffling."""
```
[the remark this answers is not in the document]
ThanatosGit: :D
That was the brain afk part where I just searched for occurences of elevator and replaced them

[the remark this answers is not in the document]
henriquegemignani: Oh, the wth isn't about your change. The original text itself doesn't make sense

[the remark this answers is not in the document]
ThanatosGit: Yes, I understood how you meant it the first time. ;)

henriquegemignani (approved):

PR #5037, opened 2023-08-13
AM2R: Add pipe rando

Miepee:
Adds teleporter pipe rando to am2r.
Needs #5036 to get merged first 

Also note that this has some buggyness due to elevator rando still being primarily designed for primes. most notably, doesnt take into account the pipes that only exist on a certain setting, as well as the lower pipes from Distribution Facility Tower East causing generation to fail.


Miepee:
Closing this, as the current state doesnt really allow for pipe rando without some major refactorings.
May revisit this in the future.

PR #5043, opened 2023-08-14
AM2R: Add some more logic to GT

Miepee:


randovania/games/am2r/json_data/Golden Temple.txt
```diff
@@ -746,7 +760,11 @@ Extra - map_name: rm_a1a08
               All of the following:
                   # Spider
                   Spider Ball
-                  Spring Ball or Mid-Air Morph (Beginner) or Can Use Any Bombs
+                  Any of the following:
+                      Spring Ball or Mid-Air Morph (Beginner) or Can Use Any Bombs
```
JeffGainsNGames, line 764: should this be an `AND can use any bombs`? assuming this is to break the upper blocks during the spider movement. As well below on line 767 there is another `can use any bombs` so it may be more efficient to group the bombs requirement with the spider requirement on line 762 since they are used in all scenarios.

Miepee: No, this is an OR.
This line is to do this (blue is shooting, yellow is movement):
grafik

JeffGainsNGames (approved):

PR #5047, opened 2023-08-14
Persist used tricks

henriquegemignani:
It's a bit slow to calculate, so let's load it directly.


randovania/games/am2r/json_data/header.json
```diff
@@ -4296,6 +4296,92 @@
             "to_shuffle_proportion": 1.0
         }
     },
+    "used_trick_levels": {
```
Miepee, line 4299: Does this mean, that as the DB is maturing, that i need to manually add the trick levels here?

henriquegemignani: No, this is updated automatically when you save.

henriquegemignani: I thought of making this only show up in the binary version of the db. Could still do that.

Miepee: Ah okay, must've missed it.

ThanatosGit (approved):

PR #5055, opened 2023-08-15
Use a proper class for Inventory

henriquegemignani:
Avoids an error that PySide 6.5 does not like non-string keys for dicts in signals

### DIFF
--- a/randovania/game_description/resources/inventory.py
+++ b/randovania/game_description/resources/inventory.py
@@ -1,0 +1,45 @@
+from __future__ import annotations
+
+import dataclasses
+from typing import TYPE_CHECKING, NamedTuple, Self
+
+from randovania.game_description.resources.resource_type import ResourceType
+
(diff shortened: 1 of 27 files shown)


randovania/game_description/resources/inventory.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class InventoryItem(NamedTuple):
+    amount: int
+    capacity: int
+
+
+@dataclasses.dataclass(frozen=True, slots=True)
+class Inventory:
+    raw: dict[ItemResourceInfo, InventoryItem]
```
duncathan, line 20: should this have a default factory to default to an empty dict? or is the explicit `empty()` initializer preferable?

henriquegemignani: I think I prefer calling `.empty`. This will need a better `__init__.py` at some point, but I'll leave that refactor to the future

PR #5056, opened 2023-08-15
AM2R: redo am2r hints

Miepee:
Stops shuffling the dna during bootstrap, and only shuffles the hints.
Should allow rdvgames to be imported again


ThanatosGit (approved):
Looks like the solution, we talked about but you didn't like :p

PR #5064, opened 2023-08-16
AM2R: add a glossary for the db.

Miepee:
I noticed a glossary is helpful for newer people going through DBs. Went through SMs DB before, and had no idea what "CWJ was supposed to mean.


Miepee:
These are indeed in comments. Sometimes writing out the full names makes them very long, or repetitive. I for example don't really like typing out `Infinite Bomb Jump` every time I need to reference an IBJ.
Granted an external file isn't probably the best place for such a glossary to be. Ideally it should available in the data visualizer.

randovania/games/am2r/json_data/_glossary.md
```diff
@@ -0,0 +1,37 @@
...
+Short for Morph Glide.
+
+### DB
+Rarely used as a short form for Damage Boost
+
+### PB
+Short for Power Bomb.
+
+### HJ or High or Hi
+Short fro Hi-Jump
```
ThanatosGit, line 27: fro!
This and DB also missing a "." at the end.

Miepee:
Gonna be drafting this, until i rewrite this to have a glossary in the visualizer itself. should also benefit other games.

Miepee:
Closing this, as I wont be getting to this anytime soon, and the rework for db editor would redo this completely anyawy

PR #5066, opened 2023-08-16
AM2R: Provide unique text for lockeed ammo

Miepee:


randovania/games/am2r/exporter/patch_data_factory.py
```diff
@@ -38,6 +38,10 @@ def _create_pickups_dict(self, pickup_list: list[ExportedPickupDetails], item_in
         for pickup in pickup_list:
             quantity = pickup.conditional_resources[0].resources[0][1]
             object_name = self.game.region_list.node_from_pickup_index(pickup.index).extra["object_name"]
+            text_index = 1 if (pickup.original_pickup.resource_lock and
```
ThanatosGit, line 41: Please use `is not None`. I prefer to be more explicit when checking for the singleton.
Writing it that way can be a gotcha as things like containers can evaluate to false if there length is zero.

randovania/games/am2r/exporter/patch_data_factory.py
```diff
@@ -38,6 +38,10 @@ def _create_pickups_dict(self, pickup_list: list[ExportedPickupDetails], item_in
...
+                               "Locked" in pickup.original_pickup.resource_lock.temporary_item.long_name and
```
ThanatosGit, line 42: Maybe use a variable for `pickup.original_pickup.resource_lock`. It it's really lengthy.

randovania/games/am2r/pickup_database/item_data.json
```diff
@@ -26,8 +26,8 @@
 	},
 	"Locked Power Bomb Expansion": {
 		"sprite_speed": 0.2,
-		"text_desc": "Power Bomb capacity {Locked Power BombsChanged} by {Locked Power Bombs}",
-		"text_header": "Got Power Bomb Tank"
+		"text_desc": "Power Bomb capacity {Locked Power BombsChanged} by {Locked Power Bombs} - Launcher still required",
+		"text_header": "Got Locked PowerBomb Tank"
```
ThanatosGit, line 30: Don't know where that will be used but it looks weird to me that it is using `PowerBomb` and `Power Bomb` in the text.
Same for super missile

Is this intentional?

ThanatosGit (changes requested):

randovania/games/am2r/pickup_database/item_data.json
Miepee: Yup. The header space is a tad limiting in-game, so this is an anti-clipping measure. also the reason why super missile is missing "tank".

randovania/games/am2r/exporter/patch_data_factory.py
Miepee: Pycharm at one pointed started yelling at me for writing stuff as `is not none` instead of like this, so through this'd be better. good to know.

ThanatosGit (approved):

PR #5068, opened 2023-08-16
Add mypy checking for a really tiny part

henriquegemignani:


ThanatosGit (approved):

randovania/game_description/resources/resource_collection.py
```diff
@@ -0,0 +1,153 @@
...
+        except IndexError:
+            self._resize_array_to(resource_index)
+            return self._resource_array[resource_index]
+
+    def __str__(self):
+        return f"<ResourceCollection with {self.num_resources} resources>"
+
+    @property
+    def _comparison_tuple(self):
+        return list(self.as_resource_gain()), self.add_self_as_requirement_to_resources
```
duncathan, line 51: while we're doing type checking, it seems a bit odd for `comparison_tuple` to be a list

randovania/game_description/resources/resource_collection.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def is_resource_set(self, resource: ResourceInfo) -> bool:
+        """
+        Checks if the given resource has a value explicitly set, instead of using the fallback of 0.
+        :param resource:
+        :return:
+        """
+        return resource.resource_index in self._existing_resources
+
+    def set_resource(self, resource: ResourceInfo, quantity: int):
```
duncathan, line 73: should there be a `__setitem__` method that calls this as well?

randovania/game_description/resources/resource_collection.py
```diff
@@ -0,0 +1,153 @@
...
+            if self._resource_array[resource_index] > 0:
+                self.resource_bitmask |= mask
+            elif self.resource_bitmask & mask:
+                self.resource_bitmask -= mask
+
+    def as_resource_gain(self) -> ResourceGain:
+        for index, resource in self._existing_resources.items():
+            yield resource, self._resource_array[index]
+
+    def remove_resource(self, resource: ResourceInfo):
```
duncathan, line 122: and `__delitem__`?

duncathan (changes requested):

randovania/game_description/resources/resource_collection.py
henriquegemignani: I just moved this to another file 😭 

Eh, kinda prefer the explicit methods?

duncathan: it's just cuz it already has `__getitem__` (and *no* explicit method) so not having the others seems kinda weird

randovania/game_description/resources/resource_collection.py
duncathan: ah i misread lol

randovania/game_description/resources/resource_collection.py
henriquegemignani: ah fair. Hm. Well mostly because adding resources should be done mostly via `add_resource_gain` I guess? I'll add some comments for these.

randovania/game_description/resources/resource_collection.py
henriquegemignani: Doing `del collection[my_item]` is such a weird API

duncathan: yeah I can live without that one lmao I don't think I've ever used `del` in python a single time

duncathan (approved):

PR #5069, opened 2023-08-16
Samus Returns - Database Cleanup #1

dyceron:
- Renames Elevators to Elevator (from Teleporter)
- Fixes Integrity Errors
- Removes incorrect Gravity Suit pickup event (Fleech Swarms will be properly handled in the patcher)

### DIFF
--- a/randovania/games/samus_returns/json_data/Area 1.json
+++ b/randovania/games/samus_returns/json_data/Area 1.json
@@ -10,7 +10,7 @@
     },
     "areas": {
         "Chozo Seal": {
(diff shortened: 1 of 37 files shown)


ThanatosGit (changes requested):
`test_integrity_checks.py` has a dict where games with database errors are defined:
```
_acceptable_database_errors = {
    RandovaniaGame.SUPER_METROID: True,
    RandovaniaGame.METROID_SAMUS_RETURNS: True,
}
```

Just remove the SR line there because all integrity checks errors are gone :)

ThanatosGit (approved):

PR #5074, opened 2023-08-17
Timeout the daily generate job after 1h

henriquegemignani:


duncathan:
this is a very funny solution to the bug of the daily generate message not displaying hours elapsed

henriquegemignani:
> this is a very funny solution to the bug of the daily generate message not displaying hours elapsed

it wont fix that, since that's the total cpu time which totally can still go over 1h :P

PR #5077, opened 2023-08-17
Micro optimizations for generation

henriquegemignani:
Very scientific benchmark:

Before
```
$ python -m randovania layout generate-from-permalink --permalink Da3N_L4qMzyCzJ0KWIRiEuMOO6C_f_zV --no-validate seed3.rdvgame
Took 23.496 seconds. Hash: ZX6L4KRT
```
After
```
$ python -m randovania layout generate-from-permalink --permalink Da3N_L4qMzyCzJ0KWIRiEuMOO6C_f_zV --no-validate seed3.rdvgame
Took 21.289 seconds. Hash: ZX6L4KRT
```

### DIFF
--- a/randovania/generator/old_generator_reach.py
+++ b/randovania/generator/old_generator_reach.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import copy
+import typing
 from typing import TYPE_CHECKING, NamedTuple
 
 from randovania.game_description.db.resource_node import ResourceNode
@@ -52,6 +53,11 @@ class GraphPath(NamedTuple):
             digraph.add_edge(self.previous_node.node_index, self.node.node_index, requirement=self.requirement)
 
 
+class _SafeNodes(typing.NamedTuple):
+    as_list: list[int]
+    as_set: set[int]
+
+
 class OldGeneratorReach(GeneratorReach):
     _digraph: graph_module.BaseGraph
     _state: State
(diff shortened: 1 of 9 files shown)


randovania/generator/old_generator_reach.py
```diff
@@ -231,21 +243,19 @@ def state(self) -> State:
...
-    def all_nodes(self) -> tuple[Node | None, ...]:
-        return self.game.region_list.all_nodes
-
     @property
     def nodes(self) -> Iterator[Node]:
-        for node in self.iterate_nodes:
-            if node.node_index in self._digraph:
+        for i, node in enumerate(self.all_nodes):
+            if i in self._digraph:
+                # assert node is not None
```
duncathan, line 250: leftover?

randovania/generator/old_generator_reach.py
```diff
@@ -60,7 +61,7 @@ class OldGeneratorReach(GeneratorReach):
     _reachable_costs: dict[int, int] | None
     _node_reachable_cache: dict[int, bool]
     _unreachable_paths: dict[tuple[int, int], RequirementSet]
-    _safe_nodes: set[int] | None
+    _safe_nodes: tuple[list[int], set[int]] | None
```
duncathan, line 64: would it hurt performance to use a NamedTuple here for readability's sake? if it would cause too much trouble then it's ok as-is tho

henriquegemignani: tbh I'm not even sure of how relevant this change is

duncathan (approved):

PR #5078, opened 2023-08-17
AM2R: Fix the locked text variables

Miepee:


randovania/games/am2r/exporter/patch_data_factory.py
```diff
@@ -94,7 +94,13 @@ def _create_hash_dict(self):
             "session_uuid": str(self.players_config.get_own_uuid())
         }
 
-    def _create_game_patches(self, item_info: dict, rng: Random):
+    def _create_game_patches(self, pickup_list: list[ExportedPickupDetails], item_info: dict, rng: Random):
+        def get_locked_ammo_text(ammo_item: str):
+            return [pickup.collection_text[0] for pickup in pickup_list if pickup.original_pickup.name == ammo_item][0]
```
ThanatosGit, line 99: Dunno by heart what's in ExportedPickupDetails. I assume only the pickups which should be exported...will this fail with an IndexError if you play without any of those three expansion because it's not present in pickup_list?

henriquegemignani: It'll for sure break in multiworld

Miepee: Oh, haven't even thought about multiworld. 
What'd be the sanest way to get a local missile expansion, and if it doesn't exist, return placeholder text?

duncathan: you probably want `default_database`

Miepee: I'll look into it.
out of curiosity, would this cause some weird side effects too?
https://github.com/randovania/randovania/pull/5078/files#diff-a3ea11958f3f07ff8043ad0e3ae2a03d10b12ff6456e8506b1cc521401e71f2fL42-L45

henriquegemignani: Oh I get the problem. The text for the expansions is set globally, not per pickup.

Should be fine to search the exported pickup list for it and default to `<MISSING TEXT, REPORT IF SEEN>`. After all, if there's no locked missile expansion in the pool, the game will never use the locked missile expansion text right?

Miepee: Yeah, it's only used in-game if a) you have ammo shuffled and b) you collect an ammo expansion before collecting its launcher.

henriquegemignani: It still needs handling the case I mentioned. Try adding a preset/rdvgame for tests that has no expansions.

henriquegemignani (changes requested):

ThanatosGit (approved):

PR #5083, opened 2023-08-18
Disallow untyped functions in mypy

henriquegemignani:


randovania/game_description/resources/resource_type.py
```diff
@@ -59,9 +59,6 @@ def non_negated_prefix(self) -> str:
         else:
             return ""
 
-    def __lt__(self, other):
```
duncathan, line 62: I take it this is rolled into `intEnum`?

duncathan (approved):

randovania/game_description/resources/resource_type.py
henriquegemignani: I imagine. I also tested in terminal and it worked just fine.

PR #5085, opened 2023-08-19
SM: add option for custom etank value

Miepee:
Can't actually test in-game but should:tm: work


Miepee:
closing until we get further SM development

PR #5092, opened 2023-08-19
SM: Allow negative ammo

Miepee:
Last time I checked, Pier mentioned this should work. Can't currently test tho.


ThanatosGit:
Should be tested or reviewed by someone who knows that this works.

Miepee:
Will merge for now. If it causes issues, we can always revert later.

PR #5093, opened 2023-08-19
SM: Expose no_ln_chozo_inventory_check in GUI

Miepee:
grafik


dyceron (approved):

dyceron:
Are these logical options in the database?

Miepee:
A misc resource exists for it, however SMs DB is currently still very basic, where almost none of the options are taken properly into account.

PR #5096, opened 2023-08-20
Bump am2r-yams to 0.0.15

Miepee:
Doing dependabots job once again, cause I don't wanna wait


ThanatosGit (changes requested):
Please update `setup.cfg`, afaik `am2r-yams>=0.0.10`, so e.g. version 0.0.10, is not compatible with the current randovnai

Miepee:
Bumping only requirements.txt is what dependabot does, so i just mimicked its behaviour
But went ahead anyway, and bumped it in setup.cfg too

ThanatosGit (approved):

PR #5100, opened 2023-08-20
Dread/move victory to after raven beak

### LINKED ISSUE #4914: Dread: Making beating Raven Beak not be a dangerous resource

Tyranisaur:
Closes #4914
This represents the solution suggested by @henriquegemignani in #4919 

This was based on the linked PR, and all the changes to the Hanubia region should be identical, as the Hanubia region also in this version does not deal with the Escape Sequence. Meanwhile, the Escape Sequence is handled in the Raven Beak Arena.


randovania/games/dread/json_data/Itorash.txt
```diff
@@ -205,11 +205,26 @@ Extra - asset_id: collision_camera_004
...
-  > Dock to Hanubia
+  > Event - Escape Sequence
+      All of the following:
+          Lay Power Bomb
+          Any of the following:
+              Space Jump or Simple IBJ
+              All of the following:
+                  Speed Booster and Spin Boost
+                  Movement (Intermediate) or Speed Booster Conservation (Beginner)
+                  Single-wall Wall Jump (Intermediate) or Wall Jump (Beginner)
```
XenoWars, line 216: I'm not too sure what the single-wall wall jump/wall jump is for. Why is this here?

PR #5103, opened 2023-08-20
AM2R: Add minimal logic info.

Miepee:
AM2R requires the following items to be beaten:
> Bombs, Speed Booster, Ice Beam, (Screw Attack or Power Bombs), 30 Missiles, 1 Super Missiles

However, since it's impossible specify the or condition to my knowledge, I'm using an and condition instead.

### DIFF
--- a/randovania/games/am2r/json_data/header.json
+++ b/randovania/games/am2r/json_data/header.json
@@ -3748,10 +3748,32 @@
         "Default": []
     },
     "minimal_logic": {
-        "items_to_exclude": [],
+        "items_to_exclude": [
+            {
+                "name": "Bombs",
+                "when_shuffled": null
+            },
+            {
+                "name": "Speed Booster",
+                "when_shuffled": null
+            },
+            {
+                "name": "Ice Beam",
+                "when_shuffled": null
+            },
+            {
+                "name": "Screw Attack",
+                "when_shuffled": null
+            }
+        ],
         "custom_item_amount": [],
-        "events_to_exclude": [],
-        "description": "Nothing is checked."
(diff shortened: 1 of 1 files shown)


ThanatosGit (commented):
I have never played with minimal logic but I have beaten my first blind playthrough without speed booster.
I assume that you need much more other items to beat the game without speed booster? But that would technically mean that there is also an `or` around speed booster.

Miepee:
I was going off of low%, which needs speedbooster, due to metroids being locked by it. Since rdv doesn't require you to kill all metroids anymore, I removed it.

randovania/games/am2r/json_data/header.json
```diff
@@ -3748,10 +3748,45 @@
...
+                "value": 10
+            }
+        ],
+        "events_to_exclude": [
+            {
+                "name": "VictoryEvent",
+                "reason": "Victory Condition"
+            }
+        ],
+        "description": "Items checked for: Bombs, Speed Booster, Ice Beam, Screw Attack, Missiles, Super Missiles, Power Bombs."
```
dyceron, line 3789: This description seems wrong. Only Ice, Bombs, and Screw are checked for. Speed isn't on the list, and neither are Missiles, Supers, or PBs (shouldn't be checked for anyway)

dyceron (changes requested):

randovania/games/am2r/json_data/header.json
Miepee: Forgot to readjust, whoops.
Fixwed

dyceron (approved):

PR #5107, opened 2023-08-21
AM2R: Fix crashes with "hide model with nothing"

Miepee:


ThanatosGit (approved):
This stuff definitely needs a refactor. `visual_etm` doesn't even make sense to me as a non-prime player and it is using a hard coded echoes game enum. Wtf :D

PR #5109, opened 2023-08-21
AM2R: Handle when trying to call dotnet, but it doesn't exist

Miepee:


randovania/games/am2r/exporter/game_exporter.py
```diff
@@ -42,8 +42,14 @@ def export_can_be_aborted(self) -> bool:
...
         # Check if dotnet is available
-        dotnet_process = subprocess.run(["dotnet", "--info"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
-        if dotnet_process.returncode != 0:
+        dotnet_ran_fine = False
+        try:
+            dotnet_process = subprocess.run(["dotnet", "--info"],
+                                            stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=False)
+            dotnet_ran_fine = dotnet_process.returncode == 0
+        except FileNotFoundError:
+            dotnet_ran_fine = False
```
henriquegemignani, line 51: Can you add a test case? if you do `mocker.patch("subprocess.run", side_effect=FileNotFoundError)` it'd work fine

henriquegemignani (changes requested):

randovania/games/am2r/exporter/game_exporter.py
Miepee: Test added

henriquegemignani (approved):

PR #5123, opened 2023-08-23
Improve instructions for new games, add rules and expectations

henriquegemignani:


docs/Glossary.md
```diff
@@ -34,6 +34,11 @@ Has two meanings:
 
 Connects two [Nodes](#node) in the same area. Contains a set of [Requirements](#requirement) that define whether the [Generator](#generator) and [Solver](#solver) are allowed to cross it.
 
+## Cosmetic Patches
+
+Settings that do not impact the permalink/generation, but can be used when exporting the game to optionally tweak certain things.
```
Miepee, line 39: I'd add either in this line, or in the line below some sentence for things that don't give unfair advantages. I.e. Cutscene skips, while not affecting permalink/generation, should *not* be cosmetic

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+In most case, this involves readings the files in
+a user-provided copy of the game and creating either a new ROM/ISO file or a mod file.
 
-**TODO**: details
+This process is split into two parts:
+
+### Patch Data Factory
+
+Each game defines a class that inherits from `PatchDataFactory`. Given a LayoutDescription, a CosmeticPatches and 
+PlayersConfiguration, it creates a JSON Object that will be passed as an argument to the Game Exporter.
```
Miepee, line 47: Would make `LayoutDescription`, `CosmeticPatches` and `PlayersConfiguration` in code blocks. Bonus points for linking to glossary.

Also technically it creates a dictionary, not a json object. Important distinction, as it does mean you can reformat it in other ways, should your patcher for some reason not deal with json nicely.

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+
+### Game Exporter
+
+Each game also defines a class that inherits from `GameExporter`. Given the JSON created by the PatchDataFactory and a
+GameExportParams, provided by your game's GameExportDialog, it's expected to create all the necessary files for the
```
Miepee, line 52: Would make the code parts here also in code blocks.

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+user to play the game.
+
+Since modifying the game files requires nothing from Randovania and can be a quite big task, a separate python package
```
Miepee, line 55: *quite a big
Also capitalize Python please

docs/Glossary.md
```diff
@@ -34,6 +34,11 @@ Has two meanings:
...
+Commonly used for custom hud color, player model/sprite, different defaults for in-game settings.
```
dyceron, line 40: Is this meant to be a new line?

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+- Rust
+- Lua
+
+Likely Good (Should integrate well, never tested):
+- Languages that compile to static library, such a Go
+
+Ok (Expect additional work to integrate properly, avoid if viable):
+- C#
+
+Avoid languages that require the user to install a runtime, such as Java.
```
Miepee, line 86: Hmm, c# *technically* requires an installed runtime too, so what's the difference to java here?

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+
+#### Standalone Use
+
+Allow your patcher to be used directly in the command line, by passing the game files and a json file.
+
+This makes it easy to test the patcher without having to do it through Randovania, and helps identify issues better.
+
+#### Json Schema
+
+Define a [JSON Schema](https://json-schema.org/) that defines the structure of the patcher data. This makes for a good
```
Miepee, line 96: Can you link https://json-schema.org/understanding-json-schema/index.html as a resource too please?

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
 
 
+## Exporting games
 
-### Exporting games
+Up to now, everything is just theoretical values inside Randovania. The next step is making these modifications be
+applied to the game. 
 
-In order to export a game, a patcher is required. Typically, the patcher will be included in Randovania for seamless exports. This is a requirement for non-experimental games. Experimental games, however, may prefer an alternative implementation: refer to Corruption as an example of exporting data to be passed to a standalone patcher.
+In most case, this involves readings the files in
```
dyceron, line 39: In most cases, this involves reading the files from a

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+
+# Rules for new games
+
+Randovania has a high bar of quality for our game integrations, and we wish to keep it that way. In order to help
+with that goal, there are rules that need to be met before we can accept a game.
+
+- An owner for the game, with an expectation to be available for a while.
+- A reviewer of Logic changes, separate from the owner.
+- An open source patcher, with a license similar to GPL or Apache.
+- Proper test coverage of new code.
```
Miepee, line 112: A) Are these *rules* or *guidelines*?

B) any further tips on how to do test coverage? this is a bit tricky, since it usually requires copyrighted files.-

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+- Python
+
+Good (Known to integrate without issues):
+- C
+- C++
+- Rust
+- Lua
+
+Likely Good (Should integrate well, never tested):
+- Languages that compile to static library, such a Go
```
Miepee, line 81: *as Go

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+PlayersConfiguration, it creates a JSON Object that will be passed as an argument to the Game Exporter.
+
+### Game Exporter
+
+Each game also defines a class that inherits from `GameExporter`. Given the JSON created by the PatchDataFactory and a
+GameExportParams, provided by your game's GameExportDialog, it's expected to create all the necessary files for the
+user to play the game.
+
+Since modifying the game files requires nothing from Randovania and can be a quite big task, a separate python package 
+responsible for this task is created. This is called the patcher for the game. You are free to implement this package
```
dyceron, line 56: Since Randovania does not modify the game files, a separate python package must be created which is responsible for these changes.

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+however you want, but we recommend the following:
+
+#### Input Data
+
+The arguments for the patcher should be:
+- A Json file, called the patcher data.
+- Paths to needed game files
```
dyceron, line 63: Paths to necessary game files (input path)

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+user to play the game.
+
+Since modifying the game files requires nothing from Randovania and can be a quite big task, a separate python package 
+responsible for this task is created. This is called the patcher for the game. You are free to implement this package
+however you want, but we recommend the following:
+
+#### Input Data
+
+The arguments for the patcher should be:
+- A Json file, called the patcher data.
```
dyceron, line 62: A json file, also called the patcher data (No period)

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+- Paths to needed game files
+- Paths to where to output goes.
```
dyceron, line 64: Paths to an output folder (output path)

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+- A status update function
+
+Nothing more. Any other parameter should come from the json file.
```
dyceron, line 67: Remove "Nothing more"

docs/New Game.md
dyceron, line 81: such as Go

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+Good (Known to integrate without issues):
+- C
+- C++
+- Rust
+- Lua
+
+Likely Good (Should integrate well, never tested):
+- Languages that compile to static library, such a Go
+
+Ok (Expect additional work to integrate properly, avoid if viable):
```
dyceron, line 83: avoid if possible

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+- C#
+
+Avoid languages that require the user to install a runtime, such as Java.
+
+#### Standalone Use
+
+Allow your patcher to be used directly in the command line, by passing the game files and a json file.
+
+This makes it easy to test the patcher without having to do it through Randovania, and helps identify issues better.
```
dyceron, line 92: without having to run it through Randovania

and helps better identify issues.

dyceron (changes requested):

docs/New Game.md
duncathan: the fact that we use it LMAO

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
 ### Multiworld
 
 **TODO**
+
+# Rules for new games
+
+Randovania has a high bar of quality for our game integrations, and we wish to keep it that way. In order to help
+with that goal, there are rules that need to be met before we can accept a game.
+
+- An owner for the game, with an expectation to be available for a while.
```
duncathan, line 109: An owner for the game. The owner is expected to be available and willing to maintain their game for the foreseeable future.

docs/New Game.md
```diff
@@ -31,14 +31,82 @@ game to have proper pickups:
...
+- A reviewer of Logic changes, separate from the owner.
+- An open source patcher, with a license similar to GPL or Apache.
```
duncathan, line 111: this should instead specify that the license must be compatible with GPLv3, in order for RDV to comply with its own license

duncathan (changes requested):

docs/New Game.md
henriquegemignani: No, it needs to be json serializable. 
A dict is considerably more powerful than just a json object.

docs/New Game.md
henriquegemignani: More precisely that YOU did the work.

But also how I think a second game using it would need some significant work nicely together so I prefer to not have have more

docs/New Game.md
henriquegemignani: A) Intentionally ambiguous. But my intention is that everything in this last section is a requirement, but we reserve the right to make exceptions.

B) There's plenty of code one adds to Randovania that doesn't need game data. But I'm trying to find a reliable way to do tests with access to game data.

docs/New Game.md
henriquegemignani: After reading a bit, if the patchers are considered plug-ins then we likely don't require them to be GPL'd due to us being GPL. I might try smmending our license to explicitly mention that/add an exception, for whatever worth changing our license is.

However the entry here is not a compliance with our license and instead a requirement for your license at all. I want all patchers to be licensed in a way that it needed we can fork and then continue development, it's needed.

docs/New Game.md
Miepee: >  Intentionally ambiguous

Currently from reading it, it very much feels like "these are set in stone rules without exceptions", hence why me asking

docs/New Game.md
henriquegemignani: Return value from PatchDataFactory. But also whatever you pass to the patcher (which I called the "patcher data") as this is what often gets saved to a patcher.json or whatever.

henriquegemignani: See how in later in the documentation, we talk about receiving a JSON as argument, as well as having a schema?

While you _could_ do that, that's kinda just real ugly and bad. Borderline acceptable if someone is integrating an existing patcher, but I'd also just recommend to refactor it to accept a better format. Worse case, just keep that _inside_ their patcher package.

Miepee: Cool. Was mainly just wondering if someone would try to write a patcher in some language whose json support isn't great for whatever reason, but now that I think about it the only languages where that *is* the case, wouldn't be accepted in RDV anyway.

docs/Glossary.md
henriquegemignani: Eh, I'll leave this for somewhere else.

But also by definition anything that doesn't affect the Permalink is Cosmetic. These are the only two kind of settings we have. So cutscene skips affect the permalink.
But Echoes has plenty cosmetic patches that can give you disadvantages in racing if disabled.

henriquegemignani:
Addressed changes, add an extra about multiworld.

docs/New Game.md
```diff
@@ -31,14 +31,85 @@ game to have proper pickups:
...
 
 
+## Exporting games
 
-### Exporting games
+Up to now, everything is just theoretical values inside Randovania. The next step is making these modifications be
+applied to the game. 
 
-In order to export a game, a patcher is required. Typically, the patcher will be included in Randovania for seamless exports. This is a requirement for non-experimental games. Experimental games, however, may prefer an alternative implementation: refer to Corruption as an example of exporting data to be passed to a standalone patcher.
+In most cases, this involves readings the files from a user-provided copy of the game and creating either
```
dyceron, line 39: reading the files

docs/New Game.md
```diff
@@ -31,14 +31,85 @@ game to have proper pickups:
...
 ### Multiworld
 
+Before a game can work in Multiworld, it must first work properly as a solo game. It is highly recommended to finish the basic integration first.
+
 **TODO**
+
+# Rules for new games
+
+Randovania has a high bar of quality for our game integrations, and we wish to keep it that way. In order to help
+with that goal, there are rules that need to be met before we can accept a game.
```
dyceron, line 110: Randovania has a high quality standard for game integration. In order for a new game to be added, the following rules must be adhered to:

dyceron (changes requested):

dyceron (approved):

docs/New Game.md
```diff
@@ -31,14 +31,85 @@ game to have proper pickups:
 2. Update `standard_pickup_configuration` and `ammo_pickup_configuration` in your game's starter preset.
 
 
+## Exporting games
 
-### Exporting games
+Up to now, everything is just theoretical values inside Randovania. The next step is making these modifications be
+applied to the game.
```
Miepee, line 37: This reads a bit weird. Maybe "The next step is applying these modifications to the game"?

Miepee (approved):

duncathan (approved):

PR #5125, opened 2023-08-23
Fix am2r and prime getting invalid ETM model

### LINKED ISSUE #5146: Prime 2: unable to export with hidden item models and Item Acquisition Popups enabled
Sentry Issue: [RANDOVANIA-CLIENT-ZQ](https://randovania.sentry.io/issues/4397127341/?referrer=github_integration)

```
KeyError: 'Unknown item'
(7 additional frame(s) were not displayed)
...
  File "randovania\exporter\pickup_exporter.py", line 208, in create_details
  File "randovania\exporter\pickup_exporter.py", line 132, in _calculate_collection_text
  File "randovania\exporter\pickup_exporter.py", line 111, in _get_all_hud_text
  File "randovania\exporter\pickup_exporter.py", line 112, in <listcomp>
  File "randovania\exporter\pickup_exporter.py", line 96, in _get_single_hud_text
```

Miepee:
This changes the `create_visual_etm` function to accept optional model_name, game and pickup_name parameters.
While I'd usually put pickup_name as the first param, AFAIK that'd mean that if one only wants to change the model name (which is much more likely), that one always has to provide `Unknown item` as the first argument.

One step forwards to seperating ETM from nothing I suppose.

Fixes #5146

### DIFF
--- a/randovania/generator/pickup_pool/pickup_creator.py
+++ b/randovania/generator/pickup_pool/pickup_creator.py
@@ -137,17 +137,19 @@ def create_nothing_pickup(resource_database: ResourceDatabase, model_name: str =
     )
 
 
-def create_visual_etm() -> PickupEntry:
+def create_visual_etm(game: RandovaniaGame = RandovaniaGame.METROID_PRIME_ECHOES,
+                      model_name: str = echoes_items.USELESS_PICKUP_MODEL,
+                      pickup_name: str = "Unknown item") -> PickupEntry:
     """
     Creates an ETM that should only be used as a visual pickup.
     :return:
     """
     return PickupEntry(
-        name="Unknown item",
+        name=pickup_name,
         progression=(),
         model=PickupModel(
-            game=RandovaniaGame.METROID_PRIME_ECHOES,
-            name=echoes_items.USELESS_PICKUP_MODEL,
+            game=game,
+            name=model_name,
         ),
         pickup_category=pickup_category.USELESS_PICKUP_CATEGORY,
         broad_category=pickup_category.USELESS_PICKUP_CATEGORY,
--- a/randovania/generator/pickup_pool/pickup_creator.py
+++ b/randovania/generator/pickup_pool/pickup_creator.py
@@ -5,8 +5,6 @@ from typing import TYPE_CHECKING
 from randovania.game_description.pickup import pickup_category
 from randovania.game_description.pickup.pickup_entry import PickupEntry, PickupGeneratorParams, PickupModel
 from randovania.game_description.resources.location_category import LocationCategory
-from randovania.games.game import RandovaniaGame
-from randovania.games.prime2.patcher import echoes_items
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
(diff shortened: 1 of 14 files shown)


randovania/generator/pickup_pool/pickup_creator.py
```diff
@@ -137,17 +137,19 @@ def create_nothing_pickup(resource_database: ResourceDatabase, model_name: str =
     )
 
 
-def create_visual_etm() -> PickupEntry:
+def create_visual_etm(game: RandovaniaGame = RandovaniaGame.METROID_PRIME_ECHOES,
+                      model_name: str = echoes_items.USELESS_PICKUP_MODEL,
+                      pickup_name: str = "Unknown item") -> PickupEntry:
```
ThanatosGit, line 142: Don't know why I look at this in a 5 minute break...
Why using the defaults?
Afaik there should be a test for am2r using this function, which now falls back to the defaults and would break if the test contains an unknown item (which it maybe should).

randovania/generator/pickup_pool/pickup_creator.py
```diff
@@ -137,17 +137,19 @@ def create_nothing_pickup(resource_database: ResourceDatabase, model_name: str =
     )
 
 
-def create_visual_etm() -> PickupEntry:
+def create_visual_etm(game: RandovaniaGame = RandovaniaGame.METROID_PRIME_ECHOES,
```
ThanatosGit, line 140: Something against renaming the function to a name which makes sense for none prime players?

ThanatosGit (dismissed):

randovania/generator/pickup_pool/pickup_creator.py
```diff
@@ -137,17 +137,22 @@ def create_nothing_pickup(resource_database: ResourceDatabase, model_name: str =
     )
 
 
-def create_visual_etm() -> PickupEntry:
+def create_visual_nothing(game: RandovaniaGame = RandovaniaGame.METROID_PRIME_ECHOES,
```
henriquegemignani, line 140: I'd make the two first arguments not have defaults

henriquegemignani (changes requested):

randovania/generator/pickup_pool/pickup_creator.py
Miepee: What values do the other games expect? AFAIK every game is still built around the idea of having an ETM passed in?
Wondering whether i should keep it that way for now and put it off to the future, or just fix it while i'm already here.

henriquegemignani: These could likely be fields in the PickupDatabase. The PatchDataFactory has a `pickup_db` field even.

If refactoring, it'd be nice to rename all these `visual_etm` arguments all over the place too, but I abstained in requesting it :P

randovania/generator/pickup_pool/pickup_creator.py
ThanatosGit: > So that all the other games using this method don't suddenly have their behaviour changed.

But the default is just wrong for other games and if you already touch it, I prefer a proper refactor.

henriquegemignani (approved):

PR #5126, opened 2023-08-23
Bump gamma health req + fix omega super

Miepee:
199 health means 298.5 for the default 1.5x multiplier
Dividing that by 2 (for one suit) gives you 149.25

### DIFF
--- a/randovania/games/am2r/json_data/header.txt
+++ b/randovania/games/am2r/json_data/header.txt
@@ -223,7 +223,7 @@ Templates
               Charge Beam and Missile-Less Metroid Fights (Advanced)
           Any of the following:
               # Energy requirements
-              Combat (Expert) or Normal Damage ≥ 99
+              Combat (Expert) or Normal Damage ≥ 199
 
 * Defeat Zeta:
       All of the following:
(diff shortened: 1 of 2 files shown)


randovania/games/am2r/json_data/header.txt
```diff
@@ -223,7 +223,7 @@ Templates
               Charge Beam and Missile-Less Metroid Fights (Advanced)
```
JeffGainsNGames, line 223: Would it be worth it bumping `missile-less metroid fights` up to expert to be in line with dual alphas? Considering shots can get interrupted or you take damage and lose the charge.

Miepee: This is for normal single dodging gammas though. Dual alphas are expert because you got two things going at you. A single gamma is much easier in that regard.

JeffGainsNGames: Right, then I guess the more appropriate comment would be should dual gammas get a new template? 
image

Miepee: Correct, but that's for the future:tm:

JeffGainsNGames: Will add this to the "to do" pile I have and close this comment.

JeffGainsNGames (approved):

PR #5131, opened 2023-08-23
Improvements for test bulk generation

henriquegemignani:
- Also use for Prime 1, 2 and Cave Story, but weekly.
- Fix total cpu time
- Covert script to python


tools/bulk_generate_for_discord.py
```diff
@@ -0,0 +1,100 @@
...
+    target_seed_count = args.target_seed_count
+    process_count = args.process_count
+
+    output_path = this_dir.joinpath("bulk")
+    rdvgame_path = output_path / "rdvgame"
+    report_path = output_path / "report.json"
+    cpu_time_path = output_path / "cpu-time.txt"
+
+    webhook_url = "fun"
+    # webhook_url = os.environ["WEBHOOK_URL"]
```
Miepee, line 33: Whats up with this?

henriquegemignani: big "oops shouldn't be here"

tools/bulk_generate_for_discord.py
```diff
@@ -0,0 +1,94 @@
...
+import sys
+import tarfile
+from pathlib import Path
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--game", required=True)
+    parser.add_argument("--preset", default="Starter Preset")
+    parser.add_argument("--target-seed-count", type=int, default=2)
```
duncathan, line 16: only 2?

.github/workflows/daily-generate.yml
```diff
@@ -25,13 +38,31 @@ jobs:
       - name: Install Python packages
         run: venv/bin/python -m pip install -c requirements.txt -e .[server]
 
-      - name: Bulk Generate
-        run: |
-          source venv/bin/activate
-          ./tools/bulk_generate_for_discord.sh
+      - name: Bulk Generate (Cave Story)
+        run: venv/bin/python tools/bulk_generate_for_discord.py --game cave_story
```
duncathan, line 42: can we run this for both Starter Preset and CSR Classic? both are supported as defaults and should both ideally be tested by this

duncathan (changes requested):

tools/bulk_generate_for_discord.py
henriquegemignani: debugging thing oops

.github/workflows/daily-generate.yml
henriquegemignani: Can we do that later? I'm not even confident I didn't mess up these cron things

duncathan: would it be anything more than just adding a second line to `run` here with `--preset "CSR Classic"`?

henriquegemignani: If you want to run both together yeah. I think CS generates faster so that might be fine? I don't want to keep the agent locked for too long as it can prevent CI for the other repositories

duncathan: yeah, running both together. CS is quick to generate and if it's really a problem I'd rather have 50 of each at the same time than 100 of each at different times

PR #5134, opened 2023-08-24
Preset patch for nerfing power bombs

steven11sjf:
*Note: as of 8/23/2023 this requires ODR to be updated/released to support the latest MEDS changes* 

Adds an option to Dread presets to nerf Power Bombs. This setting removes the ability to destroy Enkies and open Charge Beam Doors, and is accounted for in logic. Power Bombs are still logically required as they are needed to break the reactor before Raven Beak, but this setting makes Ice Missile and Charge Beam more logically useful in presets with knowledge enabled. 

It does seem that in the seeds I rolled it almost becomes another "artifact" where it gets pushed to the end of generation. I think it might make sense to make it consistently destroy diffusion blobs to return some of its strength, but that will require further testing to ensure it works. 

Setting is shown below; would appreciate any feedback on wording. There's probably a less patch-notes-y name for this than "nerf" but i'm coming up blank. 

image


randovania/layout/preset_migration.py
```diff
@@ -966,6 +966,14 @@ def _migrate_v61(preset: dict) -> dict:
     return preset
 
 
+def _migrate_v62(preset: dict) -> dict:
+    config = preset["configuration"]
+    game = preset["game"]
+
+    if game in {"dread"}:
```
henriquegemignani, line 973: It's fine to do `game == "dread"` :)

henriquegemignani:
Can call it "Reduced Power Bombs", or "Limited".

When you list the things that a PB destroy, you should also include it destroys most enemies in one hit, including bosses such Robots/Chozo Warriors.
And then when you mention the setting limits just Enkys/Charge doors.

Regarding logic: weird that's an issue. There's quite a few PB blocks around the game that locks things, as well as being an alternative to Bombs. Weird that PBs would be left for last. Can try bulk generating games with and without this change and seeing the results.

XenoWars:
Power Bomb Limitations perhaps?

Also, the long name in header.json is what appears in the database, and that seems a bit too long. Just letting it be whatever the setting is called would be fine and gets the point across, be it Reduced/Limited/Nerf Power Bombs or Power Bomb Limitations.

henriquegemignani (approved):

PR #5153, opened 2023-08-25
Dread: Adjust boss fight logic to account for Split Beams

Tyranisaur:


randovania/games/dread/json_data/header.txt
```diff
@@ -383,6 +386,31 @@ Templates
...
+* Beam Pierces Enemies:
+      Plasma Beam or Wave Beam
+
+* Beam Damage 75:
+      Any of the following:
+          Wide Beam
+          Plasma Beam and Wave Beam
+          All of the following:
+              Disabled Separate Beam Behaviors
+              Plasma Beam or Wave Beam
```
XenoWars, line 398: don't think that plasma and wave should be in the 75 if split beams is disabled.

XenoWars (changes requested):

randovania/games/dread/json_data/header.txt
duncathan: it doesn't really matter since split beams is always enabled, no matter what

Tyranisaur: If split beams is disabled, when you have a beam, you automatically also have all the previous beams. However, you don't have the previous beams as items in the logical inventory. Hence why, when you reason about what happens when split beams is disabled, having plasma beams implies that you also have wide, and having plasma beam implies that you also have wide and plasma. I don't think this template is used in any places where you don't also have a higher higher beam damage template at a lower difficulty rating. But this part encodes that you effectively satisfy a >= 75 condition when you have 150 or 240.

duncathan: now, again, split beams is *never* disabled, so this does not need to be accurate

XenoWars (approved):

PR #5156, opened 2023-08-26
Add common Mono Mac paths for echoes exporting

### LINKED ISSUE #4474: Prime 2 Exporter: Expose Mono paths for macOS
Most users start randovania on macOS by double clicking the program from their file explorer. This has the unfortunate disadvantage that RDV is spawned from a non-login non-interactive shell, and thus is not able to find the users Mono path, because those are prepended in login start scripts.
Thus I propose to whenever start mono, prepend the following paths:
- `/Library/Frameworks/Mono.framework/Versions/Current/Commands` = default installation path from mono's website
- `/usr/local/bin` = installation path for programs installed via homebrew on x64 based macs
- `/opt/homebrew/bin` = installation path for programs installed via homebrew for arm based macs.

A user *could* theoretically have installed mono somewhere else, but these three paths should cover 99% of all installation cases.

Miepee:
Fixes #4474

### DIFF
--- a/randovania/games/prime2/patcher/csharp_subprocess.py
+++ b/randovania/games/prime2/patcher/csharp_subprocess.py
@@ -9,8 +9,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
-    from os import _Environ
+    from collections.abc import Callable, Sequence
 
 IO_LOOP: asyncio.AbstractEventLoop | None = None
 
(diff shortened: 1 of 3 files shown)


test/games/patchers/test_csharp_subprocess.py
```diff
@@ -35,8 +42,13 @@ def test_is_windows(mocker, system_name, expected):
 
 
 @pytest.mark.parametrize("add_mono", [False, True])
-def test_process_command_no_thread(mock_is_windows, mocker: pytest_mock.MockerFixture, monkeypatch,
+@pytest.mark.skipif(mock_is_windows and mock_is_mac,
```
henriquegemignani, line 45: Does this actually work?

Miepee: Double checked, it doesn't. Whoops.
That currently causes it to skip the test in its entirety.

henriquegemignani: just do a `if a and b: pytest.skip()` in the function body

Miepee: Addressed now
grafik

henriquegemignani (changes requested):
Can you solve the mypy errors too?

randovania/games/prime2/patcher/csharp_subprocess.py
```diff
@@ -40,11 +46,13 @@ async def _read_data(stream: StreamReader, read_callback: Callable[[str], None])
             break
 
 
-async def _process_command_async(args: list[str], input_data: str, read_callback: Callable[[str], None]):
+async def _process_command_async(args: list[str], input_data: str, read_callback: Callable[[str], None],
+                                 envs: _Environ | dict = os.environ):
```
henriquegemignani, line 50: Change this to be a `addition_path_entries: collections.abc.Sequence[str]`, then make this function responsible for creating the new env dict.

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #5166, opened 2023-08-27
Fix mypy errors in EchoesPatchDataFactory

henriquegemignani:


randovania/games/prime2/exporter/patch_data_factory.py
```diff
@@ -235,9 +237,9 @@ def _create_elevator_scan_port_patches(
         game: RandovaniaGame,
         region_list: RegionList,
         get_elevator_connection_for: Callable[[DockNode], Node],
-        elevator_dock_types: list[DockType]
+        elevator_dock_type: DockType
```
ThanatosGit, line 240: Things like that are good examples to use it in more files.

ThanatosGit (approved):

PR #5182, opened 2023-08-30
Generator: if all actions have weight 0, use reachable node count as …

henriquegemignani:
…backup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [6.3.0] - 2023-10-??
 
-- Nothing yet.
+- Added: During generation, if no alternatives have a non-zero weight, try weighting by how many additional Nodes are reachable.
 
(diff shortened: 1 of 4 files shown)


toasterparty:
How can there ever be a non-zero set of possible actions with sum weight zero to begin with?

henriquegemignani:
> How can there ever be a non-zero set of possible actions with sum weight zero to begin with?

Weight of an action is done by:
- performing the action
- then calculating the reach
- collecting all safe resources in that reach
- **not** attempting to identify if it can calculate unsafe resources
- identifying how many _additional_ pickup indices, hint nodes, events are collectable compared to before the actions
- magical constants to transform the counts above in a weight

There are situations where some set of actions will unlock more nodes, but these nodes don't provide any new **safe** pickup/hint/event so it ends with weight 0.
For an example, see https://discord.com/channels/914291389293027329/914294505107251231/1146537960121061446. TLDR: Echoes, goes to Storage D without Dark Beam and before fighting DS1.

This PR solves that situation. Might solve others, haven't tested.

PR #5189, opened 2023-09-01
add faq about emmi bug with split weapons

duncathan:


randovania/games/dread/game_data.py
```diff
@@ -109,6 +109,11 @@ def _generator() -> game.GameGenerator:
             "Unlocking this door before turning on the power will render it unopenable.\n\n"
             "To fix this, simply save and reload the game."
         ),
+        (
+            "I received a Beam/Missile upgrade from an E.M.M.I., and now my arm cannon doesn't work!",
+            "Reload from checkpoint immediately to fix the issue. "
+            "Your checkpoint was saved after killing the E.M.M.I., so you will not lose progress."
```
Miepee, line 115:
```suggestion
            "Reload from a checkpoint to fix the issue. "
            "A checkpoint was created after killing the E.M.M.I., so you will not lose progress."
```

Sounds less intimidating and slightly nicer

duncathan: I prefer the urgency, since you're unable to shoot your beam while the glitch is active and it can crash the game if you're not careful

Miepee: Fair nuff

Miepee (approved):

PR #5193, opened 2023-09-02
AM2R: Fix most mypy errors in exporter

Miepee:
Leaves some errors as is, due to either me not knowing how to fix them, or because the error is just a consequence of something other deeply rooted


randovania/games/am2r/exporter/patch_data_factory.py
```diff
@@ -110,18 +115,18 @@ def get_locked_ammo_text(ammo_item: str):
...
-            "fusion_mode": self.patches.configuration.fusion_mode,
-            "supers_on_missile_doors": self.patches.configuration.supers_on_missile_doors,
-            "nest_pipes": self.patches.configuration.nest_pipes,
-            "softlock_prevention_blocks": self.patches.configuration.softlock_prevention_blocks,
-            "a3_entrance_blocks": self.patches.configuration.a3_entrance_blocks,
-            "screw_blocks": self.patches.configuration.screw_blocks,
+            "septogg_helpers": config.septogg_helpers,
+            "respawn_bomb_blocks": config.respawn_bomb_blocks,
+            "skip_cutscenes": config.skip_cutscenes,
+            "skip_item_cutscenes": config.skip_item_cutscenes,
```
ThanatosGit, line 121: Afaik the `skip_save_cutscene` needs to be merged here now.

randovania/games/am2r/exporter/patch_data_factory.py
```diff
@@ -77,27 +81,28 @@ def _create_room_dict(self):
...
+    def _create_hash_dict(self) -> dict:
         return {
             "word_hash": self.description.shareable_word_hash,
             "hash": self.description.shareable_hash,
             "session_uuid": str(self.players_config.get_own_uuid())
         }
 
-    def _create_game_patches(self, pickup_list: list[ExportedPickupDetails], item_info: dict, rng: Random):
-        def get_locked_ammo_text(ammo_item: str):
+    def _create_game_patches(self, config: AM2RConfiguration, pickup_list: list[ExportedPickupDetails],
```
ThanatosGit, line 103: Was there s need to pass it now as a parameter instead of accessing it via `self.configuration` ( or assign it in the method itself to a new variable `config` to use a shorter name)?

ThanatosGit (changes requested):

randovania/games/am2r/exporter/patch_data_factory.py
Miepee: I'm *slowly* trying to refactor the patch data factory a bit, to have the methods not as coupled to the factory. While poking around in dread/SM, and needing to write the test for music shuffling, I noticed that it's something that's helpful.

randovania/games/am2r/exporter/patch_data_factory.py
Miepee: Fixed via merging

ThanatosGit (approved):

PR #5209, opened 2023-09-05
More mypy typing

henriquegemignani:


randovania/bitpacking/construct_pack.py
```diff
@@ -13,17 +13,32 @@
...
 T = typing.TypeVar("T")
 
+JsonValue = str | int | float | list["JsonValue"] | dict[str, "JsonValue"]
+
+
+@typing.runtime_checkable
+class JsonEncodable(typing.Protocol):
+    @classmethod
+    def from_json(cls, value: JsonValue) -> typing.Self:
+        ...
```
Miepee, line 31: What do these 3 dots do?

randovania/game_description/data_writer.py
```diff
@@ -256,7 +259,7 @@ def write_node(node: Node) -> dict:
     :return:
     """
 
-    data = {}
+    data: dict = {}
```
Miepee, line 262: Why is the type here necessary lol. Does mypy  really not see `{}` as a dict otherwise?

randovania/games/cave_story/exporter/__init__.py
Miepee, line 1: Why not am2r?

Miepee (commented):
Only briefly skimmed through it.
Personally wish there could be a way to have less # type: ignore's

randovania/games/cave_story/exporter/__init__.py
henriquegemignani: Because I dug this commit from an old branch from before am2r was a thing :P

randovania/game_description/data_writer.py
henriquegemignani: It asks to specify the key/value types of the dict, which I'm then explicitly being generic. Could probably used `dict[str, Any]` instead. Or some `JsonObject` defintiion.

randovania/bitpacking/construct_pack.py
duncathan: nothing! it just indicates that children are expected to implement this function (consider it abstract)

henriquegemignani:
> Only briefly skimmed through it. Personally wish there could be a way to have less # type: ignore's

The ones on imports are for external packages that don't have types, or requires an additional package. These can be fixed later.
The other case is the obfuscator thing which is codegen during CI only, so uhhh yeah

The ones in bitpacking are the code being funny and doing violations. There's particularly quite some BitPackValue usage in there that is just wrong. Gonna leave for redesigning that module later.

The `RelativeData` ones I just couldn't fix. The functions are a big weird there and I couldn't figure out how to handle it.

randovania/bitpacking/construct_pack.py
henriquegemignani: I copied from typing.Protocol documentation, saw it as "dummy function thing" and seems to be it. Looking it up, called `Ellipsis` and seems to be a fancy way of doing `pass`?

Miepee: if it's the same thing as pass, then i'd rather just use pass.

duncathan:
is it possible to get reports on typing coverage, like we do for code coverage?

randovania/bitpacking/construct_pack.py
duncathan: it's functionally equivalent to `pass`, but semantically it means "this must be implemented in the children" without explicitly raising a NotImplementedError

Miepee: ah ok

duncathan: upon further inspection: it's being used here because this Protocol is serving as a sort of formalized duck typing. by using an ellipse, the meaning becomes more clear: any object with methods of this name/signature is considered to be this type, regardless of the contents of those methods. using `pass` may give the wrong impression that these methods must be no-op. ALSO, I'd bet that mypy would be unhappy about the functions not returning their assigned types, but has special handling for the ellipse here

henriquegemignani:
> is it possible to get reports on typing coverage, like we do for code coverage?

[Seems like yes](https://mypy.readthedocs.io/en/stable/command_line.html#report-generation), but I have no clue on how to add a visualisation. So far, the files listed in pyproject.toml are forced to have 0 warnings/errors.

PR #5210, opened 2023-09-05
Samus Returns - Add Hash List

dyceron:


randovania/games/samus_returns/assets/hash_words.json
```diff
@@ -0,0 +1,114 @@
...
+  "Hatchling",
+  "Arachnus",
+  "Diggernaut",
+  "Proteus",
+  "Ridley",
+  "Animworm",
+  "Autoad",
+  "Autrack",
+  "BigBug01",
+  "BigBug02",
```
Miepee, line 20: I dunno if it's that good of an idea to expose internal MSR names like this
(Same with animworm too i think?)

randovania/games/samus_returns/assets/hash_words.json
```diff
@@ -0,0 +1,114 @@
...
+  "Gunship",
+  "Door",
+  "Inactive",
+  "Closed",
+  "Green",
+  "Yellow",
+  "Map",
+  "Boost",
+  "Reserve",
+  "amiibo",
```
Miepee, line 105: ngl, I think we should just screw the normal stylization and enforce Title Case.

Miepee (commented):
`Purple` and `Liquid` are also other ideas in case you need more.

randovania/games/samus_returns/assets/hash_words.json
dyceron: Oh is the lowercase gonna cause issues? I'll fix it

Miepee: Probably wont cause issues, it just looks bad :stuck_out_tongue_closed_eyes:

randovania/games/samus_returns/assets/hash_words.json
dyceron: I thought it'd be funny tonight include them but yeah those three can go

randovania/games/samus_returns/assets/hash_words.json
```diff
@@ -0,0 +1,120 @@
...
+  "Teleport",
+  "Thrower",
+  "Tsumuri",
+  "Varia",
+  "Wallfire",
+  "Water",
+  "Wave",
+  "Yellow",
+  "Zebes",
+  "Zeta"
```
Miepee, line 119: can you add a trailing comma please, just in case that more words will get added in the future?

henriquegemignani: invalid json

Miepee: ah this is json, not python code, whoops.
Disregard

Miepee (approved):

PR #5220, opened 2023-09-06
AM2R: Add progressives

Miepee:
Adds progressive suit and jumps
None of them are enabled by default at the moment, as I'd like to only do that once I get non-placeholder sprites for them.

### DIFF
--- a/randovania/games/am2r/pickup_database/progressive_items.py
+++ b/randovania/games/am2r/pickup_database/progressive_items.py
@@ -1,7 +1,12 @@
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
 
-def tuples() -> tuple:
+if TYPE_CHECKING:
+    from randovania.games.game import Progressive_item_tuples
+
+
+def tuples() -> Progressive_item_tuples:
     return [
         ("Progressive Suit", ("Varia Suit", "Gravity Suit")),
         ("Progressive Jump", ("Hi-Jump Boots", "Space Jump")),
--- a/randovania/games/am2r/pickup_database/progressive_items.py
+++ b/randovania/games/am2r/pickup_database/progressive_items.py
@@ -3,13 +3,6 @@ from __future__ import annotations
 
 def tuples():
     return [
-        (
-            "Progressive Suit",
-            ("Varia Suit", "Gravity Suit")
-        ),
-        (
-            "Progressive Jump",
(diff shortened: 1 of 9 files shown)


randovania/games/am2r/pickup_database/progressive_items.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
+
+
+def tuples() -> tuple:
```
ThanatosGit, line 4: This should be a list of tuples, where the first element in the list is a string (the progressive name) and the second element is a tuple with an arbitrary amount of strings for the non-progressive name.
So `list[tuple[str, tuple[str, ...]]]` ?

ThanatosGit (changes requested):

randovania/games/am2r/pickup_database/progressive_items.py
henriquegemignani: Would be great to define a type alias somewhere for this mess haha

Miepee: Yeah, as all the other games will need the same type hint. Should i do that in this PR, or just add a TODO for the future?

ThanatosGit: Hmmm in feature freeze I wouldn't change anything in other games but it's also just a type hint.
I would say: Add a type alias somewhere where other games can use it in the future, which means don't just put in the AM2R specific file (I hope that there exists a nice spot for it)

randovania/games/game.py
```diff
@@ -32,6 +32,9 @@
     from randovania.layout.preset_describer import GamePresetDescriber
     from randovania.resolver.bootstrap import Bootstrap
 
+# Type alias for progressive items
+Progressive_item_tuples = list[tuple[str, tuple[str, ...]]]
```
ThanatosGit, line 36: Iirc we use PascalCase so far for a type alias. Like `TeleporterConnection` etc.

ThanatosGit (changes requested):

randovania/games/game.py
henriquegemignani: Technically it should be `Iterable[tuple[str, tuple[str, ...]]]`, and please also update GameGui.progressive_item_gui_tuples :D

ThanatosGit (approved):

PR #5223, opened 2023-09-07
Samus Returns - Add Progressive Jump

dyceron:


randovania/games/samus_returns/pickup_database/pickup-database.json
```diff
@@ -412,7 +412,7 @@
             "progression": [
                 "Boots"
             ],
-            "default_shuffled_count": 1,
+            "default_shuffled_count": 0,
```
ThanatosGit, line 415: @Miepee Déjà-vu...we better keep it at 1 and do the fancy special case stuff for the description?
Same for space

dyceron: Oh this will cause issues?

Miepee: The pickup database isnt the starter preset. defining 0 as the default shuffled count in here, will make text display weird. You can look at dreads for refernce
https://github.com/randovania/randovania/blob/main/randovania/games/dread/pickup_database/pickup-database.json#L357-L408

While you're add it, can you fix the "broken" numbers for progressive suits (and beams maybe too? havent looked at them) too please?

Miepee (approved):

PR #5234, opened 2023-09-08
AM2R: Add extra data to area transitions

Miepee:
Boy do I love unreviewable PRs


henriquegemignani (approved):
Very easy to see you added just the instance_id, source_x, source_y :)

What do mean check the values? If they're wrong that's a future you problem :P

Miepee:
Exactly xP
Not like anyone wants to go through the effort of fact checking each of them

PR #5245, opened 2023-09-10
Add bash scripts for debug client/server for Unix

Miepee:
I was too lazy to type these all out every time for MW testing.


henriquegemignani:
I don't like the start server scripts even for windows but 🤷

PR #5246, opened 2023-09-10
AM2R: Fix possible crash for MW patch data factory

Miepee:
Currently hardcoded, because I havent implemented a real system for custom models. Will fix this properly later such a system was implemented, just wanna get rid of the crash for now.

### DIFF
--- a/randovania/games/am2r/exporter/patch_data_factory.py
+++ b/randovania/games/am2r/exporter/patch_data_factory.py
@@ -48,10 +48,12 @@ class AM2RPatchDataFactory(PatchDataFactory):
                                "Locked" in res_lock.temporary_item.long_name and
                                len(pickup.collection_text) > 1) else 0
 
+            # TODO: figure out later on how to get proper sprite names and sprite animation speed for offworld items
             pickup_map_dict[object_name] = {
                 "sprite_details": {
                     "name": pickup.model.name,
                     "speed": item_info[pickup.original_pickup.name]["sprite_speed"]
+                    if not self.players_config.is_multiworld else 0.2
                 },
                 "item_effect": pickup.original_pickup.name if not pickup.other_player else "Nothing",
                 "quantity": quantity,


randovania/games/am2r/exporter/patch_data_factory.py
```diff
@@ -48,10 +48,12 @@ def _create_pickups_dict(self, pickup_list: list[ExportedPickupDetails], item_in
                                "Locked" in res_lock.temporary_item.long_name and
                                len(pickup.collection_text) > 1) else 0
 
+            # TODO: figure out later on how to get proper sprite names and sprite animation speed for offworld items
             pickup_map_dict[object_name] = {
                 "sprite_details": {
                     "name": pickup.model.name,
                     "speed": item_info[pickup.original_pickup.name]["sprite_speed"]
+                    if not self.players_config.is_multiworld else 0.2
```
henriquegemignani, line 56: This entire block you should've been looking at `pickup.model` instead of depending on the pickup name. There's already some handling to map the other pickup into something your this game, or the default offworld model.

henriquegemignani (changes requested):

randovania/games/am2r/exporter/patch_data_factory.py
Miepee: Currently there is no way for me to know a sprite's animation speed based solely on its name. It's what i meant with `Will fix this properly later such a system was implemented` :P 
I'd need to have some mapping for this somewhere, which i didnt want to do for a simple "just stop crashing"-fix

henriquegemignani: The big thing is why are you using `pickup.original_pickup.name` instead of `pickup.model.name`?

Miepee: `pickup.model.name` will give me the model name. I.e. `sItemSpeedbooster` for `Speed Booster`.
The only way for me to get the anim speed for it is via [this mapping](https://github.com/randovania/randovania/blob/main/randovania/games/am2r/pickup_database/item_data.json) (i.e. `Speed Booster` -> `0.5`) - which requires the items name.
I guess I could do something like:
- traverse all pickups
- see if model name matches
  - if yes, use that model name to get the animation speed
- if no such model is found, use default of 0.2


but i should really just refactor the whole thing to have a sprite name -> animation speed mapping somewhere. Anything else just feels very clunky, as currently the system would allow two items to have the same model but different animation speed, which is just a thing that shouldnt happen.

henriquegemignani: Very clearly seems to be you have that data in the wrong place, yeah

PR #5260, opened 2023-09-11
Fix files that would've failed line length limit with black

henriquegemignani:


randovania/games/samus_returns/layout/preset_describer.py
```diff
@@ -42,17 +41,14 @@ def format_params(self, configuration: BaseConfiguration) -> dict[str, list[str]
                     "Progressive Suit": has_shuffled_item(standard_pickups, "Progressive Suit"),
                 }
             ],
-            "Gameplay": [
-                {f"Elevators/Shuttles: {configuration.teleporters.description()}":
-                 not configuration.teleporters.is_vanilla}
-            ],
+            "Gameplay": [],
```
Miepee, line 44: any specific reason why the elevaotr/shuffles were removed here?

henriquegemignani: it is not a thing. Actually now that I think about it surprised it doesn't error? I bet the function is not being used 😅

Miepee (approved):

randovania/games/samus_returns/layout/preset_describer.py
henriquegemignani: After checking, yup `MSRPresetDescriber` is unused oops

PR #5263, opened 2023-09-11
Add AM2R to daily generate

Miepee:
grafik
am2r is relatively quick to generate, so would that like on two days
bumped the time to be earlier to hopefully be less conflicting with other schedules (and also because it aligns better with my timezone)

### DIFF
--- a/.github/workflows/daily-generate.yml
+++ b/.github/workflows/daily-generate.yml
@@ -72,6 +72,12 @@ jobs:
         env:
           WEBHOOK_URL: ${{ secrets.SR_DEV_TALK_WEBHOOK }}
 
+      - name: Bulk Generate (AM2R)
+        run: venv/bin/python tools/bulk_generate_for_discord.py --game am2r
+        if: inputs.game == 'am2r' || github.event.schedule == '0 10 * * 1,5'
+        env:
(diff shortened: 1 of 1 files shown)


dyceron (changes requested):
You're missing the section under `steps` for a2mr

Miepee:
oops

Miepee:
Would also need an orga/discord mod to add `AM2R_DEV_TALK_WEBHOOK` to the repo/server

.github/workflows/daily-generate.yml
```diff
@@ -12,13 +12,15 @@ on:
...
           - samus_returns
+          - am2r
 
   schedule:
     - cron: '0 16 * * 3'  # Cave Story
     - cron: '0 23 * * *'  # Dread
     - cron: '0 16 * * 2'  # Prime 1
     - cron: '0 16 * * 1'  # Prime 2
     - cron: '0 16 * * 4'  # Samus Returns
+    - cron: '0 10 * * 1,5' # AM2R
```
dyceron, line 23: Is 1,5 going to cause issues? I know it's a regional thing but I am curious.

Miepee: See the image posted in the PR description :)

dyceron (approved):

PR #5271, opened 2023-09-12
Generator: Fix not wanting to place etanks, when damage req would be just high enough to kill player

Miepee:
I only tested this with ~2k seeds with am2r

Should *probably* be tested for other games, can imagine that echoes could not like this much.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [6.3.0] - 2023-10-??
 
 - Added: During generation, if no alternatives have a non-zero weight, try weighting by how many additional Nodes are reachable.
+- Changed: The generator will now consider placing Energy Tanks, if there's a damage requirement that's exactly high enough to kill the player.
 - Fixed: The menu option for viewing all Randovania dependencies and their licenses has been restored.
 
 ### Metroid Dread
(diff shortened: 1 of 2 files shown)


ThanatosGit:
I think this is more important for the db maintainers and not for code maintainers.
It's about the question if db maintainers use `<` or `<=` for the damage requirements.
As I'm not maintaining any db :-), it's not up to me to decide.

Miepee:
I feel like this is more an issue on the code side. If you have 99 energy, and there is a path forward with some requirement listed in the DB that takes 100 energy , but which after multipliers/divisors end up at 99 again, and the generator doesnt consider health, then it's definitely the generators fault. Increasing health in the DB would make it technically less accurate and just to workaround  a generator bug (which even with increased health can still happen)

ThanatosGit:
I don't get that example at all.

> If you have 99 energy, and there is a path forward with some requirement listed in the DB that takes 100 energy , but which after multipliers/divisors end up at 99 again [...]

Then it's "more wrong" to use `<=` as the energy on the path has a wrong value due to floating point stuff? Before it was `99 < 99` and now it is `99 <= 99` but it should be compare with 100 e.g. `99 <= 100` from your example?

It's pretty much a db thing for me as I have already no idea if people write down the required energy or the required damage you need to take. To me: 99 energy and a path needs 99 damage means, you have 0 health after this path, which means you do not survive. 99 energy and a path requires 99 energy is fine.

henriquegemignani:
If there's a path that requires 99 damage and you have 99 health, then we _should_ place an etank as we don't have enough.

The modified check means "do we die if we take this path? If yes, then consider placing more health".

henriquegemignani (approved):

PR #5264, opened 2023-09-11
Super Metroid: Crateria Logic Update

Logarhythmic:
A minor first wave of SM logic fixes. I wanted to make sure I'm doing everything correctly before I submit a much larger PR.

PR #5273, opened 2023-09-12
SM: exclude beginner shinespark instead of specific items.

Miepee:
This is nicer than listing every single pickup.

Noticed this while investigating why tests failed for #5264


SolventMercury (approved):
This all seems correct.

SolventMercury:
New changes look good, only tests that are complaining are codecov. I think this is good to merge.

PR #5277, opened 2023-09-12
Generator: handle negative requirements a little better

Miepee:
Ran 2k of am2r+prime+echoes, couldnt see any big regressions so far.

prime and dread probably would benefit the most from this

### DIFF
--- a/randovania/generator/filler/pickup_list.py
+++ b/randovania/generator/filler/pickup_list.py
@@ -46,7 +46,9 @@ def _unsatisfied_item_requirements_in_list(
     damage = []
 
     for individual in alternative.values():
-        if individual.negate:
+        # We should only skip negative requirements, if we can not satisfy them anymore, as that means
+        # we have lost the chance to do so
+        if individual.negate and not individual.satisfied(state.resoures, state.energy, state.resource_database):
             possible = False
             break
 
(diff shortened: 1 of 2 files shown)


randovania/generator/filler/pickup_list.py
```diff
@@ -46,7 +46,9 @@ def _unsatisfied_item_requirements_in_list(
     damage = []
 
     for individual in alternative.values():
-        if individual.negate:
+        # We should only skip negative requirements, if we can not satisfy them anymore, as that means
+        # we have lost the chance to do so
+        if individual.negate and not individual.satisfied(state.resoures, state.energy, state.resource_database):
```
ThanatosGit, line 51: This looks not correct. `resoures` is a typo and maybe your result will be completly different if you use `resources`

ThanatosGit (dismissed):

randovania/generator/filler/pickup_list.py
Miepee: Whoops, no idea how that made it in.

ThanatosGit: On a second thought: This is python. Shouldn't it throw an `AttributeError` if this was ever called?

randovania/generator/filler/pickup_list.py
```diff
@@ -46,7 +46,9 @@ def _unsatisfied_item_requirements_in_list(
     damage = []
 
     for individual in alternative.values():
-        if individual.negate:
+        # We should only skip negative requirements, if we can not satisfy them anymore, as that means
+        # we have lost the chance to do so
+        if individual.negate and not individual.satisfied(state.resources, state.energy, state.resource_database):
```
henriquegemignani, line 51: You could move the existing check to after the existing satisfied check. Could just add an ` or individual.negate`
Damage requirements are never negated (there's an assert enforcing that).

henriquegemignani (changes requested):

PR #5301, opened 2023-09-14
Set negate to False not None when migrating old rdvgames

henriquegemignani:


Miepee:
What if someone migrated to a newer rdvgame version without this fix? Will their negates still have None?

henriquegemignani:
~~I think when we save an rdvgame parsed from this it writes a boolean properly.~~ Checked and it'd save whatever was in negate. Ugh. I could add another migration that fixes this, but I could just lol whatever.

The reason for fixing this is that the test files in randovania are kept in whatever old version they are, so they run this migration. I was testing a thing which made it strict and didn't like getting None instead of Bool.

Miepee:
Any complains with making a new issue/todo to fix that in the future:tm:?

henriquegemignani:
Sure, one of these forever issues :P

Miepee:
The alternative is you doing that for this PR :P

henriquegemignani:
That'd mean not opening yet another PR of the next shiny that took my attention!

Miepee (approved):

PR #5303, opened 2023-09-14
Even more typing

henriquegemignani:
And actually support for using mypyc, that'll be another PR.


randovania/game_description/requirements/resource_requirement.py
```diff
@@ -20,15 +19,27 @@
     from randovania.game_description.resources.simple_resource_info import SimpleResourceInfo
 
 
-@dataclasses.dataclass(frozen=True, slots=True)
```
duncathan, line 23: why remove the dataclass? does removing the frozen aspect affect anything?

randovania/generator/graph.py
```diff
@@ -13,28 +14,32 @@
 
 
 class BaseGraph:
-    def copy(self):
+    def copy(self) -> BaseGraph:
```
duncathan, line 17: is this not `typing.Self`?

randovania/generator/old_generator_reach.py
```diff
@@ -231,7 +231,8 @@ def connected_nodes(self) -> Iterator[Node]:
         :return:
         """
         self._calculate_reachable_paths()
-        all_nodes = self.all_nodes
+        assert self._reachable_paths is not None
+        all_nodes = typing.cast(tuple[Node, ...], self.all_nodes)
```
duncathan, line 235: what's with all the casting for this? is that really desirable?

randovania/game_description/requirements/resource_requirement.py
henriquegemignani: mypyc doesn't like this much :(

randovania/generator/old_generator_reach.py
henriquegemignani: it's a weird thing. The `all_nodes` array does in fact have holes in it, if you iterate.
But the way we do it here, we're always using indices that are guaranteed to not be None.

duncathan: huh, ok. would it be worth making a helper so that we don't need to manually cast this every time?

henriquegemignani: Maybe? Annoyingly this is all in very hot code.

duncathan: ugh, right. so the overhead actually matters. I guess this is fine then

randovania/generator/graph.py
henriquegemignani: yes

randovania/game_description/requirements/resource_requirement.py
```diff
@@ -20,15 +19,27 @@
...
+        return type(self)(self.resource, self.amount, self.negate)
+
+    def __reduce__(self) -> tuple[type[ResourceRequirement], tuple[ResourceInfo, int, bool]]:
+        return type(self), (self.resource, self.amount, self.negate)
+
+    def __init__(self, resource: ResourceInfo, amount: int, negate: bool):
+        self.resource = resource
+        self.amount = amount
+        self.negate = negate
+        self.__post_init__()
```
ThanatosGit, line 41: I'm definitely missing some knowledge: Is the assertion to an object not always true? If that's the case, what is the purpose of this?

randovania/generator/reach_lib.py
ThanatosGit, line 1: What is the reason to introduce the two types here?

test/generator/filler/test_runner.py
```diff
@@ -197,6 +198,7 @@ def test_add_relative_hint(
         )
     else:
         max_distance = 20
+        precision = HintItemPrecision.GENERAL_CATEGORY
```
ThanatosGit, line 201: I guess the type error comes from the fact that mypy infers the type from the first usage.

test/generator/filler/test_pickup_list.py
```diff
@@ -36,7 +36,7 @@ def item(name):
             (item("Space Jump Boots"), 1),
         ]
     )
-    uncollected_resources = []
+    uncollected_resources = set()
```
ThanatosGit, line 39: Can this type be more specified for mypy?

ThanatosGit (approved):

test/generator/filler/test_runner.py
henriquegemignani: oh these warnings are good catch that I forgot to add these files as required for mypy

randovania/generator/reach_lib.py
henriquegemignani: Which types?

ThanatosGit: These two: https://github.com/randovania/randovania/blob/da0abdffe1e17ab7f2e8560dce7c310247dab414/randovania/generator/reach_lib.py#L18-L19

Are these required to allow for covariance in the iterables?

henriquegemignani: I'm not 100% sure they're used (and looking at the code now, maybe they aren't), but the idea is `_filter_reachable` now returns whatever you pass to it. So if you pass a `ResourceNode` iterable, or even a `DockNode` iterable, then the result is also one of these.

PR #5331, opened 2023-09-20
Dread: Exclude some connections from trickless

XenoWars:
- Changed various trickless instances of Speed Booster connections to require Speed Booster Conservation set to Beginner, completely excluding some item locations from trickless in the process.


XenoWars:
There may be some arguments to be had for some of these here, or even some not present. And while we are at it, are there any other connections that aren't speed-related that we should consider excluding from trickless?

test/generator/test_generator_reach.py
```diff
@@ -127,6 +127,8 @@ def _compare_actions(
 _include_tricks_for_game = {
     # Some items require shinesparking to reach in vanilla, which due to varying difficulty has been made into a trick
     RandovaniaGame.AM2R: {("Shinesparking", LayoutTrickLevel.ADVANCED)},
+    # Same reasons as above, with some still trickless
+    RandovaniaGame.DREAD: {("Speedbooster", LayoutTrickLevel.BEGINNER)},
```
Miepee, line 131: This needs to be `METROID_DREAD`, not `DREAD`

Tyranisaur (approved):

PR #5338, opened 2023-09-22
Built in Change Log Improvements

Wolfmyths:
This pull request is trying to fix most of the issues at #4621 

### Here is what I fixed/added:

+ Change log widget is easier to navigate through versions via combo box (instead of tabs)
+ Links now work correctly *(Used the .setOpenExternalLinks() method on DelayedLabel)*
+ The date that releases were published now show above the patch notes (*More about that in concerns below*)

### Notes:

I used a QStackedWidget to hide/show the given patch notes. The functionality of which version is being shown is in `select_version_index_changed()`. This function is triggered from the combo box's `currentIndexChanged` event

#4621 Mentions embedded images not working. Unless we do some kind of workaround involving downloading those images, I do not think it is possible right now because of PySide6's Markdown limitations

### How I tested:

I booted up the app like in the instructions in the readme.md and tested the functionality of the changelog widget by going to the change log in `Help -> Change Log`, clicking the combo box, and clicking on external links in the patch notes. The new dates also appear as explained earlier.

### Concerns:

So this is my first time contributing to a large project so please excuse me if I messed with anything fundamental. In order to get the dates for the releases I had to create a whole dictionary containing those dates `{tag_name : date}`

I also had to include the datetime module to parse the dates. Below is a quick comparison of parsing the date times.

Before: `2023-09-16T02:39:26Z`
After: `Sept 16th, 2023`

If I mistakenly made repeated code with getting the dates or something else needs to be changed I'd be glad to cooperate 👍


Miepee:
Don't have time to fully review for now, so just a few notes:
1. This is missing a changelog entry as per https://github.com/randovania/randovania/blob/main/CONTRIBUTING.md#adding-to-changelog
2. For GUI work, it's generally nice to have screenshots and/or videos accompanying it, showing how it looks now
3. You should fix the tests. The readme explains how to run them locally 
4. Quickly ran it locally, as a fellow european, i find `YEAR-MONTH-DAY` (or `DAY-MONTH-YEAR`) much more readable, personally than the `MONTH_AS_TEXT-DAY-YEAR
5. The amount of versions that we have, makes it a little unfit for a dropdown, I feel. Their purpose is usually for a smaller list of options, not for "the amount of options are twice as high as my monitor"-list of options. I don't think there was ever a good consensus on how the versions should be selectable, might be wrong tho.
6. Can't see that yet, but your PR should have a coverage of at least 84%

duncathan:
> This is missing a changelog entry as per https://github.com/randovania/randovania/blob/main/CONTRIBUTING.md#adding-to-changelog

this may be single funniest possible review comment possible

Wolfmyths:
I adapted pytest to work with the new code, changed the time format to be a European standard format, and added entries to changelog.md

>The amount of versions that we have, makes it a little unfit for a dropdown, I feel. Their purpose is usually for a smaller list of options, not for "the amount of options are twice as high as my monitor"-list of options. I don't think there was ever a good consensus on how the versions should be selectable, might be wrong tho.

We could try a search bar that displays results as you type?

Here are some pictures of how the changelog widget looks now:

image

Untitled

Miepee:
Huh, looks different on my system. Dropdown stretches over the whole monitor
grafik

randovania/interface_common/update_checker.py
```diff
@@ -84,6 +87,9 @@ def versions_to_display_for_releases(
             log = f"## {version.tag_name}\n\n{version.change_log}"
             all_change_logs[version.tag_name] = log
 
+            published_time = datetime.strptime(version.published_at, "%Y-%m-%dT%H:%M:%SZ")
```
henriquegemignani, line 90: As much as I'd love to disrespect the terrible US date format, use the locale-dependant `%c` here.
It will make the tests a fun-experience though.

henriquegemignani (changes requested):
Some things to change. Appreciate the contribution!

Wolfmyths:
> Huh, looks different on my system. Dropdown stretches over the whole monitor. Maybe you'll need to set an explicit height limit?

I added a limit to 10 items in the combo box

randovania/interface_common/update_checker.py
Wolfmyths: I added the `%c` format in the latest commit, it looks like this:

image

I'm unfamiliar with this specific format, so I'm not sure if there's a way to further simplify this without splicing. Unless this is the result you wanted.

henriquegemignani: Oh good catch. Forgot that one includes the time, which is very not relevant here.
Docs says %x is just the date.

Wolfmyths: Okay, I updated the format

image

CHANGELOG.md
```diff
@@ -8,9 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [6.3.0] - 2023-10-??
 
 - Added: During generation, if no alternatives have a non-zero weight, try weighting by how many additional Nodes are reachable.
+- Added: Changelog now shows the date an update was publishe
```
ThanatosGit, line 11: I prefer to leave the wording to native speakers but `publishe` looks like a typo to me.

ThanatosGit (changes requested):

CHANGELOG.md
dyceron: I would say something like `The changelog now shows the date for each release`

Wolfmyths: Changed the entry to dyceron's suggestion 👍 and yes that was a typo

Miepee:
> I added a limit to 10 items in the combo box

This does not seem to work.
grafik

Wolfmyths:
>This does not seem to work.

Strange.

I use the QComboBox method `.setMaxVisibleItems(10)` which works for me. I checked the documentation and there seems to be a note for this method.

The note reads:

>This property is ignored for non-editable comboboxes in styles that returns true for SH_ComboBox_Popup such as the Mac style or the Gtk+ Style.

https://doc.qt.io/qtforpython-6/PySide6/QtWidgets/QComboBox.html#PySide6.QtWidgets.PySide6.QtWidgets.QComboBox.maxVisibleItems

I'm using windows when running the program.

I found a possible solution here, but it requires adding to the project's stylesheet:
https://stackoverflow.com/questions/11252299/pyqt-qcombobox-setting-number-of-visible-items-in-dropdown

Miepee:
> This property is ignored for non-editable comboboxes in styles that returns true for SH_ComboBox_Popup such as the Mac style or the Gtk+ Style.

That's interesting, cause I'm using neither of these. Overwriting the QT_STYLE to `Fusion`, `Oxygen` and `Breeze` they all dont seem to respect the limit. For `Windows` it does seem to respect it.

> but it requires adding to the project's stylesheet:

From what I can see that isn't required. You should be able to just do:
```py
self.select_version = QtWidgets.QComboBox(self)
self.select_version.setMaxVisibleItems(10)
self.select_version.setStyleSheet("QComboBox { combobox-popup: 0; }"); // HACK: Done to respect max items limit on Linux/macOS
```
no?

Wolfmyths:
>I tried that out, it does indeed work

Okay cool, I'll add that line in then. I just wasn't sure if you guys wanted all of your stylesheet related stuff in one spot.

>altho it has some slight issues displaying the "scrollbar"

I think I might know the issue? Because of how `setStyleSheet()` works, instead of appending the style it replaces it. So we'd need to include the scrollbar I think... here's the documentation for QComboBox stylesheets (Although this page is helpful I feel like a lot of things are missing)

https://doc.qt.io/qtforpython-6/overviews/stylesheet-examples.html#customizing-qcombobox

I'm not entirely sure what's happening in that picture but styling scroll bars in Qt are very weird, if you change one part of the scroll bar you have to change the rest of the parts

CHANGELOG.md
```diff
@@ -8,9 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [6.3.0] - 2023-10-??
 
 - Added: During generation, if no alternatives have a non-zero weight, try weighting by how many additional Nodes are reachable.
+- Added: The changelog now shows the date of each release
 - Added: Data Visualizer now has a very visible checkbox to quickly toggle if the selected trick filters are enabled.
 - Added: When trick filters are enabled, a line is added indicating how many requirements are being filtered.
 - Changed: The generator will now consider placing Energy Tanks, if there's a damage requirement that's exactly high enough to kill the player.
+- Changed: Changelog now displays update versions in a drop-down box
+- Fixed: Changelog external links now work
```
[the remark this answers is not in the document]
Wolfmyths: Does this work?

"The Changelog window has recieved a slight overhaul. The date of each release is shown, hyperlinks are fixed, and patch notes are now accessed through a drop-down box (previously used vertical tabs)."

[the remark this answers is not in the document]
Miepee: Sounds good to me

ThanatosGit:
What's status here?
I see the macOS tests are failing? Looks like it omits the 20 in `2023`? lol

Miepee:
AFAIK still needs some thoughts on how to properly limit the length for unix systems without making scrollbar dissapear. Altho I personally would be fine with no limit / the stylesheet hack from above, as long as a `TODO` is put on there.

Wolfmyths:
> AFAIK still needs some thoughts on how to properly limit the length for unix systems without making scrollbar dissapear.

There's no way for me to debug this issue as I only have a windows machine, I did recommend a fix that might work by adding a scrollbar to the stylesheet in an earlier comment.

Try copy-pasting some sample stylesheets from this link
https://doc.qt.io/qtforpython-6/overviews/stylesheet-examples.html#customizing-qcombobox

As for whatever is going on in the macOS check failing I tried to fix it in this commit, so we'll see what happens.

Miepee:
> Try copy-pasting some sample stylesheets from this link

I toyed around with the examples from there, I couldn't manage to get any styled scrollbar in there, or anything that resembles one.
So just keep it like this then.

Miepee (commented):
Can you try to fix the mypy complains?

henriquegemignani:
I have an impression that whatever locale the macOS runners are configured with have that different date format. I think the best is to just fill the test data with generating the strings using the same locale-dependant function. The only other option is mocking that call.

Wolfmyths:
Okay so far looking good from my fork's workflow. What I did was generate the test data within the test function instead of the script.

randovania/gui/widgets/changelog_widget.py
```diff
@@ -22,9 +36,22 @@ def __init__(self, all_change_logs: dict[str, str]):
...
-            self.addTab(scroll_area, version_name)
+            self.changelog.addWidget(scroll_area)
+
+            self.select_version.addItem(version_name)
+
+        self.changelog.setCurrentIndex(0)
+
+    def select_version_index_changed(self) -> None:
+        # Used "Any" type hint to bypass mypy raising "expecting QWidget"
+        selected_widget: Any = self.findChild(
```
Miepee, line 53: Why wouldn't QWidget as a type hint work here if it's whats expected? Does findChild not return that?

test/gui/widgets/test_changelog_widget.py
```diff
@@ -1,9 +1,17 @@
...
+
 from randovania.gui.widgets.changelog_widget import ChangeLogWidget
 
 
 def test_create(skip_qtbot):
+    # mypy expects a QWidget (which they aren't) from these vars
+    qScroll0: Any  # QScrollArea
+    qScroll1: Any  # QScrollArea
+    qLabel0: Any  # QLabel
+    qLabel1: Any  # QLabel
```
Miepee, line 13: Also here, why are these not QWidgets? I would expect `currentWidget` to return one at least.

henriquegemignani: In both cases the best thing to do is just `typing.cast`. `findChild` returns QWidget, but if you need a QLabel method that doesn't work.

Wolfmyths: > Also here, why are these not QWidgets? I would expect currentWidget to return one at least.

It's possible that it can return a QWidget, but 9/10 that's not the case. In most cases it returns an object inheriting QWidget. Mypy raises an issue saying it has to be specifically QWidget

> In both cases the best thing to do is just typing.cast. findChild returns QWidget, but if you need a QLabel method that doesn't work.

Using `typing.cast` has worked, thanks for the tip

henriquegemignani (approved):

PR #5343, opened 2023-09-22
Codecov: allow coverage to drop by 0.5%

henriquegemignani:


ThanatosGit (commented):
Shouldn't this pr already use the values?
Also: Is 0.5% not too high? The standard floating 0.01 % is annoying but at 0.5%, you should maybe just add some tests.

Kinda funny that the only check failing is the codecov one for the project coverage.

henriquegemignani:
I don't know if this is in the correct place, if codecov uses the config from a branch, if the syntax is correct. It's a wild guess

I don't mind if coverage actually decreases by 0.5% in a pot with code. Means you sir a bunch of tests but barely got the goal.

ThanatosGit:
> I don't know if this is in the correct place, if codecov uses the config from a branch, if the syntax is correct. It's a wild guess

I think it needs to be in a `coverage` top block and not `codecov` ? See https://docs.codecov.com/docs/codecovyml-reference#coveragestatus

> I don't mind if coverage actually decreases by 0.5% in a pot with code. Means you did a bunch of tests but barely got the goal.

Maybe I just have wrong feeling about what 0.5 % means for Randovania. I mean, for very small projects it could be like 10 LOC, while for Randovania it could mean many many LOC e.g. that you removed all gui tests.

PR #5345, opened 2023-09-23
Add checkbox to quickly toggle if trick filters are enabled

henriquegemignani:
image
image


toasterparty (approved):
Idea: Write text somewhere that says how many connections are hidden due to active filters if not 0

henriquegemignani:
I'll make another PR for it

PR #5350, opened 2023-09-25
Dread: Add ledge warp out after Diffusion Beam

Tyranisaur:


randovania/games/dread/json_data/Cataris.txt
```diff
@@ -3153,6 +3153,11 @@ Extra - asset_id: collision_camera_044
       Any of the following:
           Varia Suit
           Heat/Cold Runs (Intermediate) and Heat Damage ≥ 50
+  > Tunnel to Teleport to Dairon
+      All of the following:
+          # https://youtu.be/Y3wgVEtidi0
+          Varia Suit and Ledge Warp (Intermediate) and Disabled Door Lock Randomizer
```
XenoWars, line 3159: does it really need to check for door rando disabled though? i feel like that's kinda redundant seeing as how it can only matter if it's disabled in the first place since you can otherwise go through the blob or door.

XenoWars (changes requested):

randovania/games/dread/json_data/Cataris.txt
XenoWars: Yeah, there's also that. So it doesn't really need to check for disabling door lock randomizer.

XenoWars (approved):

PR #5358, opened 2023-09-26
Change tracker to use Nodes instead of Areas

ThanatosGit:
The best description for this PR: 
"I tried to make it less broken."


Miepee:
Dread and Echoes got fixed, so it indeed is less broken :)

You also forgot a changelog entry btw

ThanatosGit:
> Dread and Echoes got fixed, so it indeed is less broken :)

I thought it was Dread and Prime? But Prime and Echoes works for me in 6.2.0.
Was it a specific preset setting? (Which games got fixed is good to know for a changelog entry)

Miepee:
you're correct, was dread and prime1. It happened already on starter preset, not sure whether there was some setting that would allow it *not* to crash.

ThanatosGit:
In 6.2.0 the April Fools and Moderate Challenge preset works. Starter does not.

henriquegemignani (approved):

PR #5370, opened 2023-09-28
CI: Cache more pip stuff

Miepee:
I have no clue if this works.
From the documentation:
> The action defaults to searching for **a dependency file** (`requirements.txt` or `pyproject.toml` for pip, `Pipfile.lock` for pipenv or `poetry.lock` for poetry) in the repository, and uses its hash as a part of the cache key. Input cache-dependency-path is used for cases when **multiple dependency files** are used, they are located in different subdirectories or different files for the hash that want to be used.

Emphasis mine. Since we don't provide the `cache-dependency-path` path, it seems that it would only cache `requirements.txt`.

Looking at further documentation shows us this:
```yaml
# Using a list of wildcard patterns to cache dependencies

steps:
- uses: actions/checkout@v3
- uses: actions/setup-python@v4
  with:
    python-version: '3.10'
    cache: 'pip'
    cache-dependency-path: |
      **/setup.cfg
      **/requirements*.txt
- run: pip install -e . -r subdirectory/requirements-dev.txt
```

So I essentially just copied that.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -75,6 +75,9 @@ jobs:
         with:
           python-version: "${{ matrix.python }}"
           cache: "pip"
+          cache-dependency-path: |
+            **/setup.cfg
+            **/requirements*.txt
 
       - name: Apt
         run: sudo apt-get -q update && sudo apt-get -y install xvfb xserver-xephyr tigervnc-standalone-server x11-utils gnumeric libxcb-cursor0 libxkbcommon-x11-0


henriquegemignani:
I actually only want to hash requirements.txt. I configured it so that the cache is only when installing that file, and not for requirements-lint.txt. requirements-setuptools.txt. Maybe actually that one?
We don't need to hash setup.cfg/pyproject.toml because we properly locks all dependencies.

Reading the docs, it seems [caches are already runner-image specific](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) so we could skip over adding the OS to the key. Not just that, it supports restoring the cache from different keys so we could restore the cache from something that _doesn't_ match the hash which is actually great!

Miepee:
My main thought for this was to speed up `Install Python packages (source)` as it takes a bit over a minute which seemed weird to me.
But looking further at CI logs, it indeed seems like everything is cached there already, which makes this PR rather useless.

PR #5374, opened 2023-09-29
AM2R: Make super Omegas/zetas not trickless

Miepee:
Not quite sure whether Omegas have the best approach
Both zetas and omegas are harder with supers only than guardian (which is combat beginner)

but omegas specifically can be killed quicker if you only do backshots. I currently bumped this to advanced.
Combat in general needs to be bumped up and redone completely, likely.

### DIFF
--- a/randovania/games/am2r/json_data/header.txt
+++ b/randovania/games/am2r/json_data/header.txt
@@ -218,7 +218,8 @@ Templates
       All of the following:
           Any of the following:
               # Item requirements
-              Missiles ≥ 30 or Super Missiles ≥ 6
+              Super Missiles ≥ 6
+              Missiles ≥ 30 and Combat (Intermediate)
               Charge Beam and Missile-Less Metroid Fights (Advanced)
           Any of the following:
               # Energy requirements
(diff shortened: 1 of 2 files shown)


JeffGainsNGames:
Agree combat will likely need another full pass. For now I like this change and agree that a "minimum ammo with only back shots" can get bumped as consistently getting those shots can be difficult.

randovania/games/am2r/json_data/header.txt
```diff
@@ -218,7 +218,8 @@ Templates
       All of the following:
           Any of the following:
               # Item requirements
-              Missiles ≥ 30 or Super Missiles ≥ 6
+              Missiles ≥ 30
+              Super Missiles ≥ 6 and Combat (Intermediate)
```
JeffGainsNGames, line 222: Shouldn't the combat be on the missiles?

Miepee: Yes, it should be.  Fixed.

JeffGainsNGames (approved):

PR #5376, opened 2023-09-29
SM: Update Ceres room names to be according to wiki

Miepee:
https://wiki.supermetroid.run/List_of_rooms#Ceres_Station

### DIFF
--- a/randovania/games/super_metroid/json_data/Ceres Station.json
+++ b/randovania/games/super_metroid/json_data/Ceres Station.json
@@ -4,15 +4,15 @@
         "asset_id": 5553
     },
     "areas": {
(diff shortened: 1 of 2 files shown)


Miepee:
Worth noting that VARIA has some internal names for these. not sure which ones are better, probably the ones from the wiki as those are likely to be more known/findable. These are the ones from VARIA:
```
[Ceres Elevator Room]
[Ceres Jump Tutorial Room]
[Ceres Staircase Room]
[Ceres Dead Scientists Room]
[Ceres Last Corridor]
[Ceres Ridley Room]
```

Logarhythmic (approved):

PR #5379, opened 2023-09-30
SM: rename brinstar rooms

Miepee:
more renames according to sm wiki


Logarhythmic (approved):
I like these room names better.

PR #5380, opened 2023-09-30
MD Logic - Ferenia Pitfall Puzzle Room missile tank with Flash Shift

Logarhythmic:
Added:  The missile tank in the Pitfall Puzzle Room can now be obtained with Flash Shift.

Included a video.


randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -886,7 +886,12 @@ Extra - asset_id: collision_camera_013
   * Layers: default
   * Morph Ball Tunnel to Space Jump Room/Tunnel to Pitfall Puzzle Room (Upper)
   > Pickup (Missile Tank)
-      Morph Ball and After Ferenia - Pitfall Puzzle Blob and Use Spin Boost
+      All of the following:
+          Morph Ball and After Ferenia - Pitfall Puzzle Blob
+          Any of the following:
+              Use Spin Boost
+              # With Flash Shift: https://youtu.be/XBcdDthTfdg
+              Flash Shift and Movement (Intermediate)
```
Tyranisaur, line 894: This is maybe fine, but for reference, look at the item at the top of Ghavoran Spin Boost Tower. Base on that, these should have "Climb Sloped Tunnels".

Tyranisaur (changes requested):

randovania/games/dread/json_data/Ferenia.txt
```diff
@@ -886,7 +886,12 @@ Extra - asset_id: collision_camera_013
   * Layers: default
   * Morph Ball Tunnel to Space Jump Room/Tunnel to Pitfall Puzzle Room (Upper)
   > Pickup (Missile Tank)
-      Morph Ball and After Ferenia - Pitfall Puzzle Blob and Use Spin Boost
+      All of the following:
+          Morph Ball and After Ferenia - Pitfall Puzzle Blob
+          Any of the following:
+              Use Spin Boost
+              # With Flash Shift: https://youtu.be/XBcdDthTfdg
```
XenoWars, line 893: With Diffusion, you need to be able to get up the slope somewhat in order to destroy the blob. I don't think the video should be accepted or even the trick (for diffusion) until you can show you can destroy the blob this way with diffusion. It would be fine if it hard requires wave.

XenoWars (changes requested):

randovania/games/dread/json_data/Ferenia.txt
Logarhythmic: Well, the video was for `After Pitfall Puzzle Blob`. I can add more later.

Logarhythmic: If you do that with diffusion or wave is irrelevant to this trick.

XenoWars: Yeah, this was done because you used a power bomb right? It looks like you could do it with Diffusion, but still should probably have a video for the whole thing just in case.

Logarhythmic: Do we really need a video to include how to pop this blob?

PR #5381, opened 2023-09-30
Restructure some test files

Miepee:
Currently restructures some test files to be a little less confusing in where they're placed / be more consistent with the rest.

Currently these two still fail:
```
FAILED test/interface_common/test_preset_manager.py::test_get_preset_at_version - KeyError: b'prime2'
FAILED test/interface_common/test_preset_manager.py::test_history_for_file - AssertionError: assert [] == [(datetime.da...cb6ff6cd211')]

``` 

Not sure what's up with that yet, I will look into that / ask about that later, wanted to just have it pushed for now. Was originally meant as "this annoys me, lemme quickly fix this in 5 minutes before I go get food", which was definitely the amount of time spend on this.


henriquegemignani:
I prefer the folder to be named after the class, not the extension.
Also I was thinking of having some structure where every single file in there also has a matching patcher file, maybe with a cosmetic_patches.json. Then there would be a test that does every single one at once.

Miepee:
I personally, when i go through failed tests, usually just look at the file extension for failed tests, but I can see why class-named folders makes more sense from an orga standpoint

henriquegemignani:
These files got moved already.

PR #5387, opened 2023-10-01
Echoes: Fix Main PB not counting towards PBs on UI

Miepee:
Before:
grafik

After:
grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed: When Progressive Grapple is enabled, it will now show `2 shuffled copies` rather than `Shuffled` for better consistency.
 - Changed: A proper error message is displayed when mono is not found, when exporting a game on macOS and Linux.
+- Fixed: The UI will now consider the Main Power Bomb, which is set to Shuffled, as a Power Bomb source.
 
 #### Logic Database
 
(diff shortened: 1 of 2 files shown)


henriquegemignani:
Does this really cause no differences in the patcher json generated for Echoes? Or are we just missing some test? Which I wouldn't dismiss as unlikely 😅

Miepee:
:shrug: 
I feel like we're definitely missing tests for more exotic options, but I'm pretty sure this one is fine

henriquegemignani:
I find it really weird that adding a progression entry for the pickup causes no differences. That's the kind of thing that _should_ change how a pickup is exported and should need at least one test being changed.

Miepee:
So I had a look at this.
I exported a game with PB set to "shuffled" before this patch, and let the factory generate a json. I also did the same for "seed_a".
Then I ran through both my game and "seed_a" after this patch and low and behold.... only "seed_a" changed it's output.
In the resources for the Main PB, it's now given 3 times for whatever reason.
```json
"resources": [
    {
        "index": 43,
        "amount": 2
    },
    {
        "index": 47,
        "amount": 1
    },
    {
        "index": 43,
        "amount": 1
    },
    {
        "index": 74,
        "amount": 57
    }
],
```
Note how index 43 is in there 2 times.

I got curious as to why my game didn't cause any change. Apparently it's because setting it to "shuffled" caused it in the RDVgame to not exist at all. So something's just really screwy there, and this approach to fixing it is probably wrong.

Miepee:
This approach to fixing it is wrong, thus cloasing
#5410 should be merged first before reinvestigating this

PR #5389, opened 2023-10-02
Samus Returns - Add pickups to all metroids

dyceron:
- Updated `patch_data_factory` to support Metroid pickups and patcher export
- All Metroids now have pickups (extra data might be wrong)
- Baby location now has extra data
- New template for having all DNA (not added into DB fully yet since goal/pool creator is not setup yet)

The reason I'm saying that the extra data might be wrong is because I don't have the new DNA items handled anywhere other than the header file. I assume that when I make `pool_creator` then it will be more relevant? The reason I did it this way and not just list the DNA item is because this will work better for hints? Is this necessary? The patcher checks for DNA amount to beat the game, so I don't know if I should just have the extra data be that, or keep what this PR does.


randovania/games/samus_returns/json_data/Area 1.txt
```diff
@@ -416,6 +416,15 @@ Extra - asset_id: collision_camera_017
           Lay Any Bomb
           Super Jump (Intermediate) or Unmorph Extend (Intermediate) or Climb Rooms Vertically (High Jump)
 
+> Pickup (DNA); Heals? False
+  * Layers: default
+  * Pickup 176; Category? Minor
+  * Extra - pickup_type: metroid
+  * Extra - item_id: ITEM_RANDO_DNA_5
```
ThanatosGit, line 423: What's the purpose of all the `item_id` in the extra section?

dyceron: I guess nothing since it doesn't do anything. My idea is to use these custom dna pickups so they can be hinted at, and then each one grant's `ITEM_ADN` to satisfy the patcher requirement.

Miepee: Hints would come from RDV which doesnt need these kinds of item IDs. you just give it the actual pickup instead that you want the hint for.

Miepee (approved):
Logic seems fine

PR #5395, opened 2023-10-03
MSR: Add Goal Tab+ Code for the DNA

ThanatosGit:
Just for the record: I have a couple of commits and I'm going to split them.
(don't worry too much about the rdvgame because there will be a few changes in each commit)

image

### DIFF
--- a/randovania/games/samus_returns/generator/bootstrap.py
+++ b/randovania/games/samus_returns/generator/bootstrap.py
@@ -18,8 +18,8 @@ if TYPE_CHECKING:
     from randovania.layout.base.base_configuration import BaseConfiguration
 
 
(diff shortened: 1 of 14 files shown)


dyceron (commented):
Does patch data factory need to updated to include the goal (see dread)?

ThanatosGit:
> Does patch data factory need to updated to include the goal (see dread)?

Can you be more specific? The file has like 600 LOC. You mean the `objective`? That's currently only for Hints. That would be a nice  for future, if we start integrating hints.

dyceron:
Yeah I meant that. If that's for hints then we're good.

dyceron (approved):

PR #5396, opened 2023-10-03
Add Energy tab for SR

ThanatosGit:
Next one :)
image


randovania/games/samus_returns/layout/msr_configuration.py
```diff
@@ -17,6 +17,7 @@ class MSRArtifactConfig(BitPackDataclass, JsonDataclass):
 @dataclasses.dataclass(frozen=True)
 class MSRConfiguration(BaseConfiguration):
     energy_per_tank: int = dataclasses.field(metadata={"min": 1, "max": 1000, "precision": 1})
+    starting_energy: int = dataclasses.field(metadata={"min": 99, "max": 1000, "precision": 1})
```
dyceron, line 20: We can't start with less than 99 energy?

ThanatosGit: Well, you can but I thought about be even more restrictive as the game UI is bad for it.
E.g. this is 130 energy...it won't show the purple bar until you have x99 energy.
image

ThanatosGit:
> Hmm, why have both seperate here, instead of doing it like the other games and have starting energy = eTank-1?

Honestly, I was asking myself the exact opposite question: Why is the starting energy tied to the energ tank value? It's totally valid for a game to start with e.g. 150 energy and give you 50 for each health upgrade.

But regarding the in-game ui, I seriously thinking about nuking it and hard code etank = 100 and starting = 99.

dyceron:
> But regarding the in-game ui, I seriously thinking about nuking it and hard code etank = 100 and starting = 99.

Is the ui also weird here?

ThanatosGit:
These are the vanilla values, why it should be weird in that case? So, no?

dyceron:
Because you said you want to hardcode this

Miepee:
> It's totally valid for a game to start with e.g. 150 energy and give you 50 for each health upgrade.

It is yes, see my plans here https://github.com/randovania/randovania/pull/5154#issuecomment-1708328331
Having every game be consistent would make it easier IMO for me to refactor later.

ThanatosGit:
What's the conclusion? Nuke the tab / option in randovania UI because other values than vanilla behaves weird in the in-game ui or any other suggestions?

dyceron:
The only thing weird about starting with more than 99 is the 100th digit isn't shown. I'd hate to nuke an option just because of that.

ThanatosGit:
It has nothing to do with starting energy. Like I said, any max life value between x00 and x98 will not show / add a additional purple tank bar icon.
I won't mess around for that in the patcher. I never even used the option for dread to modify the energy per tank value, so I'm totally fine to nuke the tab.
Or only define energy per tank and only allow multiplies of 100.

dyceron:
So of you have an etank of 80 it won't add a bar even though you have 179?

dyceron (approved):
This can be a later project to attempt to make these customizable.

PR #5401, opened 2023-10-03
Dread: Elun Start and Cataris North refining

XenoWars:
- Added: Elun's Save Station is now a valid starting room.
- Changed: The Energy Tank in Thermal Device Room North now requires Highly Dangerous Logic from above.

I didn't actually do anything to header.json. It seems to have done that when i saved my changes.


randovania/games/dread/json_data/Cataris.txt
```diff
@@ -2906,9 +2906,9 @@ Extra - asset_id: collision_camera_042
           Any of the following:
               Varia Suit or After Cataris - Lower Lava Button
               Heat/Cold Runs (Beginner) and Heat Damage ≥ 200
-          After Cataris - Final Thermal Device or Enabled Highly Dangerous Logic
+          After Cataris - Magnet Wall Thermal Device or Enabled Highly Dangerous Logic
```
Tyranisaur, line 2909: I think one thing is to require that the item should be obtainable by lowering the magnet wall, but going to dairon this way might be considered separate. If the Obsydomithon blob has been destroyed, then it should be safe the take the path to dairon. Whether the thermal device next to the maget wall has been activated probably doesn't have much to do with whether it's safe to activate the final thermal device.

Tyranisaur (changes requested):

randovania/games/dread/json_data/Cataris.txt
XenoWars: Thinking about it, it would still be possible to lock yourself out of the item if you do the final thermal device coming out of the heated route before getting it from below, and that item locks cloak. Requiring the Magnet Wall lowered won't change this. This change isn't actually solving anything it would seem since that was what it was trying to solve in the first place too.  The only way i can think of to truly ensure these don't get locked out is to either also require Highly Dangerous Logic to go up or to just require it to go to dairon, both of which lock out vanilla-required routing. The first at least would still let you leave through the lower door.

Miepee:
> It seems to have done that when i saved my changes.

For the future,  normal/intended behaviour.

XenoWars:
Probably gonna close this and then just pr the Elun start, since this isn't really preventing the possibility of locking out those items in Cataris like I had intended.

PR #5416, opened 2023-10-05
Workaround for images not working in built-in changelog widget

### LINKED ISSUE #4621: Built in Change Log improvements
- [ ] Images aren't loaded (see v0.9.0)

- [X] Make clicking links open them in a browser (see v5.8.0)
- [X] Add tabselection to the top, as reading sideways is uncomfortable
- [X] Add release date to releases

Wolfmyths:
This PR fixes #4621

### Additions:
+ Created a function to find image tags in the markdown of changelogs and then create new labels where the images should be in the markdown.
    - This function only runs if there are image tags in a changelog.
    - In the case of an http error from getting the image, the error will display where the image would be in the GUI
+ Created a function to reuse the code for setters in DelayedTextLabels

### Notes:
+ A DelayedTextLabel is created for every block of markdown before and after an image, images are DelayedTextLabels too.
    - This is to make patch notes look as identical as possible to the original release markdown
+ Uses the requests module to obtain images
+ Imported regex to easily parse the markdown

### How I tested it:
+ Created and ran pytests for the new function in `test_changelog_widget.py`
+ Checked changelog versions 0.9.0 and 0.10.0 and they display the same as if they were on github

### Concerns:
+ Unsure what library you guys want to use for getting images (PySide6's networking, requests, urllib)
+ Exception handling for getting images from the internet
+ Code readability
+ All images are loaded during changelog's `__init__` which could cost performance in the future
+ Using requests module freezes the program especially if the user has a slow connection
+ Changelog widget does not resize automatically when there are images that are too big to show in the widget

Capture12
Capture11


Miepee:
Sorry for the late response. Don't have the time to review yet, but can you please make sure that the Executable tests don't fail?

randovania/gui/widgets/changelog_widget.py
```diff
@@ -55,3 +70,53 @@ def select_version_index_changed(self) -> None:
...
 
         self.changelog.setCurrentWidget(selected_widget)
+
+    def setup_delayed_text_label(self, label: DelayedTextLabel) -> None:
+        label.setAlignment(QtCore.Qt.AlignmentFlag.AlignTop)
+        label.setOpenExternalLinks(True)
+        label.setTextFormat(QtCore.Qt.TextFormat.MarkdownText)
+        label.setWordWrap(True)
+
+    def image_parse(self, version_text: str) -> list[DelayedTextLabel] | None:
```
Miepee, line 80: You also probably want to fix this mypy error

henriquegemignani (changes requested):
The test failures is because `requests` is only installed for the server. Client code we use `aiohttp` to have non-blocking requests.
Which is another issue in this PR: it downloads the image synchronously during widget creation, which can be quite slow!

henriquegemignani:
Which I did not see you mentioned in the PR description, ooops!

The rest seems fine for me. But adding some placeholder text while the image downloads seems best for me. The placeholder would also work to indicate if the download fails.

Wolfmyths:
Sorry for the late response, I was able to implement aiohttp and got it to work.

The only issue I have right now that I am stumped on is pytest. I'm probably missing something simple as I'm fairly new to async and pytest.

When test_changelog.py is tested, errors come up about the async function `setup_labels()` saying this:
```
asyncio:base_events.py:1771 Task was destroyed but it is pending!
```
this error is the catalyst to an assert error because the task was never finished. Other test modules also complain about the async function.

> The rest seems fine for me. But adding some placeholder text while the image downloads seems best for me. The placeholder would also work to indicate if the download fails.

I'll add that when I can

ThanatosGit:
Do you still work on this or need help?

Wolfmyths:
> Do you still work on this or need help?

I'll take a look tomorrow and see if I can fix it

Wolfmyths:
Yeah I still need help, my knowledge of pytest is very basic especially with async stuff.

My code works when I run it, for some reason pytest is giving me a hard time.

The test is setup to create the changelog widget like normal, and it will wait for the signal that is emitted when the changelog is done fetching and placing the data.

However even though this waitSignal function is passed, an assertion error is thrown on the first assert `0 == 4`

I am at a loss for what the problem is

ThanatosGit:
Uh, looks kinda weird to me, too.
I'm not 100 % sure how `waitSignal` works.
`If your program has long running computations running in other threads or processes, you can use qtbot.waitSignal to block a test until a signal is emitted`
So, maybe it is blocking the qt event loop? That would make sense why the test code doesn't work. Also I think the within the `with` block you would write code to call something which will eventuall emit the signal from `waitSignal`.

I quickly hacked something together and pushed it: Basic idea is to mock the `startFetchingData` but assert that it is called and then call the `setup_lables` manually (and get rid of the leaked not awaited function call in `test_main_window`.

ThanatosGit (changes requested):
Needs an entry in the `CHANGELOG.md` and then we are good to go, I guess.

randovania/gui/widgets/changelog_widget.py
```diff
@@ -1,16 +1,24 @@
...
-from PySide6 import QtCore, QtWidgets
+import aiohttp
+from PySide6 import QtCore, QtGui, QtWidgets
+from qasync import asyncSlot
 
 from randovania.gui.widgets.delayed_text_label import DelayedTextLabel
 
 
 class ChangeLogWidget(QtWidgets.QWidget):
+    doneFetchingData = QtCore.Signal()
```
ThanatosGit, line 14: snake _case

randovania/gui/widgets/changelog_widget.py
```diff
@@ -1,16 +1,24 @@
...
+    startFetchingData = QtCore.Signal()
```
ThanatosGit, line 15: snake_case

randovania/gui/widgets/changelog_widget.py
```diff
@@ -28,30 +36,103 @@ def __init__(self, all_change_logs: dict[str, str]) -> None:
...
+        self.changelog.setCurrentWidget(selected_widget)
+
+    def setup_delayed_text_label(self, label: DelayedTextLabel) -> None:
+        label.setAlignment(QtCore.Qt.AlignmentFlag.AlignTop)
+        label.setOpenExternalLinks(True)
+        label.setTextFormat(QtCore.Qt.TextFormat.MarkdownText)
+        label.setWordWrap(True)
+
+    @asyncSlot()
+    async def setup_lables(self) -> None:
```
ThanatosGit, line 59: It's `labels` and not `lables`, isn't it?

test/gui/test_main_window.py
```diff
@@ -197,6 +198,7 @@ def test_on_menu_action_help(default_main_window, monkeypatch):
 def test_on_menu_action_changelog(default_main_window, monkeypatch, has_changelog):
     mock_show = MagicMock()
     monkeypatch.setattr(QtWidgets.QWidget, "show", mock_show)
+    monkeypatch.setattr(ChangeLogWidget, "setup_lables", MagicMock())
```
ThanatosGit, line 201: Same function, same typo, just that this one won't get catched with (at least my) automatic refactoring.

test/gui/widgets/test_changelog_widget.py
```diff
@@ -1,33 +1,74 @@
...
+
+async def test_create(skip_qtbot: QtBot, mocker):
     # Setup
-    widget = ChangeLogWidget(
-        {
-            "1.0": "Foo",
-            "2.0": "Bar",
-        }
-    )
+    mocked_fetch = mocker.patch("randovania.gui.widgets.changelog_widget.ChangeLogWidget.startFetchingData")
```
ThanatosGit, line 31: snake_case for thr patched field.
Guess this also won't be catched by refactoring.

test/gui/widgets/test_changelog_widget.py
```diff
@@ -1,33 +1,74 @@
...
+        assert widget.select_version.count() == EXPECTED_VERSIONS_COUNT
+
+    for i in range(1, EXPECTED_VERSIONS_COUNT + 1):
+        assert widget.select_version.itemText(i - 1) == CUSTOM_LOGS_KEYS[i - 1]
 
     widget.select_version.setCurrentIndex(0)
     qScroll0 = cast(QtWidgets.QScrollArea, widget.changelog.currentWidget())
-    qLabel0 = cast(QtWidgets.QLabel, qScroll0.widget())
+    qFrame0 = cast(QtWidgets.QFrame, qScroll0.widget())
+    qLabel0 = cast(DelayedTextLabel, qFrame0.findChild(DelayedTextLabel))
```
ThanatosGit, line 47: snake_case for all the `qBlaX` variables.
`qScroll0` was already there before but would be nice to fix that

ThanatosGit (changes requested):
Oops, I'm very sorry, I was so heavily focused on the logic that I didn't notice the naming.
I must admit that's because Qt is using camelCase and therefore it is heavily mixed in the gui related parts anyways. 
But at least we want to be consistent with our own variables, fields etc.

ThanatosGit (approved):

ThanatosGit:
@henriquegemignani You wanna re-review it or should I just dismiss?

randovania/gui/widgets/changelog_widget.py
```diff
@@ -28,30 +36,103 @@ def __init__(self, all_change_logs: dict[str, str]) -> None:
...
+
+    def setup_delayed_text_label(self, label: DelayedTextLabel) -> None:
+        label.setAlignment(QtCore.Qt.AlignmentFlag.AlignTop)
+        label.setOpenExternalLinks(True)
+        label.setTextFormat(QtCore.Qt.TextFormat.MarkdownText)
+        label.setWordWrap(True)
+
+    @asyncSlot()
+    async def setup_labels(self) -> None:
+        for version_name, version_text in self.all_change_logs.items():
```
henriquegemignani, line 60: Downloading every single image is not great, but we can change that later

henriquegemignani (approved):

Miepee:
@Wolfmyths sorry, but can you rebase and put the changelog entry into 7.3.1?

ThanatosGit:
Just did it quickly. It's faster for me to click the "Update branch" button and quickly change the changelog via the github online editor than to ask, watch this pr later again + write this message here 😛

PR #5420, opened 2023-10-06
Dread: Change Charge Beam door weakness to use proper template

Miepee:
Not quite sure how to handle changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Dread
 
-- Fixed: The "Power Bomb Limitations" setting is now accounted for by logic.
 - Fixed: Custom shields now use the correct shader and texture effects and no longer a black background
 - Fixed: Issues with negative amount for ammo items. The current amount was set to a wrong value and you had to use a ammo refill station. This also caused issues with the auto tracker and multiworld.
 
 #### Logic Database
+
+- Fixed: The "Power Bomb Limitations" setting is now respected for opening Charge Beam Doors.
 
 ##### Artaria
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Dread
 
-- Fixed: The "Power Bomb Limitations" setting is now accounted for by logic.
+- Fixed: The "Power Bomb Limitations" setting is now accounted for by logic. For real this time.
```
Tyranisaur, line 21: Might as well just immediately isolate this in 6.5 and move all the other changelog entries to 6.6

Tyranisaur (changes requested):

CHANGELOG.md
Tyranisaur: Ok, so with Dakrszero's comment. I think it's reasonable to move this entry into the logic database section and rephrase it as "Fixed: The "Power Bomb Limitations" setting is now respected for opening Charge Beam Doors.".

Tyranisaur (changes requested):

ThanatosGit:
Can you address the requested changes? We should maybe give this to beta testers. Half of the month already passed again oO

CHANGELOG.md
Tyranisaur: I think we should also put in a bit about enabling pb limitations again after it was disabled in 6.4.1, but it seems kind of odd to do that before stable gets merged back into main, because I don't know how the entire changelog will read.

Miepee:
Sorry for the delay, I think I addressed them now. Not sure whether I should mention that PB limitations are enabled again, considering that it's currently a stable only thing/never was in dev and no idea how it'll look after merging.

Tyranisaur (approved):

PR #5425, opened 2023-10-07
AM2R: Update starter preset with correct super count and a3 blocks enabled

JeffGainsNGames:
- Super Launcher now correctly gives 2 instead of 5
- A3 blocks are enabled

- Include DoorLockRando in bootstrap


Miepee:
Fix tests please

Miepee (approved):

PR #5434, opened 2023-10-10
Fix map tracker with a single non-default starting location

ThanatosGit:
Missing changelog and test.
Did this because of bug report in Discord.

I'm very unsure about the `raise ValueError("Preset without a starting location.")`. Do we have a better way to deal with it without spamming sentry in this case?
Should it actually be completly ignored and be fixed in the future by simply forbid to save a preset without a starting location?


Miepee (commented):
I assume draft because you still want to add tests?

ThanatosGit:
> I assume draft because you still want to add tests?

Test, changelog and `if node_location is None:` should be the first condition instead of checking for it twice

Miepee (approved):

PR #5436, opened 2023-10-10
Sync export with MSR patcher

ThanatosGit:
Little bit wild:
Patcher does everything from Lua but has parent classes and assumes that the first item is the important main item.
Non required mains works in randovania by just adding the main item to each ammo/tank item.
=> Had to prepend the main item to the ammo in that case.

I introduced an `item_export_id` because we need different items in starting inventory and in the pickups. E.g. we need to use `ITEM_WEAPON_POWER_BOMB_MAX` in starting inventory but `ITEM_POWER_BOMB_TANK` in pickups.
The game has implicit main item, which means as soon as you have a `ITEM_WEAPON_POWER_BOMB_MAX` you have the main item => patcher locks them by using  `ITEM_POWER_BOMB_TANK` in pickups.

Please never add an option to start with ammo without the main item :)

### DIFF
--- a/randovania/games/samus_returns/exporter/patch_data_factory.py
+++ b/randovania/games/samus_returns/exporter/patch_data_factory.py
@@ -46,13 +46,20 @@ def get_item_id_for_item(item: ResourceInfo) -> str:
         raise KeyError(f"{item.long_name} has no item ID.") from e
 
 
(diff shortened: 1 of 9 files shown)


Miepee:
> Please never add an option to start with ammo without the main item :)

>:(

dyceron (approved):

PR #5437, opened 2023-10-10
Only get the dock node default connection when force enabling two-way.

henriquegemignani:


ThanatosGit (approved):
This makes only a difference when there is a DockNode where the default connection is pointing to a non DockNode.
I could have sworn that there is an integrity check which ensures it.

PR #5438, opened 2023-10-10
Allow ammo to be standalone in their category

Miepee:
AM2R "needs" (as in, I dont want to provide worse UX) ammo currently in their own category, so here's this little hack.

Hopefully very very temporary, as dunc is working on refactoring hint category from the UI category. Dunno how far away that is tho, and wanted this in the the meantime

This PR fixes the crash that would happen when having a UI category with only ammo:
grafik

### DIFF
--- a/randovania/gui/preset_settings/item_pool_tab.py
+++ b/randovania/gui/preset_settings/item_pool_tab.py
@@ -233,22 +233,17 @@ class PresetItemPool(PresetTab, Ui_PresetItemPool):
         all_categories = list(pickup_database.pickup_categories.values())
 
         categories = set()
+        standard_broad_categories = set()
         for standard_pickup in pickup_database.standard_pickups.values():
             if not standard_pickup.hide_from_gui:
                 categories.add(standard_pickup.pickup_category)
+                standard_broad_categories.add(standard_pickup.broad_category.name)
 
-        # FIXME: this really shouldn't be done this way. This is just temporary to have ammo in their own category.
-        broad_to_category = {
-            "beam_related": "beam",
-            "morph_ball_related": "morph_ball",
-            "missile_related": "missile",
-        }
         for ammo_pickup in pickup_database.ammo_pickups.values():
-            if ammo_pickup.broad_category.name in broad_to_category:
-                proper_name = broad_to_category[ammo_pickup.broad_category.name]
-                categories.add(next(filter(lambda p: p.name == proper_name, all_categories)))
-            else:
-                categories.add(ammo_pickup.broad_category)
+            # When a proper distinction between UI and hint categories exists, this needs to be changed
+            if ammo_pickup.broad_category.name in standard_broad_categories:
+                continue
+            categories.add(ammo_pickup.broad_category)
 
         for standard_pickup_category in sorted(categories, key=lambda it: all_categories.index(it)):
             category_box = Foldable(None, standard_pickup_category.long_name)
--- a/randovania/gui/preset_settings/item_pool_tab.py
+++ b/randovania/gui/preset_settings/item_pool_tab.py
@@ -230,13 +230,26 @@ class PresetItemPool(PresetTab, Ui_PresetItemPool):
 
     def _create_categories_boxes(self, pickup_database: PickupDatabase, size_policy):
         self._boxes_for_category = {}
+        all_categories = list(pickup_database.pickup_categories.values())
 
         categories = set()
         for standard_pickup in pickup_database.standard_pickups.values():
             if not standard_pickup.hide_from_gui:
                 categories.add(standard_pickup.pickup_category)
 
-        all_categories = list(pickup_database.pickup_categories.values())
+        # FIXME: this really shouldn't be done this way. This is just temporary to have ammo in their own category.
+        broad_to_category = {
+            "beam_related": "beam",
+            "morph_ball_related": "morph_ball",
+            "missile_related": "missile",
+        }
+        for ammo_pickup in pickup_database.ammo_pickups.values():
+            if ammo_pickup.broad_category.name in broad_to_category:
+                proper_name = broad_to_category[ammo_pickup.broad_category.name]
+                categories.add(next(filter(lambda p: p.name == proper_name, all_categories)))
+            else:
+                categories.add(ammo_pickup.broad_category)
+
         for standard_pickup_category in sorted(categories, key=lambda it: all_categories.index(it)):
             category_box = Foldable(None, standard_pickup_category.long_name)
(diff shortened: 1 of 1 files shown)


ThanatosGit (commented):
I like changes in parts I never looked by myself :)

Would it be less hacky to do this in `_create_ammo_pickup_boxes`?
With something like check if it is not already present with
```python
category = broad_to_category.get(ammo.broad_category.name, ammo.broad_category.name)
if self._boxes_for_category.get(category, None) is None:
   ....
```
and then create the `_boxes_for_category` thing (by refactor it out to an own method)?

Miepee:
I had the same thought, but ultimately I decided against it.
Currently, it's very clear what both functions do: `_create_categories_boxes` is supposed to fully create all the GUI categories, while `_create_ammo_pickup_boxes` is supposed to *only* add ammo GUI boxes to existing categories. If the latter function now suddenly also takes the responsibility of creating more gui categories if needed, then the functionality kinda gets intermixed and would make future refactoring slightly harder. I also feel like that approach would scale even worse, if i.e.  we'd have some other pickup type that's alone in its category, because then its function would also need to create gui categories. 
So I just think that it'd be even messier, than what I currently do.

Miepee:
> _create_categories_boxes creates category boxes but only for main items, which means to me that it already has two purpose

Not really. It creates all needed category boxes. It's sort of a "prepare" function in a way. 

> I think it would be better to just split it.

I really dislike splitting here. I don't want to start refactoring too much here for now, as with this PR I only wanted to get ammo stuff working for AM2R.
Currently I think the main argument for splitting is to get around the second pasted broad_to_category map. Which, while I can see the appeal, is soon completely redundant anyway as dunc is working on a more full refactor.
I also TBH prefer one "prepare" function, rather than each function having to check if the category it's trying to use exists, and if not add it. It just means that I have to need a little less brainpower when reading through the functions, as I can assume that everything they need already exists, rather than follow which structures they all still need to create for them to properly work.
I do share the sentiment that having the whole `category_box = Foldable(None, standard_pickup_category.long_name)` stuff seperated into another function would be neat. But I just cannot see how doing that alone would be useful right now. If each function that needs the boxes would check whether the categories it needs exists, and if not call the function, then we run into what I described above. If we split `_create_categories_boxes` so it collects all categories first, and just does a `foreach category in category_list: _create_box(category)` (which I already think would be outside of PR scope), it also isn't really any cleaner in any way IMO.

Dunno. Not really a fan of any option, but the currently implemented one just feels like the least bad one.

ThanatosGit:
I still dislike duplicating a game specific hack into a common code part. I actually want more of this removed instead of adding more.

I normally play the "I don't want to do further refactors" card, if the PR already touched like 50ish files and need to change another 10 files without being able to automatically do the refactor or if the code is a mess (greetings from the map tracker).

randovania/gui/preset_settings/item_pool_tab.py
```diff
@@ -230,13 +230,26 @@ def _on_update_maximum_starting(self, value: int):
...
+        # FIXME: this really shouldn't be done this way. This is just temporary to have ammo in their own category.
+        broad_to_category = {
+            "beam_related": "beam",
+            "morph_ball_related": "morph_ball",
+            "missile_related": "missile",
+        }
+        for ammo_pickup in pickup_database.ammo_pickups.values():
+            if ammo_pickup.broad_category.name in broad_to_category:
+                proper_name = broad_to_category[ammo_pickup.broad_category.name]
+                categories.add(next(filter(lambda p: p.name == proper_name, all_categories)))
```
ThanatosGit, line 249: Is this part really needed? It's only executed for one of `beam_related`, `morph_ball_related` or `missile_related`. So, you want to add one of those for AM2R without a main item in any of the mentioned categories?
Otherwise this only adds the only existing one.

It looks more like you want to just skip for `beam_related`, `morph_ball_related` or `missile_related`

randovania/gui/preset_settings/item_pool_tab.py
```diff
@@ -230,13 +230,26 @@ def _on_update_maximum_starting(self, value: int):
...
+            else:
+                categories.add(ammo_pickup.broad_category)
```
ThanatosGit, line 251: Is it possible to generalise it? Searching for broad_categories which are not referenced by any main item / `standard_pickups`?

randovania/gui/preset_settings/item_pool_tab.py
Miepee: > So, you want to add one of those for AM2R without a main item in any of the mentioned categories?

Not quite, this is more needed as a hack to make the other games not regress. Otherwise on echoes for example, you suddenly get new empty "beam related" or "morph ball related" categories.

randovania/gui/preset_settings/item_pool_tab.py
Miepee: Not that I know of. The issue currently is that `broad_category` *is* nowhere referenced for most ammo. I.e. for missiles, all the main items refer to it as `missile`.

randovania/gui/preset_settings/item_pool_tab.py
ThanatosGit: You translate `beam related` to `beam` and then do `categories.add(...)`. `categories` is a set and it is just adding `beam` again, which is doing nothing in a set. It was already added by code which is above (and not shown) in the snippet. That's why I said, that I how I understand the code, you just want to skip it in a sense of `continue` when the condition is true instead of adding stuff.

Miepee: Ah, completely missed that it's a set.
For am2r, we can indeed just make it a continue. There's no case yet known to me where we need that stuff, and if some future case arrives, then dunc's rewrite hopefully happened by then.

randovania/gui/preset_settings/item_pool_tab.py
ThanatosGit: I won't check it in its entirety on mobile but https://github.com/randovania/randovania/blob/cb87d71c37d1f789be933381fae97baaf162f3ff/randovania/games/am2r/pickup_database/pickup-database.json#L85-L99
at least here is definitely the broad_category present. If it's true in general, you could generate list of all broad categories from the ammo pickups and subtract all broad categories you found in the standard pickups.
Could maybe combine it into the already existing iteration, so it is not that expensive and it would remove the game specific junk.

Miepee:
Changed it according to your suggestion. It indeed does not feel like a hack now anymore :)

ThanatosGit (approved):
Looks good to me :)

PR #5441, opened 2023-10-13
Download one specific nintendont version directly

henriquegemignani:
Consistently gets the same nintendont for a given commit, and probably fails less with rate limiting.


Miepee:
How often is the nintendont fork / upstream nintendont updated?

duncathan:
the fork? basically never. upstream? not sure

Miepee (approved):

henriquegemignani:
I don't like commiting executables to the repository

PR #5443, opened 2023-10-13
PyInstaller 6.1

henriquegemignani:


henriquegemignani:
Some people are going to enjoy this
image

randovania/gui/main_window.py
```diff
@@ -764,10 +764,21 @@ async def _on_menu_action_verify_installation(self):
...
             errors = []
             if bad_files:
-                errors.append(f"- {len(bad_files)} files are missing or are incorrect")
+                errors.append(f"- {len(bad_files)} files are incorrect")
+
+            if missing_files:
+                errors.append(f"- {len(missing_files)} files are missing")
+
             if extra_files:
                 errors.append(f"- {len(extra_files)} files are unexpected")
```
Miepee, line 773: maybe its just me, but can you change these to be `* {len(extra_files=}` instead of using a hyphen (or if very fancy, use some unicode bullet character)? the first time i read `- 15 files are missing or are incorrect` i read it as `minus 15 files are missing or incorrect`.

Miepee (commented):
Am I correct in the assumption that you tested stuff that tries to reference the data folder (i.e. nintendont)?

henriquegemignani:
> Am I correct in the assumption that you tested stuff that tries to reference the data folder (i.e. nintendont)?

Turns out they did a thing and the `_MEIPASS` thing now points to the new folder. The verify installation works just fine (though it doesn't find `randovania.exe` anymore) and exporting Echoes worked I think?

CS isn't working because the `lupa` hook isn't anymore, for reasons....

randovania/gui/main_window.py
henriquegemignani: image
it doesn't look very confusing for me 🤔 
but I can use `*` or something else

duncathan:
> CS isn't working because the lupa hook isn't anymore, for reasons....

is that my problem to solve or

henriquegemignani:
> > CS isn't working because the lupa hook isn't anymore, for reasons....
> 
> is that my problem to solve or

something pyinstaller. It's not including some pyd files that lupa ships

duncathan:
> Some people are going to enjoy this image

yeah literally every person i sent a build to for testing CS multi went "oh god which file do i want" lmao

randovania/gui/main_window.py
Miepee: Its more prominent when its only one list entry
grafik

henriquegemignani: I made this: https://github.com/randovania/randovania/issues/5520

tools/create_release.py
```diff
@@ -170,10 +178,12 @@ async def main():
 
     if platform.system() == "Windows":
         create_windows_zip(package_folder)
+        write_frozen_file_list(package_folder)
     elif platform.system() == "Darwin":
         create_macos_zip(app_folder)
```
Miepee, line 183: Any reason why this cant be done for macos?

henriquegemignani: Funnily enough the feature itself is not enabled for linux, though I think it works fine there?

I have no clue how adding files to the .app works and I didn't want to look into it

Miepee: can this get some comment added then? going ahead and putting the function would also be an alternative, but i fear itd cause some whacky breakage somewhere

Miepee: actually, issue would work too

Miepee:
AM2R, Cave Story and Echoes are spewing out *quite* a few warnings. Which boil down to
`ldd: Warning: You don't have any executable permissions for <insert file here>` and `the program is not linked dynamically`
In addition, dotnet. QT and cpython are spewing out a few other warnings. Is there a way to test run CI with this to see if it needs to install additional packages? Cause it complains for me with `could not resolve 'liblttng-ust.so.0'/'libQt6VirtualKeyboard.so.6'/'libQt6Pdf.so.6'/'libmpdec.so.3'`

Miepee (approved):
lets see how this goes

PR #5468, opened 2023-10-17
MultiworldClient: Safeguard against errors when encoding inventory

henriquegemignani:


ThanatosGit (approved):
Hopefully it doesn't get triggered with new odr at all

henriquegemignani:
> Hopefully it doesn't get triggered with new odr at all

Hopefully not, but also if something else breaks or new games break all we lose is remote trackers

PR #5487, opened 2023-10-19
Allow manually running "Python" workflow

Miepee:
Helps when troubleshooting certain things / a build failed for a random reason / other things where we want to manually (re)run one


henriquegemignani:
You can always just re-run an existing the one that ran for a commit, but also why not

PR #5488, opened 2023-10-19
Add dotnet to portable installations

Miepee:
This PR adds dotnet to portable releases. This was mostly done to stop AM2R from crashing on exporting, but will be useful for planets later down the line as well.

Shale tested this on Windows where it worked, I tested this on Linux where it also worked. So far, no one has tested this on macOS yet. I also need to test whether this works in flatpak later.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -188,6 +188,11 @@ jobs:
         run: |
           mv dist/randovania-* dist/randovania
 
+      - name: Remove dotnet from package
+        shell: bash
+        run: |
+          rm -r dist/randovania/data/dotnet_runtime
+
       - name: import windows certificate
         env:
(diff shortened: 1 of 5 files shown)


henriquegemignani:
How many more MBs it is? I'm more fine with the stable releases being bigger, but dev builds are a bit annoying since we have new versions quite often.

Miepee:
With the compression we have, surprisingly not thaaat much.
|  | Current Main | Results from branch | Diff |
|---|---|---|---|
| Linux | 163 MB | 191 MB | 28 MB |
| Windows | 85.8 MB | 105 MB | 19.2 MB |
| macOS | 117 MB | 144 | 27 MB |

henriquegemignani:
Getting over 100MB for the windows builds, might be worth looking into how to make a slim version or something.

henriquegemignani (approved):

PR #5511, opened 2023-10-21
MSR - Add additional DNA placement goals

dyceron:
Allows for placing DNA on the stronger Metroid variants separately from all Metroids, as well as on Bosses. I could have combined bosses and stronger Metroids, but this allows for more customization.

image


randovania/games/samus_returns/generator/bootstrap.py
```diff
@@ -39,7 +50,6 @@ def _get_enabled_misc_resources(
         self, configuration: BaseConfiguration, resource_database: ResourceDatabase
     ) -> set[str]:
         enabled_resources = set()
-        assert isinstance(configuration, MSRConfiguration)
```
ThanatosGit, line 42: Why do you have to remove the assertion? That kinda scares me :D

randovania/games/samus_returns/generator/bootstrap.py
```diff
@@ -14,21 +15,31 @@
...
+                elif config.prefer_metroids and pickup_index not in _stronger_metroid_indices:
+                    locations.append(node)
+                elif config.prefer_stronger_metroids and pickup_index in _stronger_metroid_indices:
+                    locations.append(node)
+            # Boss pickups
+            elif config.prefer_bosses and pickup_index in _boss_indices:
                 locations.append(node)
             # DNA anywhere
-            elif not config.prefer_metroids:
+            elif not config.prefer_metroids and not config.prefer_stronger_metroids and not config.prefer_bosses:
```
ThanatosGit, line 42: I think this was "wrong" at the first place because isn't this just `else:`? There shouldn't be another case beause it's already ensured that it is a `PickupNode` and all other configuration options are handled by the cases above.

randovania/games/samus_returns/generator/bootstrap.py
```diff
@@ -87,8 +97,17 @@ def assign_pool_results(self, rng: Random, patches: GamePatches, pool_results: P
...
+
         for dna, location in zip(dna_to_assign, locations, strict=False):
             pool_results.to_place.remove(dna)
             pool_results.assignment[location.pickup_index] = dna
 
         return super().assign_pool_results(rng, patches, pool_results)
+
+
+_boss_indices = [37, 139, 171]
+_stronger_metroid_indices = [177, 178, 181, 185, 186, 187, 188, 192, 193, 199, 200, 202, 205, 209]
```
ThanatosGit, line 113: If it isn't required elsewhere in this module, just move it to the function `all_dna_locations` itself.

randovania/games/samus_returns/generator/pool_creator.py
```diff
@@ -42,6 +43,19 @@ def pool_creator(results: PoolResults, configuration: BaseConfiguration, game: G
...
+    if config.prefer_metroids:
+        max_artifacts = 25
+    elif config.prefer_stronger_metroids:
+        max_artifacts = 14
+    elif config.prefer_bosses:
+        max_artifacts = 3
+    elif not config.prefer_metroids and not config.prefer_stronger_metroids and not config.prefer_bosses:
+        max_artifacts = 39
+    elif config.required_artifacts > max_artifacts:
+        raise InvalidConfiguration("More Metroid DNA than allowed!")
```
ThanatosGit, line 57: This is definetly wrong.
E.g. you set prefer metroids and stronger metroids but this is all `elif`, so  `max_artifacts` will be set to `25`.
BUT as it also uses else if in `elif config.required_artifacts > max_artifacts:`, the check is never done.
It should be `if config.required_artifacts > max_artifacts:` and that throws an `InvalidConfiguration` if you activate more than one option.
It needs to be pretty mucht the same logic as in `def num_preferred_locations`

Maybe a good reason to not leave this line uncovered from the tests :p

ThanatosGit (changes requested):

ThanatosGit:
Another thing: Should be maybe make more clear that you are explicitly allowed to disable all options to place them "anywhere"?
image

Miepee:
> Should we maybe make more clear that you are explicitly allowed to disable all options to place them 

yes. Can see that causing confusion, because in am2r if you disable all options, then it will give you errors

dyceron:
Later on I want to revamp this so placing anywhere is logical

randovania/games/samus_returns/generator/bootstrap.py
dyceron: Without this, it would incorrectly place dna when I had an option selected (I had bosses only enabled and it place it anywhere)

randovania/games/samus_returns/generator/bootstrap.py
dyceron: Will do

randovania/games/samus_returns/generator/bootstrap.py
ThanatosGit: Oops, yes you are right. If you would use `else` and you iterate through all nodes and only have boss pickups and then of course often no condition is true and the else would catch it and add the node.
So, nvm!

ThanatosGit:
> Later on I want to revamp this so placing anywhere is logical

We could do it after this PR. We should maybe think about a more explicit option like "Prefered locations or anywhere" and the checkboxes are disabled in the anywhere case?

dyceron:
> We could do it after this PR. We should maybe think about a more explicit option like "Prefered locations or anywhere" and the checkboxes are disabled in the anywhere case?

I was thinking kinda like how Echoes does it. Have a drop down with a slider for DNA anywhere which is logical, then have other options in the dropdown for any metroids, standards only, stronger only, bosses only, etc.

randovania/games/samus_returns/generator/pool_creator.py
```diff
@@ -42,6 +43,22 @@ def pool_creator(results: PoolResults, configuration: BaseConfiguration, game: G
...
+    max_artifacts = 0
+    if config.prefer_metroids:
+        max_artifacts += 25
+    if config.prefer_stronger_metroids:
+        max_artifacts += 14
+    if config.prefer_bosses:
+        if max_artifacts == 39:
+            pass
+        else:
+            max_artifacts += 3
```
ThanatosGit, line 56: If you just `pass` when the condition is `true` and just doing things in `else`, just negate the condition in the if-clause.

randovania/games/samus_returns/gui/preset_settings/msr_goal_tab.py
ThanatosGit, line 1: Can you add some tests for the uncovered lines in this file? As we are making big steps towards a beta tester phase, it's probably reasonable to make more tests.

ThanatosGit (changes requested):

ThanatosGit:
> > We could do it after this PR. We should maybe think about a more explicit option like "Prefered locations or anywhere" and the checkboxes are disabled in the anywhere case?
> 
> I was thinking kinda like how Echoes does it. Have a drop down with a slider for DNA anywhere which is logical, then have other options in the dropdown for any metroids, standards only, stronger only, bosses only, etc.

UI design? I gotta go!
But on a more serious note: I don't think a dropdown with 8 options is the way to go. Can be confusing to have things like `DNA on stronger metroids`, `DNA on all metroids and bosses`, `DNA on weaker metroids and bosses` etc.
I think a checkbox for each option is much more easy to understand like it is now?

test/games/samus_returns/gui/preset_settings/test_msr_goal_tab.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Assert
+    if not prefer_metroids and not prefer_stronger_metroids and not prefer_bosses:
+        assert slider_value_after_first_set == 0
+    # The slider value should never increase from what it was before
+    assert slider_value_after_first_set >= tab.dna_slider.value()
+    assert tab.num_preferred_locations == expected_max_slider
+    assert tab.dna_slider.maximum() == expected_max_slider
+    assert tab.dna_slider.isEnabled() == (expected_max_slider > 0)
+    assert editor.configuration.artifacts.prefer_metroids == prefer_metroids
```
ThanatosGit, line 66: Assert that this is a `MSRConfiguration` then all the mypy errors should be gone.

ThanatosGit (changes requested):

ThanatosGit (approved):

PR #5513, opened 2023-10-21
AM2R: fix spoiler log in credits being broken

Miepee:
No clue how the other one failed cause i did test it.


Miepee:
Added test

ThanatosGit (approved):

PR #5522, opened 2023-10-22
Echoes: Make sure event nodes are reachable

Tyranisaur:
For the resolver to not explore these event nodes over and over gain, make sure they are reachable also after the event has been collected.


randovania/games/prime2/logic_database/Temple Grounds.txt
```diff
@@ -1496,7 +1496,16 @@ Extra - in_dark_aether: False
   > Pickup (Power Bomb)
       Morph Ball and Power Bomb
   > Event - Dynamo Chamber Gates
-      Scan Visor and Before Dynamo Chamber Gates
+      All of the following:
+          Scan Visor
+          Any of the following:
+              Space Jump Boots or Before Dynamo Chamber Gates
```
dyceron, line 1502: Why check for before the event on the event connection?

dyceron (changes requested):
Do we need changelog entry for this?

randovania/games/prime2/logic_database/Temple Grounds.txt
XenoWars: probably because if you've done the event, you then can't just walk on through

dyceron: Its possible I misread this

Tyranisaur:
> Do we need changelog entry for this?

On paper these changes should have no actual impact on what is logical to do, but only serve as a change of an implementation detail that happens to be useful for how the resolver works. I think my thought process was that this doesn't need a changelog entry because there's already an entry about improvements to the resolver, but since it's in the general resolver section it might not cover this. So I'm uncertain. Also, if this is now being pushed into the next release then that would probably help the case for adding a changelog entry, plus I won't need to move the entry to the next release as a quasi merge conflict :)

dyceron:
> > Do we need changelog entry for this?
> 
> On paper these changes should have no actual impact on what is logical to do, but only serve as a change of an implementation detail that happens to be useful for how the resolver works. I think my thought process was that this doesn't need a changelog entry because there's already an entry about improvements to the resolver, but since it's in the general resolver section it might not cover this. So I'm uncertain. Also, if this is now being pushed into the next release then that would probably help the case for adding a changelog entry, plus I won't need to move the entry to the next release as a quasi merge conflict :)

Makes sense. Would be nice to get this in this release

JeffGainsNGames (approved):

dyceron (approved):

PR #5529, opened 2023-10-23
AM2R: Add better input validation

Miepee:
No idea if I should write tests for that function, and if yes how.


ThanatosGit:
If you ask if you should write tests, the answer is always yes :p. Here some copy paste stuff for you: https://github.com/randovania/randovania/blob/903f07bc464922be9a419b1ffbac22cef251a5bb/test/games/dread/gui/test_dread_game_export_dialog.py#L103-L173

PR #5560, opened 2023-10-28
Add am2r to stable

Miepee:
The header in changelog should be changed to 7.0.0 at one point,but not sure if we're still waiting on a patch or two.


ThanatosGit (approved):
🥳🥳🥳

PR #5572, opened 2023-10-31
[Dread] Update to Underlava Puzzle Room 2 logic

steven11sjf:
You can break the speed blocks in Underlava Puzzle Room 2 from the right. After opening the grapple block (no changes), if you have a speedbooster upgrade you can charge a shinespark on the above-lava element, run back and morph into the grapple tunnel, and shinespark up and down to destroy the blocks. 

Screenshot of added trick connection:
image

### DIFF
--- a/randovania/games/dread/logic_database/Cataris.txt
+++ b/randovania/games/dread/logic_database/Cataris.txt
@@ -3875,7 +3875,7 @@ Extra - asset_id: collision_camera_053
   > Dock to Underlava Puzzle Room 1
       Morph Ball and After Cataris - Underlava Speed Blocks Destroyed
   > Event - Speed Blocks Destroyed
-      # Charge shinespark on above lava platform
+      # https://youtu.be/syNED-7J8HM
       Gravity Suit and Morph Ball and Speed Booster and Speed Booster Upgrade and After Cataris - Lava Grapple Block and Speed Booster Conservation (Beginner)
 
 > Tunnel to Dropdown Pit; Heals? False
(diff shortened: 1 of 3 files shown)


randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -3874,6 +3874,9 @@ Extra - asset_id: collision_camera_053
                   Heat/Cold Runs (Advanced) and Heat Damage ≥ 200
   > Dock to Underlava Puzzle Room 1
       Morph Ball and After Cataris - Underlava Speed Blocks Destroyed
+  > Event - Speed Blocks Destroyed
+      # Charge shinespark on above lava platform
+      Gravity Suit and Morph Ball and Speed Booster and Speed Booster Upgrade and After Cataris - Lava Grapple Block and Speed Booster Conservation (Beginner)
```
Tyranisaur, line 3879: Can you break the speed blocks at the dock to Underlava Puzzle Room 1?

XenoWars: i would also test being able to get to the item with just speed without breaking those blocks to the left.

XenoWars: ok, i tested, and yes you can destroy the speed blocks. and yes you can get up without destroying them (though irrelevant now). Since you need beginner conservation to get to that dock from reverse in the first place, it should be fine leaving the connection in Underlava Puzzle Room 1 as trivial.

steven11sjf: Yep

Tyranisaur (changes requested):
You should be clear to sync with main and add a changelog entry under version 7.1.0

ThanatosGit:
Looks like the videos are in it now? So, I re-request review from db maintainers.

Tyranisaur (changes requested):
Still missing a changelog entry.

Tyranisaur (approved):

PR #5582, opened 2023-11-05
Dread: Golzuna logic update

XenoWars:
- Changed: Golzuna logic has been overhauled to include Storm Missiles, Bombs, or Cross Bombs to fight it and forcing Flash Shift, Spin Boost, or Space Jump to dodge its attacks.

### DIFF
--- a/randovania/games/dread/logic_database/Ghavoran.txt
+++ b/randovania/games/dread/logic_database/Ghavoran.txt
@@ -2074,6 +2074,8 @@ Extra - asset_id: collision_camera_026
                           Any of the following:
                               Use Spin Boost
                               Flash Shift and Flash Shift Upgrade ≥ 2
+                      # Speed-only dodging, requires good rng on its attack cycle
+                      Speed Booster and Combat (Expert) and Can Slide
           Any of the following:
               # Requirements for killing Golzuna
               All of the following:
--- a/randovania/games/dread/logic_database/Ghavoran.txt
+++ b/randovania/games/dread/logic_database/Ghavoran.txt
@@ -2061,22 +2061,31 @@ Extra - asset_id: collision_camera_026
   * Extra - tile_types: ('WEIGHT',)
   > Event - Golzuna
       All of the following:
-          Any of the following:
-              # Requirements for dodging attacks. Not needed if at least beginner combat.
-              Combat (Beginner)
+          All of the following:
+              # Requirements for dodging attacks.
               All of the following:
                   # Dodge its slam attack
                   Can Slide
                   Any of the following:
                       # Dodge its zappy attacks
-                      Use Spin Boost
-                      Flash Shift and Wall Jump (Beginner)
+                      Space Jump
+                      All of the following:
+                          Combat (Beginner)
+                          Flash Shift or Use Spin Boost
           Any of the following:
               # Requirements for killing Golzuna
-              Shoot Charge Beam
+              All of the following:
+                  # Charge Beam methods
+                  Shoot Charge Beam
+                  Any of the following:
+                      # Plasma, Wave, or Power
+                      Combat (Intermediate) or Beam Damage 120 or Beam Damage 240
+                      # Wide
+                      Combat (Beginner) and Beam Damage 75
+              # Shinesparks only
               Speed Booster and Combat (Advanced)
               All of the following:
-                  Combat (Intermediate)
+                  # Missile methods
                   Any of the following:
                       All of the following:
                           # Minimum: 35 Ice Missiles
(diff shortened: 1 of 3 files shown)


randovania/games/dread/logic_database/Ghavoran.txt
```diff
@@ -2061,46 +2061,69 @@ Extra - asset_id: collision_camera_026
...
+                          Flash Shift or Use Spin Boost
           Any of the following:
               # Requirements for killing Golzuna
-              Shoot Charge Beam
+              All of the following:
+                  # Charge Beam methods
+                  Shoot Charge Beam
+                  Any of the following:
+                      # Plasma, Wave, or Power
+                      Combat (Intermediate) or Beam Damage 120 or Beam Damage 240
```
Tyranisaur, line 2082: It's redundant to say `Beam Damage 120` or `Beam Damage 240`

randovania/games/dread/logic_database/Ghavoran.txt
```diff
@@ -2061,46 +2061,69 @@ Extra - asset_id: collision_camera_026
...
+              Any of the following:
+                  # Bomb methods
+                  # Normal Bombs
+                  Combat (Intermediate) and Lay Bomb
+                  # Cross Bombs
+                  Combat (Beginner) and Lay Cross Bomb
           Any of the following:
               # Energy Requirements
-              Combat (Intermediate) or Normal Damage ≥ 699
+              Normal Damage ≥ 699
```
Tyranisaur, line 2126: There's probably room for having a combat level that doesn't require any health, though I think it could reasonably he as high as Expert.

randovania/games/dread/logic_database/Ghavoran.txt
```diff
@@ -2061,46 +2061,69 @@ Extra - asset_id: collision_camera_026
...
                   # Dodge its slam attack
                   Can Slide
                   Any of the following:
                       # Dodge its zappy attacks
-                      Use Spin Boost
-                      Flash Shift and Wall Jump (Beginner)
+                      Space Jump
+                      All of the following:
+                          Combat (Beginner)
+                          Flash Shift or Use Spin Boost
```
Tyranisaur, line 2074: My first thought here is that there might be a world where having both Flash Shift and Spin Boost could live at a lower Combat difficulty than each of them individually, but I don't think it's very important.

However, I'm not sure if simply `Flash Shift` is enough here, maybe you need to also require a certain number of Flash Shift Upgrades.

Tyranisaur (changes requested):

randovania/games/dread/logic_database/Ghavoran.txt
XenoWars: seeing as you can't survive being tapped by Golzuna with 0 tanks even with gravity, I don't see it as doable.

randovania/games/dread/logic_database/Ghavoran.txt
XenoWars: Flash Shift alone goes the whole arena. Although i suppose just having Flash Shift could mean you might only have a single flash, if you set the main to not give any upgrades.

randovania/games/dread/logic_database/Ghavoran.txt
Tyranisaur: My thinking is that there are no paths that require you to take damage, so the damage requirements are here only to act as leeway so that players don't need to perfectly avoid damage.

ThanatosGit:
> forcing Flash Shift, Spin Boost, or Space Jump to dodge its attacks

What about a cursed fight without any of those?
After the first spark on golzuna, iirc you can get lucky and golzuna does not turn around, so you can directly spark again (which I failed to store in this video at all). You could also use the second shinespark with a wall jump to get over the 4 diagonal zappy things if that's what golzuna does after the first spark.
I cannot even tell if you can beat all patterns with this strategy but maybe something for hypermode or one trick level below.

https://github.com/randovania/randovania/assets/117127188/20e4a256-ca30-4cf4-9d0b-beb0af8aad62

Tyranisaur:
Oh right, if you use the advanced combat method with 2 shinesparks, then you could skip dodging the zappy attacks. I think it's fine off the basis that it's not reasonable to get patterns that let you do that since you only need 2 sparks and one of them can be reasonably done from the start of the fight.

XenoWars:
but can you do that without morph? it's possible to get here without morph if you were to start in the save station right below, or coming from the transport there, with X pre-released of course.

ThanatosGit:
> but can you do that without morph?

Yes, you can just slide and do the first shinespark. And yes, two zappy things in a row (unless it's not two times the following single one) will be bad. That's why it is a very high trick difficulty but requires nothing besides Speed Booster (and slide).

Tyranisaur (approved):

PR #5588, opened 2023-11-05
Add Ice/Melee Clip dock to get to teleporter room from teleporter Access in area 1

MeriKatt:
Not sure if i did this entirely right, Added a Generic Node, changed it to open passage dock to the teleporter in next room.
Then added the requirements for it in the connection from the start point to it, this IS a one way connection.

### DIFF
--- a/randovania/games/samus_returns/logic_database/Area 1.txt
+++ b/randovania/games/samus_returns/logic_database/Area 1.txt
@@ -738,10 +738,10 @@ Extra - asset_id: collision_camera_028
   * Open Passage to Shaft West/Dock to One-Way Puzzle
   > Door to Shaft West
       Plasma Beam
-  > Dock to Save Station South (Top)
+  > Dock to Save Station South
       Trivial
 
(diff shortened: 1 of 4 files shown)


randovania/games/samus_returns/logic_database/Area 1.txt
```diff
@@ -1039,6 +1039,12 @@ Extra - asset_id: collision_camera_042
       Trivial
   > Tunnel to Teleporter Room
       Lay Any Bomb
+  > Dock to Teleporter Room
+      Morph Ball and Melee Clip (Intermediate) and Fully Freeze Enemy
+
+> Dock to Teleporter Room; Heals? False
+  * Layers: default
+  * Open Passage to Teleporter Room/Teleporter
```
dyceron, line 1047: You need to add a dock in the teleporter room as well, which connects to the teleporter or right dock, and then connect both new docks together

dyceron (changes requested):

dyceron (approved):

PR #5601, opened 2023-11-08
Bumped random-enemy-attributes from 1.0.3.1 to 1.0.3.3

Fantaselion:
Updated it so that it runs on the latest version of Cython, along with fixing the incorrect Homepage on Pypi.


ThanatosGit:
There was already a dependabot PR for it ;)
https://github.com/randovania/randovania/pull/5600

PR #5604, opened 2023-11-10
Update .gitignore

toasterparty:


.gitignore
```diff
@@ -85,8 +85,12 @@ cache/
 exported_videos/*
 /randovania/data/icons/executable_icon.ico
 
+# Ignore generated UI files
+/randovania/gui/ui_files/*_ui.py
```
Miepee, line 89: i thought these were already git ignored? at least my git clients never see these

.gitignore
```diff
@@ -85,8 +85,12 @@ cache/
...
+
 # Ignore certain files on root
 /*.rdvgame
 /*.rdvpreset
+/*.log
 /*.txt
-/*.log
\ No newline at end of file
+!requirements*.txt
```
Miepee, line 96: I would rather know why .txts are in the gitignore in the first place.

.gitignore
henriquegemignani: They are. It's via a git ignore in the folder itself.
This folder shouldn't have anything, but toaster's setup somehow generate files there.

.gitignore
henriquegemignani: Common mistake of adding files to root.
Since we don't plan on addingang other txt files to root we can just ignore.

.gitignore
Miepee: i feel like his setup is wrong then?
`randovania/gui/ui_files/` only has this in its gitignore:
```
*.ui.autosave
```

the generated py files should always be placed into `randovania/gui/generated`. Which does have
```
*_ui.py
```
in its gitignore

henriquegemignani: Yes, that folder shouldn't ever create py files. Somehow it does for him

ThanatosGit: Toaster, are you using vs code?
There is a option for the qt plugins which auto generated the files in the folder if you change something in a ui file. That's also why I accidentally committed one of these files but after that, I just unticked the option.

toasterparty: That was it. I don't even remember installing the qt plugin haha.

Miepee (approved):

PR #5623, opened 2023-11-17
MSR - Add Diggernaut Chase Sequence reward as boss location

dyceron:
I added the Diggernaut Chase Sequence reward (vanilla Space Jump) as a boss location to add more value to the boss goal. It fits since Diggernaut is a boss, so the chase, although removed, could be considered a "boss" or something equivalent in importance. A neat side effect of this is now every even area has a boss check (2, 4, 6, 8). This is also more relevant now since the area has be relocked by Charge, so it's less "free" to access.

image

### DIFF
--- a/randovania/games/samus_returns/generator/bootstrap.py
+++ b/randovania/games/samus_returns/generator/bootstrap.py
@@ -22,7 +22,7 @@ if TYPE_CHECKING:
 
 def all_dna_locations(game: GameDescription, config: MSRArtifactConfig) -> list[PickupNode]:
     locations = []
-    _boss_indices = [37, 139, 171]
+    _boss_indices = [37, 99, 139, 171]
     _stronger_metroid_indices = [177, 178, 181, 185, 186, 187, 188, 192, 193, 199, 200, 202, 205, 209]
 
     for node in game.region_list.all_nodes:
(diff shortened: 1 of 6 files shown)


ThanatosGit:
I really have a mixed opinion about that.
Adding something to the boss goal sounds nice and I understand why Diggernaut Chase fits. Especially with the bad distribution of the vanilla bosses having one very early and all others are very late game content (considering that diggernaut is at the very end of area 6 and after that you just rush through the shortest areas to the queen).
But at the same time it is maybe very weird to tell other players: "We consider Diggernaut Chase as a boss. Oh, and we removed Diggernaut Chase". So, it is a boss location without a boss which makes it "a location".

Hmmm...would like to know what other people think about it.

dyceron:
I didn't want to add a whole different group just for this one location. An alternate fix would be to re-enable the chase but I won't do that ever

ThanatosGit:
@MasierAzulite @Miepee @MeriKatt Any option on that?
I still understand the motivation behind it (adding more value to the goal etc.) but still weird for me to have a boss location without an actual boss. 

What are the requirements to get this item?

Miepee:
I'm indifferent about it TBH.

MeriKatt:
I like the idea personally, in vanilla I’m pretty sure that you require grapple and supers at the least to get through it, grapple can technically be skipped but in a trickless scenario it’s required. With the chase removed you still need supers to reach the item I believe as well as space jump or and ibj or spiderspark to get back out if I remember right. Either way I’m in favor of it, it may seem weird I know but the chase itself being removed just makes it so you don’t require multiple passes to get the items if you have what you need to get them

ThanatosGit:
Well, then I merge it.

ThanatosGit (approved):

PR #5635, opened 2023-11-21
AM2R: Zeta/Omega rebalances

JeffGainsNGames:
- Health and dodging requirements separated
- Health scaled for multiple difficulties
- Dodging requirements expanded

### DIFF
--- a/randovania/games/am2r/logic_database/header.txt
+++ b/randovania/games/am2r/logic_database/header.txt
@@ -226,8 +226,8 @@ Templates
               Charge Beam and Missile-Less Metroid Fights (Advanced)
           Any of the following:
               # Energy Requirements
-              Combat (Advanced) or Normal Damage ≥ 299
-              Combat (Intermediate) and Normal Damage ≥ 199
+              Combat (Expert) or Normal Damage ≥ 299
+              Combat (Advanced) and Normal Damage ≥ 199
           Any of the following:
               # Dodging Requirements
               Space Jump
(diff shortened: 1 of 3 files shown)


Miepee:
conflucts

randovania/games/am2r/logic_database/header.txt
```diff
@@ -225,11 +225,15 @@ Templates
               Missiles ≥ 30 and Combat (Intermediate)
               Charge Beam and Missile-Less Metroid Fights (Advanced)
           Any of the following:
-              # Energy requirements
-              Normal Damage ≥ 199
+              # Energy Requirements
+              Combat (Advanced) or Normal Damage ≥ 299
```
Miepee, line 229: i wonder if this should  be expert instead of advanced. cause doing both min items + hitless is way harder

JeffGainsNGames: Are these health values good, or should I bump them both up by 100
Zeta's trickless 399, Omega's 499?

Miepee: health values seem fine? sam had trouble fighting a zeta with 4 etanks on hard (299*1,5=448), and you take double damage there.

JeffGainsNGames: figured since we bumped the super-less fights it might be good to get the health up a bit. But I'm not too concerned.

PR #5641, opened 2023-11-23
Move FtpUploader to lib

ThanatosGit:
I moved the `FtpUploader` from `games.dread.gui.dialog` to `lib`.
If I see it correctly, there isn't any dialog or dread specific code in it and as it will be used in SR in the very near future, it is more like a lib for ftp uploads and could also be used for any other game in the future.


henriquegemignani (approved):

henriquegemignani:
It was there more as "it's a dread feature only used by it, stays here for now".

PR #5646, opened 2023-11-24
Item Tracker tooltip fixes

steven11sjf:
- adds a tooltip with the resource's long name to disabled images
- selects the first resource if it is a progressive item
- added a stylesheet for the QToolTip to make text clearer (black on light gray instead of gray on light gray)

image


steven11sjf:
merged main and bumped changelog to the new month since it still had 7.1.0 as the upcoming version

henriquegemignani (approved):

PR #5653, opened 2023-11-25
MSR - Simplify Surface West

dyceron:
Removes the unused connections/nodes in Surface West, since these connections are only accounted for in Surface East


MeriKatt (approved):
Definitely makes sense since this is all handled in Surface East already and Surface West is mainly accessed when going to ridley

PR #5664, opened 2023-11-27
Prime 1: Have artifacts be static assignments instead of being placed by generator

Miepee:
There was some concern that placing artifacts by the generator lead to them being placed too soon and/or too easily accessible
This is thus an attempt to place them statically, before any generation takes place.

Obviously needs heavy testing. Should maybe also be gated behind an option too.

### DIFF
--- a/randovania/games/prime1/generator/bootstrap.py
+++ b/randovania/games/prime1/generator/bootstrap.py
@@ -4,15 +4,34 @@ import copy
 import dataclasses
 from typing import TYPE_CHECKING
 
+from randovania.game_description.db.pickup_node import PickupNode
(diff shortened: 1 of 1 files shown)


Samuel6710:
Should be an option imo.

henriquegemignani:
This has the massive disadvantage of being local only.
Also it should handle, in some way or form, if certain locations are inaccessible, like excluded pickups, starting location, elevator distribution.

duncathan:
I'm not totally convinced this will provide better outcomes than messing with the minimum actions for artifact placement feature tbh

Miepee:
Closing as not deemed useful. Idea could be revisited, if there's a generic implementation of it accessible.

PR #5674, opened 2023-11-30
Multi-pickup generator settings no longer experimental

toasterparty:
It's been long enough. I still hold that this is objectively better for seed diversity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [7.1.0] - 2023-12-??
 
 - Fixed: Bug with progressive suits in the autotracker always highlighting first suit
+- Changed: "Remove redundant pickup alternatives" and "Stagger placement of pickups" are no longer experimental options and will be included in all presets moving forwards.
 
 ### AM2R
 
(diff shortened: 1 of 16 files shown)


duncathan:
Any particular reason to leave the preset options in?

toasterparty:
No, but I know historically that options removed from GUI still tend to keep their settings in the preset.

toasterparty:
@henriquegemignani I need help understanding the final failing test

henriquegemignani:
A test das depending on the single set setting being off, as the function behaves differently with that setting enabled.
Since it's now always, the test is failing

toasterparty:
> Since it's now always, the test is failing

I'm not sure how you came up with the expected data for the original implementation. It sounds like I need to do that again for the new output.

toasterparty:
Generator Function - https://github.com/randovania/randovania/blob/2d4925641e657a8ab2e7250769335abefde51e2b/randovania/generator/filler/pickup_list.py#L168-L174

Test - https://github.com/randovania/randovania/blob/2d4925641e657a8ab2e7250769335abefde51e2b/test/generator/filler/test_pickup_list.py#L119-L172

henriquegemignani (approved):

PR #5678, opened 2023-12-01
Fix exporting Echoes with Memo popup enabled, hidden collection text and cross-game multiworld

henriquegemignani:


ThanatosGit (changes requested):
Looks fine to me so far.
I'm just concerned about the failing test. There are a lot of additional (test) files which create a `PickupEntry` but using just an `int` as second parameter. Second parameter should be a `PickupModel` and now a test fails because you actually access the model now.
Isn't really a problem of this PR but kinda because now the tests are failing (though they were already wrong before).

henriquegemignani:
Fixed tests

Miepee:
changelog conflicts

ThanatosGit (approved):

PR #5679, opened 2023-12-03
Improve db video static site generator

toasterparty:
[Before.zip](https://github.com/randovania/randovania/files/13537854/Before.zip)

[After.zip](https://github.com/randovania/randovania/files/13537855/After.zip)


toasterparty:
`tools\update_requirements.py` doesn't add the new dependency to requirements.txt. Should I add it by hand or is there a way to fix this?

duncathan:
you haven't added the new dependency to `pyproject.toml`

henriquegemignani (approved):

PR #5680, opened 2023-12-03
Dread/pseudo additions

Tyranisaur:


randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -4952,7 +4952,12 @@ Extra - asset_id: collision_camera_081
   > Event - Screw Attack Water Blob
       Any of the following:
           Wave Beam or Shoot Diffusion Beam
-          Pseudo-Wave Beam (Advanced) and Shoot Beam
+          All of the following:
+              Pseudo-Wave Beam (Advanced) and Shoot Charge Beam
+              Any of the following:
+                  Slide
+                  No Gravity Suit and Enabled Highly Dangerous Logic
```
XenoWars, line 4959: why the slide req?

randovania/games/dread/logic_database/Burenia.txt
```diff
@@ -303,7 +303,12 @@ Extra - asset_id: collision_camera_002
                   Before Burenia - Dairon Hub Water Blob Right and Before Burenia - Dairon Hub Water Blob Left and Enabled Highly Dangerous Logic
                   Simple IBJ or Use Spin Boost
   > Event - Water Blob Right
-      Wave Beam or Lay Bomb or Lay Power Bomb
+      Any of the following:
+          Wave Beam or Lay Bomb or Lay Power Bomb
```
XenoWars, line 307: Unless we make it so all blobs get destroyed from PBs, this should probably account for Shoot Beam with them.

XenoWars (changes requested):

randovania/games/dread/logic_database/Artaria.txt
Tyranisaur: So this is a thing I added for the purpose of making this be non-dangerous. Since there's a lower difficulty version that requires slide, if you have slide, it's not a problem that you pick up gravity suit. Since often, and will probably always, have slide as a starting item, this allows gravity suit to not be a dangerous resource.

randovania/games/dread/logic_database/Burenia.txt
Tyranisaur: Good call.

randovania/games/dread/logic_database/Artaria.txt
XenoWars: this makes sense, although charge isn't needed for the pseudo.

XenoWars (approved):

PR #5687, opened 2023-12-05
MSR - Add support for Reserve Tanks

dyceron:
Adds support for the amiibo Reserve Tanks as shuffleable items
image
image
image


randovania/games/samus_returns/layout/preset_describer.py
ThanatosGit, line 1: Can you add / modify the tests for this file to cover the added lines? We already have a file for it, so it's hopefully a low hanging fruit: https://github.com/randovania/randovania/blob/main/test/games/samus_returns/layout/test_msr_describer.py

ThanatosGit (changes requested):

dyceron:
I changed where and how the preset changes are handled. It is now also covered by the current tests. 
image

ThanatosGit (approved):

PR #5688, opened 2023-12-06
Fix Burenia fan item logic

ArcanoxDragon:
The "Flash-only" trick to get the item in the Burenia fan requires the player to Flash Shift quickly twice to the left. The logic didn't have a Flash Shift Upgrade in the requirements, so a seed could end up being unsolvable if the player was not given enough upgrades.

### DIFF
--- a/randovania/games/dread/logic_database/Burenia.txt
+++ b/randovania/games/dread/logic_database/Burenia.txt
@@ -258,7 +258,7 @@ Extra - asset_id: collision_camera_002
           Speed Booster
           # A grounded bomb jump with cross bombs to get past the fan
           Movement (Beginner) and Lay Cross Bomb
-          Flash Shift and Movement (Advanced)
+          Flash Shift and Flash Shift Upgrade and Movement (Advanced)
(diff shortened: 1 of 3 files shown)


XenoWars (approved):

XenoWars:
just need to resolve conflicts

randovania/games/dread/logic_database/Burenia.txt
```diff
@@ -258,7 +258,7 @@ Extra - asset_id: collision_camera_002
           Speed Booster
           # A grounded bomb jump with cross bombs to get past the fan
           Movement (Beginner) and Lay Cross Bomb
-          Flash Shift and Movement (Advanced)
+          Flash Shift and Flash Shift Upgrade and Movement (Advanced)
```
Tyranisaur, line 261: I think we arrived on the conclusion that this can be lowered from advanced to intermediate. Since you're changing this line right here, it probably makes more sense to change that in this PR to avoid merge conflicts.

Tyranisaur (changes requested):

PR #5698, opened 2023-12-10
Very basic item hints for MSR

ThanatosGit:
Redone the rdvgame files in the tests because they didn't had hints before.


dyceron:
Is this PR going to enable UI to change the hint precision, or is that for another pr?

dyceron (changes requested):
Add tests

ThanatosGit:
> Is this PR going to enable UI to change the hint precision, or is that for another pr?

Nope. It's really very simple. We might want to change it in the future. That depends on whether we think the clues are too powerful or too weak, as they are now.

Removed all the unneccessary functions. The generator part (`HintDistributor`) is not tested really for specific games if I'm not mistaken.

randovania/games/samus_returns/exporter/hint_namer.py
```diff
@@ -0,0 +1,62 @@
...
+        msg = self.location_formatters[hint.precision.location].format(
+            game,
+            dataclasses.replace(pick_hint, pickup_name=pick_hint.pickup_name),
+            hint,
+            with_color,
+        )
+        return f"{msg}\n"
+
+    def format_temple_name(self, temple_name: str, with_color: bool) -> str:
+        raise RuntimeError("Unsupported feature")
```
dyceron, line 62: Since color isn't a thing, shouldn't this be removed?

ThanatosGit: In general, yes, but in this case it's implementing the methods from the base class: https://github.com/randovania/randovania/blob/main/randovania/exporter/hints/hint_namer.py

dyceron (approved):

dyceron:
We probably want to have a guaranteed hint for the baby.

ThanatosGit:
> We probably want to have a guaranteed hint for the baby

At a specific spot or just guarantee that the hint is somewhere?

dyceron:
> At a specific spot or just guarantee that the hint is somewhere?

We could have a specific spot (could do the same for ice later on) or for now just have it guaranteed period anywhere)

PR #5704, opened 2023-12-11
MSR - Update FAQ, Update some text

dyceron:
image

image


ThanatosGit (commented):
Looks good so far.
Only one thing came to my mind: Do we have something explicitly describing how you access ridley? Going east -> west and back through the same way is not really obvious.

dyceron:
Updated the `Differences` tab to include information on how the Surfaces are connected. Should I also add this to the FAQ?

image

ThanatosGit (approved):

PR #5705, opened 2023-12-11
MSR - Add spoiler to credits

dyceron:


randovania/games/samus_returns/exporter/patch_data_factory.py
```diff
@@ -275,6 +284,7 @@ def create_data(self) -> dict:
                 "reverse_area8": self.configuration.reverse_area8,
             },
             "text_patches": self._static_text_changes(),
+            "spoiler_log": self._credits_spoiler(),
```
ThanatosGit, line 287: If you want to remove the spoiler log for races, you could do 
```python 
"spoiler_log": self._credits_spoiler() if self.description.has_spoiler else {},
``` 
for now.
That would export an empty spoiler log for races but still triggers the patcher to create the headlines (but that's not a big deal)

dyceron: Will do

randovania/games/samus_returns/exporter/hint_namer.py
```diff
@@ -10,11 +10,22 @@
...
     from randovania.interface_common.players_configuration import PlayersConfiguration
 
 
+def _area_name(region_list: RegionList, pickup_node: PickupNode, hide_region: bool) -> str:
+    area = region_list.nodes_to_area(pickup_node)
+    return region_list.area_name(area)
+
+
+def colorize_text(text: str, with_color: bool) -> str:
+    return text
```
ThanatosGit, line 26: This looks useless if it only returns the first parameter without doing anything.
Could also keep it for future if we find out how to colorize the text.

dyceron: I believe I left that because the function was being called further down. Could probably simplify that or just leave it.

ThanatosGit (approved):

PR #5712, opened 2023-12-11
MSR - Update A6 Trickless Logic

dyceron:
- Improves Trickless logic for Area 6
- Adds template for defeating enemies that can't be countered

### DIFF
--- a/randovania/games/samus_returns/logic_database/Area 6.txt
+++ b/randovania/games/samus_returns/logic_database/Area 6.txt
@@ -20,7 +20,7 @@ Extra - asset_id: collision_camera_034
   > Before Diggernaut Area
       All of the following:
           Morph Ball and After Area 6 - Transport to Area 7 Grapple Block Top
-          Plasma Beam or Destroy Non-Counterable Enemies
+          Plasma Beam or Defeat Non-Counterable Enemies
 
 > Save Station; Heals? False; Spawn Point; Default Node
   * Layers: default
@@ -32,7 +32,7 @@ Extra - asset_id: collision_camera_034
   > Before Diggernaut Area
       All of the following:
           Morph Ball
-          Spider Ball or Destroy Non-Counterable Enemies
+          Spider Ball or Defeat Non-Counterable Enemies
 
 > Door to Crumble Steps; Heals? False
   * Layers: default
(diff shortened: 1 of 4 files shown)


randovania/games/samus_returns/logic_database/Area 6.txt
```diff
@@ -18,7 +18,9 @@ Extra - asset_id: collision_camera_034
   * Extra - actor_type: weightactivatedplatform
   * Extra - start_point_actor_name: ST_FromArea09
   > Before Diggernaut Area
-      Morph Ball and After Area 6 - Transport to Area 7 Grapple Block Top
+      All of the following:
+          Morph Ball and After Area 6 - Transport to Area 7 Grapple Block Top
+          Plasma Beam or Defeat Non-Counterable Enemies
```
ThanatosGit, line 23: No idea how this here actually works: The Grapple Block is already pulled by default in the starter settings iirc.  Which means you don't need to collect the event? Is this taken into account somewhere?

Also: Can you bypass `Defeat Non-Counterable Enemies` with damage boost? (This would apply to a lot of changes here)

randovania/games/samus_returns/logic_database/Area 6.txt
```diff
@@ -630,13 +662,16 @@ Extra - asset_id: collision_camera_045
...
   > Top Platform
       Any of the following:
-          Space Jump or Single-wall Wall Jump (Expert)
+          Single-wall Wall Jump (Expert)
+          All of the following:
+              Space Jump
+              Screw Attack or Movement (Intermediate)
           All of the following:
               Morph Ball
               Any of the following:
```
ThanatosGit, line 677: GH doesn't let me write a comment below this line...
 If I read that correctly it would be enough to have Morph + Spider but with Space Jump you are required to have Screw or Movement (Intermediate). That doesn't sound balanced to me. Isn't Morph + Spider much harder or do the enemies not reach you at the left or right side?

randovania/games/samus_returns/logic_database/Area 6.txt
dyceron: Yes the event is handled in the bootstrap iirc.

And yeah there's definitely room for improvement here as tricks could be added, but I wanted to at least make trickless somewhat better and more consistent.  It will have to be revisited later on

randovania/games/samus_returns/logic_database/Area 6.txt
dyceron: Actually Spider should have some movement since it involves dodging the enemies iirc

dyceron: Just tested this. All you have to do is hold up. Might not even need a trick then

ThanatosGit (approved):

PR #5715, opened 2023-12-12
Add vertical dock information

JeffGainsNGames:
Excluded from this PR:
- one way paths (BG2 shinespark, Tower Interior)
- Tower exterior

### DIFF
--- a/randovania/games/am2r/logic_database/GFS Thoth.json
+++ b/randovania/games/am2r/logic_database/GFS Thoth.json
@@ -60,7 +60,7 @@
                 ]
             },
             "nodes": {
-                "Dock to Research Site Elevator": {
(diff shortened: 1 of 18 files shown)


Miepee:
Thoth is missing

Miepee (commented):
HS Exterior -> Arachnus save station door is wrong
Varia Chamber Access -> Left water pipe dock is wrong
South east gamma nest in tower is missing



Looks good otherwise. we should probably better document the ones who're missing tho, so that it isn't needed to crawl through the whole DB for them.

JeffGainsNGames:
Will make corrections. I can add a TODO to the remaining docks, but honestly there aren't many vertical transitions in total.

Miepee (approved):
ty

PR #5726, opened 2023-12-15
MSR - Add hint tab to generated game window

dyceron:
Adds the Hint Tab to the generated game window. I don't know why there's so much padding for each hint

image

### DIFF
--- a/randovania/games/samus_returns/gui/hint_details_tab.py
+++ b/randovania/games/samus_returns/gui/hint_details_tab.py
@@ -1,0 +1,81 @@
+from __future__ import annotations
+
+import collections
+import random
+from typing import TYPE_CHECKING
+
+from PySide6 import QtWidgets
+
+from randovania.exporter.hints.hint_exporter import HintExporter
+from randovania.games.samus_returns.exporter.hint_namer import MSRHintNamer
+from randovania.gui.game_details.game_details_tab import GameDetailsTab
+from randovania.layout import filtered_database
+from randovania.lib.container_lib import iterate_key_sorted
+
+if TYPE_CHECKING:
+    from randovania.game_description.game_patches import GamePatches
+    from randovania.games.game import RandovaniaGame
+    from randovania.interface_common.players_configuration import PlayersConfiguration
+    from randovania.layout.base.base_configuration import BaseConfiguration
+
+
+class MSRHintDetailsTab(GameDetailsTab):
+    def __init__(self, parent: QtWidgets.QWidget, game: RandovaniaGame):
+        super().__init__(parent, game)
+        self.tree_widget = QtWidgets.QTreeWidget(parent)
+
+    def widget(self) -> QtWidgets.QWidget:
+        return self.tree_widget
+
+    def tab_title(self) -> str:
+        return "Hints"
+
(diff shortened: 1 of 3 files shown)


ThanatosGit:
> I don't know why there's so much padding for each hint

Uh...this is using the text how it get's exported. Atm the patcher has the assumption that every hint ends with newline character `\n`. Per message box it gathers three lines by splitting at `\n`.
a) We don't use multi hints per station. Could be removed or being kept for future 
b) Either we change this in the patcher to not force every line with `\n` or change it before showing in the UI here.

dyceron:
> Uh...this is using the text how it get's exported. Atm the patcher has the assumption that every hint ends with newline character `\n`. Per message box it gathers three lines by splitting at `\n`.
> a) We don't use multi hints per station. Could be removed or being kept for future 
> b) Either we change this in the patcher to not force every line with `\n` or change it before showing in the UI here.
> 

Ah that makes sense. I guess we could leave it for now in case we change hints later.

randovania/games/samus_returns/gui/hint_details_tab.py
```diff
@@ -0,0 +1,81 @@
...
+        per_region: dict[str, dict[str, tuple[str, str]]] = collections.defaultdict(dict)
+        namer = MSRHintNamer(all_patches, players)
+        exporter = HintExporter(namer, random.Random(0), ["A joke hint."])
+
+        for identifier, hint in patches.hints.items():
+            node = game.region_list.node_by_identifier(identifier)
+            source_region = region_list.nodes_to_region(node)
+            source_name = region_list.node_name(node)
+
+            hint_text = exporter.create_message_for_hint(hint, all_patches, players, False)
```
ThanatosGit, line 53: If you don't want to change the patcher now, use `.strip()` at the end of this line.
It should remove the new line and fix the padding.

randovania/games/samus_returns/gui/hint_details_tab.py
```diff
@@ -0,0 +1,81 @@
...
+        super().__init__(parent, game)
+        self.tree_widget = QtWidgets.QTreeWidget(parent)
+
+    def widget(self) -> QtWidgets.QWidget:
+        return self.tree_widget
+
+    def tab_title(self) -> str:
+        return "Hints"
+
+    def update_content(
```
ThanatosGit, line 33: Don't know if we have any tests but also I don't mind it for such a simple non crucial thing.
But: Minimum should be to add the return type here for mypy.

ThanatosGit (changes requested):

dyceron:
I can't seem to find the test file for this. Do I even need to test for this?

ThanatosGit:
> I can't seem to find the test file for this. Do I even need to test for this?

That's exactly what I said :p `Don't know if we have any tests but also I don't mind it for such a simple non crucial thing.`

ThanatosGit (approved):

dyceron:
> > I can't seem to find the test file for this. Do I even need to test for this?
> 
> That's exactly what I said :p `Don't know if we have any tests but also I don't mind it for such a simple non crucial thing.`

Fair enough

PR #5731, opened 2023-12-16
Add MSR exporter and activate it

ThanatosGit:


Miepee:
any specific reason why it's draft?

ThanatosGit:
Well, if you want the honest answer: If I open it without a draft and requesting review, I would get immediate review for tests, mypy etc. So, I opened it as a draft first to check it and now I get a message why it is a draft :(

ThanatosGit:
`module is installed, but missing library stubs or py.typed marker [import-untyped]` do we actually provide these for patcher? Don't think we have that for odr either

Miepee (approved):
lgtm. would be nice to get the type hints complains away, but no idea what's involved to do that.

PR #5748, opened 2023-12-19
Add tournament layouts to the auto-tracker for Prime and Echoes

EthanArmbrust:
Gives the key/artifact labels big numbers to make them easier to see on tournament restreams.
This is achieved by allowing tracker layouts to define stylesheets for labels.  

image

image

### DIFF
--- a/randovania/data/gui_assets/tracker/prime2-pixel-tournament.json
+++ b/randovania/data/gui_assets/tracker/prime2-pixel-tournament.json
@@ -176,7 +176,7 @@
             "row": 6,
             "column": 4,
             "label": "{capacity}",
-            "style": "font:18pt 'Consolas';color:white",
+            "style": "font:18pt;color:white",
             "resources": [
                 "Dark Agon Key 1",
                 "Dark Agon Key 2",
@@ -193,7 +193,7 @@
             "row": 6,
             "column": 5,
             "label": "{capacity}",
-            "style": "font:18pt 'Consolas';color:white",
+            "style": "font:18pt;color:white",
             "resources": [
                 "Dark Torvus Key 1",
                 "Dark Torvus Key 2",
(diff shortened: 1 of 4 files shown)


duncathan (commented):
I like the stylesheets change, but I don't know if I like including the new layouts. Can't you just distribute the jsons directly to racers? You'll need to do that anyway to get people using it before the next release

EthanArmbrust:
We'd need a new release to be able to use these layouts, even if it is just a 7.1.1 patch release. I'll let the tournament organizers weigh in on how they'd like these distributed.

Uncle-Reggie:
Could there perhaps be a toggle in the auto tracker options to show/hide max numbers? That way it’s not tournament specific and we could have this change available for all games with a count in their trackers.

duncathan:
> Could there perhaps be a toggle in the auto tracker options to show/hide max numbers? That way it’s not tournament specific and we could have this change available for all games with a count in their trackers. 

no, this isn't possible without a more substantial change

duncathan:
> We'd need a new release to be able to use these layouts, even if it is just a 7.1.1 patch release. I'll let the tournament organizers weigh in on how they'd like these distributed.

ah right. if there's desire for immediately usable new layouts, you could take a look at what cave story does for its puppy counter and make a layout that works similar

randovania/data/gui_assets/tracker/prime2-pixel-tournament.json
```diff
@@ -0,0 +1,244 @@
...
+            "row": 5,
+            "column": 7,
+            "resources": [],
+            "image_path": "gui_assets/tracker/pixel-icons/mp2/sky_temple_key.png"
+        },
+        {
+            "row": 6,
+            "column": 7,
+            "label": "{capacity}",
+            "style": "font:18pt 'Consolas';color:white",
```
Miepee, line 230: might be a dumb thing, but what would happen if `Consolas` is not installed as a font on the system?

Miepee (approved):

PR #5749, opened 2023-12-19
Prevent import via permalink or rdvgame of unsupported games in multiworld

ThanatosGit:


randovania/gui/multiplayer_session_window.py
```diff
@@ -778,6 +780,20 @@ async def _should_overwrite_presets(self, parameters: GeneratorParameters, perma
 
         return True
 
+    def _check_for_unsupported_games(
```
Miepee, line 783: mypy error :)

randovania/gui/multiplayer_session_window.py
```diff
@@ -787,6 +803,10 @@ async def import_permalink(self):
             return
 
         permalink = dialog.get_permalink_from_field()
+        has_unsupported_games = self._check_for_unsupported_games(permalink.parameters.presets)
+        if has_unsupported_games is not None:
+            return await has_unsupported_games
```
Miepee, line 808: i would have the function return a set of unsupported games, and have the `if len(unsupported_games) != 0` part in another function that can be called with `show_dialog_for_unsupported_games(unsupported_games)`. reads nicer IMO.

randovania/server/multiplayer/session_admin.py
```diff
@@ -217,6 +220,13 @@ def _update_layout_generation(sa: ServerApp, session: MultiplayerSession, world_
         session.save()
 
 
+def _check_preset_consistency(presets: list[VersionedPreset], session: MultiplayerSession) -> None:
+    unsupported_games = {preset.game.data.long_name for preset in presets if preset.game not in session.allowed_games}
```
Miepee, line 224: if using the above suggested aproach, then this function would simply be able to call the first `get_unsupported_games`  function

randovania/gui/multiplayer_session_window.py
```diff
@@ -778,6 +780,20 @@ async def _should_overwrite_presets(self, parameters: GeneratorParameters, perma
...
+        self, presets: list[Preset]
+    ) -> None | Coroutine[Any, Any, QtWidgets.QMessageBox.StandardButton]:
+        unsupported_games = sorted(
+            {
+                preset.game.data.long_name
+                for preset in presets
+                if not preset.game.data.defaults_available_in_game_sessions
+            }
+        )
+        if len(unsupported_games) != 0:
```
henriquegemignani, line 793: Just `if unsupported_games:`.

randovania/server/multiplayer/session_admin.py
ThanatosGit: Nah, you are mixing backend and frontend code.
Your suggestion was about frontend but this part here is the backend.

randovania/server/multiplayer/session_admin.py
```diff
@@ -243,11 +253,13 @@ def _change_layout_description(sa: ServerApp, session: MultiplayerSession, descr
             raise error.InvalidActionError("One of the worlds has undefined order field.")
 
         try:
-            description = LayoutDescription.from_bytes(
-                description_bytes, presets=[VersionedPreset.from_str(world.preset) for world in worlds]
-            )
+            presets = [VersionedPreset.from_str(world.preset) for world in worlds]
```
henriquegemignani, line 256: These checks and the error below are pointless.when uploading a Layout Description, it's encoded in a special binary format that excludes the presets. You can see above how they're fetched from the db.

We'd need to include some sort of hash of the preset to double check, but it means a data change sadly.
What can be done is check if the game matches though. That data is redundant and saved in both places.

henriquegemignani (changes requested):

randovania/server/multiplayer/session_admin.py
ThanatosGit: I assume, due to what already has been done, we already have incompatible presets in the database itself. So, the check is not really pointless even if it fetches it from the db?

The error below is not pointless. It's a little bit inaccurate, I admit, but if the preset and game is not matching it throws the `ValueError` from `LayoutDescription.from_bytes` like:
```
randovania\layout\game_patches_serializer.py", line 159, in decode_single
    raise ValueError(f"Expected '{game.game.value}', got '{game_modifications['game']}'")
```
and with my change it is catched.

Though it is already prevented by the changes in the randovania frontend to not even do that but I thought it is nicer to not only rely on the "frontend being nice to the backend".

Honestly, as the description is sent in a byte format, I haven't seen a nice way to check for a matching game in preset and layout, that's why I just catched the error it will throw.

randovania/server/multiplayer/session_admin.py
Miepee: whoops, didnt notice. thanks

randovania/server/multiplayer/session_admin.py
henriquegemignani: ah yes, the exception capturing is good. I was originally writing the message was wrong though but then saw the other issue

The layout is also saved in binary in the db, to save some space (99% of the db size was the jsons, and having 10+ copies of the preset was pretty wasteful) so the mismatch can be there. But we can't do much so we cry I guess.

The current api has a design error in that the server must trust the client to upload a layout whose presets match the current session, with no way to confirm.

At least checking the game matches is helpful, and the client refusing to do it also helps.

ThanatosGit: > At least checking the game matches is helpful, and the client refusing to do it also helps.

That's what this PR (hopefully) already does.
I think I could use `LayoutDescription.from_bytes` without the presets and then compare if layout and preset match instead of calling it with the presets and catch the error it will throw.

So, you want me to change something there or leave it?

The full crazy blown up PR would probably to fix this:
> The current api has a design error in that the server must trust the client to upload a layout whose presets match the current session, with no way to confirm.

ThanatosGit: ^ @henriquegemignani 🙋‍♂️

henriquegemignani: Calling `_check_preset_consistency` here right now is not useful. These presets are fetched from the DB so they should already be valid. Could keep the call, but can you at least add a `FIXME` note so it's clear this is wrong?

ThanatosGit: It was intended for the fact that we already have invalid presets in the db, don't we?
The thing which brought the whole thing, is still part of the db or was it removed?
Otherwise I can of course just remove it.

henriquegemignani: If we were to care about the existing wrong values, we'd be doing a data migration to fix. I don't really care about these few that are wrong.

randovania/gui/multiplayer_session_window.py
ThanatosGit: Changed it. Is this better?

randovania/server/multiplayer/session_admin.py
ThanatosGit: Alright. Removed it and just left the two invalid action things.

randovania/gui/multiplayer_session_window.py
```diff
@@ -778,6 +779,20 @@ async def _should_overwrite_presets(self, parameters: GeneratorParameters, perma
 
         return True
 
+    def _check_for_unsupported_games(self, presets: list[Preset]) -> list[str]:
+        unsupported_games = sorted(
+            {
+                preset.game.data.long_name
+                for preset in presets
+                if not preset.game.data.defaults_available_in_game_sessions
```
ThanatosGit, line 787: One thing which I just realized now:
Should this use the session's `allowed_games` ? It should only matter in a rare edge case. It would require to have an additional game in that list and only added it via permalink / rdvgame instead of creating it "normally" (because the latter would not work if it has not the `defaults_available_in_game_sessions` set to true)

ThanatosGit:
ping

randovania/gui/multiplayer_session_window.py
henriquegemignani: Would be good, yeah. Can leave a TODO, but I bet `self` has the session data which has the list?

henriquegemignani (approved):

randovania/gui/multiplayer_session_window.py
ThanatosGit: I don't like adding TODOs. I changed it. Had to change some tests because they now require to have `allowed_games` in the mock and provide something sortable for the `long_name`.
If it's fine, feel free to hit the merge button.

(Funny that the CHANGELOG had no conflict and was added to the old release which I saw last minute).

henriquegemignani (approved):

PR #5761, opened 2023-12-22
AM2R: Add starting items memo

### LINKED ISSUE #5111: AM2R: Add text on first load to display starting items
image
image

Miepee:
Fixes #5111 
Not sure if there's a nicer way to do something like:
```py
return {
...
if starting_memo:
  "starting_memo": starting_memo
}

### DIFF
--- a/randovania/games/am2r/exporter/patch_data_factory.py
+++ b/randovania/games/am2r/exporter/patch_data_factory.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from random import Random
 from typing import TYPE_CHECKING
 
-from randovania.exporter import pickup_exporter
+from randovania.exporter import item_names, pickup_exporter
 from randovania.exporter.hints import credits_spoiler, guaranteed_item_hint
 from randovania.exporter.patch_data_factory import PatchDataFactory
 from randovania.game_description.assignment import PickupTarget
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: randovania/games/am2r/exporter/patch_data_factory.py [hunk 1]
```diff
@@ -188,11 +199,15 @@ def _create_starting_location(self) -> dict:
...
     def _create_hash_dict(self) -> dict:
-        return {
+        returnDict = {
             "word_hash": self.description.shareable_word_hash,
             "hash": self.description.shareable_hash,
             "session_uuid": str(self.players_config.get_own_uuid()),
         }
+        starting_memo = self._create_starting_popup(self.patches)
+        if starting_memo:
+            returnDict["starting_memo"] = starting_memo
```
ThanatosGit, line 209: That's a type error because your `returnDict` is inferred to be a `dict[str, str]` by the definition in line 202 and following.

ThanatosGit (changes requested):
> Not sure if there's a nicer way to do something like:
> 
> ```python
> return {
> ...
> if starting_memo:
>   "starting_memo": starting_memo
> }
> ```
If your patcher is fine with it being `None` / `null` in the JSON, you could always assign it directly at the definition in line 202ff.

randovania/games/am2r/exporter/patch_data_factory.py
Miepee: hmmm how so? the return type for the function is just a dict. i don't have a `dict[str, str]` explictily mentioned anywhere.

Miepee:
> If your patcher is fine with it being None / null in the JSON, you could always assign it directly at the definition in line 202ff.

oh. good to know.

ThanatosGit (approved):

PR #5768, opened 2023-12-23
AM2R: Add Unlocked save station option.

Miepee:
Planning to later also do another configuration change for am2r (exclude certain doors in labs from being shuffled), not sure how i should handle preset migrations there. Should I just stuff the next thing also into v70? or should i make a v71 for it?

### DIFF
--- a/randovania/gui/ui_files/preset_am2r_patches.ui
+++ b/randovania/gui/ui_files/preset_am2r_patches.ui
@@ -204,14 +204,14 @@
            <item>
             <widget class="QCheckBox" name="blue_save_doors_check">
              <property name="text">
-              <string>Force blue Save Station Doors</string>
+              <string>Unlock Save Station Doors</string>
              </property>
             </widget>
            </item>
            <item>
             <widget class="QLabel" name="blue_save_doors_label">
              <property name="text">
-              <string>Ensures all Save Station doors are blue doors, even with door lock rando enabled.</string>
+              <string>Ensures all Save Station doors are normal (blue) doors, even with door lock rando enabled.</string>
              </property>
             </widget>
            </item>
--- a/randovania/gui/ui_files/preset_am2r_patches.ui
+++ b/randovania/gui/ui_files/preset_am2r_patches.ui
@@ -44,7 +44,7 @@
          <x>0</x>
          <y>0</y>
          <width>745</width>
-         <height>944</height>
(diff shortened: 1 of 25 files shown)


randovania/gui/ui_files/preset_am2r_patches.ui
```diff
@@ -201,6 +201,20 @@
...
+            <widget class="QCheckBox" name="blue_save_doors_check">
+             <property name="text">
+              <string>Force blue Save Station Doors</string>
+             </property>
+            </widget>
+           </item>
+           <item>
+            <widget class="QLabel" name="blue_save_doors_label">
+             <property name="text">
+              <string>Ensures all Save Station doors are blue dors, even with door lock rando enabled.</string>
```
XenoWars, line 214: dors typo, should be doors.

XenoWars (changes requested):

randovania/gui/ui_files/preset_am2r_patches.ui
```diff
@@ -201,6 +201,20 @@
...
+            <widget class="QCheckBox" name="blue_save_doors_check">
+             <property name="text">
+              <string>Force blue Save Station Doors</string>
+             </property>
+            </widget>
+           </item>
+           <item>
+            <widget class="QLabel" name="blue_save_doors_label">
+             <property name="text">
+              <string>Ensures all Save Station doors are blue doors, even with door lock rando enabled.</string>
```
ThanatosGit, line 214: I'm fine if you want to call it internally "Blue door" but AM2R has more than one door in the color of blue AFAIK.
Also it is called "Normal door" in the Door lock settings. If calling it blue door is not already ambiguous, it is at least questionable when calling it by a different name in another setting. 
The same applies to the changelog.

ThanatosGit (changes requested):

randovania/gui/ui_files/preset_am2r_patches.ui
Miepee: whoops, yeah, that shouldnt happen. thanks.

Miepee: Altho I just noticed that Prime has the same issue xP
grafik
grafik
But that's up to change in a later PR i guess.

ThanatosGit (approved):

PR #5778, opened 2023-12-25
AM2R: add extra data for blue lab doors

Miepee:
Part of the future "force lab doors to be blue" option. PRing this seperately, so that the future PR gets easier to review


JeffGainsNGames:
Are the crystal caves portion included due to `doors anywhere`?

Miepee:
Yup. would suck pretty hard otherwise, to kill queen, and then finding out you need to reload because there's a genesis door in your way.

JeffGainsNGames (approved):

PR #5782, opened 2023-12-26
Merge stable to main

Miepee:
WIP, will need to remove stuff / solve conflicts


Miepee:
Hopefully fixed it, really unsure about the DB exporting command.

randovania/cli/commands/export_db_videos.py
```diff
@@ -99,21 +99,26 @@
...
-</style><a href=https://www.youtube.com/embed/{}?start={}&autoplay=1>
-<img src=https://img.youtube.com/vi/{}/hqdefault.jpg alt='vid'><span>▶️</span></a>"
-frameborder="0"
-allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
-allowfullscreen
-title="vid"
-></iframe>
+        <p><i> {} </i></p>
+        <iframe
+        width="560"
```
toasterparty, line 104: I can't find a reason why this would be changed. Looking at the db output of stable and main, I prefer the larger HTML frames (720x410)

Everything else doesn't seem to matter

toasterparty (changes requested):

randovania/cli/commands/export_db_videos.py
Miepee: do you want me to just revert width/height, or the whole html snippet?

toasterparty: the whole file please

toasterparty (approved):

PR #5790, opened 2023-12-30
fix 711 changelog date

Miepee:


ThanatosGit (approved):
German umlaute in branch names are like:
image

PR #5799, opened 2024-01-02
Diffusion Abuse from below for Wide Block in Cataris

Logarhythmic:
Added a Diffusion Abuse trick for a Wide Block in Cataris.
Added a video for the trick.


randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -2814,7 +2814,12 @@ Extra - asset_id: collision_camera_040
   > Dock to Thermal Device Room North
       Can Slide
   > Event - Push Wide Beam Block
-      Wave Beam and Can Slide and Push Wide Beam Block
+      All of the following:
+          Wave Beam and Can Slide
+          Any of the following:
+              Push Wide Beam Block
+              # Diffusion Abuse: https://youtu.be/hJYbMcLUTX0
+              Wide Beam and Diffusion Abuse (Intermediate)
```
Tyranisaur, line 2822: Having this as diffusion abuse is all kinds of weird. We should introduce a new trick for using charge beam with a melee counter.

Tyranisaur (changes requested):

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -2814,7 +2814,12 @@ Extra - asset_id: collision_camera_040
   > Dock to Thermal Device Room North
       Can Slide
   > Event - Push Wide Beam Block
-      Wave Beam and Can Slide and Push Wide Beam Block
+      All of the following:
+          Wave Beam and Can Slide
```
XenoWars, line 2818: Wave + Charge doesn't actually need the slide. You would only need Slide if you are using the melee charge shot.

XenoWars (changes requested):

randovania/games/dread/logic_database/Cataris.txt
XenoWars: The new trick would only be used here and for Artaria's. Is two uses enough to add a new trick? Different story if we find other uses for it.

randovania/games/dread/logic_database/Cataris.txt
Logarhythmic: Do you not need a slide to get into that area of the room?

Logarhythmic: image

XenoWars: wave + charge you can just shoot from directly below the block, whereas without charge you'll need to slide in to melee the enemy.

Logarhythmic: updated

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -2815,11 +2815,11 @@ Extra - asset_id: collision_camera_040
       Can Slide
   > Event - Push Wide Beam Block
       All of the following:
-          Wave Beam and Can Slide
+          Wave Beam
           Any of the following:
               Push Wide Beam Block
```
XenoWars, line 2820: one last thing, this can be changed to just charge beam since you dont need wide for it.

XenoWars (changes requested):

randovania/games/dread/logic_database/Cataris.txt
duncathan: leave the wide req. there's no telling whether later patcher changes will introduce the requirement again and it shouldn't make too much difference right now

randovania/games/dread/logic_database/Cataris.txt
Logarhythmic: I asked before making the PR and the consensus was to keep it as Diffusion Abuse.

randovania/games/dread/logic_database/Cataris.txt
Logarhythmic: I think since the Beam Split bug isn't intentional and there is effort to fix it, the logic should be ignorant of the bug.

randovania/games/dread/logic_database/Cataris.txt
Tyranisaur: To me it didn't look like that discussion ended with a consensus to file this under Diffusion Abuse. But if that's the route we're taking, can we at least put this at advanced so that it's above the pre existing usage?

Tyranisaur: This is in fact not the exact same technique as the existing diffusion abuse trick. In the existing one, we use this setup to shoot an enemy and let the diffusion effect push a wide block. In this instance there is no diffusion involved, hence by reluctance to put this under diffusion abuse. Furthermore, the beginner diffusion abuse instances use the diffusion effect to trigger buttons, so the trick doesn't exclusively relate to wide beam blocks to begin with.

Up until now, the hypermode diffusion abuse trick was the only instance of using a melee counter for the purpose of shooting a charge beam without having a charge beam. If we introduce a trick for precisely that purpose, I think the old hypermode diffusion abuse trick should use both diffusion abuse and the new trick, which could be called something like "Charge from Counter". I'm not sure about how the difficulty load should be distributed between the two tricks, but I think it makes sense to take the diffusion abuse down from hypermode, while the primary difficulty level of each of these tricks can be placed on the charge from counter trick.

Also worth mentioning, by the logic that pushing wide beam blocks in exotic ways falls under diffusion abuse, the one next to the DNA hint station could also have been placed under that trick, but we have it under pseudo wave, even though it uses the wave beam.

I think it's worth it to add Charge from Counter as a new trick to accurately express the mechanisms at play instead of shoehorning different mechanics into the the same trick because they are used in tandem in some instances. If we discover more usages for using charge beam through a counter then I think this will look more clean.

Though, like I said, I can accept putting this as diffusion abuse advanced.

randovania/games/dread/logic_database/Cataris.txt
Tyranisaur: For reference, I think this would have been added as viable with just charge and wave beam, if the data visualizer map had shown the wide beam block in its correct location. At the time the split beams were added, we thought that pushing wide beam blocks with wave beam was ok from a sharp angle, so in a bunch of sitations, that was added to logic, just not here. The map looks like this 
image while it looks like this ingame 
image
However, with what we know now, I think it's for the better to just leave it as is and wait for a fix for the wideless shenanigans, as it's probably overall better for the item balance if it requires wide beam.

randovania/games/dread/logic_database/Cataris.txt
Logarhythmic: I like Counter Charge better, everyone loves alliteration. I think it probably merits its own trick, but in my experience I've been discouraged to give tricks that only get one or two usage in the database. I bumped it to advanced for now.

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed: Burenia Hub to Dairon: Getting the item in the fan with only Flash Shift now requires at least one Flash Shift Upgrade as well, and also only requires Intermediate movement (instead of Advanced).
 - Changed: Main Hub Tower Middle: Climbing out of the water from Left of Central Grapple Block without any items now requires Advanced Movement, up from Intermediate.
 
+##### Cataris
+
+- Added: The Wide Beam Block in Dairon Transport Access can now be traversed with a Diffusion Abuse trick from below.
```
Tyranisaur, line 33: This should be moved to version 7.3.0

Tyranisaur (changes requested):

CHANGELOG.md
Logarhythmic: Updated.

Tyranisaur: You didn't remove the one from 7.2.

Tyranisaur (approved):

CHANGELOG.md
Logarhythmic: My mistake, it was re-added when I pulled

PR #5804, opened 2024-01-03
MSR - Add Missile Launcher

dyceron:
- Adds Missile Launcher as an pickup. Starter Preset starts will still start with it Missiles
- `Nerf Super Missiles` is now visible again in `Optional Patches`
- Updated some descriptions in `Optional Patches`
- Utilizes temporary resources for Mains/Tanks
- Some minor logic cleanup
- Updated templates


randovania/games/samus_returns/exporter/patch_data_factory.py
```diff
@@ -42,28 +42,23 @@
...
         return item.extra["item_id"]
     except KeyError as e:
         raise KeyError(f"{item.long_name} has no item ID.") from e
 
 
-def get_export_item_id_for_item(item: ResourceInfo) -> str:
-    assert isinstance(item, ItemResourceInfo)
-    if "item_export_id" in item.extra:
-        return item.extra["item_export_id"]
-    return get_item_id_for_item(item)
```
ThanatosGit, line 57: Was there a reason to change this? My intention was to differentiate between the exported id for the patcher + temporary locked ressources and the item id which needs to be tracked if you want to know how many ammo you have.

randovania/games/samus_returns/logic_database/Area 5 - Exterior.txt
```diff
@@ -398,12 +398,9 @@ Extra - asset_id: collision_camera_008
...
   > Door to Tower
-      Any of the following:
-          Lay Bomb
-          Lightning Armor and Morph Ball
-          All of the following:
-              Lay Power Bomb
-              Baby Metroid or Power Bomb ≥ 2
+      All of the following:
+          Morph Ball
+          Lightning Armor or Bomb or Power Bomb ≥ 2
```
ThanatosGit, line 403: Was the removal of "Baby Metroid" intentional?

randovania/games/samus_returns/logic_database/Area 6.txt
```diff
@@ -1229,13 +1225,9 @@ Extra - asset_id: collision_camera_Hazard_End_A
...
       All of the following:
           # Completing the puzzle
-          Grapple Beam and Morph Ball and Screw Attack and Shoot Any Missile
+          Grapple Beam and Morph Ball and Power Bomb ≥ 3 and Screw Attack and Shoot Any Missile
           # Traversing the crumble blocks
           Phase Drift or Spider Ball or Movement (Beginner)
-          All of the following:
-              # Breaking the Bomb blocks
-              Lay Power Bomb
-              Bomb or Power Bomb ≥ 3
```
ThanatosGit, line 1238: The bomb requirement is gone here?

randovania/games/samus_returns/logic_database/Area 7.txt
```diff
@@ -163,7 +157,7 @@ Extra - asset_id: collision_camera_005
       All of the following:
           Grapple Beam
           Any of the following:
-              Power Bomb ≥ 2 and Lay Power Bomb
+              Power Bomb ≥ 2
               All of the following:
                   Lay Bomb
                   Plasma Beam or Shoot Beam Burst
```
ThanatosGit, line 163: This should be fulfilled with Grapple and 2 Power Bombs but is missing Morph Ball to actually use the Power Bombs.

randovania/games/samus_returns/logic_database/header.txt
ThanatosGit, line 1: The removal of the huge logic blocks for zeta metroids(+) was intentional?

ThanatosGit (changes requested):

randovania/games/samus_returns/exporter/patch_data_factory.py
dyceron: To me, it looks like it does what was done before but without an extra function call. It still produces a correct patcher file which is why I went with it.

randovania/games/samus_returns/logic_database/Area 5 - Exterior.txt
dyceron: The Baby is still a valid path to get to the item. All the Baby did on the return path was save a power pomb,bomb, which is kinda unnecessary.

randovania/games/samus_returns/logic_database/Area 6.txt
dyceron: Same as the above comment. Having Bombs saves one power bomb. No need to account for two items to save one ammo

randovania/games/samus_returns/logic_database/header.txt
dyceron: Yes. It was partly to simplify, and partly because that section is for fighting without missiles, so listing missiles there defeats the purpose. I could have moved those paths out of that group, but eh.

randovania/games/samus_returns/logic_database/Area 7.txt
dyceron: Good catch. Though I guess technically this is fine since you'd need Morph to get here anyway.

randovania/games/samus_returns/exporter/patch_data_factory.py
ThanatosGit: `get_item_id_for_item` was able to return the `item_capacity_id` even if it had an `item_export_id` e.g. if you want to auto track the amount of usable missiles which is`ITEM_WEAPON_MISSILE_MAX`. In that case you would need the `item_capacity_id` even an `item_export_id` exists.
Now, this function is not usable anymore for it because it will always return the `item_export_id` of such items.
If you want to go with this change and we don't want to track it anyways, you could completly remove the `item_export_id` from the db and place the string of it into the `item_capacity_id`.

dyceron: Is there a benefit to keeping track? I figured it was ok to make this change.

ThanatosGit: Well, uh, the benefit of keeping track is that you keeping track of it xD. I mean, it's the purpose of an auto tracker

ThanatosGit (approved):

PR #5806, opened 2024-01-04
AM2R: Wrap starting memo text

Miepee:
Makes the starting text wrap in AM2R. No changelog entry needed, as the starting memo feature was introduced in dev builds after the last stable release.
Before:
grafik

After:
grafik

### DIFF
--- a/randovania/games/am2r/exporter/patch_data_factory.py
+++ b/randovania/games/am2r/exporter/patch_data_factory.py
@@ -182,9 +182,21 @@ class AM2RPatchDataFactory(PatchDataFactory):
     def _create_starting_popup(self, patches: GamePatches) -> dict | None:
         extra_items = item_names.additional_starting_equipment(patches.configuration, patches.game, patches)
         if extra_items:
+            i = 0
+            items_with_length = [""]
+            # Go through the items, so that we can wrap them
+            for item in extra_items:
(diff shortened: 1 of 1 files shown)


randovania/games/am2r/exporter/patch_data_factory.py
```diff
@@ -182,9 +182,21 @@ def _create_room_dict(self) -> dict:
     def _create_starting_popup(self, patches: GamePatches) -> dict | None:
         extra_items = item_names.additional_starting_equipment(patches.configuration, patches.game, patches)
         if extra_items:
+            i = 0
+            items_with_length = [""]
+            # Go through the items, so that we can wrap them
+            for item in extra_items:
+                if len(items_with_length[i] + item) < 60:
+                    if items_with_length[i] != "":
+                        items_with_length[i] = items_with_length[i] + ", " + item
```
ThanatosGit, line 191: This creates a string with 2 characters more compared to the check done in line 189 but I assume the 2 extra characters are no problem?

Miepee: yeah the extra 2 characters are fine. The actual cutoff is more at like 70ish character, but because i wanted to still leave a bit of space, i just took it down to 60 to have some extra room there

ThanatosGit (approved):

PR #5808, opened 2024-01-05
Rebrand the Randovania icon

Miepee:
Huge credits to @Uncle-Reggie for designing these!
A seperate PR that renames the icons to not reference STKs anymore will come out after the next stable release.


henriquegemignani (approved):
Been wanting to do this for a while! :D

PR #5848, opened 2024-01-14
Use qdesktopservices instead of webbrowser

### LINKED ISSUE #5843: Discord login sometimes fails on Linux
Absolutely no clue what causes it, it doesn't print anything useful to the console
Seems to always fail as root, but that should be ignored anyway
Normal binaries sometimes fail and print different things for different users.
Flatpak seems to always work

Seems like some dependency issue?

Miepee:
Hopefully fixes #5843 
But this didn't get any public testing yet. It didn't regress for me locally at least, so there's that.
Dunno if we should get testing now and don't merge on issues, or merge first then get testing and then revert if we have any.

### DIFF
--- a/randovania/gui/lib/qt_network_client.py
+++ b/randovania/gui/lib/qt_network_client.py
@@ -4,11 +4,10 @@ import asyncio
 import datetime
 import functools
 import json
-import webbrowser
 from typing import TYPE_CHECKING
 
 import PySide6
-from PySide6 import QtCore, QtWidgets
+from PySide6 import QtCore, QtGui, QtWidgets
 from PySide6.QtCore import Signal
 
 import randovania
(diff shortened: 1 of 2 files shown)


Miepee:
This has received testing and works fine. It indeed seems to be more reliable than webbrowser.

randovania/gui/lib/qt_network_client.py
```diff
@@ -157,7 +156,7 @@ async def login_with_discord(self):
 
         sid = await self.server_call("start_discord_login_flow")
         url = self.configuration["server_address"] + f"/login?sid={sid}"
-        webbrowser.open(url)
+        QtGui.QDesktopServices.openUrl(QtCore.QUrl(url))
```
ThanatosGit, line 159: The thing returns a bool for success / failure.
I think, we should check for that and otherwise throw an error explicitly mentioning that it failed instead of silently fail like it was done and happened now.

ThanatosGit (changes requested):

randovania/gui/lib/qt_network_client.py
Miepee: We could, but dunno if it's that good of an idea:
> **Warning**: A return value of `true` indicates that the application has successfully requested the operating system to open the URL in an external application. The external application may still fail to launch or fail to open the requested URL. This result will not be reported back to the application.

But i guess better to sometimes show an error, than to never show an error?

henriquegemignani: That's preferable to returning an error when the URL was in-fact opened.

ThanatosGit (approved):

henriquegemignani (approved):

PR #5852, opened 2024-01-15
Fix pickup filtering for items that start the same

Miepee:
Before:
grafik
After:
grafik
grafik
Shouldn't cause any performance loss hopefully, as i'm pretty sure that setFixedString was also using a regex in the background.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [7.3.0] - 2024-02-??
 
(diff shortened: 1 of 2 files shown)


henriquegemignani:
`void QSortFilterProxyModel::setFilterFixedString(const [QString](https://doc.qt.io/qt-6/qstring.html) &pattern)`
> Note: Calling this method updates the regular expression, thereby breaking the binding for [filterRegularExpression](https://doc.qt.io/qt-6/qsortfilterproxymodel.html#filterRegularExpression-prop). However it has no effect on the [filterCaseSensitivity](https://doc.qt.io/qt-6/qsortfilterproxymodel.html#filterCaseSensitivity-prop) bindings.

henriquegemignani (approved):

Miepee:
I come from an old game maker background, where the docs were regurlaly incorrect xP

PR #5859, opened 2024-01-16
AM2R MW: fix progressives with random starting items

Miepee:
When geting random starting items, the game gives you the resource directly (i.e. `Hi-Jump`). Thus mixing between the normal resource and the progressive, which we need to account for.

### DIFF
--- a/randovania/game_connection/connector/am2r_remote_connector.py
+++ b/randovania/game_connection/connector/am2r_remote_connector.py
@@ -184,10 +184,16 @@ class AM2RRemoteConnector(RemoteConnector):
 
         for progressive_name, actual_items in progressives:
             quantity = inventory_dict.pop(progressive_name, 0)
-            for index in range(quantity):
+            if quantity == 0:
+                continue
+            index = 0
+            while True:
                 if index >= len(actual_items):
                     break
-                inventory_dict[actual_items[index]] += 1
+                actual_item = actual_items[index]
+                if actual_item not in inventory_dict:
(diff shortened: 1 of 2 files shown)


randovania/game_connection/connector/am2r_remote_connector.py
```diff
@@ -184,10 +184,16 @@ def new_inventory_received(self, new_inventory: str) -> None:
 
         for progressive_name, actual_items in progressives:
             quantity = inventory_dict.pop(progressive_name, 0)
-            for index in range(quantity):
+            if quantity == 0:
+                continue
+            index = 0
+            while True:
```
henriquegemignani, line 190: `for index in itertools.count():`

henriquegemignani (changes requested):

henriquegemignani:
can you be more exhaustive with the testing? Both test cases end up with both Hi-Jump and Space Jump, for example.

tbh I don't quite follow why this code works

Miepee:
I added a comment to hopefully explain it better.

duncathan:
I feel like you should probably just give the correct resource for the starting items? this seems awfully messy

tbh I don't understand why you distinguish between progressive and non-progressive items in the inventory to begin with

Miepee:
> It's kind of nonsense to keep track of the names of the Pickups you've collected

This made me realize on how to fix it. I can just overwrite the name that gets added to the inventory, so i'll just do that there for the progressive stuff, and close this PR to not make this section more unreadable.

PR #5860, opened 2024-01-16
Disable "Quick Generate" and "Generate without retry" if preset is invalid

ThanatosGit:
Also added type annotations. Just need to wait for mypy if they are correct and I wasn't bothered to add a test😶
Before:
image
After:
image
Before:
image
After:
image


henriquegemignani:
I never bothered by the "Generate without retry" because that's a dev button :p

henriquegemignani (approved):

PR #5865, opened 2024-01-16
Make drop descriptions clearer

Miepee:
IDK if it's actually clearer now. Sure hope so tho.
Could look weird with negatives, aka `Missiles refilled by -10`.
Fixes https://github.com/randovania/YAMS/issues/138


henriquegemignani (approved):

henriquegemignani:
If you really want better text for negative, you could detect when it's negative and use a different string

Miepee:
TBH i don't really care about negatives, as long as they don't sound absolutely horrible.

PR #5879, opened 2024-01-19
MSR - Add misc resource for Door Lock Rando

dyceron:
Preparing for the future and this fixes a current logic issue in Area 2 Entrance, so win-win.

- Adds misc resource for Door Lock Rando
- Fixes node connections to account for resouces
- Fixes broken logic in Area 2 Entrance

### DIFF
--- a/randovania/games/samus_returns/generator/bootstrap.py
+++ b/randovania/games/samus_returns/generator/bootstrap.py
@@ -64,6 +64,9 @@ class MSRBootstrap(MetroidBootstrap):
             if getattr(configuration, name):
                 enabled_resources.add(index)
 
+        if configuration.dock_rando.is_enabled():
+            enabled_resources.add("DoorLocks")
+
         return enabled_resources
 
     def event_resources_for_configuration(
--- a/randovania/games/samus_returns/generator/bootstrap.py
+++ b/randovania/games/samus_returns/generator/bootstrap.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING
 from randovania.game_description.db.pickup_node import PickupNode
 from randovania.games.samus_returns.generator.pool_creator import METROID_DNA_CATEGORY
 from randovania.games.samus_returns.layout import MSRConfiguration
+from randovania.layout.base.dock_rando_configuration import DockRandoMode
 from randovania.layout.exceptions import InvalidConfiguration
 from randovania.resolver.bootstrap import MetroidBootstrap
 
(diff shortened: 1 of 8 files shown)


randovania/games/samus_returns/logic_database/Area 2 - Entrance.txt
```diff
@@ -271,7 +275,7 @@ Extra - asset_id: collision_camera_005
           Lightning Armor or Use Spider Ball
           Movement (Intermediate) and Normal Damage ≥ 99
   > Door to Transport to Area 2 Exterior (Top)
-      Trivial
+      After Area 2 (Entrance) - Transport to Area 2 Exterior Unlock Door or Enabled Door Lock Randomizer
```
ThanatosGit, line 278: Like I said in the discord, I don't think that is the right way to do it. An enabled door lock randomizer doesn't force that this door is changed.
I'm not sure if this counts for individual door locks but for sure it does for type door lock rando. (where I guess it's never changed because of one sided door?)
In that case you still have to collect the event even with door lock randomizer enabled.

randovania/games/samus_returns/generator/bootstrap.py
```diff
@@ -64,6 +64,9 @@ def _get_enabled_misc_resources(
             if getattr(configuration, name):
                 enabled_resources.add(index)
 
+        if configuration.dock_rando.is_enabled() or False:
```
ThanatosGit, line 67: Can't see the purpose of `or False`.

ThanatosGit (changes requested):

randovania/games/samus_returns/generator/bootstrap.py
dyceron: Unless it was a fluke, when I didn't do this codecov/patch failed with 0% coverage somehow

randovania/games/samus_returns/logic_database/Area 2 - Entrance.txt
dyceron: Yeah this is fine for regular door lock by brekmaks for types. Maybe there's a way to specify which mode is used? Or we just check for door lock and if true, change both sides to be power regardless

randovania/games/samus_returns/generator/bootstrap.py
ThanatosGit: It is still the case and not a fluke. There is only one coverable line (ignoring the condition because it has no `else` part) for the changes and that line is not covered which makes it 0 % :D. We don't have a test for the generator of MSR with the door lock option activated. Hence the line cannot be covered by a test.
I would suggest to add tests when door lock is really added and ignore it for now.
But remove the nonsense `or False` part :p

ThanatosGit (approved):

PR #5888, opened 2024-01-20
Don't try to add duplicate entries to DB editor

### LINKED ISSUE #4506: DB editor lets you add multiple resources with the same key
- open db editor
- add an item called "foo"
- add another item called "foo"

editor will now display crashes when trying to save/close

Miepee:
I thought about display a warning popup with `Entry already exists!`, but then i'd need to figure how to properly await that function without making the setData function async.
Thus i lowered my expectations until they were fully met.
Also apparently the whole DB editor just doesnt have any tests at all????
Fixes #4506

### DIFF
--- a/randovania/gui/docks/resource_database_editor.py
+++ b/randovania/gui/docks/resource_database_editor.py
@@ -117,6 +117,9 @@ class ResourceDatabaseGenericModel(QtCore.QAbstractTableModel):
                     return True
             else:
                 if value:
+                    all_items = self._get_items()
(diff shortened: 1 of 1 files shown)


Miepee:
Since this only affects things when running from source, does this need a changelog entry?

henriquegemignani (approved):
I'd somehow color the new row red to indicate an error. There's probably a role dedicated for these things?
But I'm fine with what you did.

In general Qt things don't have much testing. And DB editing in particular is funky because the DB classes are supposedly immutable.

PR #5900, opened 2024-01-22
Multi world icons (plus small visual tweaks)

DualVission:
image
image
image


randovania/gui/main_window.py
ThanatosGit, line 1: Please also have a look at all the mypy errors and see if they can be fixed.

ThanatosGit (changes requested):

duncathan:
> Sure thing, I'll get on those this evening. I'm not sure what mypy errors are occurring though. Where you thinking a named tuple or a dictionary. If other banners are added in the future, a dictionary would be more flexible.

a NamedTuple with proper type hints is preferable

duncathan (changes requested):
changelog

randovania/interface_common/options.py
```diff
@@ -404,6 +406,14 @@ def dark_mode(self) -> bool:
     def dark_mode(self, value: bool):
         self._edit_field("dark_mode", value)
 
+    @property
+    def show_multiworld_banner(self) -> bool:
+        return _return_with_default(self._show_multiworld_banner, lambda: False)
+
+    @show_multiworld_banner.setter
+    def show_multiworld_banner(self, value: bool):
```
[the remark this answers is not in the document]
DualVission: I am unclear what this means. It matches all that is around it.

randovania/gui/main_window.py
DualVission: Beyond Cave Story's patcher not loading and an image having ICC color issues, I am not getting errors to appear in the terminal.

DualVission:
Added:
- Multiworld icon (currently represented by Aether and SR388 being divided by the RDV arrows)
- Banner icon
- Necessary functions to drive toggling visibility
- GameQtElements tuple
Changed:
- Fixed tooltip support on dark theme
- Moved old elements to new container widgets
- Modified on-hover behaviors
- Dictionary combined into single stream
Removed:
- Effect deletion upon roll-off

randovania/gui/main_window.py
ThanatosGit: Mypy is a static type checker, so this won't result in any runtime errors in the terminal when executing randovania.
Unless you mean you executed mypy, then it should give you some messages.
Check the "Files changed" tab here in Github to see the annotations.
(Also see the comment, I will write above in a few seconds)

randovania/interface_common/options.py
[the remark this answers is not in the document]
ThanatosGit: To be more clear on this specific one: See the method above this one which explicitly defined the return type as with `-> bool`.
This one here has no return, so it needs to be `-> None`.

randovania/gui/main_window.py
```diff
@@ -190,18 +204,44 @@ def __init__(
...
+            multibanner.setGraphicsEffect(multibanner.color_effect)
+            multibanner.move(0, 2)
+
+            multiicon = QtWidgets.QLabel(pack_tile)
+            multiicon.setPixmap(QtGui.QPixmap(os.fspath(multiworld_img_path)))
+            multiicon.setScaledContents(True)
+            multiicon.setFixedSize(bannerSize / 2, bannerSize / 2)
+            multiicon.setVisible(game.data.defaults_available_in_game_sessions)
+            multiicon.move(bannerSize / 4, 2 + bannerSize / 4)
+            if multiicon.isVisible:
```
henriquegemignani, line 230: Isn't this a function?

randovania/gui/main_window.py
```diff
@@ -67,6 +67,13 @@ class LayoutWithPlayers(typing.NamedTuple):
     players: list[str] | None
 
 
+class GameQtElements(typing.NamedTuple):
+    logo: QtWidgets.QLabel
+    multibanner: QtWidgets.QLabel | None
```
henriquegemignani, line 72: I'd call this `multi_banner`

henriquegemignani (changes requested):

randovania/gui/main_window.py
DualVission: I'm not sure I understand. Is what a function?

Miepee: `isVisible`

Miepee: According to qt documentation at least, it's shown as a function
https://doc.qt.io/qt-5/qwidget.html#visible-prop

duncathan (changes requested):
please add an entry in changelog.md

randovania/gui/main_window.py
DualVission: Since that change may affect behaviors, I will hold off until I get home with a computer.

DualVission:
I think that is everything. The two MyPy errors still occurring are concerning truth values and object types not having attributes where not in use.

henriquegemignani (approved):

randovania/gui/main_window.py
```diff
@@ -190,18 +206,56 @@ def __init__(
...
+                multi_banner.setGraphicsEffect(color_effect)
+                multi_banner.move(0, 2)
+
+                multi_icon = QtWidgets.QLabel(pack_tile)
+                multi_icon.setPixmap(QtGui.QPixmap(os.fspath(multiworld_img_path)))
+                multi_icon.setScaledContents(True)
+                multi_icon.setFixedSize(int(bannerSize / 2), int(bannerSize / 2))
+                multi_icon.setVisible(game.data.defaults_available_in_game_sessions)
+                multi_icon.move(int(bannerSize / 4), int(2 + bannerSize / 4))
+                if multi_icon.isVisible:
```
ThanatosGit, line 233: As already have been mentioned, this is not a bool value but a function returning a bool.
Also the check doesn't make sense at all?  The whole block is in a if case from line 214 where it checks for `game.data.defaults_available_in_game_sessions` which is also used to two lines above this one `multi_icon.setVisible(game.data.defaults_available_in_game_sessions)`.
In this code block, it should be always visible and truthy.

Which means to me: Line 231 is just `multi_icon.setVisible(True)` and the if condition in Line 233 is not needed.

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed: Searching for your own pickup in Multiworld sessions will now show only pickups which match *exactly* the name, instead of showing pickups which start with that name.
 - Fixed: The import in a multiworld session is blocked if it contains an unsupported game.
 - Fixed: Opening the webbrowser for Discord Login doesn't fail on Linux anymore.
+- Fixed: For Windows, the game select tooltip will not render as grey text on grey background in darkmode.
+- Added: Games display a banner if they are multiworld compatible.
+- Added: Option to show or hide multiworld compatibility banner.
+- Changed: Game selection on-hover behavior modified.
+- Changed: Internal references to icons renamed to reflect branding change.
```
ThanatosGit, line 18: Can be removed. The changelog is for users. Stuff specific to internals shouldn't be mentioned.

randovania/gui/main_window.py
```diff
@@ -292,9 +347,16 @@ def showEvent(self, event: QtGui.QShowEvent):
...
-        for game, logo in self._play_game_logos.items():
-            logo.setVisible(game.data.development_state.can_view())
+    def refresh_game_list(self) -> None:
+        for game, game_elements in self._play_game_elements.items():
+            game_elements.tile.setVisible(game.data.development_state.can_view())
+            if game_elements.tile.isVisible() and isinstance(game_elements.multi_banner, QtWidgets.QLabel):
+                game_elements.multi_banner.setVisible(
+                    game.data.defaults_available_in_game_sessions and self._options.show_multiworld_banner
+                )
+                game_elements.multi_icon.setVisible(
```
ThanatosGit, line 357: The mypy thing here tells that `None` has no `setVisible` method and the code will fail in that case. If `QLabel | None` is really required (and not only `QLabel`), do something like
```
label = game_elements.multi_icon
assert isinstance(label, QLabel)
label.setVisible(....
```

ThanatosGit (changes requested):

CHANGELOG.md
DualVission: > please add an entry in changelog.md

Have trimmed but it was explicitly requested by dunc.

ThanatosGit: My comment was just about the last line.
Yeah, changelog is required because it changes something for the user but not for telling something something internally changed to something.

randovania/gui/main_window.py
ThanatosGit: You did it differently but it's also fine. Just add the `isinstance` check also for `multi_icon` then mypy is quiet and everything good to go

DualVission: I think the beast is tamed. Hard to tell. Do you think the `isinstance` is still necessary?

ThanatosGit: No not necessary. As you now removed the `None` type, there is no need to check the instance anymore.

ThanatosGit (approved):
I like the solution with the `class GameQtElements(typing.NamedTuple):` much more than my suggested one 👍.

Only thing I noticed: For non-multiworld games, there is bascially an invisible banner at the same place, so you cannot click the game cover at this section. Feel free to add an else part with a `no multiworld banner` but that's for later.

PR #5902, opened 2024-01-22
Dread: Remove pushing Wide Beam Blocks with Wave and no Wide

Tyranisaur:
We have learned that pushing Wide Beam Blocks with Plasma or Wave, without Wide is possible, and at the time we thought it made sense to do so with Wave, from an angle. However, based on our current understanding, this should be considered a bug in the implementation of split beams.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Logic Database
 
+- Removed: It's no longer logical to push Wide Beam Blocks with Wave Beam without Wide Beam.
+
 ##### Artaria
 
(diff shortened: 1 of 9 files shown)


duncathan (approved):
approved in principle; would appreciate if someone can take a look over the db changes to verify they're correct

XenoWars (changes requested):
Can use this time to remove the Slide req from the Dairon Transport Access block in Cataris from below. Otherwise, i see nothing wrong.

Tyranisaur:
> Can use this time to remove the Slide req from the Dairon Transport Access block in Cataris from below. Otherwise, i see nothing wrong.

That was handled in #5799

XenoWars:
> > Can use this time to remove the Slide req from the Dairon Transport Access block in Cataris from below. Otherwise, i see nothing wrong.
> 
> That was handled in #5799

ah ok. i had looked on stable.

PR #5917, opened 2024-01-26
MSR - Rename Surface Rooms

dyceron:
Proposed room renames for Surface East. Will fix tests if names are good.

Morph Ball Room -> Morph Ball Chamber
Charge Beam Room -> Charge Beam Chamber
Scan Pulse Room -> Scan Pulse Chamber
Side Path -> Moheek Connector
One-Way Drop -> Upper Storage Access
Charge Beam Shaft -> Charge Beam Chamber Access


Miepee:
> Moheek Connector

Not a fan of the word "connector". I'd rather replace with something else, bonus points if it's an alliteration. Moheek Market?

Some rooms in surface east that should also get renamed IMO:
- Entrance pathway
- melee tutorial pathway
- charge beam storage (gets confusing with charge beam chamber)
- Alpha Arena Access (it's so far away from the actual arena, that i dont consider it "access")
- Alpha Storage
- Crumble Drop Shaft
- Landing Site Storage? With the mini puzzle in there and the spikes+crumble blocks everywhere it doesn't really fit like a "storage".
- Am unsure about "upper storage". "Upper" is very relative and if you look at it from the whole area, then it's not really that much up. It's only up in its own scan pulse subsection

In Addition, i'd like for the following rooms to get split:
- Alpha arena should be split into two: one for the arena, one for the big room with the save station before it
- Chozo seal should be split into two: one for the upper part, one for the mini room with the missile tank
- Scan Pulse Chamber should be split into two: one for the actual chamber, one for the right part.

Miepee:
> Alpha Arena also doesn't really make sense to split

It does. Currently ~70% of the room "Alpha Arena" is *not* an arena which thus sounds very misleading.
Scan Pulse Chamber has that too, but to a lesser extent. But you're right, that I completely forgot to look for this one on whether this could be sensibly split.
Alpha Arena I would just split before the missile block:
grafik

dyceron:
I've changed more names and also added two new rooms.

- Alpha Arena Access -> Energy Recharge Station Shaft
- Moheek Connector -> Moheek Meeting
- Entrance Pathway -> Hornoad Hallway
- Melee Tutorial Pathway -> Winding Path
- Charge Beam Storage -> Secret Stash
- Alpha Storage -> Cache West
- Upper Storage Access -> Split Path
- Upper Storage -> Split Path Storage
- Landing Site Storage -> Landing Site Puzzle
- Crumble Drop Shaft -> Landing Site Funnel
- Transport Storage -> Cache East

Split Alpha Arena into two

image
image

Added Gravitt Grotto (Split from Chozo Seal)

image
image

Miepee (approved):

PR #5921, opened 2024-01-26
alphabetized & expanded the hash words for Prime 1

Shaleblade:
I prioritized words that came up a lot in room/enemy names, as well as a few key Chozo Lore terms.


ThanatosGit:
Your JSON is invalid It has several occurences of the `,` being before `"` like in `"Crash,"`

Shaleblade:
Apologies - I think it should be fixed now

ThanatosGit:
> Apologies - I think it should be fixed now

Eventually, I don't like approving something for a game, I never touched.
But I can tell you that the bare minimum will be to pass the tests 😉.
Without looking at it, they are probably failing because the hash words are in some test files like rdvgame and tests for the patch data factory.

randovania/games/prime1/assets/hash_words.json
```diff
@@ -1,99 +1,138 @@
...
-    "Super",
-    "Charge",
-    "Wavebuster",
-    "Spreader",
-    "Core",
-    "Flamethrower"
-]
\ No newline at end of file
+ 	"Access",
+	"Aether",
```
Miepee, line 3: IDK if it's just me, but I associate Aether waaay more with echoes than prime 1. Where is that even used there?

Miepee (changes requested):
Missing changelog entry

randovania/games/prime1/assets/hash_words.json
Shaleblade: IIRC, it was for Research Lab Aether. I'm happy to remove or replace it, though - it's fair that it's way more associated with Echoes.

toasterparty (approved):

toasterparty:
pending the changelog entry of course

PR #5923, opened 2024-01-28
Adjust minimal logic for prime1/echoes/am2r

Miepee:
CC @Samuel6710 (for some reason, you're not a triager yet)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### AM2R
 
 - **Major** - Added: Multiworld support for AM2R. 
+- Changed: Minimal Logic has been adjusted. It now also checks for Morph Ball, Missile Launcher, the DNA and the Baby collection.
 - Changed: The Baby now checks for all DNA being collected and will display a message if not.
 
 #### Logic Database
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### AM2R
 
 - **Major** - Added: Multiworld support for AM2R. 
+- Changed: Minimal Logic has been adjusted. It now also checks for Morph Ball, Missile Launcher, Super Missile Launcher, the DNA and the Baby collection.
```
JeffGainsNGames, line 25: Do we need to check for both missiles and supers? Wouldn't missiles be enough?

Miepee: I was thinking on whether to put the reasoning into the PR description :smile: 
If we only check for Missiles, then the generator (but not the game) will start with Supers. Which means there's a fairly high chance that Missiles/Supers will be unobtainable.

Miepee: Before you ask "why not also check for bombs and PBs": there's a much lower chance that it happens there, so I decided there that I do not care

JeffGainsNGames: I mean wouldn't unobtainable supers be fine if we just check for missiles? They satisfy pretty much the same thing so I don't get why you'd need both.

Miepee: Ah wait, brainfart, you're right. Whoops.

Samuel6710 (approved):

JeffGainsNGames (approved):

PR #5925, opened 2024-01-28
AM2R: Force vanilla music for race seeds

Miepee:
This was a request, to not have accidentally someone use music shuffle in races.
grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### AM2R
 
 - **Major** - Added: Multiworld support for AM2R. 
+- Changed: Race seeds will now ignore Music Shuffle and force vanilla music.
 - Changed: Minimal Logic has been adjusted. It now also checks for Morph Ball, Missile Launcher, the DNA and the Baby collection.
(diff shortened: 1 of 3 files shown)


henriquegemignani:
Does random music provide a competitive advantage?

Miepee:
No but
1) it's slightly "buggy" (the vanilla game has some issues when a certain song is longer than it's supposed to be and rando doesn't fix it yet:tm:)
2) As a viewer it gets confusing if different runners have different songs for the same places. Racing organizers wanted at first to enforce players to choose the same music shuffle value, but to be less confusing it was later decided to just always enforce vanilla instead.

henriquegemignani:
These don't seem like great reasons to force that behaviour on whether there's a spoiler or not. I don't really like that flag being used for more things.

My suggestions would be to either move this setting to a preset setting, or have a preset setting for "allow music shuffle" or just a generic "enforce race rules" preset setting.

Miepee:
Hmm, I see.
Then I'll close this PR and instead focus the efforts on fixing it in the patcher side (which wouldn't make it into the february release).
Making a preset option for one cosmetic option feels meh to me, and this is currently the only race-enforced option that's planned. Could be revisited in the future, should there be a decision to have more enforced settings.

PR #5933, opened 2024-01-29
MSR - Support Cosmetic Options

dyceron:
Adds some color customization options. The following things can be changed:
- Free-Aim color (locked and unlocked)
- Grapple Aim color (locked and unlocked)
- Energy Tank color
- Aeion Bar color
- Ammo HUD color

image

Merge after OSRR has been updated


ThanatosGit (changes requested):
Mypy is complaining at multiple lines.
Would also be nice to raise the coverage for the  `cosmetic_patches_dialog` file.

randovania/games/samus_returns/gui/dialog/cosmetic_patches_dialog.py
```diff
@@ -1,41 +1,111 @@
...
 
     @property
     def cosmetic_patches(self) -> MSRCosmeticPatches:
         return self._cosmetic_patches
 
-    def reset(self):
+    def _on_check_update(self, check: QtWidgets.QCheckBox, field_name: str, _: None) -> None:
+        self.preferences = dataclasses.replace(
+            self.preferences, **{field_name: check.isChecked()}  # type: ignore[arg-type]
+        )
```
ThanatosGit, line 108: Is this function even used? Can't see it on a quick glance.

test/games/samus_returns/layout/test_msr_cosmetic_patches.py
ThanatosGit, line 1: I know the purpose is to exactly use a wrong type but try to use `# type: ignore` as comment above each line to tell mypy that it should ignore it at the following line.

test/games/samus_returns/gui/dialog/test_msr_cosmetic_patches_dialog.py
ThanatosGit, line 1: It should be possible to just click every color picker button then the `_open_color_picker` function would be tested, too.
Like do `skip_qtbot.mouseClick(dialog.custom_laser_locked_color_button, QtCore.Qt.MouseButton.LeftButton)` for `test_custom_laser_color` etc. maybe?

ThanatosGit (changes requested):

randovania/games/samus_returns/gui/dialog/cosmetic_patches_dialog.py
dyceron: Yeah it's definitely not used from what I can tell either. Just extra copy paste stuff

test/games/samus_returns/gui/dialog/test_msr_cosmetic_patches_dialog.py
dyceron: It seems to work but creates an other dialog which is breaking the tests here. Maybe another left click could be added for confirming the choice, if that's possible to do.

ThanatosGit: No idea on what we need to left click. I was under the impression that other games also don't test this, maybe that was the exact reason.
Then just leave it. We don't need coverage on everything and probably a waste of time here.

dyceron: Then ill revert the latest changes

ThanatosGit (approved):

dyceron:
Now supports displaying room names in game. Also fixes the default Grapple Laser Locked color.

image

Miepee (approved):

PR #5957, opened 2024-02-02
MP3: Add Rundas Quick-Kill + Other Fight Logic

SchwartzGandhi:
Collecting the Pickup in the Fuel Gel will cause the elevator to become active again. The elevator will instantly kill Rundas when he makes contact. RE: https://discord.com/channels/458432177223368704/758832290838216704/1201387901674590332

Bumped combat logic for Screw-Only method from intermediate to advanced.

### DIFF
--- a/randovania/games/prime3/logic_database/Bryyo - Fire.txt
+++ b/randovania/games/prime3/logic_database/Bryyo - Fire.txt
@@ -419,7 +419,7 @@ Extra - asset_id: 13296637487518400273
 > Pickup (Missile Expansion); Heals? False
   * Layers: default
   * Pickup 22; Category? Minor
-  > Above Ground
+  > Event - Gel Pickup
       Trivial
 
 > Above Ground; Heals? False
@@ -448,7 +448,9 @@ Extra - asset_id: 13296637487518400273
                   Combat (Beginner) and Damage ≥ 100
                   # https://youtu.be/4OggF458OMI&t=65
                   After Hangar Bay Ship Upgrade and Before Temple of Bryyo Turret and Combat (Intermediate) and Use Ship Missile (Command Visor and Ship Missile)
-          Combat (Intermediate) and Use Screw Attack (Space Jump)
+          Combat (Advanced) and Use Screw Attack (Space Jump)
+          # Collect Gel Pickup to Activate the Elevator to Kill Rundas
+          After TempleOfBryyoGelPickup and Knowledge (Expert)
   > Event - Turret
       Grapple Lasso and After Hangar Bay Ship Upgrade and Use Ship Missile (Command Visor and Ship Missile)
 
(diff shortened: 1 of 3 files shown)


randovania/games/prime3/logic_database/Bryyo - Fire.txt
```diff
@@ -419,7 +419,7 @@ Extra - asset_id: 13296637487518400273
 > Pickup (Missile Expansion); Heals? False
   * Layers: default
   * Pickup 22; Category? Minor
-  > Above Ground
+  > Event - Gel Pickup
```
dyceron, line 422: I want to say this should be left alone and have the event connect to the pickup instead? Not 100% sure but the general pattern is `event -> item`

randovania/games/prime3/logic_database/Bryyo - Fire.txt
```diff
@@ -464,6 +466,12 @@ Extra - asset_id: 13296637487518400273
   > Above Ground
       Trivial
 
+> Event - Gel Pickup; Heals? False
+  * Layers: default
+  * Event TempleOfBryyoGelPickup
+  > Above Ground
```
dyceron, line 472: If the above is true then I'd swap this to be the pickup

dyceron (changes requested):

randovania/games/prime3/logic_database/Bryyo - Fire.txt
JeffGainsNGames: So rearrange the node order to ~`above ground -> event - gel pickup -> pickup (missile exp) -> above ground`?
Will have to move all the logic from `above ground -> pickup (missile)` to `above ground -> event - gel pickup`, right?

dyceron: Correct

randovania/games/prime3/logic_database/Bryyo - Fire.txt
```diff
@@ -448,9 +437,22 @@ Extra - asset_id: 13296637487518400273
                   Combat (Beginner) and Damage ≥ 100
                   # https://youtu.be/4OggF458OMI&t=65
                   After Hangar Bay Ship Upgrade and Before Temple of Bryyo Turret and Combat (Intermediate) and Use Ship Missile (Command Visor and Ship Missile)
-          Combat (Intermediate) and Use Screw Attack (Space Jump)
+          Combat (Advanced) and Use Screw Attack (Space Jump)
+          # Collect Gel Pickup to Activate the Elevator to Kill Rundas
+          After Temple of Bryyo Gel Pickup Collected and Knowledge (Expert)
```
dyceron, line 442: I wonder if Expert is too high here. Could Advanced work?

SchwartzGandhi: My thought process for it being Expert was more due to how obscure this is rather than how difficult.
If there are more edge cases similar to this to compare to, then I'd probably have a better idea to go off of. 
Otherwise, I think Expert should be fine.

dyceron (approved):

PR #5970, opened 2024-02-05
Remove the generator log internally if it should be hidden

henriquegemignani:
It's already hidden in the UI, but too many people like looking at the file and being confused.

No change log entry being for me the rdvgame format is an internal detail.

### DIFF
--- a/randovania/generator/filler/runner.py
+++ b/randovania/generator/filler/runner.py
@@ -87,4 +87,7 @@ async def run_filler(
             unassigned_pickups=player_state.pickups_left,
         )
 
+    if any(pool.configuration.should_hide_generation_log() for pool in player_pools):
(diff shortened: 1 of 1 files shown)


randovania/generator/filler/runner.py
```diff
@@ -87,4 +87,7 @@ async def run_filler(
             unassigned_pickups=player_state.pickups_left,
         )
 
+    if not any(pool.configuration.should_hide_generation_log() for pool in player_pools):
```
duncathan, line 90: not? really?

henriquegemignani: ... copy paste is bad

ThanatosGit (approved):

PR #5987, opened 2024-02-09
MW: refactor executors to use a common socket holder

### LINKED ISSUE #5887: Refactor a common Socketholder out for Dread/Am2R/MSR MW
Dread, AM2R (and future MSR) basically all reimplement their socket from scratch. they should instead inherit it from a common socketholder.

Miepee:
Fixes #5887 
Apparently almost every executor was affected, not just dread and am2r.

### DIFF
--- a/randovania/game_connection/executor/am2r_executor.py
+++ b/randovania/game_connection/executor/am2r_executor.py
@@ -5,10 +5,11 @@ import dataclasses
 import logging
 import struct
 import typing
-from asyncio import StreamReader, StreamWriter
 from enum import IntEnum
 
 from PySide6.QtCore import QObject, Signal
+
+from randovania.game_connection.executor.common_socket_holder import CommonSocketHolder
 
 
 class AM2RConnectionException(Exception):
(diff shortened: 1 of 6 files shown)


Miepee:
CS didn't use it, but it is probably a good idea to include it.

randovania/game_connection/executor/am2r_executor.py
```diff
@@ -5,23 +5,20 @@
...
 
-@dataclasses.dataclass()
-# TODO: AM2R, Dread and future MSR have here very similar attribute. Refactor this to a common SocketHolder
-class AM2RSocketHolder:
-    reader: StreamReader
-    writer: StreamWriter
+@dataclasses.dataclass(frozen=True)
+class AM2RSocketHolder(CommonSocketHolder):
     api_version: int
-    buffer_size: int
```
duncathan, line 24: wrong line lol

Miepee: quite. removing the buffer size was intentional, am2r's executor does not use it whatsoever. Twas the classic copy paste from dread.

ThanatosGit (approved):

PR #5988, opened 2024-02-09
Show warning before genning, if item pool is more than max items.

### LINKED ISSUE #5861: Presets: check that item pool is less than or equal to pool size
When creating a preset there is no check for whether the amount of items in the pool is greater than the pool size, therefore players can save the preset and try to export with errors.

This is very noticeable with AM2R where DNA can be increased from 10 -> 46 in the Goal tab and there is no noticeable warning unless the player goes to the Item Pool tab and notices (which they might not always do).

Suggested fix: Popup on saving preset with `Warning: More items in pool than pool size, will cause generation errors.` or similar.

Miepee:
Fixes #5861 

Not sure if the base patches factory is the right place for it tbh. I just searched where the error message for starting locations was, and put it there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [7.4.0] - 2024-03-??
 
+- Added: A warning will be shown when trying to generate a game where more items are in the pool than the maximum amount of items.
 - TODO: Fill.
 
 ## [7.3.2] - 2024-02-??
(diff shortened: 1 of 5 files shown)


ThanatosGit (changes requested):
As it should be possible to save invalid presets, I guess it's fine to do it at `BasePatchesFactory` when genning the player pool.

But the tests are failing. Poor echoes has only 25 locations 🙃

> FAILED test/generator/test_generator_reach.py::test_reach_size_from_start_echoes - randovania.layout.exceptions.InvalidConfiguration: 118 items in the pool, but only 25 are allowed at maximum!

henriquegemignani:
> As it should be possible to save invalid presets, I guess it's fine to do it at `BasePatchesFactory` when genning the player pool.
> 
> But the tests are failing. Poor echoes has only 25 locations 🙃
> 
> > FAILED test/generator/test_generator_reach.py::test_reach_size_from_start_echoes - randovania.layout.exceptions.InvalidConfiguration: 118 items in the pool, but only 25 are allowed at maximum!

... that's a test that uses a hardcoded small version of the echoes db, to have a more complete integration test of a bunch of systems 😅

ThanatosGit (approved):

PR #5996, opened 2024-02-11
added speedkeep for ghav spin tower

steven11sjf:
- set to expert speed booster conservation, probably only has a 3-4 frame window to land it
- added a video for it as well

https://www.youtube.com/watch?v=17MNab3SmMc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added: Health requirements for the Balrog 1 boss fight.
 
+### Metroid Dread
+
+#### Logic Database
+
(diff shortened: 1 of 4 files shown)


Tyranisaur:
Needs changelog entry, otherwise looks good.

XenoWars (changes requested):
I wouldn't know why tests are failing, but it looks fine to me. Just needs the changelog as said

steven11sjf:
ok, added changelog and rebased on main. maybe that will fix tests? 

it's weird that they're failing on multiworld stuff though. have they failed on other PRs?

Tyranisaur (approved):

Miepee:
IIRC the failing test was the usual "something leaks in multiworld which causes a random test fail".

PR #6005, opened 2024-02-12
Dread/fix flip flopping additional requirements

Tyranisaur:


XenoWars (approved):
This should be fine. Was worried about the Dairon one, but the only ways to reach that door node are with having the power switch active, which will already be forcing Gravity or some form of cold run from the inside.

PR #6010, opened 2024-02-13
MSR: fix name of pickups, add vanilla test

Miepee:
- fix tank being named expansion
- fix aeion tank being named missile tank
- adds a test rdvgame with vanilla locations. does have progressives turned on tho.


dyceron (approved):
Wouldn't it be more accurate to use non progressives?

Miepee:
Technically yes, but progressives are more limiting.
Also I was lazy lol

dyceron:
Actually it doesn't really matter since spazer needs at least spazer to leave. It probably gets Wave first anyway and Plasma is required for a7 and Ridley so it works out anyway

PR #6020, opened 2024-02-14
Configuration cleanup

henriquegemignani:
Fix two warnings and migrate from black to ruff's formatter.


ThanatosGit (commented):
I'm not brain enought to understand the failing tests.
This PR changes from black to ruff as formater and mainly treats some long lines differently and some config is moved.
But what is the reason that only linux cannot import the `version_hash` file anymore?

henriquegemignani:
I changed how setuptools-scm is configured, apparently I broke something

ThanatosGit:
Yeah, but don't really see why it is only broken on linux.

We may want to change this:
`write_to: Pathlike[str] | Path | None = None : (deprecated) legacy option to create a version file relative to the scm root it's broken for usage from a sdist and fixing it would be a fatal breaking change, use version_file instead.`

https://github.com/pypa/setuptools_scm/blob/main/docs/config.md

ThanatosGit (approved):

PR #6021, opened 2024-02-14
MSR - Bug Fix and UI Updates

dyceron:
- Updated Differences tab to mentioned Scan Pulse don't use Aeion + rearranging other entries
image

- Updated Hints tab to mention the Hatchling Room DNA Seal is disabled until after the Queen
image

- Fixed wrong cc being used in A5 Tower Exterior Gamma Arena


Miepee:
fix test

Miepee (approved):

PR #6036, opened 2024-02-18
Add display name to templates

henriquegemignani:
No need for complex complete db operations to rename a template anymore.


randovania/gui/dialog/node_details_popup.py
```diff
@@ -303,7 +303,9 @@ def _create_connections_visualizer(
         if parent in self._connections_visualizers:
             self._connections_visualizers.pop(parent).deleteLater()
 
-        self._connections_visualizers[parent] = ConnectionsVisualizer(parent, layout, requirement)
+        self._connections_visualizers[parent] = ConnectionsVisualizer(
```
Miepee, line 306: mypy error

randovania/gui/docks/resource_database_editor.py
```diff
@@ -182,14 +182,17 @@ class TemplateEditor:
     root_item: QtWidgets.QTreeWidgetItem
     connections_item: QtWidgets.QTreeWidgetItem | None = None
 
-    def create_connections(self, tree: QtWidgets.QTreeWidget, requirement: Requirement):
+    def create_connections(
```
Miepee, line 185: mypy error

randovania/gui/docks/resource_database_editor.py
```diff
@@ -269,21 +278,32 @@ def create_template_editor(self, name: str):
...
         )
 
         return item
 
-    def edit_template(self, name: str):
-        requirement = self.db.requirement_template[name]
-        editor = ConnectionsEditor(self, self.db, requirement)
+    def rename_template(self, name: str, display_name: str) -> None:
+        if display_name == "":
+            display_name = None
```
Miepee, line 289: mypy error

randovania/gui/docks/resource_database_editor.py
```diff
@@ -269,21 +278,32 @@ def create_template_editor(self, name: str):
...
+    def edit_template(self, name: str) -> None:
+        template = self.db.requirement_template[name]
+        editor = ConnectionsEditor(self, self.db, template.requirement)
         result = editor.exec_()
         if result == QtWidgets.QDialog.DialogCode.Accepted:
             final_req = editor.final_requirement
             if final_req is None:
                 return
-            self.db.requirement_template[name] = final_req
+            self.db.requirement_template[name].requirement = final_req
```
Miepee, line 304: mypy error

randovania/game_description/resources/resource_database.py
```diff
@@ -33,6 +33,12 @@ def default_base_damage_reduction(db: ResourceDatabase, current_resources: Resou
 )
 
 
+@dataclasses.dataclass(frozen=True)
+class NamedRequirementTemplate:
+    requirement: Requirement
+    display_name: str | None
```
Miepee, line 39: I don't think i quite get this. The display name is more or less a "long key". So why make this "str | None" and deal with all the checking if its not null, instead of just forcing it to be a string?

randovania/gui/dialog/node_details_popup.py
henriquegemignani: ... these errors give me a feeling something is wrong with mypy

randovania/gui/docks/resource_database_editor.py
henriquegemignani: oh an actual error wow

randovania/game_description/resources/resource_database.py
henriquegemignani: My thought was that given a lot of existing templates use the key as a display name, no need to duplicate things. But maybe that's just better

Miepee: It does make the code easier to understand IMO.
Worst case, if someone really doesnt like their new template shortkey, they can just always write a migration to get rid of it :P

henriquegemignani:
I'm not going to fix the additional mypy errors of the functions I just added a return value.

Miepee (approved):

PR #6051, opened 2024-02-21
Add DataclassPostInitTypeCheck to StandardPickupDefinition

henriquegemignani:


ThanatosGit (changes requested):
Tests are not working.
The type check is a little bit too strict. If it resolves a type as `float` but the value type is a `int`, it should still pass because a `int` is always acceptable for a `float`.

henriquegemignani:
> Tests are not working. The type check is a little bit too strict. If it resolves a type as `float` but the value type is a `int`, it should still pass because a `int` is always acceptable for a `float`.

I was actually thinking of just fixing all the values to properly be floats.

ThanatosGit (approved):

PR #6052, opened 2024-02-21
Improve how custom shuffled/starting count works

henriquegemignani:
Fixes setting a pickup as starting effectively setting it as excluded.

I think the way this works now finally makes sense.


henriquegemignani:
Fixed progressive pickups being weird, and now the code for them is ever bettererer.

PR #6060, opened 2024-02-22
Add command for quickly benchmarking

henriquegemignani:
It also stores results into a file so it's easier to compare.


Miepee:
mypy errors :p

henriquegemignani:
> mypy errors :p

some of these type hints are sadly completely pointless :/

ThanatosGit (approved):

PR #6102, opened 2024-03-04
Standardize all patcher data tests

henriquegemignani:
Also sorts some keys for Dread and MSR so hopefully these patcher data files will stop changing randomly.
Certain tests also gain functionality by being able to set per-file cosmetic patches.


test/games/test_log_file_export.py
```diff
@@ -11,29 +12,34 @@
...
+    """
+    json_lib.write_path(data_path, data)
+
+    if cosmetic_patches != type(cosmetic_patches)():
+        # If not the default
+        json_lib.write_path(cosmetic_path, cosmetic_patches.as_json)
+    else:
+        cosmetic_path.unlink(missing_ok=True)
+
+    raise NotImplementedError("This function should not be called in normal execution.")
```
ThanatosGit, line 32: Not implemented error is not really what I would expect here.

test/test_files/patcher_data/samus_returns/samus_returns/starter_preset/cosmetic_1.json
ThanatosGit, line 1: I don't think that both MSR seeds needs a different cosmetic option (especially because they are the same), so one can be omitted.

randovania/games/dread/exporter/patch_data_factory.py
```diff
@@ -135,28 +135,19 @@ def _starting_inventory_text(self):
...
-                }
-                return new_spawnpoint_name
-            except KeyError:
-                raise self._key_error_for_start_node(node)
+            area = self.game.region_list.area_by_area_location(node.identifier.area_identifier)
+            collision_camera_name = area.extra["asset_id"]
+            new_spawnpoint_name = f"{self.spawnpoint_name_prefix}{len(self.new_spawn_points):03d}"
+            self.new_spawn_points[node] = {
+                "new_actor": {"actor": new_spawnpoint_name, "scenario": level_name},
+                "location": {"x": node.location.x, "y": node.location.y, "z": node.location.z},
```
ThanatosGit, line 147: Could assert that `node.location` is of type `NodeLocation`

ThanatosGit (changes requested):

test/games/test_log_file_export.py
henriquegemignani: I dunno what's best to actually use here. Creating a new type if pretty overkill

test/test_files/patcher_data/samus_returns/samus_returns/starter_preset/cosmetic_1.json
henriquegemignani: I just went with keeping whatever the existing tests did.

ThanatosGit (approved):

PR #6119, opened 2024-03-08
Adding additional combat requirements for several rooms.

herculesbenchpress:
Kihunter hallway spacejump changed to combat level // Added todo to Blue security room // Ability to freeze evir updated // Combat requirements added to aquarium tank // TODO added to security shaft west // TODO note added for Nightmare Nook // Simplified combat requirements in Crow's Nest from charge or missiles to beam-resistant enemies template // TODO note added to Big Shell 1 room to exit. // Combat beginner added to menu hallway

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 6 NOC.txt
+++ b/randovania/games/fusion/logic_database/Sector 6 NOC.txt
@@ -930,6 +930,7 @@ Extra - room_id: [27]
   * L0 Hatch to Big Shell 2/Door to Big Shell 1
   * Extra - door_idx: (56,)
   > Door to Shell Access
+      # TODO: Add jump extend and single wall jump template options after confirmation. 98% sure this is doable. -Herc.
       Space Jump and Can Use Power Bombs
 
 ----------------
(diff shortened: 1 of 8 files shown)


randovania/games/fusion/logic_database/Sector 6 NOC.txt
```diff
@@ -1258,12 +1259,12 @@ Extra - room_id: [40]
   * L0 Hatch to Clogged Cavern/Door to Menu Corridor
   * Extra - door_idx: (47,)
   > Door to Menu Save Room
-      Trivial
+      Combat (Beginner)
```
JeffGainsNGames, line 1262: 1) This change is likely unnecesary as accessing the save is never going to do anything logically (this could be useful in the future hypothetical entrance rando)
2) This makes it logically impossible for this path to be crossed trickless.
If you feel strongly about this add combat beginner OR beam-weak enemies.
Same comment for below.

JeffGainsNGames (changes requested):

randovania/games/fusion/logic_database/Sector 6 NOC.txt
Miepee: Isn't killing the menus in that room really trivial?

JeffGainsNGames: Yeah that's what I also meant by point 1.

herculesbenchpress: It probably is trivial. The thought here is more so with potential entrance rando, or starting in that save room. It could easily be moved back to trivial or do what Jeff suggests and make this "beam-weak enemies." I figured it's better to be thorough, at the very least. I'm not sure if anyone _would_ struggle in this room, but I also imagine the number is also not 0.

JeffGainsNGames (approved):

PR #6126, opened 2024-03-09
Added .vscode configs + how to use section in README

ThanatosGit:
Here is a cleaned up version of my vscode folder.
Mainly consists of:
- 3 tasks scrips: Create venv, build ui and run tests
- 2 launch scripts: One for the client (using build ui as prelaunch taks) and one to refresh game description.
- Recommendations on which plugins should be used.  Unsure about mypy as plugin. There are a lot of files which are missing type annotations and so it marks whole files with a red line which is very distracting.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,10 @@
+{
+    "recommendations": [
+        "ms-python.vscode-pylance",
+        "ms-python.python",
+        "ms-python.debugpy",
+        "charliermarsh.ruff",
+        "ms-python.mypy-type-checker",
+        "seanwu.vscode-qt-for-python"
+    ]
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,34 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    //  "--debug-game-backend",
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Randovania Client",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "${workspaceFolder}/randovania/__main__.py",
+            "console": "integratedTerminal",
+            "justMyCode": true,
+            "preLaunchTask": "Build UI",
+            "args": [
+                "gui", "--preview", "main"
+            ]
+           
+        },
+        {
+            "name": "Randovania Refresh-Game-Description",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "${workspaceFolder}/randovania/__main__.py",
+            "console": "integratedTerminal",
+            "justMyCode": true,
+            "args": [
+                "database", "refresh-game-description"
+            ]
+           
+        }
+    ]
+}
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,38 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            
+            "label": "Create venv with all exporters",
+            "type": "shell",
+            "linux": {
+                "command": "${workspaceFolder}/tools/prepare_virtual_env.sh ; ${workspaceFolder}/venv/Scripts/activate ; pip install -r requirements.txt",
+            },
+            "windows": {
+                "command": "@()|${workspaceFolder}/tools/prepare_virtual_env.bat ; ${workspaceFolder}/venv/Scripts/Activate.ps1 ; pip install -r requirements.txt",
+            }
+        },
+        {
+            "label": "Build UI",
+            "type": "shell",
+            "windows": {
+                "command": "${workspaceFolder}/venv/Scripts/Activate.ps1 ; python ${workspaceFolder}/setup.py build_ui"
+            },
+            "linux": {
+                "command": "${workspaceFolder}/venv/Scripts/activate ; python ${workspaceFolder}/setup.py build_ui"
+            }
+        },
+        {
+            "label": "Run all tests on all cores",
+            "type": "shell",
+            "windows": {
+                "command": "${workspaceFolder}/venv/Scripts/Activate.ps1 ; python -m pytest -n auto ${workspaceFolder}/test"
+            },
+            "linux": {
+                "command": "${workspaceFolder}/venv/Scripts/activate ;  python -m pytest -n auto ${workspaceFolder}/test"
+            }
+        }
+    ]
+}
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -17,22 +17,12 @@
         {
             "label": "Build UI",
             "type": "shell",
-            "windows": {
-                "command": "${workspaceFolder}/venv/Scripts/Activate.ps1 ; python ${workspaceFolder}/setup.py build_ui"
-            },
-            "linux": {
-                "command": "${workspaceFolder}/venv/Scripts/activate ; python ${workspaceFolder}/setup.py build_ui"
-            }
+            "command": "${command:python.interpreterPath} ${workspaceFolder}/setup.py build_ui"
         },
         {
             "label": "Run all tests on all cores",
             "type": "shell",
-            "windows": {
-                "command": "${workspaceFolder}/venv/Scripts/Activate.ps1 ; python -m pytest -n auto ${workspaceFolder}/test"
-            },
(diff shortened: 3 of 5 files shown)


.vscode/tasks.json
```diff
@@ -0,0 +1,38 @@
...
+    "tasks": [
+        {
+            
+            "label": "Create venv with all exporters",
+            "type": "shell",
+            "linux": {
+                "command": "${workspaceFolder}/tools/prepare_virtual_env.sh ; ${workspaceFolder}/venv/Scripts/activate ; pip install -r requirements.txt",
+            },
+            "windows": {
+                "command": "@()|${workspaceFolder}/tools/prepare_virtual_env.bat ; ${workspaceFolder}/venv/Scripts/Activate.ps1 ; pip install -r requirements.txt",
```
henriquegemignani, line 14: Not sure I like using these scripts for vscode.

.vscode/tasks.json
```diff
@@ -0,0 +1,38 @@
...
+            },
+            "linux": {
+                "command": "${workspaceFolder}/venv/Scripts/activate ; python ${workspaceFolder}/setup.py build_ui"
+            }
+        },
+        {
+            "label": "Run all tests on all cores",
+            "type": "shell",
+            "windows": {
+                "command": "${workspaceFolder}/venv/Scripts/Activate.ps1 ; python -m pytest -n auto ${workspaceFolder}/test"
```
henriquegemignani, line 31: you don't need to activate the venv. you can just use the python that is in the venv folder to do what you need

.vscode/launch.json
```diff
@@ -0,0 +1,34 @@
...
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    //  "--debug-game-backend",
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Randovania Client",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "${workspaceFolder}/randovania/__main__.py",
```
henriquegemignani, line 12: Wouldn't the program be the venv's python interpreter?

.vscode/tasks.json
henriquegemignani: Actually, we should use pull python from the VSCode extension. So it'd just be `{command:python.interpreterPath}`

henriquegemignani (changes requested):

README.md
```diff
@@ -226,7 +226,15 @@ the `prepare_virtual_env` scripts.
 
 Suggested IDE: [PyCharm Community](https://www.jetbrains.com/pycharm/download/)
 
-_*If using Visual Studio Code, make sure to set `"qtForPython.uic.liveExecution.enabled": false` when using the QT extension._
+## Visual Studio Code
+
+Clone this repository and open the folder in Visual Studio Code. It suggest several useful plugins for developing which you should download and install.
+
+If your python is setup properly, you can use the `Create venv with all exporters` task by pressing CTRL+SHIFT+P, type in `Task` to find + select `Tasks: Run task` and then select the task. It will create the venv with all requirements for you.
```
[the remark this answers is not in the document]
ThanatosGit: It is "find and select" but I don't like "find and select [...] and" because it reads kinda weird. 
"type in `Task` to find, select [...]" reads like I missed words because the first part is not a full sentence.
I will just remove "to find"

.vscode/extensions.json
```diff
@@ -0,0 +1,10 @@
+{
+    "recommendations": [
+        "ms-python.vscode-pylance",
+        "ms-python.python",
+        "ms-python.debugpy",
+        "charliermarsh.ruff",
+        "ms-python.mypy-type-checker",
```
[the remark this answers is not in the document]
ThanatosGit: You mean files? Mypy was introduced much later in randovania.
Atm mypy says this for me:
`Found 2748 errors in 433 files (checked 786 source files)`

I can take a look if the plugin can respect the files section in the pyproject.toml

.vscode/launch.json
ThanatosGit: No, it's a launch configuration of type `debugpy`. It starts the file listed in `program` with the workspace's default interpreter.

.vscode/extensions.json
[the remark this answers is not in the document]
ThanatosGit: Done

.vscode/tasks.json
ThanatosGit: What's the alternative besides "Call exactly this manually"?

.vscode/tasks.json
ThanatosGit: Sounds good. Then I can also remove the linux / windows specific sections except for creating the venv.

.vscode/tasks.json
henriquegemignani: The script is just a fancy eyelet with lots of checks around creating a venv and installing requirements.txt

Doesn't the python plugin create venvs automatically?

README.md
```diff
@@ -226,7 +226,15 @@ the `prepare_virtual_env` scripts.
 
 Suggested IDE: [PyCharm Community](https://www.jetbrains.com/pycharm/download/)
 
-_*If using Visual Studio Code, make sure to set `"qtForPython.uic.liveExecution.enabled": false` when using the QT extension._
+## Visual Studio Code
+
+Clone this repository and open the folder in Visual Studio Code. It suggests several useful plugins for developing which you should download and install.
+
+If your Python is setup properly, you can use the `Create venv with all exporters` task by pressing CTRL+SHIFT+P, type in `Task`, select `Tasks: Run task` and then select the task. It will create the venv with all requirements for you.
```
Miepee, line 233:
```suggestion
If your Python is setup properly, you can use the `Create venv with all exporters` task by pressing CTRL+SHIFT+P, type in `Task`, select `Tasks: Run task` and then select the task. It will create the venv with all the dependencies installed for you.
```

README.md
```diff
@@ -226,7 +226,15 @@ the `prepare_virtual_env` scripts.
...
+
+There is also a task defined to run all tests. To run individual tests you can utilise the `Testing` section of Visual Studio Code. You can simply run or run + debug a test there.
```
Miepee, line 235: I dont like the `+` but also dont have any ideas on how to get rid of it. `You can run or debug a test from there` sounds a bit weird. Does VSC call the latter `Run (Debug)`?

README.md
ThanatosGit: Well, it's not only installing dependencies but whatever.

.vscode/tasks.json
ThanatosGit: Not automatically but you can create it from there via `Python: Create environment`
But that would miss all checks for the correct python version, would ask what you want to install from the pyproject.toml (exporters, gui, server, test, typing) and would also miss the installation of the pre-commit.

README.md
Miepee: Then lets keep with the vsc terminology and also just call it "run or debug a test"

README.md
Miepee: maybe, but "requirements" sounded weird there xP

.vscode/tasks.json
ThanatosGit: ping

henriquegemignani: IMO if you're installing the entire requirements.txt, you'd also skip the python version check. My view is that someone managing the venv via VSCode is now more technical and we shouldn't restrict their choices,

ThanatosGit: There is nothing forcing you to use the tasks defined in this file, it's just more of a convenient thing if you have no idea what to do.

Hmmm but it's executed as one thing, so you can't skip the check with this task.

So... you are suggesting to remove the task and write in the readme how to create the venv within  vs code manually?

Miepee: ping

henriquegemignani: I am not sure the best things to choose. The intent of these scripts are to provide an easy way for less technical people to run Randovania, mostly focused on these who just want to edit the database.

The script clearly isn't what you're looking for because you immediately install more packages after.

Providing some documentation (and links to where read more?) about how to manage the venv might be best? Understanding what's going on is essential if you want to also go a bit deeper, such as also having the patcher installed from a git clone.

Miepee: IMO installing the exporter requirements makes things more user friendly. It avoids the following situation with:

"I cant export!"
\- "install the requirements, see the readme for details"
"What does the readme mean i have to see the tools\start_client script for info?"
....

Having a "one click button" is much nicer for that. It not like it comes with that many negatives: It works for all desktop OS, and you only have to download a few hundred MB more, which by today's standards isn't really that big of a deal.

duncathan (approved):

PR #6128, opened 2024-03-10
FUSION: clear some todo's in TRO and PYR

SchwartzGandhi:
Added a SWJ & Jump Extend method for Pyrochamber Access
Added SWJ for Level 1 Security Room
Added beginner wall jump requirement for getting back up to the door in Zazabi Speedway, as well as a beginner shinespark trick
Added SWJ template to Zazabi arena -> zazabi speedway
added shinespark tricks to level 1 Security Room to reach the door, and zazabi arena to cross without hi jump
Removed the TODO in Main Boiler as the room is not heated

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.txt
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.txt
@@ -851,8 +851,12 @@ Extra - room_id: [26]
           Any of the following:
               Space Jump or Can Single Walljump
               All of the following:
-                  Level 0 Locks and Speed Booster and Shinespark Tricks (Expert) and Disabled Door Lock Rando and Disabled Entrance Rando
-                  Level 1 Locks or Wall Jump (Advanced)
+                  Level 0 Locks and Speed Booster and Disabled Door Lock Rando and Disabled Entrance Rando
+                  Any of the following:
+                      # TODO: add trick video
+                      Level 1 Locks and Shinespark Tricks (Expert)
+                      # TODO: add trick video
+                      Shinespark Tricks (Expert) and Wall Jump (Advanced)
   > Event - L1 Locks
       Trivial
 
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.txt
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.txt
@@ -559,8 +559,6 @@ Extra - room_id: [18]
   * Layers: default
   * L0 Hatch to Zazabi Speedway/Door to Zazabi Arena
   * Extra - door_idx: (133,)
-  > Door to Zazabi Access
-      Screw Attack and Speed Booster and After Boss Zazabi Defeated and Shinespark Tricks (Expert) and Disabled Door Lock Rando and Disabled Entrance Rando
   > Arena Floor
       Screw Attack
 
@@ -853,10 +851,8 @@ Extra - room_id: [26]
           Any of the following:
               Space Jump or Can Single Walljump
               All of the following:
-                  Level 0 Locks and Speed Booster and Disabled Door Lock Rando and Disabled Entrance Rando
-                  Any of the following:
-                      Shinespark Tricks (Hypermode)
-                      Level 1 Locks and Shinespark Tricks (Hypermode)
+                  Level 0 Locks and Speed Booster and Shinespark Tricks (Expert) and Disabled Door Lock Rando and Disabled Entrance Rando
+                  Level 1 Locks or Wall Jump (Advanced)
   > Event - L1 Locks
       Trivial
 
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.txt
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.txt
@@ -854,7 +854,7 @@ Extra - room_id: [26]
                   Level 0 Locks and Speed Booster and Disabled Door Lock Rando and Disabled Entrance Rando
                   Any of the following:
                       # TODO: add trick video
-                      Level 1 Locks and Shinespark Tricks (Expert)
+                      Level 1 Locks and Knowledge (Advanced) and Movement (Beginner) and Shinespark Tricks (Expert)
                       # TODO: add trick video
                       Shinespark Tricks (Expert) and Wall Jump (Advanced)
   > Event - L1 Locks
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.txt
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.txt
@@ -560,7 +560,7 @@ Extra - room_id: [18]
   * L0 Hatch to Zazabi Speedway/Door to Zazabi Arena
   * Extra - door_idx: (133,)
   > Door to Zazabi Access
-      Impossible
+      Screw Attack and Speed Booster and Shinespark Tricks (Expert) and Disabled Door Lock Rando and Disabled Entrance Rando
   > Arena Floor
       Screw Attack
 
@@ -584,7 +584,9 @@ Extra - room_id: [18]
   > Door to Zazabi Access
       After Boss Zazabi Defeated and Can Climb High Jump Wall
   > Door to Zazabi Speedway
-      Screw Attack and Space Jump and After Boss Zazabi Defeated
+      All of the following:
+          Screw Attack and After Boss Zazabi Defeated
+          Space Jump or Can Single Walljump
   > Event - Zazabi
       All of the following:
           All of the following:
@@ -846,7 +848,9 @@ Extra - room_id: [26]
   > Door to Crumble City
       All of the following:
           Screw Attack
-          Space Jump or Wall Jump (Beginner)
+          Any of the following:
+              Space Jump or Can Single Walljump
+              Level 1 Locks and Speed Booster and Shinespark Tricks (Hypermode) and Disabled Door Lock Rando and Disabled Entrance Rando
   > Event - L1 Locks
       Trivial
 
@@ -1521,7 +1525,9 @@ Extra - room_id: [55]
   * Extra - blockx: 10
   * Extra - blocky: 26
   > Door to Zazabi Arena
-      Wall Jump (Beginner) or Can Climb High Jump Wall
+      Any of the following:
+          Wall Jump (Beginner) or Can Climb High Jump Wall
+          Speed Booster and Shinespark Tricks (Beginner)
 
 > Door to Zazabi Arena; Heals? False
   * Layers: default
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.txt
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.txt
@@ -1516,6 +1516,8 @@ Extra - room_id: [55]
   * Extra - blocky: 4
   > Door to Zazabi Arena
       Trivial
+  > Bottom of Room
+      Trivial
 
 > Pickup (Hidden Missile Tank); Heals? False
   * Layers: default
@@ -1524,19 +1526,26 @@ Extra - room_id: [55]
   * Extra - room: 55
   * Extra - blockx: 10
   * Extra - blocky: 26
-  > Door to Zazabi Arena
-      Any of the following:
-          Wall Jump (Beginner) or Can Climb High Jump Wall
-          Speed Booster and Shinespark Tricks (Beginner)
+  > Bottom of Room
+      Trivial
 
 > Door to Zazabi Arena; Heals? False
   * Layers: default
   * L0 Hatch to Zazabi Arena/Door to Zazabi Speedway
   * Extra - door_idx: (134,)
+  > Bottom of Room
+      Trivial
+
+> Bottom of Room; Heals? False
+  * Layers: default
   > Pickup (Power Bomb Tank)
       Speed Booster
   > Pickup (Hidden Missile Tank)
       Speed Booster
+  > Door to Zazabi Arena
+      Any of the following:
+          Wall Jump (Beginner) or Can Climb High Jump Wall
+          Speed Booster and Shinespark Tricks (Beginner)
 
 ----------------
 Overgrown Hallway
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.txt
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.txt
@@ -560,7 +560,7 @@ Extra - room_id: [18]
   * L0 Hatch to Zazabi Speedway/Door to Zazabi Arena
   * Extra - door_idx: (133,)
   > Door to Zazabi Access
-      Screw Attack and Speed Booster and Shinespark Tricks (Expert) and Disabled Door Lock Rando and Disabled Entrance Rando
+      Screw Attack and Speed Booster and After Boss Zazabi Defeated and Shinespark Tricks (Expert) and Disabled Door Lock Rando and Disabled Entrance Rando
   > Arena Floor
       Screw Attack
 
@@ -586,7 +586,9 @@ Extra - room_id: [18]
   > Door to Zazabi Speedway
       All of the following:
           Screw Attack and After Boss Zazabi Defeated
-          Space Jump or Can Single Walljump
+          Any of the following:
+              Space Jump
+              Wall Jump (Advanced) and Can Single Walljump
   > Event - Zazabi
       All of the following:
           All of the following:
@@ -850,7 +852,11 @@ Extra - room_id: [26]
           Screw Attack
           Any of the following:
               Space Jump or Can Single Walljump
-              Level 1 Locks and Speed Booster and Shinespark Tricks (Hypermode) and Disabled Door Lock Rando and Disabled Entrance Rando
+              All of the following:
+                  Level 0 Locks and Speed Booster and Disabled Door Lock Rando and Disabled Entrance Rando
+                  Any of the following:
+                      Shinespark Tricks (Hypermode)
+                      Level 1 Locks and Shinespark Tricks (Hypermode)
   > Event - L1 Locks
       Trivial
 
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.txt
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.txt
@@ -844,8 +844,9 @@ Extra - room_id: [26]
   > Door to Security Access (Upper)
       Space Jump
   > Door to Crumble City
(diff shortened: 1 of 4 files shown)


randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -844,8 +844,9 @@ Extra - room_id: [26]
   > Door to Security Access (Upper)
       Space Jump
   > Door to Crumble City
-      # TODO: Look into requirements further and possibly add SWJ trick. The requirement of space jump makes the advanced tricks in crumble city invalid in non-entrance rando.
-      Screw Attack and Space Jump
+      All of the following:
+          Screw Attack
+          Space Jump or Wall Jump (Beginner)
```
Miepee, line 849: is tihs supposd to be a screw single walljump? lock it behind a template please.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -844,8 +844,9 @@ Extra - room_id: [26]
   > Door to Security Access (Upper)
       Space Jump
   > Door to Crumble City
-      # TODO: Look into requirements further and possibly add SWJ trick. The requirement of space jump makes the advanced tricks in crumble city invalid in non-entrance rando.
-      Screw Attack and Space Jump
+      All of the following:
```
Miepee, line 847: RE removed todo: is it not possible to do this with HJ and WJ shenanigans, or HJ and jump extends?

randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -1520,7 +1521,7 @@ Extra - room_id: [55]
   * Extra - blockx: 10
   * Extra - blocky: 26
   > Door to Zazabi Arena
-      Trivial
+      Wall Jump (Beginner) or Can Climb High Jump Wall
```
Miepee, line 1524: Probably worth adding a seperate general node with "Top of room" so that the door and the bottom pickup can reach it both with speedbooster.
If its possible to reach the door with a shinespark, could also potentially have that.

randovania/games/fusion/logic_database/Sector 3 PYR.txt
```diff
@@ -548,13 +548,17 @@ Extra - room_id: [14]
...
+          Space Jump
+          All of the following:
+              Can Freeze Enemies
+              High Jump or Wall Jump (Beginner)
+          All of the following:
+              # Jump Extend SWJ Method
+              Jump Extends (Advanced)
+              Any of the following:
+                  Wall Jump (Advanced)
+                  High Jump and Wall Jump (Intermediate)
```
Miepee, line 561: A) use single wall jump template
b) i'm unsure i follow on this one. dont you need some way to destroy the wall-guys? I dont think you can jump high enough otherwise to jump over them to the wall to WJ. You got some videos for this?

randovania/games/fusion/logic_database/Sector 2 TRO.txt
herculesbenchpress: No, it is not. The room is too deep. Space jump or blue doors are required to leave the security room. Screw attack with single wall jumps would be the only other way to reach this, hence why I put a TODO so this does not get forgotten about. Hard requiring space jump to get to this node completely eliminates all the expert and hypermode tricks that we factored into logic for crumble city.

Miepee: > Hard requiring space jump to get to this node completely eliminates all the expert and hypermode tricks that we factored into logic for crumble city.

it does not, due to the fact that DB should assume in most cases that every room is isolated.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
herculesbenchpress: This is a template called "Can Single Wall Jump" already, but looking at the header file, it looks like what is listed under it is "Impossible."

Miepee: This is intended, because there is no final conclusion yet on how SJW behaves. Having them behind a template means we can then later fix them all easily.

randovania/games/fusion/logic_database/Sector 3 PYR.txt
herculesbenchpress: With single wall jumps you would need to have jump extends and to continue the single wall jump on the opposite wall to get enough height. I can find a clip from a recent stream in which I did this while testing. 

Are we leaving single walljump tricks under the wall jump trick umbrella? The main reason I created these TODOs in the first place was due to there not being definitive direction one way or the other yet on whether single wall jumping is being kept in. If it is going to kept, it might make sense to have SWJ as it's own trick category instead of lumping it together with wall jumps.

The only thing that is unclear in this logic is the WJ or HJ and WJ and why they would be different difficulty levels.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
herculesbenchpress: Perfect! Good to know! Then yeah, this should likely read

ALL OF:
Screw Attack
      OR
      Space Jump or Can Single Wall Jump

randovania/games/fusion/logic_database/Sector 2 TRO.txt
herculesbenchpress: Would the speedbooster requirement for the second item already cover the speedbooster exit method?

Miepee: Not sure. I was thinking that technically since the devs only intended you to shinespark the way to the item and not shinespark up, that shinespark up to the door itself should be a trick.

herculesbenchpress: I'm pretty sure you can't shinespark up the room to the door platform. In 100% it's a tight trick to keep the stored spark from the crumble blocks while falling since you need to weave back and forth to get to the bottom item.

randovania/games/fusion/logic_database/Sector 3 PYR.txt
JeffGainsNGames: RE: Herc's last point, does HJ affect height coming out of a Wall Jump?
HJ/HJless difficulty would only need to be different if that was the case. (more jumps means more harder)

herculesbenchpress: Yeah, the only difference is the height off the wall jump. With HJ you can be a bit lower and without HJ it's one or two more jumps to get the height and jump extends to make this work.

shark20061: HJ only changes the initial jump height, it doesn't impact the height from wall jumps. That said, with HJ, it's 2 wall jumps to get the right height and without HJ it's 3 (plus wall jumping at the right place to avoid bonking the ceiling)

There is another way to do this jump without the jump extend, which is turning SA off in mid-air before the last wall jump.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
herculesbenchpress: After Jagger found a way to get to the screw blocks with speed booster, we should update this to:

ALL OF:
Screw Attack
       ANY OF:
       Space Jump or Speed Booster or Can Single Wall Jump

Miepee: the speedbooster needs a shinespark, a level 1 door, no entrance rando and no door lock rando and probably other things i have forgotten

randovania/games/fusion/logic_database/Sector 2 TRO.txt
SchwartzGandhi: > Probably worth adding a seperate general node with "Top of room" so that the door and the bottom pickup can reach it both with speedbooster. If its possible to reach the door with a shinespark, could also potentially have that.

Not entirely sure what you meant by this. I just added a (Speed+Shinespark Beginner) into Bottom Pickup ->Door to Zazabi Arena and see no problems with that.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -560,7 +560,7 @@ Extra - room_id: [18]
   * L0 Hatch to Zazabi Speedway/Door to Zazabi Arena
   * Extra - door_idx: (133,)
   > Door to Zazabi Access
-      Impossible
+      Screw Attack and Speed Booster and Shinespark Tricks (Expert) and Disabled Door Lock Rando and Disabled Entrance Rando
```
Miepee, line 563: You're missing a `After Zazabi`.
And that makes me wonder, whether this is even necessary in the first place. Even if you assume starting location in speedway with the necessary items, you're forced to kill zazabi first. and from there you cant save or anything so you need HJ/SJ to get up.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -584,7 +584,9 @@ Extra - room_id: [18]
   > Door to Zazabi Access
       After Boss Zazabi Defeated and Can Climb High Jump Wall
   > Door to Zazabi Speedway
-      Screw Attack and Space Jump and After Boss Zazabi Defeated
+      All of the following:
+          Screw Attack and After Boss Zazabi Defeated
+          Space Jump or Can Single Walljump
```
Miepee, line 589: the `can single walljump` should be accomanied by a WJ trick. This is **not** trivial to do.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -844,8 +846,11 @@ Extra - room_id: [26]
   > Door to Security Access (Upper)
       Space Jump
   > Door to Crumble City
-      # TODO: Look into requirements further and possibly add SWJ trick. The requirement of space jump makes the advanced tricks in crumble city invalid in non-entrance rando.
-      Screw Attack and Space Jump
+      All of the following:
+          Screw Attack
+          Any of the following:
+              Space Jump or Can Single Walljump
+              Level 1 Locks and Speed Booster and Shinespark Tricks (Hypermode) and Disabled Door Lock Rando and Disabled Entrance Rando
```
Miepee, line 853: A) you also need L0
B) you can do this without l1 locks, which is even harder. Could probably have the l1 behind expert but maybe knowledge? IDK how known it is that you can jump out of it.

randovania/games/fusion/logic_database/Sector 3 PYR.txt
```diff
@@ -548,13 +548,17 @@ Extra - room_id: [14]
...
+      Any of the following:
+          Space Jump
+          All of the following:
+              Can Freeze Enemies
+              High Jump or Wall Jump (Beginner)
+          All of the following:
+              # Jump Extend SWJ Method
+              Jump Extends (Advanced) and Can Single Walljump
+              Any of the following:
+                  Wall Jump (Advanced)
```
Miepee, line 560: This block needs you being able to kill the fune's too. So Screw or maybe supers?
I'm unsure whether with HJ you can avoid killing the upper fune.

Miepee (commented):
Honestly, I'm considering if for newer higher tricks we should enforce adding TODO or a video link

SchwartzGandhi:
- Added 'after Zazabi' despite not being useful still better to have the connection than not imo
- added wall jump advanced to the SWJ in zazabi arena
- added 'can kill beam resistant enemy' to SWJ/Jump extend methof for pyrochamber access, tested it out, was not able to avoid the upper Fune.
- added L0 lock for the security room shinespark and separate the level 1 requirement as shine spark expert.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
SchwartzGandhi: I'm apprehensive to add a knowledge requirement here just because its hypermode/expert tbh.
If we did, whos to say that we don't add a knowledge requirement to literally every shinespark in the game?
Also chances are if you are setting shinepark trick to expert, you're going to be experimenting with it a lot

randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -560,7 +560,7 @@ Extra - room_id: [18]
   * L0 Hatch to Zazabi Speedway/Door to Zazabi Arena
   * Extra - door_idx: (133,)
   > Door to Zazabi Access
-      Impossible
+      Screw Attack and Speed Booster and After Boss Zazabi Defeated and Shinespark Tricks (Expert) and Disabled Door Lock Rando and Disabled Entrance Rando
```
Miepee, line 563: The more i think about it, the more i think itd be better to just keep this at impossible. 
You're currently missing also lv0 for that connection, as well as HJ (because i dont think you can spark with only WJ or SJ). And these are so many requirements for such a niche thing that is never in logic, that we might as well remove it.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -844,8 +848,15 @@ Extra - room_id: [26]
...
-      Screw Attack and Space Jump
+      All of the following:
+          Screw Attack
+          Any of the following:
+              Space Jump or Can Single Walljump
+              All of the following:
+                  Level 0 Locks and Speed Booster and Disabled Door Lock Rando and Disabled Entrance Rando
+                  Any of the following:
+                      Shinespark Tricks (Hypermode)
+                      Level 1 Locks and Shinespark Tricks (Hypermode)
```
Miepee, line 859: This should be changed. If both shiensaprk tricks are hypermode, this one is never chosen because it has more requirements.
it should isntead be `lv1 + shinespark (expert) + knowledge (advanced)` or something.

randovania/games/fusion/logic_database/Sector 3 PYR.txt
```diff
@@ -548,13 +548,17 @@ Extra - room_id: [14]
...
+      Any of the following:
+          Space Jump
+          All of the following:
+              Can Freeze Enemies
+              High Jump or Wall Jump (Beginner)
+          All of the following:
+              # Jump Extend SWJ Method
+              Jump Extends (Advanced) and Can Kill Tough Beam-Resistant Enemy
+              Any of the following:
+                  Wall Jump (Advanced) or Can Kill Tough Beam-Resistant Enemy
```
Miepee, line 560: if you always need to kill them, then the rquiement being here again doesnt make sense.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
herculesbenchpress: Isn't lv0 already being accounted for by the door connection itself? From what I remember all the doors have an indicator on their security level that is already being taken into account. Do we have a video of this being possible (I am assuming this is from the left to the right to get to the speedway.)? Just thinking about it, having to carry a shinecharge through 2 rooms and sparking in the third room just doesn't sound possible based on the timing alone.

Miepee: > that is already being taken into account

it usually is, in this case it isn't. Because logic looks at "can i reach the left door from the right door". if you tell it "only on non entrance rando and non door lock rando", then logic will expect you that if you spawn right at the door without lv0, that you are able to make it which you cant.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
Miepee: dunno didnt do the trick. if jagger and you say it isnt that hard, feel free to bump it down to advanced

randovania/games/fusion/logic_database/Sector 3 PYR.txt
SchwartzGandhi: this was an accident whoops. fixed

randovania/games/fusion/logic_database/Sector 2 TRO.txt
SchwartzGandhi: Having them both be hypermode was my mistake, the w/o level 1 was supposed to be expert. but know after thinking about it, I decided they should both be expert.
Don't think this should be advanced tho, ive tried this for several minutes and still cant get it
I also added a wall jump requirement for the upper method instead of knowledge

Miepee: According to jagger, the l0 was considerably more difficult than l1

Miepee: > I also added a wall jump requirement for the upper method instead of knowledge

This isnt a walljump tho. AFAIK you just hold down the jump button

SchwartzGandhi: the wall jump is in reference to getting enough height for the shine spsark, you can get climb the room in time without it

JaggerTSG: I had been trying to do it before grabbing SA, so I needed to plan around the zoro, but in retrospect, you need SA for the blocks regardless lol. It took longer, but I'm not sure if the one through the bottom door had just been a fluke or if it actually is easier.

Top requires bouncing off the right side to reach the middle platform.

JaggerTSG: Basically you have slightly more time, but the spark height is more precise.

Miepee (changes requested):
Can we please enforce either adding a TODO or directly a video link for more difficult tricks? If the video link isnt on yt, we can also just add it as a todo.
Would make things more easier to collect on which things still need properly recorded videos or not.

randovania/games/fusion/logic_database/Sector 3 PYR.txt
JeffGainsNGames: Won't this SWJ block need SWJ and Screw attack. The SA req would obsolete the `Beam-Resistant Enemy` template.'

randovania/games/fusion/logic_database/Sector 2 TRO.txt
SchwartzGandhi: @JaggerTSG @herculesbenchpress can one of you at some point get yt vids of you getting both of these sparks, *specifically with screw equipped*? I've been trying my best at it for the past few days and I'm still unable to get it.

Feel free you label all the trick requirements for them however you want, but my personal take is that `Shinespark Tricks` shouldn't be below `Expert`

Miepee (approved):

PR #6150, opened 2024-03-12
Update & refine README.md

Shaleblade:
1. Spruces up & updates the description text (trying to make it a bit more fun than earlier drafts at the request of @duncathan )
2. removes game-specific jargon
3. adds the tag info

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,15 +29,15 @@ New here or looking to install? Check [our website](https://randovania.github.io
 <!-- Begin WELCOME -->
 
 Randovania can randomize many aspects of its supported games, all while still ensuring they're completable 
-without any unintended tricks or glitches. Its features include:
+without using any glitches or exploits. Its features include:
 
-* Randomizing what can be found in each pickup location. Weapons, keys, and more can end up in
+* Randomizing what can be found in each item location. Weapons, keys, and more can end up in
   completely new places.
 
 * Use Multiworld sessions to shuffle items between multiple separate games, alone or with friends.
   All Multiworld games are compatible with each other - mix and match as you like!
 
-* Randomize how areas connect to each other, or what resources are required to travel between areas.
+* Randomize how areas connect to one another, or what resources are required to travel between areas.
   These options are highly customizable, letting you limit or unleash the chaos.
 
 * Randomize your starting equipment and location. Feeling brave? You can even shuffle items
--- a/README.md
+++ b/README.md
@@ -34,16 +34,16 @@ without any unintended tricks or glitches. Its features include:
 * Randomizing what can be found in each pickup location. Weapons, keys, and more can end up in
   completely new places.
 
-* Use multiworld sessions to shuffle items between multiple separate games, alone or with friends.
-  All multiworld games are compatible with each other - mix and match as you like!
+* Use Multiworld sessions to shuffle items between multiple separate games, alone or with friends.
+  All Multiworld games are compatible with each other - mix and match as you like!
 
 * Randomize how areas connect to each other, or what resources are required to travel between areas.
-  Both options are highly customizable. 
+  These options are highly customizable, letting you limit or unleash the chaos.
 
 * Randomize your starting equipment and location. Feeling brave? You can even shuffle items
   you normally start with.
 
-So have fun and start randomizing.
+Have fun and start randomizing!
 
 <!-- End WELCOME -->
 
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ Integration written by [Miepee](https://github.com/Miepee). Offworld sprites are
 ## Setup
 
 Getting started:
-   1. Clone this repository. You will also need to pull the tags after, which you can do via `git pull --tags`. (downloading the zip is *not* supported and will not work)
+   1. Clone this repository. (downloading the zip is *not* supported and will not work)
    2. Open a terminal in the repository root
    3. Run the following file:
       1. Windows: `tools/prepare_virtual_env.bat`
--- a/README.md
+++ b/README.md
@@ -28,19 +28,20 @@ New here or looking to install? Check [our website](https://randovania.github.io
 
 <!-- Begin WELCOME -->
 
-Here you will be able to randomize many aspects of either game, while still being ensured it's possible to
-finish without any trick or glitch! What can be randomized?
+Randovania can randomize many aspects of its supported games, all while still ensuring they're completable 
+without any unintended tricks or glitches. Its features include:
 
-* Randomize what can be found in each pickup location, including major upgrades, expansions, keys and artifacts.
+* Randomizing what can be found in each pickup location. Weapons, keys, and more can end up in
+  completely new places.
 
-* Play with multiple people, via multiworld sessions. Your pickups will be shuffled among the games of
-everyone involved, no matter what game they're playing!
+* Use multiworld sessions to shuffle items between multiple separate games, alone or with friends.
+  All multiworld games are compatible with each other - mix and match as you like!
 
-* Randomize where each teleporter goes, or what you need to unlock a translator gate. In either case,
-there's advanced options for how they're shuffled.
+* Randomize how areas connect to each other, or what resources are required to travel between areas.
+  Both options are highly customizable. 
 
-* The location you start the game in as well as the items you start with. If you're feeling brave,
-you can even shuffle items you normally start with, like the Power Beam and Scan Visor.
+* Randomize your starting equipment and location. Feeling brave? You can even shuffle items
+  you normally start with.
 
 So have fun and start randomizing.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,19 +28,20 @@ New here or looking to install? Check [our website](https://randovania.github.io
...
+Randovania can randomize many aspects of its supported games, all while still ensuring they're completable 
+without any unintended tricks or glitches. Its features include:
 
-* Randomize what can be found in each pickup location, including major upgrades, expansions, keys and artifacts.
+* Randomizing what can be found in each pickup location. Weapons, keys, and more can end up in
+  completely new places.
 
-* Play with multiple people, via multiworld sessions. Your pickups will be shuffled among the games of
-everyone involved, no matter what game they're playing!
+* Use multiworld sessions to shuffle items between multiple separate games, alone or with friends.
```
Miepee, line 37: We should probably try to treat Multiworld as a proper noun

Miepee, line 37: Hmm, I feel like it would still be worth to have a very brief explanation on how MW works.

README.md
```diff
@@ -28,19 +28,20 @@ New here or looking to install? Check [our website](https://randovania.github.io
...
+  All multiworld games are compatible with each other - mix and match as you like!
 
-* Randomize where each teleporter goes, or what you need to unlock a translator gate. In either case,
-there's advanced options for how they're shuffled.
+* Randomize how areas connect to each other, or what resources are required to travel between areas.
+  Both options are highly customizable.
```
Miepee, line 41: probably worth changing this from "both" to be more generic. since its this basically encompasses elevator shuffle, translator gates, tram shuffle, teleporer shuffle, pipe shuffle, door lock rando etc

README.md
```diff
@@ -28,19 +28,20 @@ New here or looking to install? Check [our website](https://randovania.github.io
...
-there's advanced options for how they're shuffled.
+* Randomize how areas connect to each other, or what resources are required to travel between areas.
+  Both options are highly customizable. 
 
-* The location you start the game in as well as the items you start with. If you're feeling brave,
-you can even shuffle items you normally start with, like the Power Beam and Scan Visor.
+* Randomize your starting equipment and location. Feeling brave? You can even shuffle items
+  you normally start with.
 
 So have fun and start randomizing.
```
Miepee, line 46: Excalamtion mark?

README.md
Shaleblade: Agreed - I mainly wanted to make it clear multiworld =/= multiplayer. What information do you think is missing? I'm happy to make this slightly longer for clarity.

Miepee: The `Your pickups will be shuffled among the games of everyone involved` part explained fairly quick n simple on how it works. Altho I it'd be better to say "items" than "pickups"

README.md
Miepee: This is referring to the last line: `So have fun and start randomizing.`

README.md
Shaleblade: I might be having a brain moment; doesn't "shuffle items between multiple separate games" establish the same information?

Miepee: It does, *I* was having a brain moment

README.md
```diff
@@ -180,7 +181,7 @@ Integration written by [Miepee](https://github.com/Miepee). Offworld sprites are
 ## Setup
 
 Getting started:
-   1. Clone this repository (downloading the zip is *not* supported and will not work)
+   1. Clone this repository. You will also need to pull the tags after, which you can do via `git pull --tags`. (downloading the zip is *not* supported and will not work)
```
henriquegemignani, line 184: Tags come by default when cloning. The usual issue is when people fork without tags, which is default I believe.

README.md
```diff
@@ -28,21 +28,22 @@ New here or looking to install? Check [our website](https://randovania.github.io
 
 <!-- Begin WELCOME -->
 
-Here you will be able to randomize many aspects of either game, while still being ensured it's possible to
-finish without any trick or glitch! What can be randomized?
+Randovania can randomize many aspects of its supported games, all while still ensuring they're completable 
+without any unintended tricks or glitches. Its features include:
```
Miepee, line 32: just noticed: whats the reasoning for "unintended" here?

README.md
```diff
@@ -28,21 +28,22 @@ New here or looking to install? Check [our website](https://randovania.github.io
...
 
-* Randomize what can be found in each pickup location, including major upgrades, expansions, keys and artifacts.
+* Randomizing what can be found in each pickup location. Weapons, keys, and more can end up in
```
Miepee, line 34: Pickup location or item location? Or just straight up location?

README.md
```diff
@@ -28,21 +28,22 @@ New here or looking to install? Check [our website](https://randovania.github.io
...
+  completely new places.
 
-* Play with multiple people, via multiworld sessions. Your pickups will be shuffled among the games of
-everyone involved, no matter what game they're playing!
+* Use Multiworld sessions to shuffle items between multiple separate games, alone or with friends.
+  All Multiworld games are compatible with each other - mix and match as you like!
 
-* Randomize where each teleporter goes, or what you need to unlock a translator gate. In either case,
-there's advanced options for how they're shuffled.
+* Randomize how areas connect to each other, or what resources are required to travel between areas.
```
Miepee, line 40: Having another `each other` right after the MWs `each other` reads a bit weird.

README.md
```diff
@@ -28,21 +28,22 @@ New here or looking to install? Check [our website](https://randovania.github.io
...
+  These options are highly customizable, letting you limit or unleash the chaos.
```
Miepee, line 41: the chaos or just chaos?

README.md
Shaleblade: IMO, "trick" doesn't necessarily imply to the reader that it's not developer-intended — players might still have to do maneuvers that they might consider tricks (e.g. doing speedbooster puzzles).

Miepee: Some games have developer intended tricks behind logic settings. I.e. fusion's secret message.

Shaleblade: Are there any other examples of this? Specifically with things 100% known to be dev-intended. (Or maybe I should just drop the term "trick" and say "without any glitches or exploits."

Miepee: Spider boosting in MSR, am2r with basically everything, super metroid with WJ and shinesparks and soon some shinesparks in dread too? It definitely varies on game per game basis on whats considered "trickless". Glitches would still imply that they might need to do those tricks. 
Could just maybe have something like "Note: what is considered a trick varies per game"?

steven11sjf: Better examples for dread might be the early grapple and bombs, since both require coyote jumps and bombs gives a special kill for kraid.

README.md
Shaleblade: Item location is probably better - location isn't strictly true as not every game has "items in every room" like Prime, and "pickup" feels more jargon-y.

README.md
Miepee: > I think it makes it clear that you won't have to do anything dev-unintended

It technically doesn't. There's no specification there on whether it's game dev intended or rdv dev intended. Technically every trick in the DB is "intended".

README.md
Shaleblade: changing the second one to `one another`

README.md
Shaleblade: I think "the chaos" reads best with the parallelism of "limit the chaos" and "unleash the chaos" vs. "limit chaos" and "unleash chaos." The former sounds more fun & lighthearted, which was the aim.

README.md
Shaleblade: Should I remove this addition, then?

README.md
Miepee: Might be a non-native thing then. "limit the chaos" sounds weird to me personally.

README.md
steven11sjf: I might say "transporter" or "region transport", or something else. In both echoes, dread, SR teleporters all mean different things lol.

Shaleblade: I'm confused - the new version doesn't use the word "teleporter" at all. I deliberately want to avoid naming what kind of thing links areas (transporter, teleporter, door, boat, whatever) to keep it more game-neutral

Miepee: > I'm just stupid and was reading the red text :DSLOL:

README.md
Shaleblade: is "glitches or exploits" safe, then?

Miepee: I feel like this:
> Could just maybe have something like "Note: what is considered a trick varies per game"?

Would still be the clearest IMO.

Shaleblade: I'm kinda concerned this will create more uncertainty than clarity. I think the point of this line is just dispelling the notion some users have that, to play a randomizer, you need to learn things the game doesn't teach you. Making users wonder what is and isn't a trick feels distracting from the central point.

Miepee: Fair enough.
So 
> without any tricks or glitches.

Then?

Miepee: Ignore, didnt see you already made a commit. looks fine

Miepee (approved):

README.md
henriquegemignani: If tags are going to be mentioned, I'd prefer to have an explicit guide for forking the repository

Shaleblade: OK, I'll take it out then. Someone who knows what they're talking about (not me) should be the one to write the forking guide :p

PR #6151, opened 2024-03-12
Fusion: add audio options to cosmetics

JeffGainsNGames:
Adds the following audio options:
image

### DIFF
--- a/test/games/fusion/gui/test_fusion_cosmetic_patches_dialog.py
+++ b/test/games/fusion/gui/test_fusion_cosmetic_patches_dialog.py
@@ -57,12 +57,8 @@ def test_change_music_option(skip_qtbot, music_start_value: bool, option_to_clic
 
     dialog = FusionCosmeticPatchesDialog(None, cosmetic_patches)
     skip_qtbot.addWidget(dialog)
-    str_to_option_map = {
-        "stereo_option": dialog.stereo_option,
-        "mono_option": dialog.mono_option,
-    }
-    radio_button = str_to_option_map[option_to_click]
 
+    radio_button = getattr(dialog, option_to_click)
     skip_qtbot.mouseClick(radio_button, QtCore.Qt.MouseButton.LeftButton)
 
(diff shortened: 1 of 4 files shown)


Miepee:
There should be some kind of graphical seperaet between the default stuff and the disabling stuff. Otherwise people might think that disabling is a default and it can be enabled back in-game

JeffGainsNGames:
Would have a label for 
```
in-game:
(  ) mono (x) stereo
some-name-here:
[  ] disable x
[  ] disable y
```
would it be good to add another box around it, or is that too many boxes.

test/games/fusion/gui/test_fusion_cosmetic_patches_dialog.py
```diff
@@ -41,3 +43,26 @@ def test_color_space(skip_qtbot):
...
+)
+def test_change_music_option(skip_qtbot, music_start_value: bool, option_to_click: str, music_end_value: bool) -> None:
+    cosmetic_patches = FusionCosmeticPatches(stereo_default=music_start_value)
+
+    dialog = FusionCosmeticPatchesDialog(None, cosmetic_patches)
+    skip_qtbot.addWidget(dialog)
+    str_to_option_map = {
+        "stereo_option": dialog.stereo_option,
+        "mono_option": dialog.mono_option,
+    }
```
Miepee, line 63: you should be able to use `getattr` for this. `getattr(dialog, option_to_click)`.

Miepee:
> Would have a label for
> 
> ```
> in-game:
> (  ) mono (x) stereo
> some-name-here:
> [  ] disable x
> [  ] disable y
> ```
> 
> would it be good to add another box around it, or is that too many boxes.

How about:
```
Default Audio settings:
( ) mono (x) stereo

Audio volume:
- [ ] Mute SFX
- [ ] Mute Music
```
?

JeffGainsNGames:
image

Miepee:
Can you add a little bit of spacing between them?

Miepee:
grafik

Miepee (approved):

PR #6155, opened 2024-03-13
changes for website

duncathan:
- command to extract readme sections
- command to extract db videos in yaml format
- command to extract game data
- games have a new field for website info


duncathan:
let's not worry about the fact that i somehow based this branch on a commit from exactly three months ago

pyproject.toml
```diff
@@ -65,8 +65,9 @@ dependencies = [
     # error reports
     "sentry-sdk",
 
-    # Database Video html Site Generator 
+    # Database Video html Site Generator
     "htmlmin",
+    "ruamel.yaml",
```
henriquegemignani, line 70: I wonder if these two could be an extra which we don't include in the executable.

randovania/cli/commands/export_db_videos.py
```diff
@@ -283,6 +286,59 @@ def filename_friendly_game_name(game: RandovaniaGame):
...
+                if sorted_node:
+                    add_entry(sorted_area, node, sorted_node)
+
+    output = {"regions": output}
+
+    tr = lambda s: s
+    fmt = "yml"
+    if as_frontmatter:
+        tr = lambda s: f"---\n{s}\n---"
+        fmt = "yml"
```
henriquegemignani, line 335: Is this supposed to be the same extension?

randovania/games/prime2/game_data.py
```diff
@@ -111,6 +111,19 @@ def _exporter():
...
+    web_info=game.GameWebInfo(
+        what_can_randomize=[
+            "All items including Temple Keys",
+            "Elevator destinations",
+            "Starting locations",
+            "Door locks",
+            "Translator gate requirements",
+        ],
+        need_to_play=[
+            "An ISO of any Gamecube release of the game",
```
henriquegemignani, line 123: NTSC and PAL regions only. No Dark Echoes allowed.

randovania/games/game.py
```diff
@@ -96,6 +96,21 @@ class GameGenerator:
     """Use AllJokesDistributor if not using hints."""
 
 
+@dataclass(frozen=True)
+class GameWebInfo:
+    what_can_randomize: Iterable[str] = (
+        "Everything",
```
henriquegemignani, line 102: These values fit best as values for Blank Game, instead of being default values.

henriquegemignani (changes requested):

randovania/games/prime2/game_data.py
Miepee: Huh, i thought JP was also allowed.

randovania/cli/commands/export_db_videos.py
```diff
@@ -283,6 +286,59 @@ def filename_friendly_game_name(game: RandovaniaGame):
     return "".join(x for x in game.long_name if x.isalnum() or x in [" "])
 
 
+def export_as_yaml(game: RandovaniaGame, out_dir: Path, as_frontmatter: bool):
```
Miepee, line 289: Not gonna point out every single one of them: are all the mypy errors intended?

Miepee (commented):
Our CLI commands usually have unit tests.

randovania/cli/commands/export_db_videos.py
duncathan: no oops

randovania/cli/commands/export_db_videos.py
duncathan: yes, insofar as virtually every other CLI function would have the same errors and it seems out of scope to fix them all

pyproject.toml
duncathan: certainly could be, although out of scope for this pr imo

henriquegemignani: I'd prefer that done before the next release though.

duncathan: kk. you'll have to show mii how to do it

henriquegemignani (approved):

PR #6160, opened 2024-03-13
Fusion: higher resolution cover

JeffGainsNGames:


dyceron (commented):
aspect ratio 🤮

duncathan:
> aspect ratio 🤮

what's wrong with the aspect ratio? it's 3:4, just like all the other games and like the smaller version

dyceron:
> > aspect ratio 🤮
> 
> what's wrong with the aspect ratio? it's 3:4, just like all the other games and like the smaller version

It looks like it was crunched down to make it fit the aspect ratio, which makes it look wrong. Instead, the image itself should've been edited (preferably on the top and bottom) to retain the original aspect ratio of the cover.

duncathan:
> > > aspect ratio 🤮
> > 
> > 
> > what's wrong with the aspect ratio? it's 3:4, just like all the other games and like the smaller version
> 
> It looks like it was crunched down to make it fit the aspect ratio, which makes it stretched out horizontal which is wrong. Instead, the image itself should've been edited (preferably on the top and bottom) to retain the original aspect ratio of the cover.

ah. well, that's out of scope of this pr imo but can be changed later

PR #6171, opened 2024-03-14
MP3: Gel Refinery Site Z Axis (video provided)

SchwartzGandhi:
You can screw attack from the Refinery Access door to the Gel Purification Site door using a z-axis (shoutout to Reggie for showing me)
https://youtu.be/66_fmkS3vOk
also added a node next to the Refinery Access door to account for this new method
also added an SSJ method to get up the ledge


dyceron (changes requested):
Why are Prime 2 changes in this?

randovania/games/prime3/logic_database/Bryyo - Fire.txt
```diff
@@ -246,13 +239,33 @@ Extra - asset_id: 691790544662692104
...
   > Door to Gel Purification Site
       Trivial
 
+> Ledge by Refinery Access Door; Heals? False
+  * Layers: default
+  > Door to Refinery Access
+      Plasma Beam
+  > Door to Gel Purification Site
+      # https://youtu.be/66_fmkS3vOk
+      Screw Attack at Z-Axis (Intermediate) and Slope Jump (Beginner) and Use Screw Attack (Space Jump)
```
dyceron, line 265: I don't agree with using this trick, as it's intended for NSJ Screw Attack. I'd just throw it under Movement.

dyceron (changes requested):

dyceron (approved):

PR #6172, opened 2024-03-14
Fusion: S6: Big Shell 1: Resolve TODO comment with wall jump and single wall jump logic

shark20061:
Resolving a TODO item in Sector 6 (NOC): Big Shell 1

This climb is rather easy, as the left wall of the vertical shaft is reachable with a normal jump

The walls are also close enough to gain height by performing regular wall jumps between them. This is easy enough that a video is likely not warranted

No jump extend was necessary to do either of these, so that trick condition was not added


randovania/games/fusion/logic_database/Sector 6 NOC.txt
```diff
@@ -930,8 +930,14 @@ Extra - room_id: [27]
...
   * Extra - door_idx: (56,)
   > Door to Shell Access
-      # TODO: Add jump extend and single wall jump template options after confirmation. 98% sure this is doable. -Herc.
-      Space Jump and Can Use Power Bombs
+      All of the following:
+          Can Use Power Bombs
+          Any of the following:
+              Space Jump
+              # Possible with Single Wall Jumping Physics going up the left side.
+              Wall Jump (Intermediate) and Can Single Walljump
```
SchwartzGandhi, line 938: With Hi-Jump this could probably be bumped down to beginner.
NHJ can stay intermediate though.

randovania/games/fusion/logic_database/Sector 6 NOC.txt
```diff
@@ -930,8 +930,14 @@ Extra - room_id: [27]
...
+              # Possible without Single Wall Jumping physics by jumping off both walls. Height gain per jump is small.
+              Wall Jump (Advanced)
```
SchwartzGandhi, line 940: This needs a Hi-Jump requirement
NHJ you'll need to do jump extends

SchwartzGandhi (changes requested):
Will need Hi-Jump requirements

randovania/games/fusion/logic_database/Sector 6 NOC.txt
SchwartzGandhi, line 940: Disregard my previous thought; you can easily do this without jump extends. I was just being bad at the game, sorry about that.
Still might be worth bump this wall jump down to intermediate/beginner with Hi-Jump though.

Miepee: Is t his easier with jump extends? might be worth making something like
ANY:
- WJ advanced
- ALL
  - WJ Intermediate
  - ANY
    - HJ
    - Jump extends

SchwartzGandhi: eh, not really, if anything I find it harder to do wall jumps when I jump extend, don't think it should affect any of the logic here

SchwartzGandhi (approved):

PR #6173, opened 2024-03-15
Fusion: S1: Add damage values for lava rooms

shark20061:
These are as of the current (60% reduction) implementation of the suit logic, which seems unlikely to change now.


randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -574,12 +574,17 @@ Extra - room_id: [17]
   > Door to Lava Lake Annex
       Any of the following:
           Gravity Suit and Varia Suit
-          Damage Runs (Intermediate) and Lava Damage ≥ 70
+          Damage Runs (Intermediate) and Lava Damage ≥ 60
+          Varia Suit and Damage Runs (Beginner) and Lava Damage ≥ 21
+          Gravity Suit and Damage Runs (Beginner) and Lava Damage ≥ 30
   > Door to Zebesian Tower
```
Miepee, line 580: What about jump extends?

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -588,12 +593,22 @@ Extra - room_id: [17]
   * Extra - room: 17
   * Extra - blockx: 44
   * Extra - blocky: 19
+  > Door to Lava Lake Annex
+      All of the following:
+          Morph Ball
+          Any of the following:
+              Gravity Suit and Varia Suit
```
Miepee, line 600: grav+varia is unnecesary, as you can already reach the door by going to upper door first and then free lava diving to the lower left

randovania/games/fusion/logic_database/Sector 1 SRX.txt
shark20061: Getting regular wall jump physics means that jump extends don't help anymore - unspinning will negate the extra horizontal movement the wall jump grants.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
shark20061: That's true, will remove

randovania/games/fusion/logic_database/Sector 1 SRX.txt
Miepee: There isnt any mention of wall jumps in there tho? Unless they also apply to normal jumps

shark20061: Oof, this terminology is confusing. I should have said "kick jump physics", which is any jump that involves kicking off of the wall. In addition to the "wall jump", there is also the kick jumps off of ladders and  from ledge grabs, which use the same physics as a wall jump. In this case, the jump used here is a kick jump by hanging off the cliffside ledge.

Miepee: > the jump used here is a kick jump by hanging off the cliffside ledge.

Then this should probably have beginner movement attached to it IMO.

shark20061: Sure, and added damage values for not getting the jump to the platform

JeffGainsNGames (approved):

PR #6184, opened 2024-03-17
Add option to create new preset

Miepee:
Partly addresses #6183 
grafik

Probably has mypy errors and likely needs test adjustments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [7.5.0] - 2024-04-??
 
 - Added: Command line arguments for exporting games. These commands are intended for advanced uses only.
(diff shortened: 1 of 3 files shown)


ThanatosGit:
What about this thing here and about the idea to not hide it in the right click menu like mentioned  in https://github.com/randovania/randovania/issues/6183 ?

Miepee:
I wanted to have both: an entry in the context menu, and a button.

duncathan (commented):
I would rather have both Import and Create New as proper buttons, and leave the context menu for actions on specific presets tbh

Miepee:
Hmm, considering how (IIRC) there are already quite a few exceptions solely to handle the import action, yeah.

Miepee:
Gonna close this PR for now. The next PR that wants to address the issue should make a button that creates a new preset and not have special handling for it like the import action currently does.

PR #6190, opened 2024-03-18
Fusion: DLR gui and db updates

JeffGainsNGames:
add DLR tab
add locked hatches
fix missing/incorrect room ids
remove many unnecessary `not DLR` checks and add some hatch requirements where it is necessary
fix some minor template usage in zebesians waters and security shaft west in s5


Miepee (commented):
you're missing the "this is safe" comment for a few. would be nicer to get changed in

Miepee (approved):

PR #6191, opened 2024-03-18
Dread: Remove Yuzu from FAQ

JeffGainsNGames:


JeffGainsNGames:
correct me if wrong but assume a changelog isn't necessary

Miepee:
Teeeeechnically, it would. But I don't think we gain anything from having that mentioned publically.

Miepee (approved):

PR #6203, opened 2024-03-21
New Artaria SSCs and ADCs

steven11sjf:
Screw Attack Room:
- intermediate SSC from teleportal to plasma door https://youtu.be/PCSBYWfYCOU
- intermediate SSC from plasma door to upper freezer door https://youtu.be/BuXn_Mkep20
- intermediate ADC from upper freezer door to lower freezer door https://youtu.be/FvCqCrC-LhM

Charge Beam Access
- intermediate ADC from intro room door to pickup https://youtu.be/ea_Ny5vwIQ0
- someone double check the logic i changed for making sure it doesn't think you can clip in there and spark out the other end lmao

Teleport to Dairon
- intermediate SSC from thermal door to bottom half of room

### DIFF
--- a/randovania/games/dread/logic_database/Artaria.txt
+++ b/randovania/games/dread/logic_database/Artaria.txt
@@ -4849,7 +4849,7 @@ Extra - asset_id: collision_camera_081
   > Door to Freezer (Grapple)
       All of the following:
           # https://youtu.be/FvCqCrC-LhM
-          Aim Down Clip (Expert) and Can SSC
+          Aim Down Clip (Intermediate) and Can SSC
           Speed Booster Upgrade ≥ 4 or Disabled Door Lock Randomizer
   > Door to Shortcut to Screw Attack
       Screw Attack
--- a/randovania/games/dread/logic_database/Artaria.txt
+++ b/randovania/games/dread/logic_database/Artaria.txt
@@ -758,6 +758,8 @@ Extra - asset_id: collision_camera_008
       Trivial
   > Door to EMMI First Chase End (Top)
       Trivial
+  > Lower Room
+      Before Artaria - Thermal Device and Shinesink Clip (Intermediate) and Can SSC
 
 > Dock to Arbitrary Enky Room (Lower); Heals? False
   * Layers: default
@@ -1443,7 +1445,10 @@ Extra - asset_id: collision_camera_013
   > Door to Cold Introduction (Missile)
       Trivial
   > Pickup (Missile Tank)
-      Lay Power Bomb
+      Any of the following:
+          Lay Power Bomb
+          # https://www.youtube.com/watch?v=ea_Ny5vwIQ0
+          Speed Booster Upgrade and Aim Down Clip (Intermediate) and Can SSC
   > Tunnel to Charge Beam Room
       Trivial
 
@@ -1484,8 +1489,12 @@ Extra - asset_id: collision_camera_013
   * Extra - actor_def: actordef:actors/items/item_missiletank/charclasses/item_missiletank.bmsad
   > Door to Intro Room
       Lay Power Bomb
-  > Tunnel to Cold Introduction
-      Lay Any Bomb
+  > Event - Tunnel Speed Booster Blocks (Right)
+      All of the following:
+          Ballspark and Lay Power Bomb
+          Any of the following:
+              Speed Booster Upgrade
+              Charge Beam and Disabled Door Lock Randomizer
 
 > Event - Charge Beam Access Blob (Above); Heals? False
   * Layers: default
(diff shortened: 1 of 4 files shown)


randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -758,6 +758,8 @@ Extra - asset_id: collision_camera_008
       Trivial
   > Door to EMMI First Chase End (Top)
       Trivial
+  > Lower Room
+      Before Artaria - Thermal Device and Shinesink Clip (Intermediate) and Can SSC
```
Tyranisaur, line 762: Why does activating the thermal device remove your ability to move between these nodes?

randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -1484,8 +1489,12 @@ Extra - asset_id: collision_camera_013
...
   > Door to Intro Room
       Lay Power Bomb
-  > Tunnel to Cold Introduction
-      Lay Any Bomb
+  > Event - Tunnel Speed Booster Blocks (Right)
+      All of the following:
+          Ballspark and Lay Power Bomb
+          Any of the following:
+              Speed Booster Upgrade
+              Charge Beam and Disabled Door Lock Randomizer
```
Tyranisaur, line 1497: I'm entirerly convinced that the intention for this connection has been that you can also charge speed in the energy recharge room, effectively leading to a situation where, if you can get into this room, you could do the ballspark. Granted, I think it would make sense to add a speed booster conservation trick regardless.

Also, you probably want to have a connection from the pickup to the tunnel that requires having the event. Because, if you break the speed blocks from the left, then it would be easier to back.

Tyranisaur (changes requested):

randovania/games/dread/logic_database/Artaria.txt
XenoWars: yeah, you are guaranteed to be able to charge speed if you can get into this room, so the extra reqs are not necessary.

randovania/games/dread/logic_database/Artaria.txt
XenoWars: Yeah, not sure why that's there. that thermal door has no bearing on being able to charge speed.

randovania/games/dread/logic_database/Artaria.txt
XenoWars: However, I disagree with requiring conservation. it removes the logic of breaking the speed block from the other side, then being able to come back through (only applicable in door lock rando)

randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -4852,7 +4872,14 @@ Extra - asset_id: collision_camera_081
...
   > Door to Freezer (Power)
-      Screw Attack
+      Any of the following:
+          Screw Attack
+          All of the following:
+              # https://www.youtube.com/watch?v=BuXn_Mkep20
+              Speed Booster Upgrade ≥ 3 and Shinesink Clip (Intermediate) and Can SSC
+              Any of the following:
+                  Speed Booster Upgrade ≥ 3
+                  Plasma Beam and Disabled Door Lock Randomizer
```
XenoWars, line 4882: As this is, that plasma req will never be seen since you have the speed upgrades in the "All"

XenoWars (changes requested):

randovania/games/dread/logic_database/Artaria.txt
steven11sjf: this is because you need to have a wall to sink off of. the other door is a door frame so you couldn't if the door is opened. 

though now that I think about it, it shouldn't logically matter. if the thermal device is open you can access the room to the left, and if you would be expected to sink and go to the right exits, either the right door is open and you can just  spark up the wall or it's a locked door and you can sink off that.

randovania/games/dread/logic_database/Artaria.txt
steven11sjf: the intention is that without DLR, you can charge speed in energy recharge room, go through the missile door and ballspark. if DLR is enabled you can charge in the bottom of the charge access room with 1+ upgrades. though now that I think about it again it's tight enough that you need 2 if there are shields down there.

randovania/games/dread/logic_database/Artaria.txt
steven11sjf: yeah, I just forgot to delete the SB+3 when i added the vanilla doors requirements oops. anyways, going to close this one and redo it when may merges the tricks.

randovania/games/dread/logic_database/Artaria.txt
XenoWars: The intention is that you can charge speed no matter what if you're in this room, since it requires either coming from above, coming from intro room, or just using the speed block tunnel which at that point charging the speed is useless.

PR #6204, opened 2024-03-21
Frigate Crash Site and Ruined Courtyard Beginner Logic Changes/Addtions

Samuel6710:


Samuel6710:
I'm hesitant to add the `Before Thardus` dash, especially since it's a beginner trick and the generator might not like more dangerous actions.

toasterparty (approved):

PR #6212, opened 2024-03-22
Remove pre-placed indices from DNA locations

ThanatosGit:
Fix for https://discord.com/channels/914291389293027329/1220476257557545011

Bug TL;DR: E.g.: Use 11 DNA, set Phantom Cloak to vanilla. bootstrapping can decide to also place DNA on Corpius. Game cannot be generated because you can not leave Corpius. Add Door lock to it such that Sensor lock doors gets removed => Can be generated  + exported and played. but importing the rdvgame leads to an error.

(If this is OK, I add the same fix to other games).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Dread
 
+- Fixed: DNA placement respects vanilla item placement settings to not assign two items to one location
+
 #### Logic Database
 
 - Added: New trick, Cross Bomb Launch.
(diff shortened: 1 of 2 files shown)


Miepee:
yeah adding it to the other games would be nice as well.
I know am2r, msr and fusion should have something similar. Unsure about corruption and echoes.

henriquegemignani:
The specific fix here is fine, especially because the case of not enough locations is already handled.

But instead of just fixing all the other games, I'd prefer to finally create a component for handling this kind of key placement. 2 games with copy paste was terrible but 4 is unacceptable anymore.

ThanatosGit:
That's a good summary of why I dislike copy and paste code so much.
But I always fear the killer argument of "it's out of scope because I had to touch other games" when pointing to such stuff.

This PR doesn't make this specific design issue worse. Eventually, I leave it to you guys if you want to merge it or not.
Anyways, I will have look later how easy or hard it is to move this pre-place key/dna/artifact thing into a common component for all games.

henriquegemignani (approved):

PR #6222, opened 2024-03-23
Database Editor: Least disruptive space-saving change to the json logic databases.

shark20061:
Randovania's logic databases are written to files through the same method that all json files produced by the system are. While this is nice for making them human-readable (for spoilers, game seeds, etc.), it isn't really necessary for the logic databases to be output in the same way, for a few reasons:
 - Most (or all) review of logic is between the text files and not the json.
 - The deeply nested structures of the logic database result in a lot of indentation, which bloats the file
 - The logic editor mostly makes modifying the json directly pointless (with a few exceptions)

To avoid making a very drastic change to how to interact with the logic databases (especially in regards to git), this PR makes the simplest change that it can, which is to reduce the indentation for the logic databases to a single space per level. This still results in a significant (nearly 50% reduction) in the size of the logic database folders:

```
Size of logic_database folder per game
AM2R: 4.49 MB -> 2.42 MB
Blank: 64 KB -> 39 KB
Cave Story: 1.07 MB -> 610 KB
Dread: 7.35 MB -> 4.15 MB
Fusion: 3.45 MB -> 1.88 MB
Prime 1: 5.35 MB -> 2.60 MB
Prime 2: 7.32 MB -> 3.47 MB
Prime 3: 2.88 MB -> 1.58 MB
Samus Returns: 5.71 MB -> 3.07 MB
Super Metroid: 2.29 MB -> 1.28 MB

TOTAL: 39.97 MB -> 21.09 MB (47.2% reduction)
```

This also speeds up the time it takes the editor to save the databases by the same factor, as there is less for the database editor to have to write.

The json files could be further compressed by removing _all_ unnecessary whitespace, which would push the savings closer to 66%, but this could severely impact the review process and how to do conflict resolution on the DBs

This PR does not commit the changes to the logic DBs themselves, those are in a separate branch for the time being (https://github.com/shark20061/randovania/tree/compress-json-logic-databases-lite-resave-all)

### DIFF
--- a/randovania/game_description/data_writer.py
+++ b/randovania/game_description/data_writer.py
@@ -449,12 +449,6 @@ def write_as_split_files(data: dict, base_path: Path) -> None:
     for region in regions:
         name = REGION_NAME_TO_FILE_NAME_RE.sub(r"", region["name"])
         data["regions"].append(f"{name}.json")
-        json_lib.write_path(
-            base_path.joinpath(f"{name}.json"),
-            region,
-        )
+        json_lib.write_path(base_path.joinpath(f"{name}.json"), region, compress=True)
 
-    json_lib.write_path(
-        base_path.joinpath("header.json"),
-        data,
-    )
+    json_lib.write_path(base_path.joinpath("header.json"), data, compress=True)
--- a/randovania/lib/json_lib.py
+++ b/randovania/lib/json_lib.py
@@ -32,9 +32,9 @@ def read_dict(path: Path) -> dict:
     return result
 
 
-def write_path(path: Path, data: Any) -> None:
+def write_path(path: Path, data: Any, *, compress: bool = False) -> None:
     path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(json.dumps(data, indent=4, separators=(",", ": ")))
+    path.write_text(json.dumps(data, indent=1 if compress else 4, separators=(",", ": ")))
 
 
 async def read_path_async(path: Path, *, raise_on_duplicate_keys: bool = False) -> dict | list:


duncathan:
Interesting. I'll have to think about how I feel about this. 
In the meantime, an important detail: this doesn't actually save space significantly. The JSON dbs are only used when running from source, which requires a git clone (and therefore the full history of the larger files). The performance improvement when saving the DB may be worthwhile. Do you have any profiling?

shark20061:
I don't, it's anecdotal right now, but I could probably set up some way to benchmark the saving times.

duncathan:
Yeah, please do. That's really the main benefit here. 
(In case anyone is wondering, we store the databases in much more optimized binary format for releases including dev builds, which is why this only affects running from source)

Miepee:
My thoughts:
- Removing all whitespace I would definitely not like. Not only because it would make DB conflicts 100x more annoying, but also because currently there has to be some manual editing / review of the json files for a few things which would make them impossible
- How much does 2 space indentation loose over 1 space indentation? Or a tab character?
- As dunc said, for the normal release builds they use a compressed(/binary?) form of the DB, so the benefits would only be for the running-from-source builds. You are also thus commiting a 19MB diff to the repo. Unsure whether this one big diff now would cause the future smaller diffs to eventually even out


Also CC @henriquegemignani because you mentioned big database files recently.

MayberryZoom:
> (In case anyone is wondering, we store the databases in much more optimized binary format for releases including dev builds, which is why this only affects running from source)

To me, this is the issue with this. That extra time to save doesn't matter much in practice, and running from source we shouldn't care too much about saving disk space.

Personally, I would gladly take the performance penalty to keep the JSON more readable; it's still very useful. And as Miepee said, some things still have to be done manually via the JSON as well (notably adding new areas and the first dock of each region). This would definitely make getting new games working more annoying.

henriquegemignani:
My only concern for the files in the repository are readability, followed by minimising conflicts and allowing manual changes as needed. As the others have mentioned, these files aren't included in the release packages, instead a format focused on minimal space is used.

Performance wise, the concern would be in load speed of the distribution format.

shark20061:
Re: Removing all the whitespace
Conflict resolution is exactly why I didn't want to go that far

Re: Alternative indentation
`dumps` doesn't do non-space indentation out of the box. 2 spaces indentation would cut the savings amount by exactly 1/3 (to about 36%)
Although this would just concern what is up in the repo, the indentation can always be changed locally then resave the DB through the editor to format it, which is a thing that probably already has to happen to generate the txt files.

Re: Compressed release DBs
This fact would limit the benefit this provides

Re: Readability
Examples are in the branch in the description, but it still looks readable to me personally
I think the existing 4 space indentation makes the json more cumbersome to deal with given how nested it is, but I mostly use the breadcrumbs in VS Code to navigate around it anyway, and so it mostly doesn't matter how much it's indented until it requires horizontal scrolling (especially with split view and diffs)

Re: Manual changes
Still possible, this would not impact that

Re: Loading times
I don't think this impacts the loading times nearly as much as the save times

Ultimately, I know I haven't been around here very long, so I definitely don't want to rock the boat too hard here, but my thought is that there's more benefits with doing this change compared to not
Will get some hard numbers to back that up though, and I can table this until I get them, just wanted to pitch the idea

duncathan:
Yeah, it's a worthwhile pitch. If it has a meaningful impact on saving/loading the db when running from source, it may be worthwhile. Otherwise its benefits are gonna be limited - it's actually very likely that this change will *increase* the size of the repo for several years to come

henriquegemignani:
I am concerned how big of an increase in the repository size this would cause, since the biggest advantage of this is smaller checkout size. 
Increasing repository size with new changes is work as usual, but seems counterproductive to do so with a change intended to decrease the size. Since everyone who would use this gain also has the `.git` folder, the effective gain for now is negligible.
This seems like something to do when doing some other bulk change to the db format. But would be worth considering other changes at once. Maybe something that isn't a json file?

Regarding performance improvement, it'd be good to have hard data. It's it the file II that gains the improvements, or json processing?
There are many json libraries for python that are significantly faster and is even dabbled into using one of these (didn't merge yet for reasons I can't remember ugh).

ThanatosGit:
As I'm scrolling through the PRs: Don't know what's the conclusion here. I only see value in it if the amount of games being added keeps at the current pace. Then it would probably save some space in the near future and outweighs the addition of 21 MB atm fast.

duncathan:
I don't think we can accept this until we at least have some profiling data about the speedup while saving. even then, I'm not convinced this is worthwhile

PR #6223, opened 2024-03-23
Refactor pre-place of artifacts into Bootstrap

ThanatosGit:
A try to remove all copy paste code.
Every game needs to provide a game specific check which should return true for every PickupNode where a pre-placed artifact can be.
Also every game needs to provide the category of the item which should be pre-placed.


randovania/resolver/bootstrap.py
```diff
@@ -222,6 +227,53 @@ def assign_pool_results(self, rng: Random, patches: GamePatches, pool_results: P
...
+        self,
+        rng: Random,
+        patches: GamePatches,
+        pool_results: PoolResults,
+        game_specific_check: Callable[[PickupNode, BaseConfiguration], bool],
+        artifact_category: PickupCategory,
+    ) -> None:
+        pre_placed_indices = list(pool_results.assignment.keys())
+
+        locations = self.all_artifact_locations(
```
duncathan, line 259: I would prefer if we can pass an arbitrary list of locations here if desired

ThanatosGit: Instead of passing the `game_specific_check` as argument?
To pass locations, ofc every game needs to create them.
Hmmm I don't want to copy paste the iteration over the nodes back to every game. 
I could copy paste a call to the common method `all_artifact_locations` to every game specific part and pass the result as argument but is there a reason for it?
I mean, it has to be a `PickupNode`, not an already assigned one by the preset and the rest logic comes from game specific argument. The only reason I see atm, that you maybe have a faster method to get the locations for one game instead of iterating over all nodes, which shouldn't be sooo important in the bootstrap part.

duncathan: the main use case I expect to be useful is when a game would distribute multiple such sets with virtually the same logic. consider e.g. a zelda game distributing dungeon map and compass within each dungeon

henriquegemignani:
I did a look earlier but need to read again a bit more carefully. Wanted to post this with a full review but I'll just leave this here for now.

One main feature I'd like to gain from common code for this final-key-distribution is being able to optionally let these be shuffled in the multiworld.
Not necessary for this PR, but it's the design accommodating for such a feature?

randovania/resolver/bootstrap.py
ThanatosGit: The game specific part could just make multiple calls to `self.pre_place_artifacts` with a different check function. E.g. first call with indices of dungenon A and category of boss key dungeon A, second time with dungeon B etc.
Writing it down by myself, we may want to change it to specifiy the items instead of the item category in the future.

Anyways, I added a second commit to provide the locations instead of a check function but ofc it adds the boilerplate of calling `all_artifact_locations` to every game.

randovania/resolver/bootstrap.py
```diff
@@ -222,6 +226,45 @@ def assign_pool_results(self, rng: Random, patches: GamePatches, pool_results: P
...
+    ) -> list[PickupNode]:
+        locations = []
+
+        for node in game.region_list.all_nodes:
+            if isinstance(node, PickupNode) and game_specific_check(node, config):
+                locations.append(node)
+
+        return locations
+
+    def pre_place_artifacts(
```
Miepee, line 243: since these are supposed to be generic, can we rename the functions to not reference artifacts?
`all_preplaced_item_locations` and `pre_place_items` perhaps?

henriquegemignani: I would prefer for us to have a proper name for these end-game locking key items. Don't care which name, but some.

Tyranisaur: MacGuffin?

ThanatosGit:
> One main feature I'd like to gain from common code for this final-key-distribution is being able to optionally let these be shuffled in the multiworld. Not necessary for this PR, but it's the design accommodating for such a feature?

I don't even understand your use case.
You either lock the item to specific locations (in your world) or place them anywhere (potentially across all worlds).
Both is already a thing.
I can only think of stuff like "lock them to specific locations across worlds" (like DNA can be in any dread world but only on bosses/emmis) ???

randovania/resolver/bootstrap.py
Miepee: If I understood dunc correctly he also wanted to potentially use these for other items, hence being very generic with them.

duncathan: I would suggest something like `dumb_distribute_items()`. `all_artifact_locations()` is more specifically about artifacts. if we want to make it more generic the only thing I can think of is `all_macguffin_locations()` like Ty suggested, but I think that's better suited for another pr

duncathan:
> > One main feature I'd like to gain from common code for this final-key-distribution is being able to optionally let these be shuffled in the multiworld. Not necessary for this PR, but it's the design accommodating for such a feature?
> 
> I don't even understand your use case. You either lock the item to specific locations (in your world) or place them anywhere (potentially across all worlds). Both is already a thing. I can only think of stuff like "lock them to specific locations across worlds" (like DNA can be in any dread world but only on bosses/emmis) ???

Suppose you have two Dread players in a multi. At the moment, each player's DNA will only ever be in their own world. Dark would like it to be possible eventually for them to be in other worlds, but *only* on indices that are normally allowed to have DNA.

ThanatosGit:
I actually need to think more about what is required to support this.
Somehow the locations for the Artifacts needs to be at a better place instead of buried in each game's bootstrap.py.

Especially, I need to think about the expected outcome.
If you mix a dread world with dna only on bosses + emmis and a msr world with dna anywhere, does the dread dna gets pre-placed across the 12 dread locations and all msr locations? How would you be able to not do this and use the current behaviour of forcing dna only in your own world?
It gets even more funny, if you think about the fact that a "dna anywhere" multiworld might have locations which are not accessible trickless and potentially be always out of logic.

I hate these words but I really think that this is way out of scope of my "I eliminate duplicated code" PR here

randovania/resolver/bootstrap.py
ThanatosGit: If the intention is really to use it also for non dna/keys/artifacts whatever, then we also shouldn't use macguffin.

Miepee:
As dark said earlier:

> **Not necessary for this PR**, but it's the design accommodating for such a feature?

ThanatosGit:
I don't really no what you want me to tell with that. The bold part was only about implementing the feature itself in this PR but as you can see from my last message: I have no idea how I could prepare for a feature where I have no idea how it would like and what is the expected outcome.

There are endless question how it would like in detail like already mentioned: How would you use the current behaviour? Adding a "force dna in own world" to all presets? What happens if other world's uses "anywhere" (like the dread/msr example) and it requires to use logic db? But this PR is not really the place to further discuss it

Miepee:
Yeah, just wanted to basically say "no need to hate the words, it is most definitely out of scope for this PR, and nothing bad about it only living on the discussion block for a while."

ThanatosGit:
Anyone opposed on adding a changelog entry and then merge it? Or is this bugfix now nuked by an additional feature (which still has a lot of open issues for me).

Miepee:
Does this need a changelog entry? I thought behaviour was the same.

I personally dont mind merging, but IMO it would be _nice_ to rename that `all_artifact_locations` function.

ThanatosGit:
> Does this need a changelog entry? I thought behaviour was the same.

How should I know what I have done more than a month ago 😛.
But seriously: IIRC this fixes the issue when you mark a macguffin location as "vanilla". Like we had in Dread with `Corpius => vanilla` but also rolled as a DNA location...just now for every game by respecting the pre-placed indices / locations.

> 
> I personally dont mind merging, but IMO it would be _nice_ to rename that `all_artifact_locations` function.

If it will be used for other things, then it's "all_pickup_nodes_for_condition" or something? I mean, it just returns a list of pickup nodes where the last parameter is a callable defining the condition for what to include. Or TL;DR: It's just a fancy filter function?

ThanatosGit:
I used Miepee's first suggestion and renamed it to `all_preplaced_item_locations` because afaik there is interest to use such a thing for non-MacGuffin items.
As it fixes a bug, it should be good to go and improved later for different use cases.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [8.1.0] - 2024-06-??
 
+- Fixed: MacGuffin placement respects vanilla item placement settings to not assign two items to one location.
```
Miepee, line 10: I'm unsure if we really want to phrase the changelog like this, but I also dont want the PR to block solely on it.
I feel like adding it to each affected game with `{macguffin name} placement respects vanilla item placement settings to not assign two items to one location.` would be  more understandable for end users. Because we don't have the term `macguffin` anywhere referenced and users may not necessarily be familiar what its supposed to mean.

Miepee (approved):

CHANGELOG.md
ThanatosGit: Then I need to think about all the games I don't play :p.
Prime 1: Artifacts
Prime 2: Sky Temple Keys
AM2R: DNA
MSR: DNA
(Dread was already fixed before)
Does it effect Cave Story?

Miepee: affects CS with puppies and a weapon/arthurs key/5hp life capsule

ThanatosGit: What?

ThanatosGit: What is the changelog for CS then?
`Fixed: "Weird thigns" placement respects vanilla item placement settings to not assign two items to one location.` :p ?

duncathan: I don't think it should be game specific tbh. Just say "Items placed before standard generation now respect..."

ThanatosGit: Ok, good to go?

Miepee: ye

PR #6224, opened 2024-03-23
Prime: Add power beam only doors to DLR

JeffGainsNGames:
expose the power beam only doors to Randovania. People shuffle power beam, think this is a good addition.

### DIFF
--- a/randovania/games/prime1/logic_database/header.json
+++ b/randovania/games/prime1/logic_database/header.json
@@ -1461,7 +1461,7 @@
                         },
                         "lock": null
                     },
-                    "Power Door": {
+                    "Power Beam Only Door": {
                         "extra": {
                             "shieldType": "Power Beam Only"
                         },
@@ -1885,8 +1885,8 @@
                         "Normal Door",
                         "Permanently Locked",
                         "Plasma Door",
+                        "Power Beam Only Door",
                         "Power Bomb Blast Shield",
-                        "Power Door",
                         "Super Missile Blast Shield",
                         "Wave Door",
                         "Wavebuster Blast Shield"
--- a/randovania/games/prime1/logic_database/header.json
+++ b/randovania/games/prime1/logic_database/header.json
@@ -1461,6 +1461,16 @@
                         },
                         "lock": null
                     },
+                    "Power Door": {
+                        "extra": {
+                            "shieldType": "Power Beam Only"
+                        },
+                        "requirement": {
+                            "type": "template",
+                            "data": "Shoot Power Beam"
+                        },
+                        "lock": null
+                    },
                     "Missile Blast Shield": {
                         "extra": {},
                         "requirement": {
@@ -1876,6 +1886,7 @@
                         "Permanently Locked",
                         "Plasma Door",
                         "Power Bomb Blast Shield",
+                        "Power Door",
                         "Super Missile Blast Shield",
                         "Wave Door",
                         "Wavebuster Blast Shield"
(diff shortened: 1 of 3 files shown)


randovania/games/prime1/logic_database/header.txt
```diff
@@ -82,6 +82,13 @@ Dock Weaknesses
       No lock
 
 
+  * Power Door
```
Miepee, line 85: IMO `Power Beam Door` sounds nicer

JeffGainsNGames: yes, but it doesn't fit with all the others `wave door, plasma door, ice door`

Miepee: The existing ones should be renamed too IMO

duncathan (changes requested):
I've mentioned my position on this before: I'm perfectly happy with it so long as they cannot be included unless power beam is shuffled. Same goes for e.g. charge doors in echoes. Until such a system is implemented I'm against it, though

duncathan:
One possible solution might be to exclude weaknesses that can be opened with the starting equipment. If variety is a concern, you could pick from those weaknesses any time a blue door would be placed.

JeffGainsNGames:
Is it really such a bad thing to have it available always?
What if someone wants to play with vanilla item pool but shuffle the doors anyways?
Is there also no responsibility of the player to ensure they pick doors that make sense with there item pool?

Miepee:
> I'm perfectly happy with it so long as they cannot be included unless power beam is shuffled

With the way RDV is laid out, I very much dislike options changing based on what items are in the pool. IMO, if someone wants to shuffle power beam-only doors even tho they start with it, then they should be allowed to do so.

duncathan:
> > I'm perfectly happy with it so long as they cannot be included unless power beam is shuffled
> 
> With the way RDV is laid out, I very much dislike options changing based on what items are in the pool. IMO, if someone wants to shuffle power beam-only doors even tho they start with it, then they should be allowed to do so.

In terms of UI, you mean? Do you have any thoughts on my suggested implementation?

duncathan:
> Is it really such a bad thing to have it available always? What if someone wants to play with vanilla item pool but shuffle the doors anyways? Is there also no responsibility of the player to ensure they pick doors that make sense with their item pool?

I pretty firmly believe that we should avoid giving players the ability to pick settings that aren't appropriate. I see it as a failing of some of our design that it's so easy to do atm. As for your first point, did you have any thoughts about my suggested implementation?

Miepee:
> In terms of UI, you mean? 

Not just in terms of UI, but also in general. It feels very unintuitive that Door Lock related features are linked to what is in the Pool. I personally would see having that much more as a design flaw than player being able to do whatever unappropriate settings they want to do.

> Do you have any thoughts on my suggested implementation?

How would that play out with Door Type Rando?

randovania/games/prime1/logic_database/header.json
```diff
@@ -1439,6 +1439,16 @@
                         },
                         "lock": null
                     },
+                    "Power Door": {
```
toasterparty, line 1442: Power Door would imply that canonically, Power Beam is the *minimum* requirement to open such a door. Future maintainers of this file may get this confused with vanilla blue doors.

Also, Power Door could be confused with the "Powered" doors in crashed frigate etc.

Please use `Power Beam Only` for any and all code/UI which references this new door

toasterparty (changes requested):

toasterparty (approved):

PR #6226, opened 2024-03-24
Fusion: Remove open hatches from change_from list

shark20061:
Per request in fusion-dev: https://discord.com/channels/914291389293027329/1215675687936196608/1217168246080274552

Open hatches would need to use an open clipdata hatch slot to work correctly, and some open hatch rooms don't have any hatch slots unused, so it's safer to remove open hatches from the door randomization.

(Doors can still become open hatches)

### DIFF
--- a/randovania/games/fusion/logic_database/header.json
+++ b/randovania/games/fusion/logic_database/header.json
@@ -1984,8 +1984,7 @@
                         "L1 Hatch",
                         "L2 Hatch",
                         "L3 Hatch",
-                        "L4 Hatch",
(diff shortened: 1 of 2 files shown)


JeffGainsNGames (approved):
Huh, thought I'd done this. Good catch.

Miepee:
Currently can not be done, because DLR needs some completely open type that is able to get changed from.

Miepee:
I'll close this, as this is currently necessary right now due to how DLR works, and won't be necessary later when fusion's door code has been rewritten.

PR #6227, opened 2024-03-24
Fusion: S6: Fix logic from top crumble block in xbox access becoming a bomb block

shark20061:
Per request in fusion-dev: https://discord.com/channels/914291389293027329/1215675687936196608/1217936379497943190

In entrance randomizer, this room can become a trap if the player unknowingly falls through the crumble blocks without the ability to fight the X-controlled Security Robot (aka XBOX).

To avoid this trap, two changes were made to the entrance in the top left:
 - The door stays unlocked
 - The first crumble block is now a bomb block

The logic therefore needs an update here as the player must break a bomb block now.

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 6 NOC.txt
+++ b/randovania/games/fusion/logic_database/Sector 6 NOC.txt
@@ -633,7 +633,7 @@ Extra - room_id: [16]
   > Door to X-B.O.X. Garage (Upper)
       Screw Attack and After Boss X-B.O.X. Defeated
   > Arena
-      Trivial
+      Can Break Single Bomb Blocks
 
 > Arena; Heals? False
   * Layers: default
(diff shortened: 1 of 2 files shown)


randovania/games/fusion/logic_database/Sector 6 NOC.txt
```diff
@@ -633,7 +633,7 @@ Extra - room_id: [16]
   > Door to X-B.O.X. Garage (Upper)
       Screw Attack and After Boss X-B.O.X. Defeated
   > Arena
-      Trivial
+      Can Break Single Bomb Blocks
```
Miepee, line 636: I may misremember, but didn't anty want to rethink on what this block should become?

JaggerTSG: I mentioned I prefer it being a shot block for future-proofing (I forget if I only imagined it or actually did see a vid of the pb obstruction being ice-clipped through). I forget if there'd been any response.

Miepee:
now that this is planned to become a shot block, this PR is redundant

PR #6235, opened 2024-03-25
MF: clear todos

SchwartzGandhi:
- Added a cold damage category in the resource database
- NOC: Nocturnal Playground - Clogged Cavern -> Nocturnal Shaft now has added requirements since you will have to dodge blue X
- PYR: Removed todo message in Deserted Runway that was already finished
- TRO: Added Jaggers trick videos for the Security Room shinesparks
- AQA: Moved the nodes in Pump Control Unit for better readability, this room will get a overhaul in the future to account for the shinespark.

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.txt
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.txt
@@ -464,14 +464,14 @@ Extra - room_id: [14]
   * L0 Hatch to Central Shaft/Door to Dessgeega Dormitory
   * Extra - door_idx: (33,)
   > Door to Collapsed Shaft
-      Trivial
+      Screw Attack or Combat (Intermediate) or Normal Damage ≥ 199
 
 > Door to Collapsed Shaft; Heals? False
   * Layers: default
   * L0 Hatch to Collapsed Shaft/Door to Dessgeega Dormitory
   * Extra - door_idx: (34, 100)
   > Door to Central Shaft
-      Trivial
+      Screw Attack or Combat (Intermediate) or Normal Damage ≥ 199
 
 ----------------
 Southern Save Access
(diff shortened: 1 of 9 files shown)


randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -464,14 +464,14 @@ Extra - room_id: [14]
...
-      Trivial
+      Screw Attack or Combat (Intermediate) or Normal Damage ≥ 199
 
 > Door to Collapsed Shaft; Heals? False
   * Layers: default
   * L0 Hatch to Collapsed Shaft/Door to Dessgeega Dormitory
   * Extra - door_idx: (34, 100)
   > Door to Central Shaft
-      Trivial
+      Screw Attack or Combat (Intermediate) or Normal Damage ≥ 199
```
Miepee, line 475: Does it make sense to use `Can Kill Tough Beak-Weak Enemy` instead?

randovania/games/fusion/logic_database/Sector 6 NOC.txt
```diff
@@ -346,7 +346,7 @@ Extra - room_id: [8]
   * L0 Hatch to Clogged Cavern/Door to Nocturnal Playground
   * Extra - door_idx: (16,)
   > Door to Nocturnal Shaft
-      Trivial
+      Varia Suit and Combat (Intermediate) and Cold Damage ≥ 299
```
Miepee, line 349: I thought this was supposed to stay trivial, since you traverse this path in vanilla?

JeffGainsNGames: I'm leaning towards this stays trivial. 
I could understand that going IN is the vanilla path, and returning on the way OUT is not intended vanilla (given the speedblocks in the next room).
My main point would be that this room becomes a bit confusing in a hypothetical Entrance Rando where all the other connections are considered "trivial" in regards to the blue-X.
image

SchwartzGandhi: this is a fair argument. reverted

randovania/games/fusion/logic_database/Sector 2 TRO.txt
SchwartzGandhi: added the template, but still kept the normal damage threshold as an option

Miepee (approved):

PR #6249, opened 2024-03-28
Add credits elements function

Miepee:
Adds a function that gives some building blocks for someone to make their own credits spoiler function. Originally written because Fusion needed something like that.

Unsure if and what the best way to test this is. The other spoiler functions don't have any tests, and this will later get indirectly tested by Fusion's PDF.

Example output:
grafik

### DIFF
--- a/randovania/exporter/hints/credits_spoiler.py
+++ b/randovania/exporter/hints/credits_spoiler.py
@@ -5,6 +5,7 @@ import math
 from typing import TYPE_CHECKING, NamedTuple
 
 from randovania.exporter.hints.hint_namer import HintNamer, PickupLocation
+from randovania.game_description import default_database
 
 if TYPE_CHECKING:
     from randovania.game_description.game_patches import GamePatches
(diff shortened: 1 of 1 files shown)


ThanatosGit:
I'm not a big fan of adding dead code for "maybe future"™

Miepee:
Talked with this with darkszero already. Plan to add tests soon:tm:. Fusion is guaranteed (technically does already) to use this in its PDF.

ThanatosGit:
I don't know...still not a big fan of it because it's dead code now in this PR and could be added when it's needed + it's like the problem with copy & paste code reversed: Adding a generic function which no one uses except one game.

I haven't looked into it but isn't it possible to refactor something like the `generic_credits` thing to use this? Then it would be used by....uh 3?...games and game specific parts could also access this helper function here directly.

randovania/exporter/hints/credits_spoiler.py
```diff
@@ -74,6 +75,41 @@ def sort_pickup(p: PickupEntry):
...
+    all_patches: dict[int, GamePatches],
+    players_config: PlayersConfiguration,
+) -> dict[str, list]:
+    results = {}
+
+    major_pickup_name_order = {
+        pickup.name: index for index, pickup in enumerate(standard_pickup_configuration.pickups_state.keys())
+    }
+
+    def sort_pickup(p: str) -> (float, str):
```
ThanatosGit, line 89: Syntax error

ThanatosGit (changes requested):

Miepee:
IIRC there was something that was holding me back, but can't remember anymore.

duncathan:
I think this needs to be a proper refactor instead of implemented like this

PR #6250, opened 2024-03-28
Replace configurable_node in GamePatches with game_specific

henriquegemignani:
This opens path for a game to shuffle something unique to it and have it be part of the layout file.

Translator Gates are a perfect case for this!


ThanatosGit (changes requested):
Can you merge it with the current main?

ThanatosGit (approved):

PR #6252, opened 2024-03-28
MF: bump TRO data hub speed trick

SchwartzGandhi:


randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -1013,7 +1013,8 @@ Extra - room_id: [31, 7]
   > Event - Bomb Block
       Any of the following:
           Can Break Single Bomb Blocks
-          Speed Booster and Knowledge (Intermediate) and Movement (Intermediate) and Disabled Door Lock Rando and Disabled Entrance Rando
+          # Jump into the hole with Speed Booster
+          Speed Booster and Knowledge (Intermediate) and Movement (Advanced) and Disabled Door Lock Rando and Disabled Entrance Rando
```
Miepee, line 1017: why is this movement to begin wtih and not shinesparking?

SchwartzGandhi: It's under movement because you're *technically* not shine sparking, you jump right into the hole.
Afaik this type of thing is not used anywhere else, so I find movement to be pretty appropriate

Miepee: Hmmm, fair

JeffGainsNGames: Yeah if I read it as a shinespark trick, I may think I have to store the spark which isn't the case. That being said this is a perfect case to say video when 👀

JeffGainsNGames (approved):

PR #6277, opened 2024-04-01
MF: Zoro zig zag rework

SchwartzGandhi:
made the connection from bottom to top nicer and added trick video for the jump bomb jump method

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.txt
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.txt
@@ -258,6 +258,8 @@ Extra - room_id: [9]
               # Stand on Frozen Reo
               Combat (Beginner) and Stand On Frozen Enemies (Beginner) and Can Freeze Enemies
   > Door to Maintenance Annex
+      After Sector 2 Zoro Zig-Zag Bomb Blocks Broken and Can Bounce in Ball
+  > Event - Break Bomb Blocks
       Can Bounce in Ball and Can Use Any Bombs
 
 > Door to Maintenance Annex; Heals? False
@@ -266,20 +268,29 @@ Extra - room_id: [9]
   * Extra - door_idx: (21,)
   > Door to Data Hub
       All of the following:
+          Morph Ball and After Sector 2 Zoro Zig-Zag Bomb Blocks Broken
+          Any of the following:
+              Space Jump
+              Hi-Jump and Wall Jump (Intermediate)
+  > Event - Break Bomb Blocks
+      Any of the following:
           Morph Ball
           Any of the following:
-              All of the following:
-                  # Sane method
-                  Any of the following:
-                      # Climbing Requirements
-                      Space Jump
-                      Hi-Jump and Wall Jump (Intermediate)
-                  Any of the following:
-                      # Bomb Block Requirements
-                      Power Bomb Data
-                      Screw Attack and Knowledge (Beginner)
-              # Cursed method https://youtu.be/Cqf5mZg1NZc
-              Hi-Jump and Morph Ball Bomb Data and Jump Bombjump (Hypermode) and Mid-Air Morph (Intermediate) and Wall Jump (Intermediate)
+              # Climbing Requirement
+              Space Jump
+              Hi-Jump and Wall Jump (Intermediate)
+          Any of the following:
+              # Break Blocks
+              Can Use Power Bombs
+              Screw Attack and Knowledge (Beginner)
+              # Cursed Method: https://youtu.be/Cqf5mZg1NZc
+              Hi-Jump and Jump Bombjump (Hypermode) and Can Use Bombs
+
+> Event - Break Bomb Blocks; Heals? False
+  * Layers: default
+  * Event Sector 2 Zoro Zig-Zag Bomb Blocks Broken
+  > Door to Maintenance Annex
+      Trivial
 
 ----------------
 Cultivation Station
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.txt
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.txt
@@ -284,7 +284,7 @@ Extra - room_id: [9]
               Can Use Power Bombs
               Screw Attack and Knowledge (Beginner)
               # Cursed Method: https://youtu.be/Cqf5mZg1NZc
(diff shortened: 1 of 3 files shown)


randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -269,23 +269,17 @@ Extra - room_id: [9]
...
-                      All of the following:
-                          Morph Ball Bomb Data
-                          Any of the following:
-                              Jump Bombjump (Advanced)
-                              Screw Attack and Knowledge (Beginner) and Wall Jump (Intermediate)
+                      # Bomb Block Requirements
+                      Power Bomb Data
+                      Screw Attack and Knowledge (Beginner)
+              # Cursed method https://youtu.be/Cqf5mZg1NZc
+              Hi-Jump and Morph Ball Bomb Data and Jump Bombjump (Hypermode) and Mid-Air Morph (Intermediate) and Wall Jump (Intermediate)
```
shark20061, line 282: 🔧 Common `(Hi-Jump + Wall Jump) OR Space Jump` (the space jump part is missing on the cursed logic here) can be pulled out to Morph Ball's level

shark20061 (commented):
One note here, but otherwise looks better than it did before 😄 

This room also needs to make all the bomb blocks (collectively) into an event because they are all non-reforming (which would cut out the JBJ requirement on returning up this room if you went down it already), but that can be a separate change if you want

randovania/games/fusion/logic_database/Sector 2 TRO.txt
SchwartzGandhi: afaik the cursed method hard requires hi jump to get the jump bomb jumps to break the blocks. 
you can't get it NHJ with just space. so its not included here (also why HJ+walljump) isn't grouped with the morph ball

SchwartzGandhi: unless you mean pulling specifically (walljump OR space jump). I think that could work.

SchwartzGandhi: nvm I missed that I could jump keep hi jump in the jbj method. fixed

randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -266,26 +268,29 @@ Extra - room_id: [9]
   * Extra - door_idx: (21,)
   > Door to Data Hub
       All of the following:
+          Morph Ball and After Sector 2 Zoro Zig-Zag Bomb Blocks Broken
+          Any of the following:
+              Space Jump
+              Hi-Jump and Wall Jump (Intermediate)
+  > Event - Break Bomb Blocks
+      Any of the following:
```
shark20061, line 276: Top level here seems to have accidentally changed to an OR, should be AND

shark20061 (approved):

randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -258,34 +258,41 @@ Extra - room_id: [9]
...
+              Space Jump
+              Hi-Jump and Wall Jump (Intermediate)
+          Any of the following:
+              # Break Blocks
+              Can Use Power Bombs
+              Screw Attack and Knowledge (Beginner)
+              # Cursed Method: https://youtu.be/Cqf5mZg1NZc
+              Hi-Jump and Jump Bombjump (Hypermode) and Mid-Air Morph (Intermediate) and Can Use Bombs
+
+> Event - Break Bomb Blocks; Heals? False
```
Miepee, line 289: Which bomb block is this referring to. All?
Also why is Bounce In Ball required? Looking at the map, doesnt look like you need that to *break* the blocks. Not even to leave.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -258,34 +258,41 @@ Extra - room_id: [9]
...
-                              Screw Attack and Knowledge (Beginner) and Wall Jump (Intermediate)
+              # Climbing Requirement
+              Space Jump
+              Hi-Jump and Wall Jump (Intermediate)
+          Any of the following:
+              # Break Blocks
+              Can Use Power Bombs
+              Screw Attack and Knowledge (Beginner)
+              # Cursed Method: https://youtu.be/Cqf5mZg1NZc
+              Hi-Jump and Jump Bombjump (Hypermode) and Mid-Air Morph (Intermediate) and Can Use Bombs
```
Miepee, line 287: You broke the bomb blocks in the video with spring ball. I assume that laying a bomb -> get hit by explosion -> lay bomb at the peak also works to destroy that block?

randovania/games/fusion/logic_database/Sector 2 TRO.txt
SchwartzGandhi: yes, this is referring to all the blocks. thought about having each one be independent but decided against it since you gain nothing from having only some of them broken.

randovania/games/fusion/logic_database/Sector 2 TRO.txt
SchwartzGandhi: not quite sure what you mean by this. The first bomb block was destroyed exactly as you explained, it wasn't done with just spring ball.
If you are implying that these should be separated or something, I don't think thats really worth doing, considering both methods are still technically Jump Bomb Jumps

Miepee: sorry, meant to clarify that one specific bomb block, but forgot xP

SchwartzGandhi: oh I see now what you mean. I guess maybe you could? I tried it for a little bit and couldn't get it though.

It shouldn't really change anything considering the fact that you need hi jump anyway for the last JBJ.

Miepee: If you explicitly need spring ball, please mark it by using the template.

SchwartzGandhi: added

Miepee (approved):

PR #6285, opened 2024-04-02
Update Prime FAQ with non-Superheated Rooms in Magmoor info

Samuel6710:
Could be worded a little better


randovania/games/prime1/game_data.py
```diff
@@ -95,6 +95,12 @@ def _exporter():
             "break fixes present on that version are undone. This reverts the game to be functionally equivalent to "
             "NTSC 0-00, meaning that all versions of Prime are guaranteed to be logically completable when randomized.",
         ),
+        (
+            "Why are some rooms in Magmoor Caverns not superheated?",
+            "Elevator rooms, Storage Cavern, and every room between Twin Fires and the Phazon Mines and "
+            "Phendrana Drifts elevators (Far Magmoor) are not superheated, not including Burning Trail "
+            "and Lake Access which can be encountered without Varia Suit in a vanilla playthrough.",
+        ),
```
JeffGainsNGames, line 103: You phrased it as a why then didn't say why.

If keeping it this way, you may want to say `In the vanilla game, only certain areas of magmoor are superheated. The far side beyond Twin Fires Tunnel as well as Storage Caverns are exceptions which can be traversed without heat protection.`
I don't know if it's worth mentioning burning trial etc, unless that last sentence makes sense only with the missing reason.

If you rephrase to `What rooms in Magmoor are superheated?`, you could just list the unheated rooms similar to above.

Samuel6710: I mainly phrased the question that way since I imagine that would be the general question players would ask. I'll rephrase the answer to work with it better.

Miepee: bump?

randovania/games/prime1/game_data.py
```diff
@@ -95,6 +95,11 @@ def _exporter():
             "break fixes present on that version are undone. This reverts the game to be functionally equivalent to "
             "NTSC 0-00, meaning that all versions of Prime are guaranteed to be logically completable when randomized.",
         ),
+        (
+            "What rooms in Magmoor Caverns not superheated?",
```
Miepee, line 99: TBH i like this question better phrased as something more along the lines of "Why don't I take heat damage in Magmoor Caverns, despite not having varia suit?"

randovania/games/prime1/game_data.py
```diff
@@ -96,9 +96,9 @@ def _exporter():
             "NTSC 0-00, meaning that all versions of Prime are guaranteed to be logically completable when randomized.",
         ),
         (
-            "What rooms in Magmoor Caverns not superheated?",
+            "Why do I not take heat damage in some rooms in Magmoor, despite not having Varia Suit?",
             "Elevator rooms, Storage Cavern, and every room between Twin Fires and the Phazon Mines and "
-            "Phendrana Drifts elevators (Far Magmoor), Burning Trail, and Lake Access.",
+            "Phendrana Drifts elevators (Far Magmoor), Burning Trail, and Lake Access are not coded to be superheated.",
```
Miepee, line 101: I'd rephrase the answer here a little. Something like "It is vanilla behaviour, that only some rooms are superheated. Here is a list of them: 
- entry1
- entry2"?

Miepee (changes requested):
Needs a changelog entry

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Resolver
 
+### Metroid Prime
+
+- Added: FAQ Entry about non-Superheated rooms
```
Miepee, line 14: wrong place. youre now suggesting that the seeds entry from below is a prime thing

Miepee (changes requested):

randovania/games/prime1/game_data.py
```diff
@@ -95,6 +95,15 @@ def _exporter():
             "break fixes present on that version are undone. This reverts the game to be functionally equivalent to "
             "NTSC 0-00, meaning that all versions of Prime are guaranteed to be logically completable when randomized.",
         ),
+        (
+            "Why do I not take heat damage in some rooms in Magmoor, despite not having Varia Suit?",
+            "Some rooms in Magmoor are not coded to be superheated. These include: "
+            "• Elevator Rooms "
```
Miepee, line 101: use minus here instead of unicode

Miepee (approved):

PR #6294, opened 2024-04-03
Add field in rdvgames that overrides Patch Data Factory output

MayberryZoom:
- Game specific PDFs now override `create_game_specific_data`, which is then invoked by the common PDF
- Adds a `custom_patcher_data` field to GamePatches
- Adds a package requirement for `json-delta`

Conflicts with #6250. Unsure if PDF changes conflict with others.

### DIFF
--- a/randovania/exporter/patch_data_factory.py
+++ b/randovania/exporter/patch_data_factory.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 from random import Random
 from typing import TYPE_CHECKING
 
+from json_delta import patch
+
 from randovania.game_description import default_database
 from randovania.layout import filtered_database
 
@@ -48,5 +50,10 @@ class PatchDataFactory:
     def game_enum(self) -> RandovaniaGame:
         raise NotImplementedError
 
-    def create_data(self) -> dict:
+    def create_game_specific_data(self) -> dict:
         raise NotImplementedError
+
+    def create_data(self) -> dict:
+        game_data = self.create_game_specific_data()
+
+        return patch(game_data, self.patches.custom_patcher_data, False)
--- a/randovania/exporter/patch_data_factory.py
+++ b/randovania/exporter/patch_data_factory.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from random import Random
 from typing import TYPE_CHECKING
 
-from json_delta import patch
+import json_delta
 
 from randovania.game_description import default_database
 from randovania.layout import filtered_database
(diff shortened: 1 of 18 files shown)


randovania/exporter/patch_data_factory.py
```diff
@@ -48,5 +50,10 @@ def __init__(
...
         raise NotImplementedError
 
-    def create_data(self) -> dict:
+    def create_game_specific_data(self) -> dict:
         raise NotImplementedError
+
+    def create_data(self) -> dict:
+        game_data = self.create_game_specific_data()
+
+        return patch(game_data, self.patches.custom_patcher_data, False)
```
Miepee, line 59: Can you write a test please, that tests that this behaves as it should with custom extra data?

randovania/games/prime2/exporter/patch_data_factory.py
```diff
@@ -939,7 +939,7 @@ def generate_patcher_data(
     :param cosmetic_patches:
     :return:
     """
-    return EchoesPatchDataFactory(description, players_config, cosmetic_patches).create_data()
+    return EchoesPatchDataFactory(description, players_config, cosmetic_patches).create_game_specific_data()
```
henriquegemignani, line 942: should be create_data

henriquegemignani: actually delete this function 🤣

henriquegemignani (changes requested):
You're missing including the value in `serialize_single` too.

And I agree that some test should use this. It can easily be changing, or adding, a new rdvgame to the test suite. Could be even your plando :P

randovania/exporter/patch_data_factory.py
```diff
@@ -3,6 +3,8 @@
 from random import Random
 from typing import TYPE_CHECKING
 
+from json_delta import patch
```
henriquegemignani, line 6: I prefer just `import json_delta` and then you use `json_delta.patch`

randovania/games/prime2/exporter/patch_data_factory.py
MayberryZoom: Delete `generate_patcher_data`?

henriquegemignani: yeah, it's an unused function

MayberryZoom:
> You're missing including the value in `serialize_single` too.

Wouldn't that make it write it to the rdvgame? I thought the idea was that it had to be manually added.

> And I agree that some test should use this. It can easily be changing, or adding, a new rdvgame to the test suite. Could be even your plando :P

Should there be one for each game, or will just one do?

henriquegemignani (approved):
Not sure if this need a changelog entry

PR #6308, opened 2024-04-05
MSR: fix spider boost hint

Miepee:
Currently this causes the text to be two hints. THis PR fixes it to be one hint.


dyceron (commented):
I was literally told not to do this. Also this is already displayed as one hint. There's no command outside the quotes here.

Miepee:
Oh didnt see the missing comma, that's confusing.
This should be cleaner, and still evaluates as one string:
grafik

Miepee:
Sorry, I can't count lol
grafik

dyceron (approved):

PR #6329, opened 2024-04-06
MF: More room renames

SchwartzGandhi:
ALL:
- `Entrance Hub` ->`Entrance Lobby`
- `Connection to [Sector]` -> `Glass Pipe to [Sector]`

SRX:
- `Zebesian Tower` and `Shaft South` -> `Zebesian Zig` and `Zebesian Zag` respectively
- `Shaft North` -> `Northwestern Intersection`
- `Short Shaft` -> `Checkpoint Intersection`

PYR:
- `Namihe's Abode` -> `Namihe's Lair`
- `Bob Room` -> `Bob's Abode`

AQA:
- `Owtch-Couch` -> `Owtch Atrium`
- `Owtch Cache` -> `C-Cache`

ARC:
- `Aerie` -> `Training Aerie`
- `Transformation Trial` -> `Transmutation Trial`

NOC: 
- `Menu Corridor` -> `Cavern Save Access`
- `Menu Save Room` -> `Cavern Save Room`

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 1 SRX.json
+++ b/randovania/games/fusion/logic_database/Sector 1 SRX.json
@@ -5121,7 +5121,7 @@
                                 ]
                             }
                         },
-                        "Door to Connection to Sector 3 (PYR)": {
+                        "Door to Glass Tube to Sector 3 (PYR)": {
                             "type": "and",
                             "data": {
                                 "comment": null,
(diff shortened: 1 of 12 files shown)


Miepee:
> Glass Pipe to [Sector]

Wasn't there an official term for those? 


> Zebesian Zig and Zebesian Zag respectively

I appreciate the pun, but these are so close to each other that I fear they'll get mistaken often

> Bob's Abode

NGL, prefer `Bob Room`. Wouldn't block this though.

> Owtches

IMO should be Owtch C-Cache. As for atrium, wondering whether we can have some reference with the L's (a reverse cyrillic g and normal cyrillic g) in there, but wouldnt be the end of the world if we dont


---

rest looks fine to me

Miepee (approved):

PR #6331, opened 2024-04-07
Discord Bot: mention game when describing preset

### LINKED ISSUE #3449: Discord Bot: when describing a preset, explicitly mention which game it belongs to
This can be done directly on the bot, or in the preset describer itself. Either way, a self contained change that won't touch many files.

Miepee:
Fix #3449

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [7.5.1] - 2024-04-??
 
+- Added: The Discord bot now mentions the game when describing a preset.
 - Changed: The output after verifying the installation has been made less misleading.
 - Nothing yet. Space reserved in case needed.
 - Changed: Show better errors on Linux and macOS when something goes wrong while trying to open a directory.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [7.5.1] - 2024-04-??
 
-- Added: The Discord bot now mentions the game when describing a preset.
 - Changed: The output after verifying the installation has been made less misleading.
-- Nothing yet. Space reserved in case needed.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [7.5.1] - 2024-04-??
 
+- Added: The Discord bot now mentions the game when describing a preset.
```
ThanatosGit, line 16: I thought about if this is relevant because it doesn't touch the applicaiton.
Looking into the logs, it seems to be mentioned but we had a `Discord Bot` section often but sometimes not.
Dunno if we want to reintroduce that.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Changed: The output after verifying the installation has been made less misleading.
 - Nothing yet. Space reserved in case needed.
```
ThanatosGit, line 18: Someone forgot to remove that in a PR

CHANGELOG.md
Miepee: I thought might be better just in case:tm:. Since it would affect people hosting the bot

CHANGELOG.md
Miepee: the whole changelog thing will need to be changed in the future anyway, since the next release is 8.0.0. So I'd only then would like to remove 7.5.1 and 7.6.0 and their placeholder text.

ThanatosGit: That doesn't sound like a reson to keep a placeholder if it already contains entries.

CHANGELOG.md
ThanatosGit: I already mentioned that it was done in before but what about the second part?

henriquegemignani: I'd say a section "Discord Bot" similar to the games. IMO these are more so our users will know differences in the bots.

henriquegemignani (approved):

PR #6334, opened 2024-04-07
MSR - Update Minimum Logic

dyceron:
image

### DIFF
--- a/randovania/games/samus_returns/logic_database/header.json
+++ b/randovania/games/samus_returns/logic_database/header.json
@@ -4893,7 +4893,7 @@
     "minimal_logic": {
         "items_to_exclude": [
             {
-                "name": "Baby",
+                "name": "Morph",
                 "when_shuffled": null
             },
             {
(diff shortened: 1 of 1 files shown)


Miepee (commented):
you got a reasoning for these items?

dyceron:
Bombs, Spider, and SJ are all items that can be used to reach Ridley. I wasn't sure if this was the best way to go about it, but you need one of those three to finish. So I just do this to ensure that they all aren't locked inside a Missile, Super Missile, Bomb, or Power Bomb block.

randovania/games/samus_returns/logic_database/header.json
```diff
@@ -4937,7 +5093,7 @@
                 "reason": "Victory Condition"
             }
         ],
-        "description": "Items checked for: Baby Metroid."
+        "description": "Items checked for: Morph Ball, Bomb, Spider, Space Jump, and Metroid DNA."
```
Miepee, line 5096: This should probably say `Spider Ball`

Miepee (approved):

PR #6354, opened 2024-04-09
Move hashes into GameData

### LINKED ISSUE #2853: Add hash words to GameData
The list of words provided by a game should come from the `GameData` class.

Words should be restricted to this regex: `^[A-Za-z-]+$`
Maybe enforce that it doesn't end or start in `-`?

Miepee:
Fix #2853


Miepee:
For me it 
1) was consistent with how we handle joke hints
2) means it's directly embedded into the executable for frozen builds. So we avoid the slowness of the json loading and deserialization, and makes it impossible for the user to mess with it.

henriquegemignani:
I like that it makes it less likely that users will mess with a thing that can cause issues.

randovania/games/game.py
```diff
@@ -171,6 +170,9 @@ class GameData:
     web_info: GameWebInfo = GameWebInfo()
     """Contains a handful of fields displayed primarily on the website."""
 
+    hash_words: list[str] | None = None
```
henriquegemignani, line 173: They way you ended up doing, hash words is not None for all our games and new games end up copying Blank's word list. So you can make this required.

henriquegemignani (changes requested):

randovania/games/game.py
duncathan: Yeah, I prefer new games copying Blank than the old behavior of combining the other lists

Miepee: > hash words is not None for all our games

maybe misunderstand, but that's not true right now. Corruption, super metroid and maybe some other games, have currently None for the hashwords: 
grafik


Or do you mean that I should change it so they also copy blanks words list and then remove the `| None` from here?

henriquegemignani: I have lost count of the number of games, great. Probably a good idea to copy then.

randovania/games/game.py
```diff
@@ -171,6 +170,9 @@ class GameData:
     web_info: GameWebInfo = GameWebInfo()
     """Contains a handful of fields displayed primarily on the website."""
 
+    hash_words: list[str] = None
```
henriquegemignani, line 173: Have no default value. Gotta move it up to make it work though

henriquegemignani (changes requested):

randovania/games/game.py
Miepee: oop

henriquegemignani (approved):

PR #6357, opened 2024-04-10
Map tracker: show if node is PonR

### LINKED ISSUE #4577: Make map tracker display dangerous actions
Idea I had earlier, might implement based on interest
For each node you can reach, calculate if from that node you can get back to the current node, and if not display a red "dangerous action" or something. This could help newbies using the map tracker to know what they should try to avoid.

Miepee:
fix #4577


Miepee:
grafik

grafik

Miepee:
Unsure if i should display it for every node or only resource nodes

Its nice to know when exactly you enter a ponr, but calculating that for ~half of am2rs total nodes takes around 20 seconds

randovania/gui/tracker_window.py
```diff
@@ -435,6 +437,20 @@ def update_locations_tree_for_reachable_nodes(self):
                         is_visible = False
 
                     node_item.setHidden(not is_visible)
+                    node_text = self._item_to_node[node_item].name
+                    # Check if you can reach back to current node from that node
+                    if is_visible:
```
duncathan, line 442: calculating safe nodes is awfully expensive but also *surely* there should be a dedicated function for it rather than duplicating the logic like this?

duncathan (changes requested):

Miepee:
will redo some day

PR #6365, opened 2024-04-11
Add automated report for how many tricks are documented

henriquegemignani:
image

### DIFF
--- a/randovania/cli/database.py
+++ b/randovania/cli/database.py
@@ -417,7 +417,7 @@ def _find_tricks_usage_documentation(requirement: Requirement) -> typing.Iterato
 
 
 def _flat_trick_usage(requirement: Requirement) -> dict[str, bool]:
-    doc = {}
+    doc: dict[str, bool] = {}
     for usage, documented in sorted(set(_find_tricks_usage_documentation(requirement))):
         doc[usage] = documented and doc.get(usage, True)
     return doc
--- a/randovania/cli/database.py
+++ b/randovania/cli/database.py
@@ -18,6 +18,8 @@ from randovania.lib import json_lib
 from randovania.lib.enum_lib import iterate_enum
 
 if typing.TYPE_CHECKING:
+    from argparse import _SubParsersAction
+
     from randovania.game_description.db.area import Area
     from randovania.game_description.game_description import GameDescription
     from randovania.game_description.requirements.base import Requirement
@@ -389,7 +391,7 @@ def pickups_per_area_command(sub_parsers):
     parser.set_defaults(func=pickups_per_area_command_logic)
 
 
-def _find_tricks_usage_documentation(requirement: Requirement) -> typing.Iterator[str, bool]:
+def _find_tricks_usage_documentation(requirement: Requirement) -> typing.Iterator[tuple[str, bool]]:
     from randovania.layout.base.trick_level import LayoutTrickLevel
 
     if not isinstance(requirement, RequirementArrayBase):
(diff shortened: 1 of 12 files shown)


toasterparty:
Sometimes, if multiple tricks within an "or" block are documented by the same video, we put the video higher up in the logic expression tree. Is this case handled by your counting method?

Miepee:
IMO if the same video covers multiple tricks in an OR block, then each thing should still have a comment but appropriately timestamped

Also, tests fail.

henriquegemignani:
> Sometimes, if multiple tricks within an "or" block are documented by the same video, we put the video higher up in the logic expression tree. Is this case handled by your counting method?

It does not. I meant to open this PR a month ago but kept struggling with these cases, so I'm going with the simplest possible.

randovania/cli/database.py
```diff
@@ -388,7 +389,7 @@ def pickups_per_area_command(sub_parsers):
     parser.set_defaults(func=pickups_per_area_command_logic)
 
 
-def _find_uncommented_tricks(requirement: Requirement):
+def _find_tricks_usage_documentation(requirement: Requirement) -> typing.Iterator[str, bool]:
```
Miepee, line 392: mypy

randovania/cli/database.py
```diff
@@ -397,63 +398,74 @@ def _find_uncommented_tricks(requirement: Requirement):
...
+                    for req in trick_resources
+                )
+            ),
+            requirement.comment is not None,
         )
 
 
-def uncommented_trick_usages_logic(args):
+def _flat_trick_usage(requirement: Requirement) -> dict[str, bool]:
+    doc = {}
```
Miepee, line 418: mypy

randovania/cli/database.py
```diff
@@ -397,63 +398,74 @@ def _find_uncommented_tricks(requirement: Requirement):
...
+    for source, connections in area.connections.items():
+        paths[source.name] = {}
+        for target, requirement in connections.items():
+            trick_documentation = _flat_trick_usage(requirement)
+            if trick_documentation:
+                paths[source.name][target.name] = trick_documentation
+    return paths
+
+
+def trick_usage_documentation_logic(args):
```
Miepee, line 435: mypy

randovania/cli/database.py
```diff
@@ -397,63 +398,74 @@ def _find_uncommented_tricks(requirement: Requirement):
...
+                        for target_name, trick_documentation in connections.items():
+                            lines.append(f"### {source_name} -> {target_name}:")
+                            for it, used in trick_documentation.items():
+                                lines.append(f"- [{'X' if used else ' '}] {it}")
 
     output_path.write_text("\n".join(lines))
 
 
-def uncommented_trick_usages_command(sub_parsers):
+def trick_usage_documentation_command(sub_parsers):
```
Miepee, line 457: mypy

randovania/cli/database.py
```diff
@@ -397,66 +400,77 @@ def _find_uncommented_tricks(requirement: Requirement):
...
+                        for target_name, trick_documentation in connections.items():
+                            lines.append(f"### {source_name} -> {target_name}:")
+                            for it, used in trick_documentation.items():
+                                lines.append(f"- [{'X' if used else ' '}] {it}")
 
     output_path.write_text("\n".join(lines))
 
 
-def uncommented_trick_usages_command(sub_parsers):
+def trick_usage_documentation_command(sub_parsers: _SubParsersAction) -> None:
```
Miepee, line 459: Do we not have a test for this subcommand? Codecv doesn't seem to report anything at all.

henriquegemignani: probably not, a lot of the CLI things does not have tests

Miepee: would be nice to have tho.

henriquegemignani: Get my very low effort test, all I can have for you today

Miepee (approved):

PR #6371, opened 2024-04-12
Add basic node list caching for speedup

henriquegemignani:
Importing 1000 permalinks now takes about 7.5s vs 9.7s.


henriquegemignani:
These annoying tests that fake the db :(

PR #6375, opened 2024-04-12
Use orjson

henriquegemignani:
`randovania database refresh-game-description` now takes 2.2s instead of 3.2s. However on main it takes 4.1s so a nice chunk is by having indent=2. See #6222.

Keeping this as draft because I'm not completely happy. This might be as good as it gets with JSON files, but we could experiment with other file formats.


henriquegemignani:
The PR being in history is enough of a reference point.

PR #6378, opened 2024-04-12
Updated Echoes database for new patcher integration

ShinWolford10:
Added the schema from the new patcher for the 'extra' field entry in item pickup locations.

### DIFF
--- a/randovania/games/prime2/logic_database/Agon Wastes.json
+++ b/randovania/games/prime2/logic_database/Agon Wastes.json
@@ -625,9 +625,9 @@
                     "node_type": "pickup",
                     "heal": false,
                     "coordinates": {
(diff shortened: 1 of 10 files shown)


henriquegemignani:
I think I'm gonna improve the human readable version of the `extras` dict to handle nested dicts

ThanatosGit:
> I think I'm gonna improve the human readable version of the `extras` dict to handle nested dicts

Is this something you want to do before merging it or just a thing which should be done in the future?

Miepee:
DZ tried, and ultimately it was decided to not do that.
The thing that's stopping this PR is dunc looking over this to check that the data is correct

duncathan (approved):

PR #6379, opened 2024-04-12
Get sentry url from configuration

henriquegemignani:
This means sentry is only initialized in builds provided via our CI, and the official server

---

I'm not worried about these values being secret, but I do want to stop reports from people who make their own executables.

### DIFF
--- a/randovania/monitoring/__init__.py
+++ b/randovania/monitoring/__init__.py
@@ -13,9 +13,11 @@ import sentry_sdk.scrubber
 
 import randovania
 
-_CLIENT_DEFAULT_URL = "https://44282e1a237c48cfaf8120c40debc2fa@o4504594031509504.ingest.sentry.io/4504594037211137"
-_SERVER_DEFAULT_URL = "https://c2147c86fecc490f8e7dcfc201d35895@o4504594031509504.ingest.sentry.io/4504594037276672"
-_BOT_DEFAULT_URL = "https://7e7607e10378497689b443d8922870f7@o4504594031509504.ingest.sentry.io/4504606761287680"
+
+class HomeEventScrubber(sentry_sdk.scrubber.EventScrubber):
+    def scrub_dict(self, d):
+        super().scrub_dict(d)
+        _filter_user_home(d)
 
 
 def _filter_data(data, str_filter: typing.Callable[[str], str]) -> typing.Any | None:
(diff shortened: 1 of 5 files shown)


randovania/monitoring/__init__.py
```diff
@@ -126,25 +123,22 @@ def traces_sampler(sampling_context):
...
-def server_init(sampling_rate: float):
-    return _init(True, _SERVER_DEFAULT_URL, sampling_rate=sampling_rate)
+def server_init(sampling_rate: float) -> None:
+    return _init(True, "server", sampling_rate=sampling_rate)
 
 
-def bot_init():
-    return _init(False, _BOT_DEFAULT_URL)
+def bot_init() -> None:
+    return _init(False, "bot")
```
ThanatosGit, line 141: `bot` needs to be a key in `sentry_url = configuration["sentry_urls"][url_key]` but I cannot find any place where this is set.

randovania/monitoring/__init__.py
```diff
@@ -126,25 +123,22 @@ def traces_sampler(sampling_context):
...
 
     # Ignore the "packet queue is empty, aborting" message
     # It causes a disconnect, but we smoothly reconnect in that case.
     sentry_sdk.integrations.logging.ignore_logger("engineio.client")
 
 
-def server_init(sampling_rate: float):
-    return _init(True, _SERVER_DEFAULT_URL, sampling_rate=sampling_rate)
+def server_init(sampling_rate: float) -> None:
+    return _init(True, "server", sampling_rate=sampling_rate)
```
ThanatosGit, line 137: `server` needs to be a key in `sentry_url = configuration["sentry_urls"][url_key]` but I cannot find any place where this is set.

randovania/monitoring/__init__.py
```diff
@@ -13,9 +13,11 @@
 
 import randovania
 
-_CLIENT_DEFAULT_URL = "https://44282e1a237c48cfaf8120c40debc2fa@o4504594031509504.ingest.sentry.io/4504594037211137"
-_SERVER_DEFAULT_URL = "https://c2147c86fecc490f8e7dcfc201d35895@o4504594031509504.ingest.sentry.io/4504594037276672"
-_BOT_DEFAULT_URL = "https://7e7607e10378497689b443d8922870f7@o4504594031509504.ingest.sentry.io/4504606761287680"
+
+class HomeEventScrubber(sentry_sdk.scrubber.EventScrubber):
+    def scrub_dict(self, d):
```
ThanatosGit, line 18: `-> None`

ThanatosGit (changes requested):
Maybe I'm completly missing a step with the setup but I somehow expect the urls to be present anywhere in this repo.
Like the server url in the `configuration.json` for the docker build of the server?

henriquegemignani:
Configuration comes from a json file. The values in question will show up nowhere because they're being treated as secrets.

ThanatosGit:
My concern was not about the values themself but more about the fact that I cannot even see where the secret is added because there is (at least I cannot find) any reference to such a secret. Not in the ci configs nor in a dockerfile.

But as I don't have any access to any secrets, I just trust that everything is set properly.

henriquegemignani:
You do because I shared access a long ago :P

But it's directly in the servers as a docker secret. I fixed staging's but not prod still because it involves some downtime ugh

henriquegemignani:
But they come from here: https://github.com/randovania/randovania/blob/main/tools%2Fserver-docker%2Fdocker-compose.yml#L84-L87

ThanatosGit:
Hmmm GitHub secrets are definitely not visible and the only thing I have in my keepass for randovania servers is the sqlite web view thing and a ssh login url for the ci worker without a username xD.
Must have missed it but it's totally off topic, so it's fine.

ThanatosGit (approved):

PR #6382, opened 2024-04-14
Fusion: nightmare nook, s2 data hub, reverse charge core arena item videos

bariasu:


SchwartzGandhi (commented):
If you could truncate these links to be `youtu.be/[vidID]` that would be great.
This isn't an explicit requirement but it just helps to be consistent with the rest of the links.

SchwartzGandhi (approved):

PR #6392, opened 2024-04-15
AM2R: Fix tank rename in MW

Miepee:


CHANGELOG.md
```diff
@@ -39,7 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### AM2R
 
-- Changed: All ammo tanks are now consistently being referred to as "Tanks" rather than "Expansions".
+- Changed: All ammo tanks are now consistently being referred to as "Tanks" rather than "Expansions". If you wish to continue older Multiworld sessions, you will need to re-export and delete the save file.
```
henriquegemignani, line 42: What would be necessary to keep old save files compatible?

Miepee: thinking about it, just a quick change, lemme do that rq.

henriquegemignani (approved):
All that's left is figuring out a way to migrate the inventory of existing sessions. I guess this can be a server side migration, need to think about that one

PR #6394, opened 2024-04-15
Require Missiles fireable for dmg templates

JaggerTSG:
Assumes Missile Data enables firing, regardless.


shark20061 (approved):
Seems fine to me

PR #6396, opened 2024-04-16
MF: serris escape rework

SchwartzGandhi:
add a missile data requirement to break the two missile blocks. (previously just accounted for the ammo requirement)

also added a new node `Beside other pickup` for betterreadability


shark20061:
I thought that the logic would consider you having 0 missiles until you can use them.  If that is not the case, there is a lot of logic missing a missile data check.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -504,10 +504,19 @@ Extra - room_id: [13]
   * Layers: default
   > Pickup (Hidden Missile Tank)
       Trivial
-  > Pickup (Missile Tank)
-      Missiles ≥ 2
   > Door to Escape Save Room
       Can Use Any Bombs
+  > Beside Other Pickup
+      Missile Data and Missiles ≥ 2 and Morph Ball
```
Miepee, line 510: This is wrong. Using a missile resource already implies some way to fire them (which right now is having missile data).

Miepee (changes requested):

randovania/games/fusion/logic_database/Sector 4 AQA.txt
SchwartzGandhi: ah I see, Sorry, I wasn't aware of this.

Miepee (approved):

PR #6409, opened 2024-04-19
Dread: Fix DNA slider

Miepee:


henriquegemignani (approved):

henriquegemignani:
Wonder if we can make a helper widget class for this kind of behavior 🤔

PR #6419, opened 2024-04-20
AM2R: Fix mypy errors

Miepee:


henriquegemignani (changes requested):
Add the relevant files to `pyproject.toml`.

randovania/games/am2r/exporter/patch_data_factory.py
```diff
@@ -131,7 +133,7 @@ class AM2RPatchDataFactory(PatchDataFactory):
     def _create_pickups_dict(
         self, pickup_list: list[ExportedPickupDetails], text_data: dict, model_data: dict[str, int], rng: Random
     ) -> dict:
-        pickup_map_dict = {}
+        pickup_map_dict: dict[str, dict[str, Any]] = {}
```
henriquegemignani, line 136: Make a TypedDict

henriquegemignani (changes requested):

randovania/games/am2r/exporter/hint_namer.py
```diff
@@ -102,6 +102,7 @@ def format_guaranteed_resource(
         )
 
     def format_location_hint(self, game: RandovaniaGame, pick_hint: PickupHint, hint: Hint, with_color: bool) -> str:
+        assert hint.precision is not None
```
duncathan, line 105: I think errors like this need to be fixed properly or at least have the root cause marked with a FIXME comment. it is a good thing that mypy is alerting us to problems this one

duncathan (changes requested):
in general I think if we're going to start actively going after mypy errors, we need to settle on some best practices to adhere to

randovania/games/am2r/exporter/hint_namer.py
henriquegemignani: `Hint.precision` can be None during the generator, because it's assigned later. When it reaches the exporter it should never be None.
Probably case of needing a different type for during generator and afterwards, but that's very annoying code.

duncathan: I don't mind that particular case, I suppose. I'm more bothered by the possibility of it being None because it isn't a normal hint to begin with. The ambiguity between those cases is a lot more problematic imo

Miepee:
will just close for now. conflcits + needs some other discussion on how to properly handle types

PR #6421, opened 2024-04-20
fix CS mypy errors

duncathan:
TODO: pdf


henriquegemignani:
I'm fine with everything up to now

Miepee:
What do we do about this PR? Do we just let it rot until PDF get properly done into subclasses, or only merge typing for everything-but-pdf and do pdf later?

duncathan:
i mean it's just not finished and it's marked as a draft idk what needs to be done with it besides mii finishing it

Miepee:
I just wanted to reduce the amount of open draft PRs, that's all.

duncathan:
all done, requires https://github.com/cave-story-randomizer/cave-story-randomizer/pull/181 first

pyproject.toml
```diff
@@ -75,7 +75,7 @@ exporters = [
     "am2r-yams>=2.8.1",
 
     # Cave Story
-    "cave-story-randomizer>=2.3.3",
+    "cave-story-randomizer>=2.4.2",
```
Miepee, line 78: changelog?

randovania/games/cave_story/exporter/hint_namer.py
```diff
@@ -64,6 +64,8 @@ def format_area(self, location: PickupLocation, with_region: bool, with_color: b
         return area.name
 
     def format_location_hint(self, game: RandovaniaGame, pick_hint: PickupHint, hint: Hint, with_color: bool) -> str:
+        assert hint.hint_type == HintType.LOCATION
```
Miepee, line 67: iirc @henriquegemignani wanted to have a fixme for this, but could be wrong.

randovania/games/cave_story/exporter/patch_data_factory.py
```diff
@@ -42,12 +45,50 @@ def game_enum(self) -> RandovaniaGame:
...
+        self.maps["Start"]["pickups"]["0201"] = self._create_starting_script()
+
+        data: CaverData = {
+            "maps": self.maps,
+            "other_tsc": {"Head": self._head_tsc_edits()},
+            "mychar": self.cosmetic_patches.mychar.mychar_bmp(Random(self.seed_number)),
+            "hash": get_ingame_hash(self.description.shareable_hash_bytes),
+            "uuid": f"{{{self.players_config.get_own_uuid()}}}",
+        }
+        return typing.cast(dict, data)
```
Miepee, line 60: is CaverData not inheriting from dict? is an explicit cast really necessary here?

randovania/games/cave_story/exporter/patch_data_factory.py
```diff
@@ -158,24 +215,79 @@ def create_game_specific_data(self) -> dict:
...
                 continue
 
             if item.short_name == "lifeCapsule":
-                life = starting_items[item]
+                self.life = self.starting_items[item]
                 continue
 
             if item.short_name == "puppies":
-                num_puppies = starting_items[item]
+                num_puppies = self.starting_items[item]
```
Miepee, line 290: why do these variables need to be self?

randovania/games/cave_story/exporter/patch_data_factory.py
```diff
@@ -204,128 +316,63 @@ def create_game_specific_data(self) -> dict:
             if (item_num is None) == (arms_num is None):
                 raise ValueError(f"{item.long_name} must define exactly one of item_num and arms_num.")
 
-            equip_num |= item.extra.get("equip", 0)
-            items_extra += item.extra.get("extra", "")
+            self.equip_num |= item.extra.get("equip", 0)
```
Miepee, line 319: ditto

pyproject.toml
duncathan: we currently use 2.4.1 (see requirements.txt). the only change is the typing backend which doesn't affect users

randovania/games/cave_story/exporter/patch_data_factory.py
duncathan: it is necessary, because dict type parameters are invariant. removing the cast causes an error

randovania/games/cave_story/exporter/patch_data_factory.py
duncathan: because they're used in two functions

randovania/games/cave_story/exporter/hint_namer.py
duncathan: the fixme is on the `Hint` class

randovania/games/cave_story/exporter/patch_data_factory.py
Miepee: Ah hm. Well then.

pyproject.toml
Miepee: should this not create a conflict then if we're currently using 2.4.1? :thinking:

randovania/games/cave_story/exporter/patch_data_factory.py
Miepee: can you then at least define them under class level, rather than in a function?  Feels easier for me to follow.

pyproject.toml
duncathan: I mean... it doesn't? so who cares? why would this create a conflict, anyway?

Miepee: I assumed this line was 2.4.1 on main, but shows 2.3.3 on the diff, thus should lead to a conflict.
But it's not this line that's 2.4.1 on main, it's the requirements.txt as you pointed out.

It doesn't matter, was just trying to understand it for myself.

Miepee (approved):

PR #6422, opened 2024-04-21
Dread: Add Pseudo Wave Beam to break Blob above Blue teleportal in Cataris

Tyranisaur:


CHANGELOG.md
```diff
@@ -59,6 +59,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed: Customizing a preset where EMMI and Bosses were turned off for DNA placement won't have the DNA slider be initially enabled.
 - Removed: The FAQ entry stating that only the English language is supported has been removed, as all languages are patched since version 7.4.0.
 
+#### Logic database
+
+##### Cataris
+
+- Added: Pseudo Wave Beam(Beginner) to break the blob above Blue Teleportal to Artaria, from the left side of the wall.
```
Miepee, line 66:
```suggestion
- Added: Pseudo Wave Beam (Beginner) to break the blob above Blue Teleportal to Artaria, from the left side of the wall.
```

Logarhythmic (approved):
Cool pseudo, that teleporter doesn't get used enough.

dyceron (approved):
I wasn't sure if DT Pseudos should be Beginner, but it looks like they do exist in the DB already so this is fine.

PR #6427, opened 2024-04-21
Bump OSRR to v1.2.0

dyceron:


requirements.txt
```diff
@@ -178,7 +178,7 @@ open-dread-rando==2.9.0
     # via randovania (setup.py)
 open-prime-rando==0.12.0
     # via randovania (setup.py)
-open-samus-returns-rando==1.1.1
+open-samus-returns-rando==1.2.0
```
Miepee, line 181: Since this is our version we're planning to use for stable, lets bump the version in pyproject.toml too

dyceron: Done

Miepee (approved):

PR #6435, opened 2024-04-22
Prime3: Add Landing Site quick fill option

### LINKED ISSUE #2335: Prime 3: add Quick Fill option for starting location with all ship landing sites in the game
Like the title says, but maybe we could go a step further. Currently, Quick Fill options are buttons that automatically fill in choices in the list below, right? What if, at least for Prime 3 since it has multiple landing sites, that was made into a tick option itself so you could have save stations, landing sites or both?

Miepee:
Fixes #2335

### DIFF
--- a/randovania/games/prime3/gui/preset_settings/__init__.py
+++ b/randovania/games/prime3/gui/preset_settings/__init__.py
@@ -9,17 +9,17 @@ if TYPE_CHECKING:
 
 
 def prime3_preset_tabs(editor: PresetEditor, window_manager: WindowManager) -> list[type[PresetTab]]:
+    from randovania.games.prime3.gui.preset_settings.corruption_starting_area_tab import PresetCorruptionStartingArea
     from randovania.gui.preset_settings.generation_tab import PresetGeneration
     from randovania.gui.preset_settings.location_pool_tab import PresetLocationPool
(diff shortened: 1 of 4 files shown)


dyceron:
Could you add a screenshot of this?

Miepee:
grafik

dyceron (changes requested):
This is missing some non excluded Landing Sites.

Miepee:
Which ones?

dyceron:
Reference this list https://metroid.fandom.com/wiki/List_of_Landing_Sites

Miepee:
I did
I excluded G.F.S Olympus, All the Seeds, and Phaaze.
See the explanation blurb at the top

dyceron:
The blurb does not mention the ones you have to unlock. Regardless, it'd be future proofing in case we ever get an option to start at those landing sites which need to be unlocked first.

Miepee:
> The blurb does not mention the ones you have to unlock. 

1. Because the DB doesn't even mark those as potential starting areas, so you can't even manually select them. Notice how Bryyo's Hangar Bay, or Elysia's Federation Landing Site are completely missing as a room.
grafik
grafik

2. If those areas mark their landing sites as viable starts, then they will be automatically checked here as well.

dyceron:
Norion Cargo Dock B needs to be removed then since that one has to be unlocked first

dyceron (approved):

PR #6444, opened 2024-04-23
Fix options migration for Dread's export params

henriquegemignani:


Miepee (approved):

randovania/interface_common/persisted_options.py
```diff
@@ -196,7 +202,7 @@ def _only_new_fields(options: dict) -> dict:
     _only_new_fields,  # added DebugConnectorBuilder's layout_uuid
     _only_new_fields,  # added Dread's music sliders
     _only_new_fields,  # added Dread's "Access Permanently Closed" alt-shield
-    _only_new_fields,  # added Dread Ryujinx export to Unix systems
+    _dread_linux_ryujinx_path,  # added Dread Ryujinx export to Unix systems
```
Miepee, line 205: actually, will this not screw people over who already migrated?

henriquegemignani: Yes, but dev build ahoy

henriquegemignani: Though only if they've migrated _and_ made some change that saves.

Miepee: fair nuff

PR #6450, opened 2024-04-23
CS: Clean up last cave

Miepee:
Draft until next release
I really would like to have health requirements for final bosses, but they're difficult to figure out properly, because you dont have any opportunity to full heal after misery/doctor/undead core and heavy press/ballos.

That said, I'm pretty sure the only thing where something like that might matter would be entrance rando. So could maybe theoreticaly only have the health reqs on undead core + ballos.


Miepee:
Note to self: don't use damage reqs for bosses, use the health item requirement instead.

Miepee:
i'll go revisit this some day. in the meantime it can be found in the history

PR #6454, opened 2024-04-24
Fix Tracker for not-Echoes

henriquegemignani:


duncathan (dismissed):
I'd rather at *least* put this in a dedicated function and calling it if game is echoes rather than returning early if it isn't

henriquegemignani:
I don't want to spend any brainpower fixing this.

Miepee:
`# TODO`? :stuck_out_tongue_closed_eyes:

duncathan:
if nothing else you *need* to slap a FIXME comment on it lol

henriquegemignani:
> I'd rather at _least_ put this in a dedicated function and calling it if game is echoes rather than returning early if it isn't

Eh, it's already called `update_translator_gates`.

henriquegemignani:
I'm merging because the tracker is broken otherwise

ThanatosGit:
> I'm merging because the tracker is broken otherwise

Then you need to dismiss requested changes.

henriquegemignani:
> > I'm merging because the tracker is broken otherwise
> 
> Then you need to dismiss requested changes.

Reviews don't normally block merging, weird.

PR #6465, opened 2024-04-30
AM2R: add msr joke hint

Miepee:


randovania/games/am2r/exporter/joke_hints.py
```diff
@@ -8,6 +8,7 @@
     "*Adam has not yet authorized the use of this hint.*",
     "*Did you know that Frog Legs are delicious with some Goat Butter, Garlic and Parsley?*",
     "*If you see an Omega Metroid shinesparking towards your general direction, run.*",
+    "*For the hint to Return... we must get rid of Proteus Ridley, find the Metroid DNA and repair the Chozo Seals...*",
```
dyceron, line 11:
```suggestion
    "*For the hint to Return... we must get rid of Proteus Ridley, find the Metroid DNA, and repair the Chozo Seals...*",
```

dyceron (changes requested):

randovania/games/am2r/exporter/joke_hints.py
Miepee: If I do this, then the line gets 1 character over the limit :'(

dyceron: Instead of `get rid of`, could do eradicate, remove, eliminate, defeat, destroy

Miepee: `get rid of` is part of the original quote.
> For peace to return... we must get rid of Smithy, find the Star Pieces, and repair the Star Road...

dyceron: Oh

Miepee: Could. I thought it'd be better to be completely on the nose just to make sure people don't think there's some ridley fight somewhere.

dyceron: For hints to Return?

dyceron (approved):

PR #6480, opened 2024-05-03
Dont hardcode prime 3 export message

Miepee:
Should keep the same behaviour as before. Maybe the window is a little bigger, but i dont really care about that.

grafik

### DIFF
--- a/randovania/games/prime3/exporter/game_exporter.py
+++ b/randovania/games/prime3/exporter/game_exporter.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import dataclasses
-from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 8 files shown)


ThanatosGit (changes requested):
Don't hardcode it, is a good idea.

I don't get why there is the whole game options added with input_path and output_path because nothing is used and required. It all ends up in the `NotImplementedError` and is only for "whenever anyone implements a corruption patcher".... but I still prefer to not add dead code in advance.

Miepee:
> I don't get why there is the whole game options added with input_path and output_path because nothing is used and required

Cause I copy pasted from Blank Dev Game lol. That's the default implementation basically.

Miepee:
addresed

ThanatosGit (approved):

PR #6481, opened 2024-05-03
Fusion: remove export accept button

Miepee:
The accept buttons starts exporting which errors due to not implemented.

So we just remove the accept button.


ThanatosGit (approved):
Love to see people asking why the Accept button doesn't show up 🙃

PR #6483, opened 2024-05-03
Refactor to provide custom ui names for elevators from DB

ThanatosGit:


ThanatosGit:
Should be OK?
It adds `ui_custom_name` (if anyone has a better name, i'm fine with find&replace) to the db for dock nodes, which can be used to represent a different name to the user facing UI.

I nuked the short names from Prime 1, which are only used on the tracker. Besides that the short names are gone, it shouldn't be an issue because the state.json of the tracker uses fully qualified node location (region,area,node).

The changes in `randovania/gui/lib/node_list_helper.py` are meh. Before, it was replacing the area name with a custom name because elevators were defined by region+area. That did work because of default_node but is conceptually totally wrong now beause default_node is deprecated and the custom names are defined on the node, not an area.
Now, I need the node's `ui_custom_name`.

dyceron (dismissed):
This looks good to me, but more eyes on this would be good. Missing a changelog entry.

Miepee:
What actually needs the changelog here? removal of UI names for prime1?

ThanatosGit:
The only thing changes for the users, would be that the short names for prime1 map tracker changes to be the full custom name.
It will show the `UI_CUSTOM_NAMES` now instead of the `SHORT_UI_CUSTOM_NAMES` from `randovania/games/prime1/patcher/prime1_elevators.py`.
Most of it means, that the word `Elevator` is now added, two times additonally the word `Magmoor`, `Exterior Docking Hangar` instead of `Frigate Destroyed` and `Essence Dead Cutscene` instead of `Essence Dead`.
If Toaster wants to keep the short names in the map tracker, I guess we could still add this a special case for prime1 and add it to the extra section of the few nodes.

randovania/game_description/game_migration.py
```diff
@@ -299,6 +299,23 @@ def _migrate_v23(data: dict) -> dict:
...
+    # find all dock nodes
+    all_nodes = [
+        node
+        for region in regions_data
+        for area_name, area in region["areas"].items()
+        for node_name, node in area["nodes"].items()
+        if node["node_type"] == "dock"
+    ]
+    for node in all_nodes:
+        node["ui_custom_name"] = None
```
Miepee, line 314: Will this not cause issues for games that actually *do* define this now? i.e impact crater in prime1

ThanatosGit: Not really but you are right that this isn't a good way.
It isn't an issue with the databases because I migrated them and then edited the custom names, like you can see in the json files. 
I should take the current state of the databases in the PR as blue print and add the ui_custom_names in the migration files with `region/area/node => custom name`, so the migration would create exactly the state of how it is in the PR.

Also: Is the custom name something for the pretty print iff != null/None?

Miepee: Oh I didn't even think of that.
Yeah would be nice to have it in the pretty print.

Miepee:
Mypy fails

ThanatosGit:
> Mypy fails

Yes, but AFAIK it's because of the funny thing I posted into the discord a couple days ago:`reviewdog: post failed for mypy: fail to parse diff: GET https://api.github.com/repos/randovania/randovania/pulls/6483: 406 Sorry, the diff exceeded the maximum number of lines (20000) [{Resource:PullRequest Field:diff Code:too_large Message:}]`

Miepee:
Oh whoops, youre right, sorry.

randovania/game_description/game_migration.py
ThanatosGit: I rebased and moved the stuff to the migration file. Like before: First commit is relevant stuff, second commit is only `database refresh-game-description`

Miepee (commented):
Changelog entry that mentions the changed elevator names for prime1 map tracker and this is fine to merge IMO

ThanatosGit:
> Changelog entry that mentions the changed elevator names for prime1 map tracker and this is fine to merge IMO

Added a changelog entry. I hope that's the least confusing explanation of what happens on the user side.

Miepee (approved):

PR #6484, opened 2024-05-04
Fix wrong init / visual for NTSC when no settings are in user folder

ThanatosGit:


Miepee (changes requested):
Classic

Missing changelog entry btw

ThanatosGit:
ngl, this is so niche that I intentionally didn't add one...but here you go.

Miepee (approved):

PR #6485, opened 2024-05-05
Dread: Add usage of Ice Missiles in Dairon - Big Hub

Tyranisaur:


randovania/games/dread/logic_database/Dairon.txt
```diff
@@ -285,6 +285,14 @@ Extra - asset_id: collision_camera_004
               All of the following:
                   Flash Shift
```
XenoWars, line 286: Unrelated but this could also use a flash upgrade.

XenoWars (approved):

randovania/games/dread/logic_database/Dairon.txt
Tyranisaur: Out of scope.

PR #6507, opened 2024-05-09
Remove Door as start location

ThanatosGit:
Dunno if this goes to `Logic database` section in the changelog.
Logically it is fine to start anywhere but at the same time it's just wrong to start in a door and it fails during patch time not generaton time.


Miepee (approved):
We should ensure that this isnt the only mistake like this

PR #6512, opened 2024-05-11
MSR: small export window niceties

### LINKED ISSUE #6511: MSR export dialog niceties
- Change buttons to say "Select Folder" instead of "Select File"
- Detect a modified romfs by checking if `romfs/custom_names.json` exists.

Miepee:
Fixes #6511

### DIFF
--- a/randovania/games/samus_returns/gui/dialog/game_export_dialog.py
+++ b/randovania/games/samus_returns/gui/dialog/game_export_dialog.py
@@ -81,14 +81,17 @@ def romfs_validation(line: QtWidgets.QLineEdit) -> bool:
         return True
 
     path = Path(line.text())
-    return not all(
-        p.is_file()
-        for p in [
-            path.joinpath("system", "files.toc"),
-            path.joinpath("packs", "system", "system_discardables.pkg"),
-            path.joinpath("packs", "maps", "s000_surface", "s000_surface.pkg"),
-            path.joinpath("packs", "maps", "s010_area1", "s010_area1.pkg"),
-        ]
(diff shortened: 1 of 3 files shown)


randovania/games/samus_returns/gui/dialog/game_export_dialog.py
```diff
@@ -81,15 +81,17 @@ def romfs_validation(line: QtWidgets.QLineEdit) -> bool:
...
+        not all(
+            p.is_file()
+            for p in [
+                path.joinpath("system", "files.toc"),
+                path.joinpath("packs", "system", "system_discardables.pkg"),
+                path.joinpath("packs", "maps", "s000_surface", "s000_surface.pkg"),
+                path.joinpath("packs", "maps", "s010_area1", "s010_area1.pkg"),
+            ]
+        )
+    ) and all(p.is_file() for p in [path.joinpath("custom_names.json")])
```
ThanatosGit, line 94: Uh, I'm always confused by this logic. Just to double check: Did you test that?
Iirc this should return `true` for "this is invalid, set the error border" and `false` for "there is no error, don't set error border".
If all of the L88-91 files exists, `all` returns `true` and the `not` in front of it makes it `false`.
With `and` for the `custom_names.json` you have no way to flip that boolean but you need to flip it to `true` if the file exists.

Miepee: You were right, whoops.
Fixed + tested

ThanatosGit (approved):

PR #6514, opened 2024-05-11
AM2R: Expose Queen locked doors

Miepee:


JeffGainsNGames (approved):
Another use case I thought about for this was for those people who like to 100% seeds, they may be forced to fight Queen then loop back.

PR #6515, opened 2024-05-11
Ignore derived nodes for documentation

henriquegemignani:


Miepee (commented):
Pretty sure needs changelog, as the command is available externally

henriquegemignani:
I don't really consider the cli as part of the user interface of rdv as it's pretty rare but 🤷

duncathan (approved):
yeah, I don't consider cli changes to need a changelog either

PR #6519, opened 2024-05-11
Add SR to mypy files

ThanatosGit:


randovania/games/samus_returns/exporter/game_exporter.py
```diff
@@ -73,7 +73,7 @@ def _do_export_game(
         monitoring.set_tag("msr_target_platform", export_params.target_platform.value)
         monitoring.set_tag("msr_target_version", export_params.target_version.value)
 
-        from open_samus_returns_rando.version import version as open_samus_returns_rando_version
+        from open_samus_returns_rando.version import version as open_samus_returns_rando_version  # type: ignore
```
Miepee, line 76: I feel like if we're adding mypy, we might as well use typings from the patcher. Should just be adding one file somewhere.

ThanatosGit: Yes, but I won't update the patcher now for this.

Miepee: eh fair enough

Miepee (approved):

randovania/games/samus_returns/exporter/game_exporter.py
ThanatosGit: I open an issue for it in the patcher repo

PR #6522, opened 2024-05-12
Dread/slope climb

Tyranisaur:


randovania/games/dread/logic_database/header.json
```diff
@@ -1225,6 +1225,12 @@
                 "description": "Many floors can be clippped through with some or all of the following items: Speed Booster, Flash Shift, and Grapple Beam. See https://dreadwiki.hijumpboots.com/en/tricks/clipping",
                 "require_documentation_above": 0,
                 "extra": {}
+            },
+            "SlopeClimb": {
+                "long_name": "Climb Sloped Surfaces",
+                "description": "With precise movement, Flash Shift can be used to climb Sloped surfaces.",
```
MayberryZoom, line 1231: I don't think this description should only mention Flash Shift, as this trick would definitely include things like Artaria Water Slope Skip.

This should also mention the fact you can use free-aim to jump higher off of slopes. I'd recommend something like:
`It is possible to gain height on sloped surfaces with good movement. This may require usage of techniques/items such as free-aim, Flash Shift, or Phantom Cloak.`

randovania/games/dread/logic_database/Ghavoran.txt
```diff
@@ -1508,7 +1508,7 @@ Extra - asset_id: collision_camera_021
           Diagonal Bomb Jump (Intermediate) and Lay Normal Bomb
           Flash Shift and Speed Booster and After Ghavoran - Ammo Recharge Enky Above and Speed Booster Conservation (Intermediate)
           # https://www.youtube.com/watch?v=rQiQbXvvCSA
-          Flash Shift and Climb Sloped Tunnels (Advanced) and Wall Jump (Advanced)
+          Flash Shift and Climb Sloped Surfaces (Advanced) and Wall Jump (Advanced)
```
MayberryZoom, line 1511: Should the Wall Jump here be advanced? I feel like that's the easy part of this trick.

randovania/games/dread/logic_database/Ferenia.txt
```diff
@@ -904,7 +906,12 @@ Extra - asset_id: collision_camera_013
   * Layers: default
   * Morph Ball Tunnel to Space Jump Room/Tunnel to Pitfall Puzzle Room (Upper)
   > Pickup (Missile Tank)
-      Morph Ball and After Ferenia - Pitfall Puzzle Blob and Use Spin Boost
+      All of the following:
+          Morph Ball and After Ferenia - Pitfall Puzzle Blob
+          Any of the following:
+              Use Spin Boost
+              # https://youtu.be/mw12gft21hc
+              Flash Shift and Flash Shift Upgrade and Climb Sloped Surfaces (Intermediate)
```
MayberryZoom, line 914: I'm not totally convinced this should only be Intermediate, since as far as I can see, it requires flash chaining. Advanced seems more appropriate imo.

MayberryZoom (changes requested):
Please reclassify `Artaria Waterfall - Tower Middle -> Door to Behind Waterfall`'s `Movement (Advanced)` requirement to use this trick. Imo it would be either intermediate or advanced with Phantom Cloak and expert without.

randovania/games/dread/logic_database/Ferenia.txt
XenoWars: Intermediate seems fine to me. Can retry easily and it's a very small slope. Especially if the Ghavoran one is advanced.

Tyranisaur: I'll get back to addressing there rest of this on Monday, but for this one specifically, I'm not sure how I would feel about putting this at the same difficulty level as the artaria waterfall slope trick. Maybe it's just that I haven't learnt the latter because it's not as relevant to learn as in the speedrun, but I suspect that the flash chaining might be easier. Also, maybe I've learned the flash chaining trick by fluke, because I seem to have learned to do it here, in the speedboost slope(which isn't required), and in the burenia. Howerever, I had major trouble utilizing it in the spin boost tower.

MayberryZoom: They're both a lot of inputs, so I'm still not totally sure, but if you both agree it's Intermediate than that's probably fine. I don't wanna push it too hard.

randovania/games/dread/logic_database/Ghavoran.txt
Tyranisaur: I reduced the Wall Jump to Intermediate.

randovania/games/dread/logic_database/header.json
Tyranisaur: I ended up doing this, but I also added climbing the slopes in EMMI Zone Spinner and the bottom part of the Waterfall. The former was already under a trick with Spin Boost, while the former was previously trickless. Both remain trickless with Space Jump.

randovania/games/dread/logic_database/header.json
```diff
@@ -1228,7 +1228,7 @@
             },
             "SlopeClimb": {
                 "long_name": "Climb Sloped Surfaces",
-                "description": "With precise movement, Flash Shift can be used to climb Sloped surfaces.",
+                "description": "It is possible to gain height on sloped surfaces with good movement. This may require usage of techniques/items such as free-aim, Flash Shift, Spin Boost, or Phantom Cloak.",
```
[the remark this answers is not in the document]
Tyranisaur: It might be slightly misleading, but it stems from the two places I applied this trick, where there's limited space above the slope. The fact that one of them had a trick on it before seems to indicate that the change itself is fine, and the confusion has more to do with the trick description. I'm not sure if there's a better way to write it without elaborating a lot more.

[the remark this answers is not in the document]
MayberryZoom: Yeah, that's what I meant. The trick description could be a little misleading. I'm not sure either if there's a better way to write it, but we could also just improve it later if it proves to be a problem. I think it's fine for now.

PR #6524, opened 2024-05-12
Bump am2r_yams to 2.1.0

Miepee:

PR #6525, opened 2024-05-12
AM2R: Have seperate shuffled and required artifacts

Miepee:
grafik

Only merge after #6524 please


henriquegemignani:
Instead of "is placed", what about "exists" or "can be collected" or some variant?

henriquegemignani (changes requested):

Miepee:
Rephrased
grafik

henriquegemignani (approved):

PR #6526, opened 2024-05-12
Add 1 shinespark tricks to ARC

JeffGainsNGames:
from fusion discord
~~transmutation trial~~ and nightmare playground

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 5 ARC.txt
+++ b/randovania/games/fusion/logic_database/Sector 5 ARC.txt
@@ -100,6 +100,9 @@ Extra - room_id: [3, 6]
   * Layers: default
   * L3 Hatch to Gerubus Gully/Door to Nightmare Training Grounds
   * Extra - door_idx: (7,)
+  > Door to Training Aerie
+      # https://discord.com/channels/119665192110915584/1043264314955542578/1239308298830741628
+      Level 0 Locks and Level 3 Locks and Speed Booster and Shinespark Tricks (Beginner) and Wall Jump (Advanced) and Disabled Door Lock Rando and Disabled Entrance Rando
   > Door to Arctic Containment
       Trivial
   > Door to Entrance Lobby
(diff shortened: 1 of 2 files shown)


Miepee (commented):
Can we get these as youtuibe videos pleas

SchwartzGandhi (approved):

randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -100,6 +100,9 @@ Extra - room_id: [3, 6]
   * Layers: default
   * L3 Hatch to Gerubus Gully/Door to Nightmare Training Grounds
   * Extra - door_idx: (7,)
+  > Door to Training Aerie
+      # https://youtu.be/Qe9eMPmoUcU
+      Level 0 Locks and Level 3 Locks and Speed Booster and Shinespark Tricks (Beginner) and Wall Jump (Advanced) and Disabled Door Lock Rando and Disabled Entrance Rando
```
shark20061, line 105: ⛏️ Would argue "intermediate" on the shinespark difficulty only because of reset time
⛏️ Unsure if the L0 requirement is needed since all the doors on the shinespark path are L3s

shark20061 (approved):
Just a couple nitpicks is all

randovania/games/fusion/logic_database/Sector 5 ARC.txt
JeffGainsNGames: our definition for shinespark beginner is just a simple shinespark of the ground which this fits so I think that it's safe as is (unless the window to charge it is rather tight then happy to bump it up).
Yep I just misremembered the colors, will update.

shark20061: Nah, we're good on the difficulty then

SchwartzGandhi (approved):

PR #6565, opened 2024-05-20
metroid fusion: Sector 1 trick documentation

cluekitty:
Documents and updates several requested tricks:
* Kill with shinespark from Checkpoint: Updated youtube video
* Zebezian Zig: Documented damageless climb from bottom to top.
* Zebezian Zag: Documented damageless climb from bottom to top.
* Charge Core Arena E-tank: Documented vanilla intended collection using missiles + speed booster, and with/without highjump before/after boss.
* Ridley Arena Access: Documented damageless room navigation and eye-door defeat using only charge beam and normal missiles
* Tourian Entrance: Documented Left to Right damageless room navigation with an ice weapon (Ice Weapon + Screw Attack + morph + wave beam)

Note, the charge-core tricks are all relatively similar in terms of video footage. It may be worth consolidating the footage to having one or two videos for the variety of circumstances.


randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -935,6 +938,8 @@ Extra - room_id: [26]
               Screw Attack or Can Freeze Enemies
               # DMG boost - TODO: this damage value is wrong, get proper values
               Normal Damage ≥ 50
+              # Navigate the room Damageless: https://youtu.be/6LnIEbN2deg
+              Movement (Advanced)
```
Miepee, line 942: this should be combat, not movement.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -1251,6 +1256,8 @@ Extra - room_id: [35]
               Space Jump or Movement (Beginner)
           # Shinespark from Atmospheric Stabilizer Room
           Speed Booster and Disabled Door Lock Rando and Disabled Entrance Rando
+          # Navigate the room Damageless: https://youtu.be/Lh9D7Ye3rw8
+          Movement (Beginner)
```
Miepee, line 1260: this should be the combat intermediate from above.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -1432,12 +1440,12 @@ Extra - room_id: [40]
...
               Any of the following:
-                  # Pre-Fight, with Hi-Jump
+                  # Pre-Fight, with Hi-Jump: https://youtu.be/q7HKdMt4v3g
                   Hi-Jump and Shinespark Tricks (Intermediate)
-                  # After Fight, Trickless
+                  # After Fight, Trickless: https://youtu.be/Z6_miH4qoQY
                   Hi-Jump and After Boss Charge Core-X Defeated
-                  # No Hi-Jump
-                  Shinespark Tricks (Advanced)
+                  # After Fight, No Hi-Jump: https://youtu.be/U1nM5zVh_N0
```
Miepee, line 1447: why has this been chagned to after fight? is pre-fight NHJ not possible?

SchwartzGandhi: it is possible, not sure why it isn't listed here. that might've been my fault,
Not sure what difficulty it should be at tho, Pre fight NHJ can't be advanced atm due to it being logically redundant with Post fight NHJ, but Pre NHJ doesn't really scream Expert to me tbh.
You could instead maybe bump post NHJ down to intermediate and Pre w/ HJ down to beginner, that way you could have pre NHJ be at advanced, but idk if that totally fitting either.

Miepee: the reason why it isnt listed is because clue changed it to include `After Fight`

SchwartzGandhi: i mean, yeah both both pre fight and post fight NHJ shouldve been on here in the first place

Miepee: They were.
`Shinespark Tricks (Advanced)` implies 
1. No HiJump
2. Post OR Pre fight

cluekitty: I changed it mostly to match the video upload. There wasn't a separation that I saw in main. I can adjust it back. Though the trick itself is basically the same otherwise.

SchwartzGandhi: > They were. `Shinespark Tricks (Advanced)` implies
> 
>     1. No HiJump
> 
>     2. Post OR Pre fight

I guess that makes sense but imo, the softlock potential makes me want to differentiate the two

cluekitty: I've added a 4th section for: Pre-fight, no high jump.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -1244,12 +1249,14 @@ Extra - room_id: [35]
   > Door to Checkpoint Intersection
       Any of the following:
           # Deal with the Pirates
-          Charge Beam or Missiles ≥ 10 or Combat (Intermediate) or Can Freeze Enemies or Can Use Power Bombs
+          Charge Beam or Missiles ≥ 10 or Can Kill Super Geron or Can Use Power Bombs
```
Miepee, line 1252: this should get changed back to `Can Freeze Enemies`

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -1244,12 +1249,14 @@ Extra - room_id: [35]
...
           All of the following:
               Screw Attack
               # Screw Attack to defeat the Pirates can also (accidentally) destroy the platforms, unless cancelling spins
               Space Jump or Movement (Beginner)
-          # Shinespark from Atmospheric Stabilizer Room
+          # Navigate the room Damageless: https://youtu.be/Lh9D7Ye3rw8
+          Combat (Intermediate)
+          # Shinespark from Atmospheric Stabilizer Room: https://youtu.be/_PmmHN9YvKw
           Speed Booster and Disabled Door Lock Rando and Disabled Entrance Rando
```
Miepee, line 1260: while you're there doing changes: this should have l0 and the stabilizer event added.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -1432,12 +1440,14 @@ Extra - room_id: [40]
               # Vanilla Method - Shinespark from Hornoad Housing -> Access
               Speed Booster and Disabled Door Lock Rando and Disabled Entrance Rando and Can Kill Eye Door
               Any of the following:
-                  # Pre-Fight, with Hi-Jump
+                  # Pre-Fight, with Hi-Jump: https://youtu.be/q7HKdMt4v3g
                   Hi-Jump and Shinespark Tricks (Intermediate)
-                  # After Fight, Trickless
+                  # Pre-Fight, No Hi-Jump: https://youtu.be/Fo8-jjuPdik
+                  Shinespark Tricks (Intermediate)
```
Miepee, line 1446: this doesnt make sense.
if pre-NHJ is on a lower or equal trick level than post-NHJ, then logic will always choose this path and the `After Core X` requirement becomes redudant.
This either needs to be expert (which doesnt feel right to me), or we axe the boss requirement.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
cluekitty: Odd, I don't recall changing this. Perhaps I accidentally scrolled over the field. When looking through the logic

randovania/games/fusion/logic_database/Sector 1 SRX.txt
cluekitty: The trick for each 4 of the variety is pretty much the same imo. The only difference I feel is that doing this before the boss fight has the slight chance of softlock if no progression is in the etank location. The only difficulty changes when your only way to kill the hornoads is an without beam upgrades, which is possible, just annoying

cluekitty: For now, I'll match the other combat difficulties so they should be equally likely, but if you have another way you would want this handled I'm more than willing to make further adjustments.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
Miepee: by checking the DB where they're used. Altho ideally the long names are changed in a later PR too.
Central - 5
northeast - 2
northwest - 1
southeast - 4
southwest - 3

cluekitty: That seems a bit counter-intuitive to need to check the DB to know which is which. That being said, I probably could have noticed that if I poked around a bit more. Thanks for the info!

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -1244,13 +1249,15 @@ Extra - room_id: [35]
   > Door to Checkpoint Intersection
       Any of the following:
           # Deal with the Pirates
-          Charge Beam or Missiles ≥ 10 or Combat (Intermediate) or Can Freeze Enemies or Can Use Power Bombs
+          Charge Beam or Missiles ≥ 10 or Combat (Intermediate) or Can Use Power Bombs
```
Miepee, line 1252: you put combat intermediate into here instead of `can freeze enemies`.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
Miepee: if the trick for all of them is the same then they're not worth listing. they all get useless and logic will always choose prefight NHJ.
Please do what i said in the first comment of this thread: either adjust the shinespark requirements for all to be unique, or axe the boss requirements for ones where you feel are too similar.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -1432,12 +1440,10 @@ Extra - room_id: [40]
...
               Speed Booster and Disabled Door Lock Rando and Disabled Entrance Rando and Can Kill Eye Door
               Any of the following:
-                  # Pre-Fight, with Hi-Jump
-                  Hi-Jump and Shinespark Tricks (Intermediate)
-                  # After Fight, Trickless
-                  Hi-Jump and After Boss Charge Core-X Defeated
-                  # No Hi-Jump
-                  Shinespark Tricks (Advanced)
+                  # Trickless: https://youtu.be/Z6_miH4qoQY
+                  Hi-Jump
```
Miepee, line 1444: you considerably changed logic here. please revert this back to what it was:
HJ + boss killed => trickless
HJ + boss alive => intermediate (softlock potential)
HJless => advanced (softlock potential + less lenient)


Alternatively I would also accept
HJ + boss alive =>  intermediate (softlock potential)
HJless + boss killed => intermediate (less lenient)
HJless + boss alive => advanced (softlock potential + less lenient)

Miepee (changes requested):

randovania/games/fusion/logic_database/Sector 1 SRX.txt
cluekitty: Reverted the logic and added the videos for the existing logic.

Miepee (approved):

PR #6567, opened 2024-05-20
Fix mypy with NodeRequirement pr

henriquegemignani:


Miepee (approved):

randovania/gui/dialog/connections_editor.py
```diff
@@ -302,9 +302,11 @@ def current_requirement(self) -> RequirementArrayBase | None:
...
             comment = None
 
-        nested = [editor.current_requirement for editor in self._editors]
-        if any(it is None for it in nested):
-            return None
+        nested: list[Requirement] = []
+        for it in nested:
+            if it is None:
+                return None
+            nested.append(it)
```
ThanatosGit, line 309: With every OS failing on the tests, I guess they need to be fixed.
Is it intentional that `nested` is now always an empty array? That doesn't look right to me - without looking into the rest of the context - and can't see how this loop is every executed.

ThanatosGit (changes requested):

randovania/gui/dialog/connections_editor.py
henriquegemignani: WTF DID I WRITE HERE

ThanatosGit (approved):

PR #6568, opened 2024-05-20
Rename more fields for new game

henriquegemignani:
More files no longer are named "blank" something for new games.


henriquegemignani:
Ah, it renames the references to the UI files that end up being Blank only. Ideally we'd also copy the .ui files and tweak these. That's quite a bit more, gonna do that later.

PR #6577, opened 2024-05-22
Dolphin connector: nicer error messages

Miepee:


henriquegemignani (approved):

ThanatosGit:
BTW: This thing here is stuck atm.

As I'm already writing here: Can we add a FAQ entry or similiar to point where to find the lock file? From time to time, the file doesn't get deleted for whatever reason and it's always a pain to remember where it is located. Thought about adding it to the error message but that might be too verbose?

Miepee:
ty, just need to rereun the tests. something about codecov just introduces random failings.

>  Can we add a FAQ entry or similiar to point where to find the lock file? [...] Thought about adding it to the error message but that might be too verbose?

I personally dont want to see it in the error message because it will lead to people just deleting the file while they have multiple rdv's trying to connect. The folder where the pid-file can be found is here: https://github.com/trbs/pid/blob/c9d1550ba2ee73231f8e984d75d808c8cc103748/pid/utils.py#L17

PR #6578, opened 2024-05-22
MSR: exclude a3 mc door from being wave

Miepee:
Doesnt need changelog entry, as its part of the DLR change.


dyceron (changes requested):
Shouldn't both sides be excluded? Also tests already failing.

Miepee:
tests are failing because they are flaking, unrelated. 
AFAIK both sides dont need to be excluded. because we have force_two_way defined in the header.json, DLR will never make one side wave and the other side something else.

Miepee:
excluded the other door just in case :tm:

dyceron (approved):

PR #6580, opened 2024-05-22
Dread: Improve wording of text under environmental damage in preset energy tab

MayberryZoom:
| Previous | New |
| --- | --- |
| Logic always assume this setting is **disabled**. | Logic will **never** require heat/cold runs. |

The previous wording was misleading, as it implied that heat/cold runs are in logic for exponential damage ticks, but not linear, when in reality neither of them are in logic.


Miepee (changes requested):
cbhangelog entry

MayberryZoom:
Maybe it should be "Currently, logic will **never** require heat/cold runs."?

henriquegemignani:
I wouldn't add a currently, especially as it implies it's a temporary thing.

Tyranisaur:
Or simply "Taking any heat/cold damage will not be required."

MayberryZoom:
> Or simply "Taking any heat/cold damage will not be required."

Isn't that equivalent to what I said though?

Tyranisaur:
Possibly splitting hairs, but being specific about the fact that it has to do with any amount of damage would possibly be better at addressing the myths that small heat/cold runs are somehow logical without the trick. Also, I'm taking of the "currently", while still moderating out the "never", as we could in theory make this available eventually™️ if we solve the damage logic issues in the generator.

MayberryZoom:
> while still moderating out the "never"

This purpose of this was to say that regardless of settings, no seed would be generated requiring it. And this piece of text would be removed if this ever gets fixed.

Maybe the best wording would be "Logic will **never** require taking heat/cold damage."

Tyranisaur:
There's a double meaning to "never", that I don't like. One part is that it can refer to any seed you could make, but the other part is that it can also refer to the dev team having taken an absolute stance on this never changing in the future. In the interests of precise language, I think it would be best to avoid communicating the latter. Sure the text would change in the future, but we're also potentially communicating about the future.

MayberryZoom:
Well in that case I'm not sure how to word it without using "never". I'm not opposed to it, we just need something clear and concise.

Miepee:
Any objections to using Tyr's suggestion, potentially slightly reworded to `Regardless of trick settings, logic will not expect you to take any heat/cold damage`?

MayberryZoom:
> `Regardless of trick settings, logic will not expect you to take any heat/cold damage`

I think this is somewhat misleading because heat/cold runs *are* a trick, you just can't turn them on. But I don't know if there's really a way around that.

Maybe we could try one of these:
- `Regardless of this setting, you will never be required to take any heat/cold damage.`
- `Regardless of this setting, you will never be required by logic to take any heat/cold damage.`
- `Regardless of this setting, logic will never expect you to take any heat/cold damage.`

Miepee:
then you introduce the "never" problem again :P

henriquegemignani:
Assume the heat/cold damage trick doesn't exist and won't exist. If you think that's confusing if someone were to look at the Visualizer and see the trick there, then I'd rather hide the trick in the database.

MayberryZoom:
> then you introduce the "never" problem again :P

- `Regardless of this setting, you will not be required to take any heat/cold damage.`
- `Regardless of this setting, you will not be required by logic to take any heat/cold damage.`
- `Regardless of this setting, logic will not expect you to take any heat/cold damage.

> If you think that's confusing if someone were to look at the Visualizer and see the trick there, then I'd rather hide the trick in the database.

To be fair, I do think that we should do that in non-dev builds.

duncathan (approved):

Miepee (approved):

duncathan:
conflicts

PR #6586, opened 2024-05-23
Dread: Force transport room names on HUD if transport rando is enabled

MayberryZoom:
As described in https://github.com/randovania/open-dread-rando/issues/323, this will prevent players from missing out on the large advantage from having room names on HUD enabled with transport rando. This is accomplished by forcing the option in the patcher.json to `ALWAYS` if it is disabled with transport rando, then supplying only the transport room names.

As a result of this PR, the collision camera dictionary will no longer be included in the patcher.json unless room names on HUD or transport rando is enabled.

Requires open-dread-rando to be bumped before merging.


Miepee (commented):
mypy is complaining, otherwise this looks fine to me. Don't think fixing mypy here is that easy due to how cosmetic_dict is structured.

Miepee (commented):
Since this PR is about changed behaviour on transport rando, can you add at least one more elevator rando file with the cosmetic option set to either FADE or ALWAYS?

randovania/games/dread/exporter/patch_data_factory.py
```diff
@@ -338,6 +338,24 @@ def _static_room_name_fixes(self, scenario_name: str, area: Area):
 
         return cc_name, area.name
 
+    def _build_teleporter_name_dict(self, cc_dict: dict = {}) -> dict[str, dict[str, str]]:
```
henriquegemignani, line 341: First, don't ever use mutable objects as default arguments. The default argument is shared between all calls and it's not recreated for every call, for fun(tm) bugs.

Second, let's not have this argument.

randovania/games/dread/exporter/patch_data_factory.py
```diff
@@ -351,22 +369,11 @@ def _build_area_name_dict(self) -> dict[str, dict[str, str]]:
...
-                and node.dock_type in self.game.dock_weakness_database.all_teleporter_dock_types
-            ):
-                src_region, src_area = self.game.region_list.region_and_area_by_area_identifier(
-                    node.identifier.area_identifier
-                )
-                src_cc = src_area.extra["asset_id"]
-                dest_name = _get_destination_room_for_teleportal(connection)
-                all_dict[src_region.extra["scenario_id"]][src_cc] = f"Transport to {dest_name}"
-        return all_dict
+        return self._build_teleporter_name_dict(all_dict)
```
henriquegemignani, line 372: Merge the result of `_build_teleporter_name_dict` into the all_dict here

henriquegemignani (changes requested):
Other than these comment, looks good.

henriquegemignani (approved):

PR #6590, opened 2024-05-24
Bump ODR to 2.11.0

Miepee:


MayberryZoom (changes requested):
I think this needs a changelog entry for the Z-57 cooldown flag being removed.

Suggestion: `Changed: The morph launcher leading to Experiment Z-57 will no longer cause doors to be temporarily frozen.`

Unsure if there's better wording or if this should be Fixed rather than Changed.

MayberryZoom (approved):

PR #6592, opened 2024-05-24
Bump am2r_yams to 2.4.0

Miepee:
Breaks presets that were made between 9848a1abcd60fc1a9dbeef161dbd8234f90acf97 and this commit. But since its a relatively short timeframe and only within dev builds, I don't care about ensuring that these migrate properly.


JeffGainsNGames:
should this bump straight to 2.4.0 or is the pb drops change not making it in?

henriquegemignani (approved):

PR #6607, opened 2024-05-25
Properly introduce PickupIndex age

### LINKED ISSUE #4496: Generator: per-pickup impact on considered count
The weighting of pickup indices in `_calculate_uncollected_index_weights` depends on how pickups were placed when that index was accessible.

This proposes that the consideration count to not be incremented by 1 for each pickup, but instead a value that depends on the pickup itself. The intent is that it'll be incremented by smaller values for things like Missile Expansions.
Expectation is since Tower of Light requires the placement of 6+ expansions, the considered count of everything is incremented 6 times while unlocking nothing, which causes the next pickup to be placed to highly prefer Tower of Light.

henriquegemignani:
Pickups have an impact in the age when they're assigned. The weight of an index depends on it's age, not the considered count.

Fixes #4496

### DIFF
--- a/randovania/game_description/pickup/pickup_entry.py
+++ b/randovania/game_description/pickup/pickup_entry.py
@@ -67,6 +67,7 @@ class PickupGeneratorParams:
     probability_offset: float = 0.0
     probability_multiplier: float = 1.0
     required_progression: int = 0
+    index_age_impact: float = 1.0  # by how much the PickupIndex's age is incremented when this pickup is placed
 
 
 @dataclass(frozen=True)
--- a/randovania/generator/filler/retcon.py
+++ b/randovania/generator/filler/retcon.py
@@ -39,16 +39,24 @@ _VICTORY_WEIGHT = 1000
 def _calculate_uncollected_index_weights(
     uncollected_indices: Set[PickupIndex],
     assigned_indices: Set[PickupIndex],
-    considered_counts: Mapping[PickupIndex, int],
+    index_age: Mapping[PickupIndex, float],
     indices_groups: list[set[PickupIndex]],
 ) -> dict[PickupIndex, float]:
+    """
+    Calculates a weight
+    :param uncollected_indices
+    :param assigned_indices
+    :param index_age: the higher the age for an index, the smaller it's weight
+    :param indices_groups: indices separated in distinct groups.
+    In practice, one group for each region respecting preset exclusions
+    """
     result = {}
 
     for indices in indices_groups:
         weight_from_collected_indices = math.sqrt(len(indices) / ((1 + len(assigned_indices & indices)) ** 2))
 
         for index in sorted(uncollected_indices & indices):
-            weight_from_considered_count = min(10, considered_counts[index] + 1) ** -2
+            weight_from_considered_count = min(10.0, index_age[index] + 1.0) ** -2
             result[index] = weight_from_collected_indices * weight_from_considered_count
             # print(f"## {index} : {weight_from_collected_indices} ___ {weight_from_considered_count}")
 
@@ -175,12 +183,15 @@ def select_weighted_action(rng: Random, weighted_actions: Mapping[Action, float]
         return rng.choice(list(weighted_actions.keys()))
 
 
-def increment_considered_count(locations_weighted: WeightedLocations) -> None:
+def increment_index_age(locations_weighted: WeightedLocations, increment: float) -> None:
+    """
+    Increments the pickup index's age for every already collected index.
+    """
     for player, location, _ in locations_weighted.all_items():
-        was_present = location in player.pickup_index_considered_count
-        player.pickup_index_considered_count[location] += 1
+        was_present = location in player.pickup_index_age
+        player.pickup_index_age[location] += increment
         if not was_present:
-            # if it wasn't present, thenwe get to log!
+            # if it wasn't present, then we get to log!
             filler_logging.print_new_pickup_index(player.index, player.game, location)
 
 
@@ -346,7 +357,7 @@ def _assign_pickup_somewhere(
         index_owner_state, pickup_index = usable_locations.select_location(rng)
         index_owner_state.assign_pickup(pickup_index, PickupTarget(action, current_player.index))
 
-        increment_considered_count(all_locations)
+        increment_index_age(all_locations, action.generator_params.index_age_impact)
         all_locations.remove(index_owner_state, pickup_index)
 
         # Place a hint for the new item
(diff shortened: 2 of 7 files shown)


Miepee:
Very quick test on this PR. Left sets the age for everything to 1.0, right sets the age for ammo only to 0.25
grafik

ToL got slightly lowered. I'll run some more experiments with lowering E-Tanks as well, as they're considered standard pickups and not ammo

randovania/generator/filler/retcon.py
```diff
@@ -175,12 +183,15 @@ def select_weighted_action(rng: Random, weighted_actions: Mapping[Action, float]
...
-def increment_considered_count(locations_weighted: WeightedLocations) -> None:
+def increment_index_age(locations_weighted: WeightedLocations, increment: float) -> None:
+    """
+    Increments the pickup index's age for every already collected index.
+    """
     for player, location, _ in locations_weighted.all_items():
-        was_present = location in player.pickup_index_considered_count
-        player.pickup_index_considered_count[location] += 1
+        was_present = location in player.pickup_index_age
+        player.pickup_index_age[location] += increment
```
Tyranisaur, line 192: Did you consider not incrementing the age at all when there are no new pickups?

henriquegemignani: That function is only called when assigning a pickup

duncathan:
Should this be an experimental generator setting for some time?

henriquegemignani:
Maybe, but this is rather easy to measure since it affects only specific pickups.

randovania/generator/filler/retcon.py
Tyranisaur: I was under the impression that the challenge is that when the generator has to place 5 missile expansions to unlock Tower of Light, then when assigning the first 4 expansions, there are no new locations being unlocked, but all open locations get their age incremented every time, so after assigning the fifth one, then with Tower of Light being unlocked, that leaves every other location with a comparatively much higher age compared to when most items are assigned.

Further, when there are multiple missile requirements in the world, with different numbers of required missiles, then it's going to be hard to balance the location weights based on assigning a value to the item type.

randovania/game_description/pickup/pickup_entry.py
```diff
@@ -67,6 +67,7 @@ class PickupGeneratorParams:
     probability_offset: float = 0.0
     probability_multiplier: float = 1.0
     required_progression: int = 0
+    index_age_impact: float = 1.0  # by how much the PickupIndex's age is incremented when this pickup is placed
```
Miepee, line 70: Opening a thread here so that my testing results can get hidden easier when navigating on this PR

Miepee: attempt 1: see here https://github.com/randovania/randovania/pull/6607#issuecomment-2131307449

randovania/generator/filler/retcon.py
henriquegemignani: Ah, the issue is that discovering there are no new pickups unlocked comes much later. In particular, in the next time this world is considered for assigning a pickup.

Not impossible to do, but it'll be a significant redesign.

Tyranisaur: I see. Another angle would be whether there is some information available when deciding to place a missile expansion, or any other item, about whether that pickup by itself opens anything, or if it already knows that it will need more of them.

randovania/game_description/pickup/pickup_entry.py
Miepee: attempt 2: change etank age impact and ammo age impact to 0.2
(left: all 1.0, right: etank and ammo 0.2)
grafik
artifacts seem more balanced around lower mines, but FHA and PMT are still outsiders. As was theorized earlier, their issue seems to be unrelated to this one.
Comparing the progression probability column is also interesting (yellow highlighted big changes):
grafik

Miepee: attempt 3: change etank and missiles to 0.1.
left: all 1.0, middle: etank and ammo 0.2, right: etank and ammo 0.1
artifacts:
grafik


progression probability:
grafik

---

I thought the results would be worse than attempt 2, but at first glance, they seem nicer?

I had a look at phazon suit and it seems here a bit more evened out. But that could also just be within margin of error (still only 1k seeds)
grafik

I also had a look at the bosses which... are basically unchanged.
In fact, if I look at a table that shows me everything that has a bigger difference than 2%. The results are:
- Artifacts. Few changes in transport tallon B (tallon), life grove (tunnel), storage/security cave, early magmoor, plasma processing, ToL
- Etank placement. the only noteworthy one that ToL now has an 11% decrease in giving E
- missile expansions. Life grove tunnel, storage cave, chapel of the elders, lower mines noticeably have lower chance of giving missiles, ToL chance got bumped by ~14%
- phazon suit: ~3%decrease on OP
- progression probability: life grove, phen transport access, control tower, main quarry, magmoor workstation, elder chamber got noticeably bumped. research core, PCA, elite quarters got noticeably nerfed. ToL got decreased by 17.9%

---

I currently like attempt 3 the most. Unsure of what else should be tried. I could try lowering etanks even more in order to get them more into the ~14%ish range? Have the feeling that that's something else tho.

Miepee: I'm sending my games here off in the meantime + csv's if someone else wants to look at the data.
All starter preset, 1k games each.
1.0 value for everything:
[all-one.zip](https://github.com/randovania/randovania/files/15444352/all-one.zip)

0.25 for ammo only (aka this PR):
[ammo-025.zip](https://github.com/randovania/randovania/files/15444354/ammo-025.zip)

0.2 for ammo and etanks:
[etank-and-ammo-02.zip](https://github.com/randovania/randovania/files/15444355/etank-and-ammo-02.zip)

0.1 for ammo and etanks:
[01-for-etank-and-ammo.zip](https://github.com/randovania/randovania/files/15444358/01-for-etank-and-ammo.zip)

Miepee: Just for confirmation's sake, i tried setting etank and ammo to 4.0
grafik

toasterparty: I find it easier to understand what's changed between builds by using conditional formatting on delta

image

I'm worried this is just shifting the hotspot to a different part of the game, rather than fully diffusing the placement probability.

Honestly I thought the only way to fix this was to add a feature where game maintainers could pick some items at their discretion to get randomly pre-filled; this is a good close 2nd

henriquegemignani:
I'll leave the branch up, but I don't want to fix these conflicts myself.

PR #6609, opened 2024-05-25
Fix elevator two-way shuffle (for MSR)

ThanatosGit:
Fixes a typo and changes the check in L88-89.
From my understanding this wants to check if the teleporters are connected or reachable from the other.
a) The two teleporters `cteleporter2` and `cteleporter1` are in the same region => reachable
or
b) `cteleporter2`'s area has the same name as `cteleporter1`'s destination area.

b) was intended to express `Is cteleporter2 and cteleporter1 connected` but as MSR uses the same area names multiple times e.g. `Transport to Area 7` it breaks. Therefore I added the check of the destination region name.

(Waits for new release? => then needs a changelog entry)

### DIFF
--- a/randovania/generator/teleporter_distributor.py
+++ b/randovania/generator/teleporter_distributor.py
@@ -53,10 +53,10 @@ def try_randomize_teleporters(
     rng: Random,
     teleporters: tuple[TeleporterHelper, ...],
 ) -> list[TeleporterHelper]:
-    telepoter_database: list[TeleporterHelper] = list(teleporters)
-    assert len(telepoter_database) % 2 == 0
+    teleporter_database: list[TeleporterHelper] = list(teleporters)
+    assert len(teleporter_database) % 2 == 0
 
-    teleporter_list = copy.copy(telepoter_database)
+    teleporter_list = copy.copy(teleporter_database)
     teleporters_by_region: dict[str, list[TeleporterHelper]] = defaultdict(list)
     for teleporter in teleporter_list:
         teleporters_by_region[teleporter.region_name].append(teleporter)
@@ -76,8 +76,7 @@ def try_randomize_teleporters(
         teleporter_list.remove(source_teleporter)
         teleporter_list.remove(target_teleporter)
 
-    # TODO
-    list3 = copy.copy(telepoter_database)
+    list3 = copy.copy(teleporter_database)
     cteleporter_list3 = [list3[0]]
     while list3:
         cteleporter_list1 = []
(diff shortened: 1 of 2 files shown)


randovania/generator/teleporter_distributor.py
```diff
@@ -76,18 +76,17 @@ def try_randomize_teleporters(
...
         for cteleporter1 in cteleporter_list3:
             index = 0
             while index < len(list3):
                 cteleporter2 = list3[index]
-                if (
-                    cteleporter2.region_name == cteleporter1.region_name
-                    or cteleporter2.area_name == cteleporter1.destination.area
+                if cteleporter2.region_name == cteleporter1.region_name or (
+                    cteleporter2.region_name == cteleporter1.destination.region
+                    and cteleporter2.area_name == cteleporter1.destination.area
```
Miepee, line 89: Would be nice to comment this, so that the next person looking into this doesnt have to find this PR to figure out what this is supposed to achieve.

Miepee (commented):
IMO it'd be nicer if this gets into the june release rather than july.

randovania/generator/teleporter_distributor.py
ThanatosGit: Added a few comments + gave variables a proper name instead of `list1,2,3` and `cteleporter1,2,3` etc.

randovania/generator/teleporter_distributor.py
```diff
@@ -53,53 +53,65 @@ def try_randomize_teleporters(
...
-                    cteleporter2.region_name == cteleporter1.region_name
-                    or cteleporter2.area_name == cteleporter1.destination.area
+            # check all remaining teleporters
+            while index < len(remaining_teleporters):
+                teleporter_to_check = remaining_teleporters[index]
+                # teleporter_to_check is considered reachable if it is in the same region as a reachable_teleporter
+                # or if it is in the region+area of the reachable_teleporter's destination
+                if teleporter_to_check.region_name == reachable_teleporter.region_name or (
+                    teleporter_to_check.region_name == reachable_teleporter.destination.region
+                    and teleporter_to_check.area_name == reachable_teleporter.destination.area
```
Miepee, line 100: these two lines can be simplified with `teleporter_to_check.area_location == reachable_teleporter.destination.area_identifier`, no?

ThanatosGit: yes

Miepee (approved):
Ran  ~100 games for prime, dread and echoes

Prime and dread had no changes that I could observe
Echoes did have changes (2/152), because as it turns out this bug was also affecting it as well. Which means, this should probably get a changelog entry :)

ThanatosGit:
Somehow I was under the impression that reusing area names in different regions was a very rare case. Looking through my stupid iteration over the DB it turns out it is done for like every game.
Prime defines `Transport to Magmoor Caverns South` and `Transport to Tallon Overworld South` twice. With only two duplicates, I guess it's either a very rare case or impossible but don't want to think about that too much.
No changes for Dread is interesting. It defines nearly everything multiple times `Transport to Dairon, Transport to Cataris, Transport to Burenia, Transport to Artaria, Transport to Ghavoran, Transport to Hanubia, Transport to Ferenia`

Miepee:
May have been just rare enough, considering my trick settings (starter)

PR #6610, opened 2024-05-26
Locked door: check if you can escape back

### LINKED ISSUE #4781: Disabled door placement does not check if you can escape the back
Dock Lock Rando checks if you can reach the back of the door in question. However, you might be stuck there unable to leave. Examples of this are when the backside of the door is:
- Hive Chamber area
- Leaf rooms (such as Storage D), but there's an one-way elevator into it

Miepee:
Fixes #4781 

Looking at it now, I'm not quite sure whether just setting the node of the state is enough. Thought about using `act_on_node` but that one only works with resource nodes. I think this one should be fine tho except for maybe a few health edge cases?


henriquegemignani:
The issue of this is performance. How much slower is it? Tbh the cost of just one calculate reach might be irrelevant compared to the bulk of full resolvers the DLR does

Miepee:
Before this PR:
grafik

After this PR:
grafik


On average seems to be slower by ~20seconds, but we also have quite a few less impossible seeds which is nice.

The preset I took was am2r starter preset + individual door + everything in door_from, Normal door + Locked door in door_to

duncathan (approved):

duncathan:
I don't mind a single extra reach check on doors that can possibly be disabled. preferable to the alternative of impossible seeds

PR #6613, opened 2024-05-26
Remove prime 3 from readme

Miepee:
Having Prime 3 in the experimental is misleading for people considering it's currently essentially dead.
Gollop credits is still there tho.


Miepee:
Forgot that we have some other games that have this issue too.
Closing due to ineffectiveness

PR #6632, opened 2024-05-27
MSR - DLR Updates

dyceron:
- Adds a new door type `Access Open` which is added to most 3-tiles high open docks
- Ensures that patched doors are only exported once in the pdf

Requires https://github.com/randovania/open-samus-returns-rando/pull/358 to be merged along with a new OSRR release.

Will add a changelog entry after the next release.

image


ThanatosGit (changes requested):
lgtm.
Besides the things you mentioned, the tests are failing.

randovania/games/samus_returns/exporter/patch_data_factory.py
```diff
@@ -449,6 +492,9 @@ def _door_patches(self) -> list[dict[str, str]]:
                 print(f"Invalid door (no actor): {node}")
                 continue
 
+            if any(entry["actor"] == self._teleporter_ref_for(node) for entry in result):
```
ThanatosGit, line 495: On a side note, I wonder how slow these `any` loops are.
I definitely will mess it up on mobile but given we change 200 doors, it will always need to run to the end of the list. For the first one there are 0 entries, for the second 1 entry...so the sum from 0 to 199 = 19900 iterations. As every door is twice in the list, but the second time it will not be added because it will be found by the loop, it runs approximately through half of the list, so additional 9950 iterations. Unless it's sorted then it will always run again to the end of the list and adding 19900 iterations.

Worst-case is like 39800 iterations and compares between two dicts.

randovania/games/samus_returns/logic_database/Area 2 Dam Interior.txt
```diff
@@ -635,6 +635,25 @@ Extra - asset_id: collision_camera018
   > Pickup (Missile Tank)
       Morph Ball
 
+> Door to Dam Basement (DLR Left); Heals? False
```
Miepee, line 638: I dont like these names. I'd rather call them `Empty Hatch Left/Right` instead.

randovania/games/samus_returns/logic_database/header.txt
```diff
@@ -465,6 +465,7 @@ Dock Weaknesses
       Unlocked: Power Beam Door
```
Miepee, line 465: Unrelated to this file, but area 6 omega arena should be `Blocked Open Passage`

randovania/games/samus_returns/logic_database/Area 4 Central Caves.txt
```diff
@@ -696,12 +696,17 @@ Extra - asset_id: collision_camera_013
   * Pickup 85; Category? Minor
```
Miepee, line 696: I assume A4 **Crystal Mines** :
- Green crystal dugout/mines intersection terminal is too high?
- gemstone gorge / pink crystal preserve is too weird with the slope?

Miepee, line 696: What about area 8:
- nest nodule/nest network


Also queen arena should be a blocked open passage

randovania/games/samus_returns/logic_database/Area 1.txt
```diff
@@ -1267,7 +1306,7 @@ Extra - asset_id: collision_camera_044
               Mid-Air Morph (Beginner) or Simple IBJ or Use Spider Ball or Use Spring Ball
   > Tunnel to Inner Temple West Hall
       Morph Ball
-  > Dock to Metroid Caverns Alpha Arena North East Access
+  > Door to Metroid Caverns Alpha Arena North East Access
       Trivial
```
Miepee, line 1311: Inner Temple Teleporter Access should have blocked open passage.

ThanatosGit (changes requested):
It came to my mind that this will be merged after door lock is stable, so it requires a migration.

Miepee:
> so it requires a migration

Adding new door types doesn't require a migration, does it? I can't remember needing to to do them for am2r's queen door at least.

ThanatosGit:
It's the fact that I at least saw one rename, which requires a migration

Miepee:
Ah, right.

randovania/games/samus_returns/logic_database/Area 4 Central Caves.txt
dyceron: Both are too high

dyceron: There's no open passages in a8 except for after queen

Miepee: Database disagrees

dyceron:
> It came to my mind that this will be merged after door lock is stable, so it requires a migration.

I didn't think this needed a migration since it's just a new type

randovania/games/samus_returns/logic_database/Area 4 Central Caves.txt
dyceron: Wow I completely forgot about that one

Miepee: TBF, its logical implications in vanilla are completely useless.
It would enable tho for the bottom door to be a permanently locked door :eyes:

randovania/games/samus_returns/logic_database/Area 1.txt
```diff
@@ -146,24 +149,27 @@ Extra - asset_id: collision_camera_005
...
   > Dock to Chute Leech Cabin
       Trivial
-  > Dock to Water Maze
+  > Door to Water Maze
       Trivial
 
 > Door from Water Maze; Heals? False
   * Layers: default
-  * Access Closed to Water Maze/Door to Gullugg Gangway
+  * Access Closed to Water Maze/Door to Gullugg Gangway (Bottom)
```
ThanatosGit, line 161: For reference: It needs a migration because of this renamed node because it will break the rdvgames of next month stable.
I think that's also the only one.

randovania/games/samus_returns/exporter/patch_data_factory.py
ThanatosGit: I only had a few seconds and did a super fast and rough test with `time.perf_counter()`: Before it took 0.436 ms and now 11.966 ms.
That is almost 30 times slower.
12 ms is not super slow, but for what it does, it's not really fast either.
Also I know this code isn't called very often and compared to how long it takes to export the game it's like nothing, but since this is something that is called on the server for multiworld (with door types rando) I wanted to at least mention it.

dyceron: Sets can't have duplicate entries right? Couldn't we just have `result` be a set and then append to that which would handle the duplicates automatically, then convert the set to a list at the end?

test/test_files/log_files/samus_returns/door_lock.rdvgame
```diff
@@ -1,19 +1,19 @@
 {
```
Miepee, line 1: Instead of modifying this door lock rdvgame, I would add a second rdvgame (i.e. door_lock_with_access_open.rdvgame)

test/test_files/log_files/samus_returns/door_lock.rdvgame
```diff
@@ -508,7 +508,7 @@
                     "type": "door",
                     "name": "Power Beam Door"
                 },
-                "Area 1/Water Maze/Door to Gullugg Gangway": {
+                "Area 1/Water Maze/Door to Gullugg Gangway (Bottom)": {
```
Miepee, line 511: revert this change. the migration should handle the test not failing, not the modification of the rdvgame.

dyceron:
Assuming the rest is good and the door patches is fine, all this needs is an OSRR update to be merged.

Miepee:
DB looked fine last time i looked at it.

dyceron:
Ugh I must've lost the changelog when I related

ThanatosGit (approved):

PR #6649, opened 2024-05-29
Support pickups with multiple original locations

henriquegemignani:


CHANGELOG.md
```diff
@@ -79,6 +79,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Dread
 
+- Added: Progressives pickups can now be configured to be placed vanilla.
```
ThanatosGit, line 82: It's `Progressive pickups` and not `ProgessiveS pickups`, isn't it?

ThanatosGit (changes requested):
lgtm. Can be implemented later for the other games.

Miepee:
Do you want to get that in before the june release?

toasterparty:
Does this mean you could put all missiles, power bomb expansions, etanks, etc. to their vanilla locations? I feel like that would finally end the confusion as to what exactly "major/minor" means by letting the user configure it themselves instead of checking a box.

duncathan:
> I feel like that would finally end the confusion as to what exactly "major/minor" means by letting the user configure it themselves instead of checking a box.

I'm not sure what you mean here?

henriquegemignani:
> Does this mean you could put all missiles, power bomb expansions, etanks, etc. to their vanilla locations? I feel like that would finally end the confusion as to what exactly "major/minor" means by letting the user configure it themselves instead of checking a box.

AmmoPickupDefinition don't support this actually. And the later part has nothing to do with this feature.

toasterparty:
What I'm saying is if `AmmoPickupDefinition` did support this, then you can replace the major/minor functionality with this because they are functionally equivalent.

henriquegemignani:
No? This is just assigning certain pickups to an specific location when creating the pool. It doesn't even touch the generator, like major/minor does.

duncathan:
> What I'm saying is if `AmmoPickupDefinition` did support this, then you can replace the major/minor functionality with this because they are functionally equivalent.

I still don't follow. What exactly would be replaced? And for what purpose? I'm confused what this PR is supposed to have to do with major/minor

CHANGELOG.md
ThanatosGit: It's just the changelog and ready to go?

ThanatosGit (approved):

PR #6660, opened 2024-06-01
Fix editing map min/max crashing

Miepee:
grafik


randovania/gui/data_editor.py
```diff
@@ -304,10 +304,17 @@ def _on_image_spin_update(self):
         w = self.current_region
         if not isinstance(w.extra, dict):
```
henriquegemignani, line 305: check instead why this if is not happening

henriquegemignani (changes requested):

PR #6668, opened 2024-06-03
MP3: hangar access a fix

SchwartzGandhi:
Added the SSJ for crossing both ways.

### DIFF
--- a/randovania/games/prime3/logic_database/GFS Valhalla.txt
+++ b/randovania/games/prime3/logic_database/GFS Valhalla.txt
@@ -41,7 +41,7 @@ Extra - asset_id: 5398996635579069682
   > Door to Repair Bay
       Any of the following:
           Grapple Swing or Use Screw Attack (Space Jump)
-          # https://www.youtube.com/watch?v=l8ICCVjhcq0
+          # https://youtu.be/l8ICCVjhcq0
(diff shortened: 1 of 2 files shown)


randovania/games/prime3/logic_database/GFS Valhalla.txt
```diff
@@ -41,15 +41,18 @@ Extra - asset_id: 5398996635579069682
...
 > Door to Repair Bay; Heals? False
   * Layers: default
   * Normal Door to Repair Bay/Door to Hangar A Access
   * Extra - dock_index: 1
   > Door to Docking Bay 5
-      Grapple Swing or Use Screw Attack (Space Jump)
+      Any of the following:
+          Grapple Swing or Use Screw Attack (Space Jump)
+          # https://youtu.be/l8ICCVjhcq0
+          Boost Ball and Space Jump Boots and Morph Ball and Bomb/Spring Space Jump (Advanced)
```
Samuel6710, line 55: Not a maintainer, but IMO this SSJ should be expert. The angle is much more obtuse and less forgiving than the one from Docking Bay 5.

SchwartzGandhi: if just assumed that they were the same difficulty both ways, but if you think otherwise I can test it myself here in a few minutes

dyceron (approved):

PR #6676, opened 2024-06-06
Allow 'Open Map' cosmetic option for Prime to be enabled with Random Elevators

Nystrata:
The reasoning for disabling the 'Open Map' cosmetic for Prime 1 for Elevator Randomizer was that players supposedly could just look at the room name to see the Elevator destination.  
However, in Prime 1, room names don't appear for not-visited rooms, even after using a Map Station .  
Unless there are plans for a patch that shows room names for not-visited rooms, players should be allowed to use Open Map regardless of whether elevators are randomized or not.  

* Visited Room (Name is revealed)
  * image
* Not-visited Room (Name is not revealed)
  * image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [8.2.0] - 2024-07-??
 
 ### Metroid Prime
-- Changed: Open Map can be used with Elevator Randomization (Unvisited rooms do not reveal their name. Unvisisted Elevators do not reveal their destination.)
+- Changed: Open Map can be used with Elevator Randomization (Unvisited rooms do not reveal their name. Unvisited Elevators do not reveal their destination.)
 
 ## [8.1.0] - 2024-06-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [8.2.0] - 2024-07-??
 
-- Nothing yet! Add changes here.
+### Metroid Prime
+- Changed: Open Map can be used with Elevator Randomization (Unvisited rooms do not reveal their name. Unvisisted Elevators do not reveal their destination.)
 
 ## [8.1.0] - 2024-06-04
 
--- a/randovania/games/prime1/exporter/patch_data_factory.py
+++ b/randovania/games/prime1/exporter/patch_data_factory.py
@@ -813,7 +813,7 @@ class PrimePatchDataFactory(PatchDataFactory):
         else:
             starting_memo = None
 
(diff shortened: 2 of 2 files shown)


henriquegemignani:
No for me. It's very intentional that we don't let that go. Randomprime needs to not reveal the elevator rooms when elevator rando is enabled, but it doesn't offer that afaik.

duncathan:
> No for me. It's very intentional that we don't let that go. Randomprime needs to not reveal the elevator rooms when elevator rando is enabled, but it doesn't offer that afaik.

afaict the point Nyss is making is that room names are not currently revealed *at all* until they've been visited, unlike the behavior in echoes

Nystrata:
> > No for me. It's very intentional that we don't let that go. Randomprime needs to not reveal the elevator rooms when elevator rando is enabled, but it doesn't offer that afaik.
> 
> afaict the point Nyss is making is that room names are not currently revealed _at all_ until they've been visited, unlike the behavior in echoes

Yes, this is what I mean

toasterparty (approved):
I don't remember who was doing the map patcher stuff in the first place

duncathan (changes requested):
changelog entry

henriquegemignani:
> > No for me. It's very intentional that we don't let that go. Randomprime needs to not reveal the elevator rooms when elevator rando is enabled, but it doesn't offer that afaik.
> 
> afaict the point Nyss is making is that room names are not currently revealed _at all_ until they've been visited, unlike the behavior in echoes

I know. Between spoiling the elevator destinations and not revealing the map, my choice is not revealing the map. I'm firmly opposed to spoiling the elevator destinations.

Nystrata:
> I know. Between spoiling the elevator destinations and not revealing the map, my choice is not revealing the map. I'm firmly opposed to spoiling the elevator destinations.

My point is that,   
In Echoes, Open Map shows room shapes and room names  
In Prime, Open Map shows *only room shapes* and does not show room names   
If you look at the first post, you can see that in the screenshots - enabling Open Map for Prime 1 *does not reveal elevator destinations*, it only reveals the shapes of the rooms and item dots

duncathan:
yeah nobody is advocating for spoiling the destinations. that'd be a non-starter for mii as well

henriquegemignani:
If revealing the map does not show room names for Prime 1, then that's a regression for Prime 1. A long time ago that was working.

JeffGainsNGames:
> If revealing the map does not show room names for Prime 1, then that's a regression for Prime 1. A long time ago that was working.

iirc it's been like this since ~5.0.0 at least

Nystrata:
Oh huh,
I wasn't aware revealed map names was a feature in earlier versions

henriquegemignani:
image
It works in 3.0.3

ThanatosGit:
What's the conclusion? Should not visited rooms show the room name or should this one here be merged?
The later would require to fix the tests. (I know they are flaky but given that all of them are failing, I guess there is so something to fix).

PR #6682, opened 2024-06-08
Add "Export All Presets" feature for multiworld sessions

ArcanoxDragon:
This adds a button (hidden behind "Show Experimental Settings") to the Multiworld Session window to export the preset of every world in the session to a folder. This is extremely helpful for troubleshooting large multis that are having generation problems, or to verify that all users are using settings that abide by whatever session rules were established by the host.

I'm not sure if "Show Experimental Settings" is necessarily the correct choice for toggling the visibility, but I didn't think it made sense for it to show up for all users by default.

### DIFF
--- a/randovania/gui/lib/common_qt_lib.py
+++ b/randovania/gui/lib/common_qt_lib.py
@@ -177,6 +177,16 @@ def prompt_user_for_preset_file(window: QtWidgets.QWidget, new_file: bool, name:
     )
 
 
+def prompt_user_for_preset_folder(window: QtWidgets.QWidget) -> None | Path:
+    """
+    Shows a QFileDialog asking the user for a directory in which to save multiple Randovania preset files
+    :param parent:
+    :return: A string if the user selected a directory, None otherwise
+    """
+
+    return _prompt_user_for_directory(window, caption="Select a directory in which to place preset files")
+
+
 def set_default_window_icon(window: QtWidgets.QWidget):
     """
     Sets the window icon for the given widget to the default icon
--- a/randovania/gui/multiplayer_session_window.py
+++ b/randovania/gui/multiplayer_session_window.py
@@ -205,10 +205,12 @@ class MultiplayerSessionWindow(QtWidgets.QMainWindow, Ui_MultiplayerSessionWindo
         self.rename_session_action = QtGui.QAction("Change title", self.advanced_options_menu)
         self.change_password_action = QtGui.QAction("Change password", self.advanced_options_menu)
         self.duplicate_session_action = QtGui.QAction("Duplicate session", self.advanced_options_menu)
+        self.export_all_presets_action = QtGui.QAction("Export all presets", self.advanced_options_menu)
 
         self.advanced_options_menu.addAction(self.rename_session_action)
         self.advanced_options_menu.addAction(self.change_password_action)
         self.advanced_options_menu.addAction(self.duplicate_session_action)
+        self.advanced_options_menu.addAction(self.export_all_presets_action)
         self.advanced_options_tool.setMenu(self.advanced_options_menu)
 
         # Generate Game Menu
@@ -252,6 +254,7 @@ class MultiplayerSessionWindow(QtWidgets.QMainWindow, Ui_MultiplayerSessionWindo
         self.rename_session_action.triggered.connect(self.rename_session)
         self.change_password_action.triggered.connect(self.change_password)
         self.duplicate_session_action.triggered.connect(self.duplicate_session)
+        self.export_all_presets_action.triggered.connect(self.export_all_presets)
         self.copy_permalink_button.clicked.connect(self.copy_permalink)
         self.view_game_details_button.clicked.connect(self.view_game_details)
         self.everyone_can_claim_check.clicked.connect(self._on_everyone_can_claim_check)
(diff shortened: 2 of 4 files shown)


duncathan:
> I'm not sure if "Show Experimental Settings" is necessarily the correct choice for toggling the visibility, but I didn't think it made sense for it to show up for all users by default.

lock it to session admins?

ArcanoxDragon:
> > I'm not sure if "Show Experimental Settings" is necessarily the correct choice for toggling the visibility, but I didn't think it made sense for it to show up for all users by default.
> 
> lock it to session admins?

I considered that too, but since "Export Preset" is available for all players, it seemed reasonable to expose the "Export All Presets" option as well.

Also, I'm not sure why that one test crashed in the workflow check...I ran all the non-resolver tests locally before committing and it was fine 🤔

ThanatosGit:
> I considered that too, but since "Export Preset" is available for all players, it seemed reasonable to expose the "Export All Presets" option as well.
> 
What's the downside of showing it to everyone then? There is nothing experimental about that feature and hiding it behind an unrelated flag is really bad IMO.

> Also, I'm not sure why that one test crashed in the workflow check...I ran all the non-resolver tests locally before committing and it was fine 🤔

Because they are flaky af. I restart it.

ArcanoxDragon:
Maybe it should go under "Advanced Options" instead? That way it's not front-and-center with the generate/export buttons, but it doesn't need to be restricted to an option.

randovania/gui/multiplayer_session_window.py
```diff
@@ -578,6 +582,30 @@ async def duplicate_session(self):
...
+
+        if path is None:
+            return
+
+        illegal_chars_pattern = r"^[ .]|[/<>:\"\\|?*]+|[ .]$"
+        world_owners = {world_uid: user for user in self._session.users.values() for world_uid in user.worlds.keys()}
+        world_index = 1
+        world_count = len(self._session.worlds)
+
+        for world in self._session.worlds:
```
henriquegemignani, line 596: `for world_index, world in enumerate(self._session.worlds):`

randovania/gui/multiplayer_session_window.py
```diff
@@ -578,6 +582,30 @@ async def duplicate_session(self):
...
+            owner_name = world_owners[world.id].name if world.id in world_owners else "Unclaimed"
+            preset = world.preset
+            filename = f"World{world_index}-{preset.game.short_name}-{owner_name}-{world.name}-preset.json"
```
henriquegemignani, line 599: `...-.{VersionedPreset.file_extension()}`

(aka `rdvpreset`)

henriquegemignani (changes requested):

henriquegemignani:
Plus changelog entry

randovania/gui/lib/common_qt_lib.py
```diff
@@ -177,6 +177,16 @@ def prompt_user_for_preset_file(window: QtWidgets.QWidget, new_file: bool, name:
     )
 
 
+def prompt_user_for_preset_folder(window: QtWidgets.QWidget) -> None | Path:
```
Miepee, line 180: this function being in common_qt_lib is weird, is there a specific reason why it needs to be here?
Considering its only used once, can you not just inline it?

randovania/gui/multiplayer_session_window.py
```diff
@@ -578,6 +582,30 @@ async def duplicate_session(self):
         if new_name is not None:
             await self.game_session_api.duplicate_session(new_name)
 
+    def export_all_presets(self):
+        path = common_qt_lib.prompt_user_for_preset_folder(self)
+
+        if path is None:
+            return
+
+        illegal_chars_pattern = r"^[ .]|[/<>:\"\\|?*]+|[ .]$"
```
Miepee, line 591: you can use re.compile which should give a small performance boost

randovania/gui/multiplayer_session_window.py
ArcanoxDragon: Ah, I was wondering if Python had something like that!

randovania/gui/multiplayer_session_window.py
ArcanoxDragon: I didn't even think about that, lol. I wrote all this initially as a once-off thing for BashPrime's async and totally forgot about the file extension

ArcanoxDragon:
> Plus changelog entry

I added one already :)

randovania/gui/lib/common_qt_lib.py
ArcanoxDragon: That's where all the other ones were. The _prompt_user_for_directory function is prefixed with an underscore suggesting it's not supposed to be consumed as a public API, so I followed the pattern of the rest of the file.

Miepee: it is not tho?
grafik


I dont see any leading underscore.

henriquegemignani:
> > Plus changelog entry
> 
> I added one already :)

I looked for one and failed to find it. Good job me.

randovania/gui/multiplayer_session_window.py
ArcanoxDragon: Ended up using the already-existing `sanitize_for_path` function in `string_lib` to avoid duplication of efforts. It looks like it'll be SLOWER since it's iterating each character manually, but maybe *that* function could be optimized to use a regex some time in the future.

henriquegemignani (approved):

PR #6684, opened 2024-06-08
Slightly optimize add_ammo_pickups

henriquegemignani:
The test was working by magic before, because apparently `range(MagicMock())` is `range(0, 1)`??

### DIFF
--- a/randovania/generator/pickup_pool/ammo_pickup.py
+++ b/randovania/generator/pickup_pool/ammo_pickup.py
@@ -5,8 +5,6 @@ from typing import TYPE_CHECKING
 from randovania.generator.pickup_pool.pickup_creator import create_ammo_pickup
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


randovania/generator/pickup_pool/ammo_pickup.py
```diff
@@ -15,13 +13,15 @@
 def add_ammo_pickups(
     resource_database: ResourceDatabase,
     ammo_configuration: AmmoPickupConfiguration,
-) -> Iterator[PickupEntry]:
+) -> list[PickupEntry]:
```
Miepee, line 16: where is this function used? does that not need to be adjusted?

henriquegemignani: No. There's one place and it only cares to iterate over the result, which you can with a list result.

Miepee (approved):

PR #6686, opened 2024-06-08
Change CLI to allow better creating crossgame permalinks

henriquegemignani:
Game is now part of the the preset name, via <game name>/<preset name> Add --preset-file to specify files directly

### DIFF
--- a/randovania/cli/commands/permalink.py
+++ b/randovania/cli/commands/permalink.py
@@ -2,12 +2,19 @@ from __future__ import annotations
 
 import asyncio
 import random
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from randovania.games.game import RandovaniaGame
+from randovania.layout.versioned_preset import VersionedPreset
 
 if TYPE_CHECKING:
     from argparse import ArgumentParser
+
+
+def _error(msg: str) -> None:
+    print(msg)
+    raise SystemExit(1)
 
 
 def create_permalink(args):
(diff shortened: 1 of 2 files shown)


randovania/cli/commands/permalink.py
```diff
@@ -54,14 +75,15 @@ def permalink_command(args):
...
-        help="The name of the game of the preset to use.",
+        "--preset-name",
+        type=str,
+        help="The name of the presets to use. Should be named <game>/<name>",
+        nargs="*",
+        default=[],
     )
-    parser.add_argument("--seed-number", type=int, help="The seed number. Defaults to 0.")
-    parser.add_argument("--preset-name", required=True, type=str, help="The name of the presets to use", nargs="+")
+    parser.add_argument("--preset-file", type=Path, help="The path of preset file to use", nargs="*", default=[])
```
ThanatosGit, line 86: The help text should also use the plural because it accepts multiple files like e.g. `preset-name` accepts multiple names.
...which makes me wonder if it is also worth using the plural in both argument names itself.

ThanatosGit (changes requested):

randovania/cli/commands/permalink.py
henriquegemignani: It's actually an argument you need to specify multiple times. Will change the name to be shorter and mention in the help.

ThanatosGit (approved):

PR #6696, opened 2024-06-10
Compile structs in construct_pack.encode

henriquegemignani:
Not sure I like hacking construct like that, but it's slightly faster


ThanatosGit (approved):
Nothing beats strings containing code where the string is partially created by code.

PR #6706, opened 2024-06-12
AM2R: Add liquid info as extra data

Miepee:


JeffGainsNGames (commented):
Is Liquid level drawn distance X from the top of the sceen, so bigger # = lower?

Miepee:
distance y, but yes
The exception is level 0, which means disabled

JeffGainsNGames:
X was just a dumby variable, not refering to X,Y

Also what is should_move_horizontally?

Miepee:
whether the liquid is still, or slowly moves on the horizontal axis
Example room where you can see it very clearly is water turbine station vs inner alpha nest south

JeffGainsNGames (approved):

PR #6711, opened 2024-06-12
Extra safety in broswer dialog

henriquegemignani:
Was causing crashes on macOS, but I don't think it's the reason tests are flaky.


Miepee (approved):
is it this specific test that's flaky, or something else that just causes this test to sometimes crash?

henriquegemignani:
> is it this specific test that's flaky, or something else that just causes this test to sometimes crash?

I believe what the code is doing is inherently unsafe: `self._selection_model().selection()` creates a new list, while `.first` creates a view of that list. However the list is not stored anywhere, so it's garbage. If the GC collector were to run between these two lines it could crash.

I suspect the actual reason for flaky tests is similar, but somewhere else.

PR #6717, opened 2024-06-13
Split mypy GHA job

henriquegemignani:
Makes it reasonable to mark one as required


ThanatosGit:
Since this get's changed: When I opened the Connector branch there were mypy errors on the test files with annotations here in GH.
Later, after I pushed a commit which still haven't touched that part, the annotations from mypy were gone. Is there any option to readd the annoations for a following push to a pr / branch?
Don't even know why they were gone. Does it only checks the modified files between two commits instead of comparing modified files to upstream?

henriquegemignani:
> Since this get's changed: When I opened the Connector branch there were mypy errors on the test files with annotations here in GH. Later, after I pushed a commit which still haven't touched that part, the annotations from mypy were gone. Is there any option to readd the annoations for a following push to a pr / branch? Don't even know why they were gone. Does it only checks the modified files between two commits instead of comparing modified files to upstream?

Good question. Not a clue. I didn't even plan on having the "run for modified files" version but since it just happened to work, I left it in.

ThanatosGit (approved):

PR #6718, opened 2024-06-13
Dodge around construct-typing and mypy not liking hacks

henriquegemignani:


randovania/bitpacking/construct_pack.py
```diff
@@ -16,23 +16,25 @@
 if typing.TYPE_CHECKING:
     from _typeshed import DataclassInstance
 
+    from randovania.lib.construct_stub import CodeGen
```
ThanatosGit, line 19: Uh, that doesn't exist?
And Mypy still fails because of this.

ThanatosGit (changes requested):

randovania/bitpacking/construct_pack.py
henriquegemignani: Ugh forgot to add the file

ThanatosGit (approved):

PR #6724, opened 2024-06-16
Add requirement_to_collect to ResourceNode

henriquegemignani:
Might change slightly how translators are placed for Echoes, but I didn't test too much.

### DIFF
--- a/randovania/game_description/db/resource_node.py
+++ b/randovania/game_description/db/resource_node.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 from randovania.game_description.db.node import Node, NodeContext
 
 if TYPE_CHECKING:
+    from randovania.game_description.requirements.base import Requirement
     from randovania.game_description.resources.resource_info import ResourceGain, ResourceInfo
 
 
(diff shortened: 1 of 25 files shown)


randovania/game_description/db/resource_node.py
```diff
@@ -18,9 +19,13 @@ def is_resource_node(self) -> bool:
     def resource(self, context: NodeContext) -> ResourceInfo:
         raise NotImplementedError
 
-    def can_collect(self, context: NodeContext) -> bool:
+    def requirement_to_collect(self) -> Requirement:
```
ThanatosGit, line 22: On a quick look, it looks like all except one implement it with `Requirement.trivial()`. So, I thought about make it the default and only let the one class override the method. But that's not really necessary IMO.

ThanatosGit (approved):

randovania/game_description/db/resource_node.py
henriquegemignani: There's likely others that could change, but I didn't try a lot. I'm doing this PR more so another PR can hopefully have no visible results.

PR #6735, opened 2024-06-18
Dread: Exclude door to Orange EMMI from Door Lock Rando

Tyranisaur:


ThanatosGit:
I stumbled across this PR: Is it consens to exclude this door? I think it's a very small change with high impact.
What's the reason?

Miepee:
emmi breaks

Miepee:
See the discussion here: https://discord.com/channels/914291389293027329/914508458739511358/1251733140230377533

steven11sjf:
I would suggest locking the door when it closes instead. And unlock after emmi. Should be trivial through lua.

Miepee:
If you can manage to do so in the next few days, sure.
Otherwise, this feels like a safer bet to get into the next release.

MayberryZoom:
> If you can manage to do so in the next few days, sure. Otherwise, this feels like a safer bet to get into the next release.

I agree. Imo, let's just merge this now and revisit a proper solution in the future.

MayberryZoom (approved):

ThanatosGit:
> I would suggest locking the door when it closes instead. And unlock after emmi. Should be trivial through lua.

I really prefer this because I can't express how much I hate that players are forced to remember which doors are excluded and which not. 
But it's already merged, so whatever 🤷

duncathan:
> I can't express how much I hate that players are forced to remember which doors are excluded and which not

this is a fundamental problem with door types rando and i don't think one extra door really makes much difference

duncathan:
to be more clear: door exclusions were implemented before door types were. the exclusions are necessary to prevent bugs and we should not allow the fact that door type rando cannot handle the existing system properly to hold us back from making use of it for its intended purpose

ThanatosGit:
> > I can't express how much I hate that players are forced to remember which doors are excluded and which not
> 
> this is a fundamental problem with door types rando and i don't think one extra door really makes much difference

I'm quiet late to the party: Why is this a problem with door types rando? The exclusion is for all door lock randomization modes.
I know that not all doors are randomized in Dread but it's stil a big advantage to know that a given door is never randomized.
What makes it for me so bad, is the fact that it was like that for...uh... 1 1/2 years now? So, it doesn't look like this bug is a big issue and I would have preffered a different solution instead of messing around with it that way.

But also maybe my feeling about the change is wrong. Casual players maybe don't care and racers hopefully read the changelog.

duncathan:
> Why is this a problem with door types rando? 

exclusions break the fundamental rules of types rando. casual players can and do notice the effect exclusions have and it generally feels bad. in individual DLR any one door remaining vanilla is basically impossible to notice, especially in dread since it leaves a significant portion of doors vanilla every seed anyway

> it's stil a big advantage to know that a given door is never randomized

is it? I'd be astounded if that knowledge ever made a meaningful difference in a race. 

> What makes it for me so bad, is the fact that it was like that for...uh... 1 1/2 years now? So, it doesn't look like this bug is a big issue and I would have preffered a different solution instead of messing around with it that way.

We received multiple reports of this bug in the week leading up to this pr. It needed fixing. If you prefer another solution, I would be very happy to merge it and revert this change once the other fix has been implemented

ThanatosGit:
> > it's stil a big advantage to know that a given door is never randomized
> 
> is it? I'd be astounded if that knowledge ever made a meaningful difference in a race.
> 

Of course. This one here can lock DNA. Before, I would go for the gamble that this door / the path is randomized instead of searching for wave beam. Now, I don't even need to go there which others maybe not know.
Another example: "Door to Chain Reaction Room" is excluded and is a charge beam door. The other way back is behind a trick which is not in race preset. Without being able to open charge beam doors, the only progress could be something to open charge beam doors otherwise it's not logical to go all the way up. So, I skip it and do other things first. Could be the wrong decision but it still effects how I play.


> It needed fixing. If you prefer another solution, I would be very happy to merge it and revert this change once the other fix has been implemented

Given what I know about the game, I don't think it's rocket science to disable the door if you enter the room until red chozo is dead. But having it enabled for 1 1/2 years, disable it for 1 month and then enable it again doesn't sound any better to me 😅

MayberryZoom:
Maybe we should've just forced it to be a one sided power door, considering there's a lock outside. For types, that would mean it would have zero difference, and for full DLR, it would just reduce the potential items needed on average.

henriquegemignani:
If you're that bothered by this, please go fix the issue entirely in ODR and we remove the exclusion.

duncathan:
> Maybe we should've just forced it to be a one sided power door, considering there's a lock outside. For types, that would mean it would have zero difference, and for full DLR, it would just reduce the potential items needed on average.

this is not something we have the option to do without patcher changes so it doesn't really change the decision here

PR #6764, opened 2024-06-22
MSR - Support changing environmental damage

dyceron:
Heat and Lava damage can now be configured to be constant instead of scaling

image

image


ThanatosGit (approved):
lgtm but fixing mypy is now required before merge :)

PR #6786, opened 2024-06-24
AM2R: Override light and liquid info with default levels in export

Miepee:
Might sound weird, but otherwise the game defaults are used which we do not want as they don't handle things properly.


henriquegemignani (approved):
I'll trust you that all these values for all areas is needed.

PR #6795, opened 2024-06-25
Bump OSRR to v2.2.0

dyceron:


ThanatosGit (approved):
Just to double check: No Changelog is correct? Afaik all changes should be already in the changelog or they are not active (e.g. baby hint).
Only tests needs to be fixed then.

dyceron:
Yeah the fixes were related to the item icons which is new, and multiworld, or internal stuff that isn't exposed in rdv (the negative Missile fix)

Miepee:
The OSRR schema is using a wrong default:
```json
 {
     'default': {},
     'description': 'Hint for where the Baby Metroid is located.',
     'type': 'string'
 }
```
Your default is an empty dictionary. It should be an empty string (`""`) instead.

PR #6799, opened 2024-06-25
Fix + Mypy for data_editor

ThanatosGit:
Closes #6660

### DIFF
--- a/randovania/gui/data_editor.py
+++ b/randovania/gui/data_editor.py
@@ -2,13 +2,15 @@ from __future__ import annotations
 
 import dataclasses
 import re
+import typing
 from pathlib import Path
-from typing import TYPE_CHECKING, Self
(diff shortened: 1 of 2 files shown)


randovania/gui/data_editor.py
```diff
@@ -240,14 +242,15 @@ def update_game(self, game: GameDescription):
             self.focus_on_connection(current_connection)
 
     @classmethod
-    def open_internal_data(cls, game: RandovaniaGame, edit_mode: bool) -> Self:
+    def open_internal_data(cls, game: RandovaniaGame, edit_mode: bool) -> DataEditorWindow:
```
henriquegemignani, line 245: could change the return to use `cls` instead of explicitly using `DataEditorWindow`

henriquegemignani (approved):

PR #6801, opened 2024-06-25
Bump am2r_yams to 2.7.0

Miepee:
Nothing relevant in the changelog that needs to be mentioned, as they're all just changes compared to previous dev builds.


dyceron (changes requested):
Tests

dyceron (approved):

PR #6807, opened 2024-06-26
Sort by pretty name

ThanatosGit:
Tbh, if we add more and more it needs a new UI anyways.
Currently it is sorted like "when we added" something which is like the worst, so I decided to at least sort it by the pretty name.

image


henriquegemignani (approved):

henriquegemignani:
yeah we need an "auto" thing. just gotta figure out if there's a better way that isn't "try everything"

PR #6814, opened 2024-06-27
Prime 1: add nerf_oob option

Miepee:
I wasn't quite sure where to put the option in the GUI. I don't think right now that it is (currently) affecting logic in any way, since the only oob tricks are all single room. However, it didn't really seem fit for the "Quality of Life" tab, and I didn't want to create a new tab with only this as the option. So Chaos Options it is.

UI screenshot:
grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Prime
 
+- Added: Added an option to nerf out of bounds movement.
 - Fixed: Typo on the Quality of Life preset tab.
 
 #### Logic Database
--- a/randovania/games/prime1/exporter/patch_data_factory.py
+++ b/randovania/games/prime1/exporter/patch_data_factory.py
@@ -991,6 +991,7 @@ class PrimePatchDataFactory(PatchDataFactory):
                 },
                 "requiredArtifactCount": (12 - self.configuration.artifact_target.value)
                 + self.configuration.artifact_required.value,
+                "patchWallcrawling": self.configuration.nerf_oob,
             },
             "tweaks": ctwk_config,
             "levelData": level_data,
--- a/randovania/games/prime1/gui/preset_settings/prime_patches_chaos.py
+++ b/randovania/games/prime1/gui/preset_settings/prime_patches_chaos.py
@@ -41,6 +41,7 @@ class PresetPrimeChaos(PresetTab, Ui_PresetPrimeChaos):
             self._add_persist_option(getattr(self, f"{f}_check"), f)
         signal_handling.on_checked(self.small_samus_check, self._on_small_samus_changed)
         signal_handling.on_checked(self.large_samus_check, self._on_large_samus_changed)
+        signal_handling.on_checked(self.nerf_oob_check, self._on_nerf_oob_changed)
         self.superheated_slider.valueChanged.connect(self._on_slider_changed)
         self.submerged_slider.valueChanged.connect(self._on_slider_changed)
 
@@ -74,6 +75,10 @@ class PresetPrimeChaos(PresetTab, Ui_PresetPrimeChaos):
             editor.set_configuration_field("large_samus", value)
             if value:
                 editor.set_configuration_field("small_samus", False)
+
+    def _on_nerf_oob_changed(self, value: bool):
+        with self._editor as editor:
+            editor.set_configuration_field("nerf_oob", value)
 
     def _on_room_rando_changed(self, value: RoomRandoMode):
         with self._editor as editor:
--- a/randovania/games/prime1/gui/ui_files/preset_prime_chaos.ui
+++ b/randovania/games/prime1/gui/ui_files/preset_prime_chaos.ui
@@ -43,8 +43,8 @@
         <rect>
          <x>0</x>
          <y>0</y>
-         <width>487</width>
-         <height>1190</height>
+         <width>482</width>
+         <height>1440</height>
         </rect>
        </property>
        <layout class="QVBoxLayout" name="scroll_layout">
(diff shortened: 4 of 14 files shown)


JeffGainsNGames:
is there any reason for this option?

Shaleblade:
> is there any reason for this option?

Races, primarily - either because two runners are doing "gentleman's agreements," or because the TO doesn't want racers to be able to use OoB. (That feels plausible for beginner-friendly tourneys & gimmick weeks.)

Miepee:
Essentially what Shale said. I remembered some conversation somewhere where people didn't like OoB, and made it a preset setting so that TOs can actually enforce it.

Granted, this affects Prime 1 *much* less than Echoes, but it still disallows a few things, like the upper dynamo peek, skipping some Ice Beam doors, IBBF, Power Bombs Before Bombs, the Training Chamber WC and probably a few more situational things.

I don't *think* that this affect Arbor Chamber, but I haven't actually tested it.

JeffGainsNGames:
If you are trying to add an option to enforce a rule to dis-allow OoB, this doesn't do it. 

Sure "nerf", but I could still easily do the main plaza tree OoB, life grove spinner skip, etc. in many single room situations.

If a player is actively going OoB, they certainly KNOW they are doing it (echoes accidental SA's into tunnels excluded) and are actively going against the rules if that's the terms. If this is done in a tournament setting it'll be obvious and a TO can react accordingly.

I don't see the benefit of a setting that doesn't completely solve the issue. IMO.

Miepee:
This isn't intended to dis-allow OoB. Just to nerf it.

A huge part of the reason why I opened this PR, is that it was really easy to do since randomprime supports it.
If Toaster and TO's don't want it, I'm fine with closing it. But I thought I might as well open it, in case it does become useful.

dyceron:
This feels more like a plando feature imo

Miepee:
Hmm, I just realized that it would be a very easy change to additionally scale the aether box up by a lot as well. Which would bring more value to the nerf than just "can't WC" (and would then also actually justify being in chaos options, as it probably breaks every single room out of bounds trick).

Would the thoughts be the same, if the above is done as well with the option?

JeffGainsNGames:
Scaling the Aether Box probably still doesn't guarantee a full option against OoB (see remastered), but would make most tricks either not possible (main plaza tree needs to jump off the aether floor) or not worth the investment (can just aether jump to the top of the box regardless of size).

I'm sure someone could still find scenarios to work around this setting if they tried hard enough.

At the end of the day if someones OoB it's obvious, and if it's a tourney match that's broadcasted (and the rules dis-allow OoB) then it's easy to give that player the boot.

Jason-Bee:
speaking for myself here, i don't really see value in using a preset option to nerf oob, its either one or the other. meaning either we allow oob for tournaments or we ban it. since this only prevents door transitions iam not too inclined to add it to presets.

Miepee:
> Scaling the Aether Box probably still doesn't guarantee a full option against OoB (see remastered), but would make most tricks either not possible or not worth the investment

So... the option would do its job then. Since it wasn't planned to be full protection against OoB, just heavily discourage **current** methods (e.g, get oob, and infinitely gain height via aether jumping, or infinitely loose height via morph. Both of which are now too bothersome/difficult to do)

Nystrata:
I put in my opinion that this could be a cool addition, but I'm leaning for it to be a Chaos Option

toasterparty (changes requested):
I'm pretty sure TO's have said that if an RA is not a preset option, they can't enforce adherence to it (presumably because there's risk of hearsay). If this is still true, the option actually has utility.

Chaos settings is categorically the wrong place for this. It also nullifies any argument for tournament usage because settings on that tab are typically banned outright.

This is actually a logic option and would need to go in the logic tab because of tricks like this one which would need to be updated to check for the setting:
https://www.youtube.com/watch?v=raoSUAIfIUk&t=10s

I also don't like the label for it. I'd prefer something like `Restrict Wallcrawling - Modifies most doors to prevent room transitions while out-of-bounds, thus drastically reducing sequence breaking potential.` I really just don't like the word "nerf"

It might even be worth mentioning explicitly that it does not:
- Prevent door warps
- Prevent "true" single-room oob tricks
- Prevent wallcrawl transitioning circle frigate doors or morph ball tunnel doors

toasterparty:
I think it might actually prevent you from doing some tricks like arbor chamber because idk if you could transition the room to shoot the blue side, actually

Miepee:
> I really just don't like the word "nerf"

The reason for the term "nerf" is that I knew Darkszero had a method to limit OoB via scaling up the aether box.

Hence the other comment for me: I'm slightly unsure now how this option should exactly be handled (with the exception of making it logical):
- should it only limit wallcrawls?
- should it prevent "aether-walking" by scaling up the aether box really really high?
- should it do both?
- should it do something else as well?

Miepee:
> I'm pretty sure TO's have said that if it's not a preset option, they can't enforce adherence to it

Not quite: TOs can enforce anything they want, but RAs need to be a preset option.

toasterparty:
I dislike scaling the aether box because:
1. There's a reason it exists and is the size it is in the first place, it's a performance operation
2. Most sequence breaking wallcrawls avoid aether, stuff like main plaza tree is the *exception*
3. A couple rooms actually have aether inside the room, for better or worse

Miepee:
So far the things I have heard.
Wall Crawl ban only:
- Makes things confusing because it sounds like "some OoB works, some OoB doesn't"

Aether box only:
- People will just use standables as a replacement for everything

Both:
- Completely kills OoB

Miepee:
> I think it might actually prevent you from doing some tricks like arbor chamber

I was able to do arbor chamber, but I also accidentally had "no doors" option enabled (because apparently that was hiddenly enabled in that preset...)

toasterparty:
The door in arbor chamber is kind of important for that test

Miepee:
Ok tried it again, the Arbor Chamber heist still works.
I could get the item fine, but triggering the door from the other side was slightly weird. I couldnt touch the dock trigger while in standing form, only while walking into the door and morphing I was able to trigger it.

MrMiguel211:
> Ok tried it again, the Arbor Chamber heist still works.
> I could get the item fine, but triggering the door from the other side was slightly weird. I couldnt touch the dock trigger while in standing form, only while walking into the door and morphing I was able to trigger it.

Sounds about right

Miepee:
> I'm pretty sure TO's have said that if an RA is not a preset option, they can't enforce adherence to it (presumably because there's risk of hearsay)

Judging by the echoes-no-oob tournaments (and RA's afaik) this seems to no longer be the case.
Do you still think this has utility? From what I remember in conversations, a lot of racers think this doesn't, as OoB isn't nowhere near as powerful in prime 1 compared to echoes. 
The uses for me seem limited; most of the time in races you dont really want to do oob (unless you're in a multi and bk'd), so this'd basically only forbid specific quickish peeks (arbor chambor, upper dynamo) and some "single room" oobs like the ice ruins west and exo skip. And the latter two are rarely a deciding factor in a race anyway.

Samuel6710:
> The uses for me seem limited; most of the time in races you dont really want to do oob (unless you're in a multi and bk'd), so this'd basically only forbid specific quickish peeks (arbor chambor, upper dynamo) and some "single room" oobs like the ice ruins west and exo skip. And the latter two are rarely a deciding factor in a race anyway.

Working on confirming our stance on a no OOB RA, which I believe we can enforce given that we established strict ruling on what is considered OOB in the Echoes tournament and Prime has less edge cases like with SAing into a tunnel. (Can confirm we can expect it to be a valid RA in the future)

I think for Prime though, in terms of tournament utility, it is a little niche but this setting would be much more preferable than having a no OOB RA since we don't have to do a significant amount of work checking that a player didn't go OOB. Additionally, if you're trying to do races that appeal to more beginners, this setting limits more complex OOB in favor of "simpler" ones that can encourage newer players to play.
A bit of a silly case but I can also see this getting use just from people trying to work around the setting for fun. I'm sure I will if this gets merged.

PR #6816, opened 2024-06-27
Add documentation on how we make a new release

Miepee:
Mostly copy pasted from the pin in #randovania-dev


Miepee:
Some other things probably worth noting here:
- Don't keep a release in pre-release state, as that will cause a server-client mismatch
- Don't delete releases, for the above reason. Instead, make another release that's identical to the previous one
- Always release on 1st, even when we have our "rc" late (except for very special exceptions which should be announced)
- Encourage pinging beta testers multiple times instead of waiting to group pings

Maybe some more things I forgot.

henriquegemignani:
Can you also mention the release schedule aka #2531?

Miepee:
Mentioned the release schedule, and also incorperated my other notes from above.

docs/New Release Process.md
```diff
@@ -0,0 +1,37 @@
...
+- Get the fixes to the `stable` branch however you want (commit directly, cherry pick).
+- Make sure the changelog is correct.
+- Check the build is working.
+- Make a (signed) tag, push, wait for CI.
+- Copy-paste the changelog from CHANGELOG.md into the GH Release.
+- Website gets automatically updated. For Flathub, [merge the PR](https://github.com/flathub/io.github.randovania.Randovania/pulls) that will get automatically created.
+- Make a PR that merges stable into main.
+
+Some other miscellaneous notes:
+- The server version always has to be in sync with the client version. For this reason, do **never** delete releases, nor keep releases in the pre-release state for longer than they're needed. If you want to delete a release, make a new release that reverts the changes.
```
henriquegemignani, line 36: "do never delete" I think this need fixing

Miepee: leftover from a previous draft whooops

henriquegemignani (approved):

henriquegemignani (changes requested):
Call the file `Release Process.md`, not `New Release Proccess`

henriquegemignani (approved):
nvm I can do that easily

Miepee:
I was just attempting to do that and then got the "canot push because you need to pull before hand" error lol

PR #6828, opened 2024-06-29
MSR <-> Prime 1/2 SJ for Offworld Progressive Jump

dyceron:


ThanatosGit (approved):
Ok, but tests needs to be fixed.
I guess it's because it is used in a "multiworld patch data thing" test.

dyceron:
Yeah I always forget to check the tests.

PR #6830, opened 2024-06-29
DRAFT: copy on game details

Miepee:
Someone please check this branch out and report how the button looks on non-linux.

Here's how it looks for me:
grafik

Current known issue that the icon does not change color on light theme:
grafik

### DIFF
--- a/randovania/gui/game_details/game_details_window.py
+++ b/randovania/gui/game_details/game_details_window.py
@@ -87,6 +87,7 @@ class GameDetailsWindow(CloseEventWidget, Ui_GameDetailsWindow, BackgroundTaskMi
         self._tool_button_menu.addAction(self._action_view_trick_usages)
 
         # Signals
+        self.permalink_copy_button.clicked.connect(self._permalink_copy_clicked)
         self.export_log_button.clicked.connect(self._export_log)
         self.export_iso_button.clicked.connect(self._export_iso)
         self._action_open_tracker.triggered.connect(self._open_map_tracker)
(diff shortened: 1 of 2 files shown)


ThanatosGit:
> does not change color on light theme

It's the opposite on windows
image
image

Miepee:
Is your system/Windows theme light or dark?
For me in the above screenshots, I was on dark mode. If I switch my system theme to light mode (I have never seen my desktop blinding me so much before), I get the same results as you:
image
image

ThanatosGit:
> Is your system/Windows theme light or dark?

What is this question 😛?
I use "Dunkel" on Windows 10 in these settings here and everything else like Github etc. also appears in dark mode. Should be the system settings. Just saying this because windows is good in hiding the same or the nearly the same settings in 2 different menus when they switched the whole control panel. 
image

Miepee:
Does the clipboard icon change, if you change that setting to "Hell"?

ThanatosGit:
image
image

Doesn't change in light mode.

Miepee:
Weird.
Wonder if it would be better to just ship our own icon.
Cause writing logic for
```py
if linux:
  if system_theme is not randovania_theme:
    invert_image_colors()
 elif windows:
   if randovania_theme is dark:
     invert_image_colors()
 elif macos:
   # ???
   pass
 ```
 doesn't sound like fun.
 
 
 I wouldn't be surprised if QtQuick had a proper solution to this, but I also am not keen on just rewriting everything into qtquick.

ThanatosGit:
I switched to Windows 11. lol
image
image

Miepee:
Gonna close this PR, as every OS coloring the icon in its own way is a nightmare.
No clue how to do it better, we probably need to provide our own icon and then recolor it based on whether dark mode gets (de)actiavted? Which also doesn't sound as nice. IDK.
Advice QT Experts welcome

PR #6853, opened 2024-07-04
MSR - Add more starting locations

dyceron:
Most Chozo Seals, all major locations (minus Power Bomb), most Metroid checkpoints, and some other misc spawns have been added as valid starting locations.


Miepee (approved):
hope you tested that these work

dyceron:
I may or may not have found some problematic ones 🙃

PR #6854, opened 2024-07-04
Fix some CI things

Miepee:
- should fix CI not upload *any* artifacts when creating a release
- should fix CI not being able to make a push to update website automatically.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -563,7 +563,7 @@ jobs:
         if: ${{ startsWith(github.ref, 'refs/tags/v') }}
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
-          file: packages/Randovania * *
+          file: packages/Randovania*
           file_glob: true
           tag: ${{ github.ref }}
           overwrite: true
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-publish.yml
```diff
@@ -647,15 +647,16 @@ jobs:
         env: 
           TAG_NAME: ${{ github.ref }}
         run: |
-          # Tag is of format vX.Y.Z, cut the v off.
-          echo "NEW_VERSION=$(${TAG_NAME:1})" >> $GITHUB_OUTPUT
+          # Tag is of format refs/tags/vX.Y.Z, cut everything before X.Y.Z off.
+          echo "NEW_VERSION=$(${TAG_NAME:11})" >> $GITHUB_OUTPUT
```
ThanatosGit, line 651: Can you please double check if this is correct? The CI in stable gave a command not found thing here.
I think it should be like I wrote in the discord without the double `$` and only the one with curly braces. 
`echo "NEW_VERSION=${TAG_NAME:11}" >> $GITHUB_OUTPUT`

Miepee: oh im dumb. the `$()` is a leftover from when there actually *was* a command in there. thanks for noticing

PR #6864, opened 2024-07-05
Web API: handle if sessions don't exist

henriquegemignani:


Miepee:
Is this related to #1775, or is this the admin panel?

henriquegemignani:
This is for admin panel, yes

randovania/server/multiplayer/web_api.py
```diff
@@ -123,7 +127,10 @@ def admin_session(user: User, session_id: int) -> ResponseReturnValue:
 
 
 def download_session_spoiler(user: User, session_id: int) -> ResponseReturnValue:
-    session: MultiplayerSession = MultiplayerSession.get_by_id(session_id)
+    try:
+        session: MultiplayerSession = MultiplayerSession.get_by_id(session_id)
+    except MultiplayerSession.DoesNotExist:
+        return "Session not found", 404
```
Miepee, line 133: the same string+error code is used 3 times, does it make sense to just have a shorthand attribute for them in case we ever e.g. change the error message?

henriquegemignani: The message's not really intended for end user and I'd rather figure out a way to avoid the try/except being repeated

Miepee: then i'll merge.

Miepee (approved):

Miepee:
clicked the wrong button, whoops

PR #6871, opened 2024-07-06
MSR - Add cosmetic option to shuffle music

dyceron:
Supports shuffling music via a cosmetic option. Can shuffle music by type or shuffle all music together.

Requires https://github.com/randovania/open-samus-returns-rando/pull/398 to be merged and a new patcher release.

image


ThanatosGit:
I don't know if I understand correctly but full mode can shuffle jingle music to exploration and runs it in a loop? 
Ngl, if this can play endless times a 5 seconds jingle melody while exploring an area, that is an anti-cosmetic or uglify option, which sounds so bad that I don't even like adding it at all.

It doesn't need to be only the types option but I would prefer a full option which at least trys to avoid make everything garbage.

dyceron:
If you're biggest issue with this is the jingles in full, then I can change how it is. I agree it'd be annoying, but also at the same time, full would already be semi annoying since having a boss theme as an exploration theme would get old. I think thats just kinda how it is with full. Types is the best option to use.

dyceron:
I'll change this later. I'll probably change the jingles to only shuffle amongst themselves regardless of the mode.v

Miepee:
> is just hot garbage

There are a few folks out there, who like to use Full Garbage modes. You could maybe hide that setting behind experimental or something, so that its only chosen by people who know what they're doing. 
 
But ignoring ~~the masochists~~ those people to have no chance of someone even accidentally enabling it does sound safer.

randovania/games/samus_returns/exporter/patch_data_factory.py
```diff
@@ -103,6 +105,97 @@ def get_resources_for_details(
...
+    excluded_list = [
+        "m_movie_ship_landing99",
+        "m_staffcredit99",
+    ]
+
+    if music_mode == MusicMode.VANILLA:
+        return {}
+
+    # Music is now either TYPE or FULL
+    assert music_mode in (MusicMode.TYPE, MusicMode.FULL)
```
ThanatosGit, line 178: Is there a reason I'm missing for this assertion?
For people messing around with this in the rdvgame file and with the `else` part they would get types mode if it is not a value of the enum?

randovania/games/samus_returns/exporter/patch_data_factory.py
```diff
@@ -103,6 +105,97 @@ def get_resources_for_details(
...
+
+    total_orig = combat_list + exploration_list + fanfare_list
+
+    if music_mode == MusicMode.FULL:
+        total_orig += excluded_list
+        total_new = random_lib.shuffle(rng, total_orig)
+    else:
+        # MusicMode is TYPE
+        # TODO: copying is not necessary anymore, clean this up in the future.
```
ThanatosGit, line 187: Could do it now :p

randovania/games/samus_returns/exporter/patch_data_factory.py
```diff
@@ -103,6 +105,97 @@ def get_resources_for_details(
...
+    ]
+
+    fanfare_list = [
+        "event_jingle_placeholder",
+        "k_matad_jinchozo",
+        "matad_jintojo_32728k",
+        "special_ability2_32",
+        "sphere_jingle_placeholder",
+        "tank_jingle",
+    ]
```
ThanatosGit, line 167: Like already discussed, please shuffle fanfares even in full mode only with one of the same list.

ThanatosGit (changes requested):

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added: More starting locations have been added for all areas.
 - Added: Progressive pickups can now be configured to be placed vanilla.
+- Added: New cosmetic to shuffle music either by song type or full shuffle.
```
ThanatosGit, line 22: I'm not sure if this should have the word `[New cosmetic] option` added, so I just leave it as a comment.

ThanatosGit (approved):

CHANGELOG.md
dyceron: I'll fix it. Does this pr also need the persisted options updated?

Miepee: yes

PR #6872, opened 2024-07-07
AM2R: Add Area rando

Miepee:
This feels kinda hacky (but it works). Comments very appreciated on how to do this with less copy-paste. Feels like I can probably use lambdas in a few places, but no idea how.

Will fix patcher tests and changelog entry after initial core review is done.

### DIFF
--- a/randovania/games/am2r/exporter/patch_data_factory.py
+++ b/randovania/games/am2r/exporter/patch_data_factory.py
@@ -487,6 +487,26 @@ class AM2RPatchDataFactory(PatchDataFactory):
             if (
                 isinstance(node, DockNode)
                 and node.dock_type in self.game.dock_weakness_database.all_teleporter_dock_types
+                and node.dock_type.short_name == "teleporter"
+            )
+        }
+
+        entrances = {
+            str(node.extra["instance_id"]): {
+                "dest_id": connection.extra["instance_id"],
+                "dest_direction": connection.extra["facing"],
+                "dest_room": self.game.region_list.area_by_area_location(connection.identifier.area_identifier).extra[
+                    "map_name"
+                ],
+                "force_idle_after_transition": (node.extra["facing"] == connection.extra["facing"])
+                or (node.default_dock_weakness != connection.default_dock_weakness),
+            }
+            for node, connection in self.patches.all_dock_connections()
+            if (
+                isinstance(node, DockNode)
+                and isinstance(connection, DockNode)
+                and node.dock_type in self.game.dock_weakness_database.all_teleporter_dock_types
+                and node.extra.get("is_area_transition", False)
             )
         }
 
@@ -498,6 +518,7 @@ class AM2RPatchDataFactory(PatchDataFactory):
             "rooms": self._create_room_dict(),
             "game_patches": self._create_game_patches(self.configuration, pickup_list, text_data, self.rng),
             "pipes": pipes if self.configuration.teleporters.mode != TeleporterShuffleMode.VANILLA else {},
+            "entrances": entrances if self.configuration.areas.mode != TeleporterShuffleMode.VANILLA else {},
             "door_locks": self._create_door_locks(),
             "hints": self._create_hints(self.rng),
             "cosmetics": self._create_cosmetics(self.description.get_seed_for_player(self.players_config.player_index)),
--- a/randovania/games/am2r/game_data.py
+++ b/randovania/games/am2r/game_data.py
@@ -20,7 +20,7 @@ def _gui() -> game.GameGui:
         cosmetic_dialog=gui.AM2RCosmeticPatchesDialog,
         export_dialog=gui.AM2RGameExportDialog,
         progressive_item_gui_tuples=progressive_items.tuples(),
-        spoiler_visualizer=(gui.AM2RTeleporterDetailsTab,),
+        spoiler_visualizer=(gui.AM2RTeleporterDetailsTab, gui.AM2RAreaRandoDetailsTab),
     )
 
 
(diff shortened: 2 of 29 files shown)


Miepee:
TODO: needs preset describer.

ThanatosGit:
I just had a quick look over this PR and it looks to me that the PR has a huge conceptual problem.
The smaller issue is that it adds the `is_teleporter` flag to all Doors and then it adds an extra guard with `is_area_transition` to ignore most of them. I was first thinking that it needs a different type for such transitions but that would probably mess around with DLR (and custom docks in open transitions ?) such that it would require similiar hacks in the code for other randomizations. But maybe anyone come up with something better.

The much bigger issue is that due to it being added with `is_teleporter`, it's now in the `all_teleporter_dock_types` but every piece of code is working around the fact that it is in this list.
E.g. `locations = TeleporterList.nodes_list(self.game_enum)` was replaced with something custom to remove the `door` type from it.
`AM2RTeleporterDetailsTab` now inherits from a more general class just to replace the `_fill_per_region_connections` to once again work around the issue that there is more in `all_teleporter_dock_types`. Same for `AM2RAreaRandoDetailsTab` but just that it "kicks" the type for the pipe rando (basically the opposite of the previously named `AM2RTeleporterDetailsTab`).
The PDF and `base_patches_factory` all suffer from the same problem. Same for why the changes in `teleporter_distributor.py` were necessary. Like I said, it's like in every file changed.

Back when I was thinking about how to add Dread's `Teleportal`s as `Teleporter` and allow them to shuffle independently from the Elevator and Shuttles, I though about adding something like a "teleporter group". Everything would default to "group A", which means no changes for the existing games (but maybe for Portal rando in Echoes?) but you can specify a different group on the dock type. I never thought about that in detail and it would require to change the helper functions to take an optional parameter for the group but it would be a much better solution which also can be used by any game.

Miepee:
> I was first thinking that it needs a different type for such transitions but that would probably mess around with DLR (and custom docks in open transitions ?) such that it would require similiar hacks in the code for other randomizations. 

Yeah, the door type can't change for these. They need to be X Door / Open Transition in order to work with DLR properly. I could *theoretically* change the type for some of the non 4-height transitions. But I already kinda need to workaround the types not being a teleporter type, so doing that for only a few other types doesn't really matter too much. I'm sure that not many games actually would run into this issue.

> I though about adding something like a "teleporter group"

I like this in concept. Could introduce a new field to the DB to be something like:
```json
"teleporter_rando": {
  "default": [
    "Tram",
    "Elevator"
  ],
  "portals": [
    "Teleportal"
  ]
}
```

or for am2r's case: [teleporter-pipes], [door, dock]. Then modify all the functions to use `default`, with it being overridden. For me that still remains:
- No idea how entrance rando should handle that, I will just leave it as a problem for the future:tm:
- AM2R would still need to override its list to check for "is_area_transition" 
- What about the fact that the default teleporter configuration field is `teleporters`? Currently in the Area rando tab, i just explicitly overwrite the necessary function to use the `areas` field instead, but this feels kinda ugly if multiple games would start doing that.

ThanatosGit:
> * AM2R would still need to override its list to check for "is_area_transition"

Maybe it would be possible to also add a (optional) game specific check? E.g. like I did with the `all_preplaced_item_locations` function which takes callable.
So you could tell the function for which teleporter group it should return the nodes and also what game specific check it should take care of.


> * What about the fact that the default teleporter configuration field is `teleporters`? Currently in the Area rando tab, i just explicitly overwrite the necessary functions to use the `areas` field instead, but this feels kinda ugly if multiple games would start doing that.

If I understand correctly that's about the parent class `PresetTeleporterTab`. Isn't it possible to simple replace the calls to the property method `editor.layout_configuration_teleporters` by a `TeleporterConfiguration` you pass to the parent class from the child class?


Oh boy, I really should take some time in the future and read the generator code. While scrolling through occurences of `all_teleporter_dock_types`, I saw this one: https://github.com/randovania/randovania/blob/1e5f349ffbd176b27f51069e3512738ab087fe01/randovania/generator/filler/player_state.py#L190-L209

Isn't something like the check in L193 already broken because it doesn't take care of your two different groups and also not for your `is_area_transition` thing? If that's true, I don't even see a way around first refactoring the whole stuff before this PR.

Miepee:
>  Isn't it possible to simple replace the calls to the property method editor.layout_configuration_teleporters by a TeleporterConfiguration you pass to the parent class from the child class?

It could be that I'm not understanding things correctly, but I don't see it that way. I don't see how the hardcoded `layout_configuration_teleporters`  to `self.configuration.teleporters` can be easily replaced with `self.configuration.areas` from a childs class.

> Isn't something like the check in L193 already broken

Kinda? That specific instance only seems to be used for "logging". But yeah that definitely should be fixed first.

ThanatosGit:
> > Isn't it possible to simple replace the calls to the property method editor.layout_configuration_teleporters by a TeleporterConfiguration you pass to the parent class from the child class?
> 
> It could be that I'm not understanding things correctly, but I don't see it that way. I don't see how the hardcoded `layout_configuration_teleporters` to `self.configuration.teleporters` can be easily replaced with `self.configuration.areas` from a childs class.

Wouldn't it be something like that (just cleaner of course :p)? And you just call `super().__init__` with `areas` as the last parameter in Area tab thing?
https://github.com/randovania/randovania/compare/main...ThanatosGit:please-no?expand=1

Miepee:
Oh yeah I'm dumb lol

duncathan (changes requested):
I haven't looked very closely at any code yet but *please* do not call this "Area rando". immensely confusing since it contradicts the meaning of `Area` in the rest of RDV

Miepee:
Am aware, I haven't found a better name for it tho :cry:

Areas is how they're known (/called? Need to double check in-game). Region rando thus sounds a bit confusing.

Miepee:
This needs a few backend changes and so i'll close this until those are resolved.
Being in PR history is enough for a reference.

PR #6873, opened 2024-07-07
Am2r: Add tileset and background rotation

Miepee:
Only merge after yams bump please.
grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added: Progressive pickups can now be configured to be placed vanilla.
 - Added: The first log entry in the game now displays a history of the collected items.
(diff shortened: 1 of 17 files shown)


Miepee:
No idea why that other test fails :cry:

Miepee:
Doing some debugging, for some reason on type->vanilla / full->vanilla, the `_on_music_option_changed` is never even triggered. no idea how.

Miepee:
After trial and error: *Something* in the UI file causes the test to fail. No idea what.

Miepee:
If either of the `tileset_rotation_spinner_max` or `background_rotation_spinner_max` items is in the UI file, it causes the other test to fail.

Miepee:
Seems to be specifically, if they are in the layout.

Miepee:
Fixed the test.

dyceron (approved):
Lgtm

PR #6874, opened 2024-07-07
Bump am2r_yams to 2.8.0.dev3

Miepee:


dyceron (approved):
Interesting release version, but I'm assuming there's a reason for it.

Miepee:
Yes. Don't want to mark this as a stable release, but want to get this into rdv for testing.

PR #6899, opened 2024-07-13
Resolver: Fix wrong additional requirement when rolling back progress…

Tyranisaur:
…ives

If one progressive item is located in a position where a later item in the same progressive chain is required to escape, and the resolver finds that item first, then the resolver must not set the later progressive item as an additonal requirement when rolling back obtaining the earlier progressive item.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Filtering in the Generation Order tab is now case-insensitive.
 - Changed: The Area View of the Data Visualizer now always has a dark gray background to help with readability.
 
+### Resolver
+
+- Fixed: Some seeds being considered impossible when finding a progressive item in an area where a later item in the progressive chain is required to leave.
+
 ### AM2R
 
 #### Logic Database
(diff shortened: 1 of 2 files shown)


henriquegemignani (commented):
I'm not entirely sure I like this solution. The function name in Node, the complex implementation that is repeated compared to other methods.

Everything here feels like plugging a hole instead of trying to solve the problem at the root.

Tyranisaur:
> I'm not entirely sure I like this solution. The function name in Node, the complex implementation that is repeated compared to other methods.
> 
> Everything here feels like plugging a hole instead of trying to solve the problem at the root.

Maybe I came up with a less elegant solution because I'm not intimately familiar with the architecture here. It looks kind of ugly because of how in this step of the resolver, it needs to check if we're rolling back obtaining an item in a progressive chain, and do some logic with that. I ended up adding the function in Node to get code that's generic. Maybe I misunderstood, but I tought using instanceof was discouraged.

To some extent, having to account for progressive items in this way feels like plugging a hole, but primarily because it deals in terms that are usually abstracted away into the resource gain abstraction. If I put the entire solution in resolver.py does that constitute fixing the problem at the root?

Tyranisaur:
I moved everything into resolver.py, renamed some things and move some things into their own functions.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Reduced some visual noise in the main window and customize preset window.
 - Changed: Don't show unnecessary lines on Door Lock and Teleporter rando customization tabs.
 
+### Resolver
```
Miepee, line 17: can you rebase this to not appear in 83x?

henriquegemignani (approved):

PR #6925, opened 2024-07-19
Add more sentry metrics

Miepee:
See title.
No changelog cause internal stuff.

### DIFF
--- a/randovania/interface_common/generator_frontend.py
+++ b/randovania/interface_common/generator_frontend.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import multiprocessing
+import operator
 from concurrent.futures import ProcessPoolExecutor
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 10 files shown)


randovania/interface_common/generator_frontend.py
```diff
@@ -60,6 +62,50 @@ def generate_layout(
             if manager.is_included_preset_uuid(preset.uuid):
                 span.set_tag("builtin_preset", f"{preset.name} ({preset.game.short_name})")
 
+        for preset in parameters.presets:
+            monitoring.metrics.incr(
+                "generation_trick_amount_disabled",
+                value=operator.countOf(
+                    preset.configuration.trick_level.specific_levels.values(), LayoutTrickLevel.DISABLED
```
henriquegemignani, line 69: This can definitely be a for loop instead of the same block of code 6 times.

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #6934, opened 2024-07-20
Dread: fix hide scans with nothing data

### LINKED ISSUE #6658: Option to hide item scans doesn't work in Dread
image
I was under the impression that the option to hide an item's scan translates to the name that appears on the map when you look at the item. 
image
It seems odd that we would have support for hiding the item name when picking up the item, but not when looking at the map, given that the opposite way around seems like the most natural configuration to use when playing with this option.
image

Miepee:
Fixes #6658

### DIFF
--- a/test/games/test_log_file_export.py
+++ b/test/games/test_log_file_export.py
@@ -109,6 +109,7 @@ def pytest_generate_tests(metafunc: _pytest.python.Metafunc) -> None:
         "dread/custom_start.rdvgame",  # crazy settings
         "dread/custom_patcher_data.rdvgame",  # custom patcher data
         "dread/all_settings.rdvgame",  # all settings enabled
+        "dread/hide_all_with_nothing.rdvgame",  # Model+scan+name hidden with nothing data
         # Prime 1
         "prime1_crazy_seed.rdvgame",  # chaos features
         "prime1_crazy_seed_one_way_door.rdvgame",  # same as above but 1-way doors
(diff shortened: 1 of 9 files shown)


test/games/test_log_file_export.py
```diff
@@ -66,7 +66,7 @@ def test_layout_patch_data_export(
 
     data = factory.create_data()
 
-    # _update_committed(data_path, cosmetic_path, data, cosmetic_patches)
+    _update_committed(data_path, cosmetic_path, data, cosmetic_patches)
```
dyceron, line 69: Nice

Miepee: ops

test/games/test_log_file_export.py
```diff
@@ -109,6 +109,7 @@ def pytest_generate_tests(metafunc: _pytest.python.Metafunc) -> None:
         "dread/custom_start.rdvgame",  # crazy settings
         "dread/custom_patcher_data.rdvgame",  # custom patcher data
         "dread/all_settings.rdvgame",  # all settings enabled
+        "dread/hide_all_with_nothing.rdvgame",  # Model+scan+name hidden with nothing data
```
dyceron, line 112: This test seems a but unnecessary the only difference is two lines are slightly different. The rest of the rdvgame is standard.

Miepee: this test is to ensure that nothing breaks this in the future.
Its the easiest way to test things, ideas welcome for another test.

Miepee: also unsure where you got your two lines from?
every time `"map_icon": {`is used, its different now

henriquegemignani (approved):
YOUR NOTHING

PR #6939, opened 2024-07-20
Default InfiniteTimer to not strict

henriquegemignani:
Dunc had the correct idea when hii made the CS connector use a non-strict timer: if `target` runs for longer than `interval`, then `_current_task` will be not-None in `_on_timeout`. Since we use this timer to make network calls, it's perfectly reasonable for them to take forever.

For an infinitely repeatable timer, we don't really mind skipping a few triggers so defaulting strict to false.

No changelog because I'm not even sure this error shows up visibly for users. And the message would be something like "Fixed: An invisible error that could happen in rare cases with is now fixed."

### DIFF
--- a/randovania/lib/infinite_timer.py
+++ b/randovania/lib/infinite_timer.py
@@ -15,7 +15,7 @@ class InfiniteTimer(QtCore.QObject):
     should_start_timer: bool = False
     _current_task: asyncio.Task | None = None
 
-    def __init__(self, target: Callable[[], Awaitable[None]], interval: float, *, strict: bool = True):
+    def __init__(self, target: Callable[[], Awaitable[None]], interval: float, *, strict: bool = False):
         super().__init__()
         self._dt = interval
         self.target = target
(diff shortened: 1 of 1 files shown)


Miepee:
Its supposed to make these not show in sentry, right?
https://randovania.sentry.io/issues/4227860587/?end=2024-07-19T23%3A59%3A59&environment=production&environment=staging&project=4504594037211137&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D+timer&referrer=issue-stream&sort=freq&start=2024-07-18T00%3A00%3A00&stream_index=0

henriquegemignani:
Yes

Miepee (approved):
I'll just trust you on the fact that this error is never visible for users

henriquegemignani:
I'm mostly guessing, because I don't have a clue how it shows up. Random popup at weird times?

Miepee:
I'd assume that if it shows up, that we wouldve had a few reports in #bug-reports
But since no one ever has reported this... :shrug:

PR #6940, opened 2024-07-20
Pure asyncio timers for InfiniteTimer

henriquegemignani:
I always hated using Qt for the Game Connector. If we also stop using signals we can have a non-qt build of Randovania that can do multiworld things.

Plus if we're calling the timeout via the event loop, hopefully https://randovania.sentry.io/issues/3920644751/ won't happen anymore.

Leaving for next release because I want a full month of testing for this.

### DIFF
--- a/randovania/lib/infinite_timer.py
+++ b/randovania/lib/infinite_timer.py
@@ -4,45 +4,46 @@ import asyncio
 import logging
 from typing import TYPE_CHECKING
 
-import shiboken6
-from PySide6 import QtCore
-
 if TYPE_CHECKING:
     from collections.abc import Awaitable, Callable
 
 
-class InfiniteTimer(QtCore.QObject):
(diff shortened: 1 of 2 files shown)


Miepee:
Based on the failing tests, this PR seems unfinished?

henriquegemignani:
Forgot about fixing tests. Needed to be redone because they cared about tweaking the internal timer

Miepee (approved):

Miepee:
Since we have at least one occurrence of this error appearing to the end user, i guess this means that a changelog entry should now be added? :P

henriquegemignani:
The fix is the other pr :p
If you can come up with a changelog entry that isn't "fixed a bug", sure

Miepee:
Too lazy to come up with a proper changelog entry, so just merging xP

PR #6950, opened 2024-07-22
AM2R: update trick documentation requirements

JeffGainsNGames:
copying MSR and doing this

### DIFF
--- a/randovania/games/am2r/logic_database/header.json
+++ b/randovania/games/am2r/logic_database/header.json
@@ -826,19 +826,19 @@
             "Combat": {
                 "long_name": "Combat",
                 "description": "If this trick is enabled, the player may be expected to defeat enemies and bosses with fewer items and less health.",
-                "require_documentation_above": 0,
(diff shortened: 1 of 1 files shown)


randovania/games/am2r/logic_database/header.json
```diff
@@ -904,7 +904,7 @@
             "MissileLessMetroids": {
                 "long_name": "Missile-Less Metroid Fights",
                 "description": "If this trick is enabled, the player may be expected to kill Metroids with only Charge Beam.",
-                "require_documentation_above": 0,
+                "require_documentation_above": 1,
```
Miepee, line 907: wonder if this should be 0 instead? alphas all use one template, adding one video there wouldnt hurt.

JeffGainsNGames: Were dodging vs non-dodging not split?
honestly if we just reuse the same video can just stay 0 then.

Miepee (approved):

PR #6960, opened 2024-07-23
Enable sentry's asyncio support

henriquegemignani:
This seems to automatically make a function span for each asyncio task. It'll make a bunch of stuff show there, but I'm more curious about transactions not leaking into unrelated issues. Hopefully this fixes that.


Miepee (approved):
> It'll make a bunch of stuff show there

What kind of stuff?

PR #6966, opened 2024-07-23
More user logging, better scope management for sentry

henriquegemignani:


randovania/monitoring/__init__.py
```diff
@@ -150,8 +150,9 @@ def traces_sampler(sampling_context):
 def client_init() -> None:
     _init(False, "client", exclude_server_name=True)
 
-    sentry_sdk.set_tag("frozen", randovania.is_frozen())
-    sentry_sdk.set_tag("cpu.architecture", platform.machine())
+    global_scope = sentry_sdk.Scope.get_global_scope()
+    global_scope.set_tag("frozen", randovania.is_frozen())
+    global_scope.set_tag("cpu.architecture", platform.machine())
```
Miepee, line 155: Thoughts on logging the CPU name too? Seems useful for our performance measuring.

henriquegemignani: I thought that'd be hard but turns out `platform.processor()` exists

randovania/monitoring/__init__.py
```diff
@@ -153,6 +154,8 @@ def client_init() -> None:
     global_scope = sentry_sdk.Scope.get_global_scope()
     global_scope.set_tag("frozen", randovania.is_frozen())
     global_scope.set_tag("cpu.architecture", platform.machine())
+    global_scope.set_tag("cpu.processor", platform.processor())
```
Miepee, line 157: i thought you wanted to revert this?

henriquegemignani: It works on Windows, 🤷 about not having data for linux

Miepee: Our linux userbase is only ~2% so fair enough. Worst case can always change it in the future to pull data from lscpu or something.

Miepee (approved):

PR #6970, opened 2024-07-24
When exporting, hash inputs when an exception happens

henriquegemignani:


randovania/exporter/game_exporter.py
```diff
@@ -58,18 +67,67 @@ def export_game(
...
+                finally:
+                    self._after_export()
+
+
+def _hash_file_internal(path: Path) -> typing.Any:
+    sha = hashlib.sha1()
+
+    with path.open("rb") as f:
+        while True:
+            data = f.read(65536)
```
Miepee, line 104: why this number specifically?

randovania/exporter/game_exporter.py
```diff
@@ -58,18 +67,67 @@ def export_game(
...
+                    scope.capture_exception(e, contexts=input_hash)
+                    # TODO: compare `input_hash` with known good values and if different,
+                    # wrap into a "User Error" exception
+                    raise
+
+                finally:
+                    self._after_export()
+
+
+def _hash_file_internal(path: Path) -> typing.Any:
```
Miepee, line 99: any? :cry:

henriquegemignani: I tried but couldn't find a decent hint. Best has `_hashlib.HASH` but at least pycharm didn't like

randovania/exporter/game_exporter.py
henriquegemignani: random example I found online. Seemed like a decent chunk to read, being 64kb

Miepee: If we only read 64kb, isnt the system then useless for e.g the gamecube games where there were dumping errors happening after 64kb?

henriquegemignani: There's a `while True`?

Miepee: nvm whoops.

PR #6981, opened 2024-07-26
MSR - Move `boss_go` and `boss_mae_32` to the correct type in PDF

dyceron:
These tracks were in the wrong section.


Miepee (commented):
What somgs are these? Since they contain the word "boss" it sounds very weird if they're in exoloration.

dyceron:
The ambient theme that plays after defeating a metroid (briefly) or boss. After resisting to them after making this, I was having doubts on moving them since there are less than 1 miuln long each. Might just close this pr.

Miepee:
Id rather have them as exploration than as boss music.
Im fine with listening to a short calm ambient soundtrack while exploring, but not while fighting a boss.
The alterantive is also to move htem to excluded

dyceron:
Ok last but, both are pretty identical, but one has extra beeps in it (the classic item room theme). Maybe the other one should be moved to excluded, or just keep both in exploration anyway.

Miepee:
Having one in excluded and one in exploration feels weird. Either both excluded or both shuffled.

dyceron:
Then I'll just leave it as is.

Miepee (approved):

PR #6462, opened 2024-04-27
Prime: Logic Fixes and Root Cave Additions

Samuel6710:

PR #6993, opened 2024-07-28
Prime: re-add Mines Security Station barrier item reqs

JeffGainsNGames:
accidently removed in #6462.


JeffGainsNGames:
shows a diff in some of the combat logic but that's because of the change to the and over the whole requirements. I did not modify any of the combat reqs.

henriquegemignani (approved):

PR #7022, opened 2024-07-30
Fix MW tab selection

### LINKED ISSUE #7012: Multiworld session: "Users and Worlds" tab acts very strange in regards to activeness
For some reason that tab seems to decide randomly on when you can actually click it / when its considered being hovered over

e.g. i have it here in the middle of the tab, but i cant click it 
grafik

From basic hovering around, seems like it only gets active in the red area somehow
grafik

Miepee:
Fixes #7012


dyceron (approved):
I'm assuming this just makes the clickable area bigger

Miepee:
Ends up having that effect. But actually this removes an invisible UI component that was overlaid there, which didn't need to be one

PR #7036, opened 2024-08-03
Prime 1: Add Unlimited Missiles+PBs as items

### LINKED ISSUE #7006: Prime1: Fake Unlimited missiles/power bombs
A fake unlimited missiles/power bombs, which is just an expansion which gives 99999 ammo. Model should be GCN.

Miepee:
Tested this in-game, works as expected.

Technically its not "unlimited" but you'd need 10 minutes to get rid of all of your PBs and ~5minutes to get rid of missiles with constant wavebuster.

Model is gamecube because toaster said so. Dunno what the reasoning is for not choosing a big missile expansion / big pb expansion, but I would assume its because just scale is less recognizable at first glance.

Fixes #7006


UltiNaruto:
Feels rather hacky

Miepee:
how so

UltiNaruto:
it's not really unlimited + in Echoes unlimited missiles doesn't give the item if you don't own at least one missile

toasterparty:
Why stop at 99 and 9999. If there's a possibility of running out just 100x both

Miepee:
Because randomprime seems to cap. The 9999 for Missiles doesnt even work right now, it only gives 999.

toasterparty:
@Miepee randovania is telling randomprime what the max capacity should be, double check the max cap

Miepee:
Ooh, I'll toy around with it later. Does that do some weird stuff on the hud if they're trying to display really large numbers?

PR #7038, opened 2024-08-03
MSR - A6 Crumbling Stairwell Door Exclusion

dyceron:
Makes going for the DNA not dangerous


Miepee (approved):
Needed to look at the room layout but sounds good to me. I think ideally, we'd only exclude it in individual DLR, but this works.

dyceron:
> Needed to look at the room layout but sounds good to me. I think ideally, we'd only exclude it in individual DLR, but this works.

The door only gets shuffled in individual anyway since it's a power closed door, so that's fine

PR #7039, opened 2024-08-04
Revert "Bump cryptography from 42.0.8 to 43.0.0"

henriquegemignani:
Reverts randovania/randovania#6947

in case it's breaking things


Miepee (approved):

duncathan:
this fixes a warning but does not affect the actual error so I don't think it's necessary

ThanatosGit:
Yeah, the matplotlib import error needs to be fixed first before using the merge queue.

PR #7047, opened 2024-08-05
Add seperate teleporter rando field for DB

Miepee:
This adds a seperate `teleporter_rando` field into the DB, that allows to group dock types together. 
This is in preperation for AM2R's #6872 and Dread's future Teleportal rando, where both games need teleporters shuffled in different pools.


henriquegemignani (changes requested):
If you're going to change this, can you please rename a bunch of things to be clearer, possibly with comments?

`DockWeaknessDatabase` is definitely more than just about _weakness_.
`dock_rando_config` is more about _lock_ rando. Plus I kinda dislike adding "rando" to name of things when possible.

Also lastly, I just dislike the `all_teleporter_dock_types` method period. Every single usage feels like we're missing something more appropriate.

Miepee:
I'm fine with renaming things. Doesnt sound *too* hard to come up with better names.

Unsure about that last method, what exactly is bugging you about it? Haven't looked much at the code of where it's used, but would it suffice to instead have it as "get dock types for teleporter group"? That's at least what am2r/dread will eventually use, and most of the time (if at all?) there isnt really a case to get *all* types from *all* groups.

duncathan:
we can't seem to decide whether teleporter randomization is game-specific code or generic. if we're going to treat it like it's generic, there are some serious refactors we need to do before we can really justify much more generic work on it imo

henriquegemignani:
For me "Teleporter" is not a term that should show up in any generic code, as it's game specific. `all_teleporter_dock_types` exists solely so the game-specific `PresetTeleporterTab` can know which dock type is the one that is a teleporter. For me since it's a game specific tab it just knows the correct dock type and could get it by name or something. It's even weird it returns a list, because prime 1/2/dread only has one type. Even Dread, if it were to do teleporter rando it'd be completely different.

Big issue is that we do have a big number of games that uses teleporter rando and the configuration window is rather complex. But they're all _very_ different!

Miepee:
5/6 stable games have some sort of teleporter randomizer, most of them also sharing very similar (if not the same) settings/code. The things that stand out to me that the games have exclusive being
- Prime 1 (and echoes?) having a separate mode with 1way anywhere where there are extra targets
- Echoes having its custom "rotation" mode
- AM2R and Dread having separate pools planned.

I know Fusion also has both separate pools planned (elevator + sector tubes) *and* will also have the same generic two-way/one-way code.


So I feel like we should at least have *some* generic code available. Because very broadly summed down, all games basically have "here's the pool of sources, here's the pool of targets, shuffle them with the modes".



> all_teleporter_dock_types exists solely so the game-specific PresetTeleporterTab can know which dock type is the one that is a teleporter

Why does it need to know the type, does it just go through the whole DB, and create the checkboxes for every dock that is of that type?

ThanatosGit:
> For me "Teleporter" is not a term that should show up in any generic code, as it's game specific. 

Well, I made it all "teleporter" because nobody had a better name. No matter if you call it elevator, portal, shuttle, pipe, teleporter or whatever, all of this is just picking you up at x, make some fancy animation and throws you at y. Totally fine to name it teleporter.

> It's even weird it returns a list, because prime 1/2/dread only has one type. 

No, Dread has two types: "elevator" and "shuttle" and both are in the same pool to randomize.

duncathan:
also for the record this PR isn't really applicable to dread at all. teleportal randomization is going to share little if any code with other forms of transport rando, and this PR will only complicate the other planned feature of optionally randomizing shuttles and elevators separately from each other.

Miepee:
Ah good to know. I thought it'd be similar to the current shuttle/elevator, where you can have the typical two way/one way connections on teleportals too.

duncathan:
no. teleportals will only ever be two-way, and will never allow picking and choosing individual pairs to randomize.

Miepee:
So what I get from this discussion, is to scrap the idea of having the groups defined in the database, and instead define them in the PresetTeleporterTab-thingy instead. That way, each game has its own game-specific way of how they define the pool(s), and the generic part can keep being the "shuffle the pool based on certain mode". And then also have each tab define on which configuration field it gets saved to, rather than a hardcoded "teleporter".
And the spoiler tabs in the game details window can then use the same function to determine which docks to show that the presetTeleporterTab uses.

Does that sound better?

henriquegemignani:
> > For me "Teleporter" is not a term that should show up in any generic code, as it's game specific.
> 
> Well, I made it all "teleporter" because nobody had a better name. No matter if you call it elevator, portal, shuttle, pipe, teleporter or whatever, all of this is just picking you up at x, make some fancy animation and throws you at y. Totally fine to name it teleporter.

Teleporter is fine to show up as a common term, just not in classes that are intended to be "for every game". But this does mean that not every game needs to use the term. Probably what we need is better tooling to let each game do their own thing better.

ThanatosGit:
> [...] is to scrap the idea of having the groups defined in the database, and instead define them in the PresetTeleporterTab-thingy instead.

Even after discussion, I still don't see the reason. If you move it out of the database and move it into the `PresetTeleporterTab`, it won't work because there are multiple parts in the code which needs to access it. E.g. the PDFs because all this stuff lands in `all_dock_connections` or the common `GameDescription` class...unless you also want to rewrite all that.

> And the spoiler tabs in the game details window can then use the same function

Well, then it's not defined in the `PresetTeleporterTab` but in a common function which needs to be overriden by all specific games, which was already the conclusion from the first part for me. And that is the part where I don't see any advantage of defininig it in code instead of in the db. In fact, I would omit defining arbitrary names from the db in the code as much as possible because the names can be changed. (Even I never have seen someone renaming a teleporter type). Defining it in the db can be checked by the integrity but if you don't know the names are used in code, you just break it.

Miepee:
Urgh, I really should've looked at all the places where it's actually used before thinking about this.

So lemme try to clean this up here, cause otherwise it gets messy in my head.
Requirements:
- Games need a way to define different pools and their associated dock types
- The pool definition needs to be easily accessible as its used by quite a lot of things 
- Games should have a way of modifying the pools based on certain options



Which makes me think of a few implementation options:
- Define the pools in the DB, provide a function to get the types for a pool. Give each game the ability to overwrite the function
- Same as above, but define the pools in code
- Don't provide a way to overwrite the "get pool" function. (Sounds bad and reads to me like the above but with unnecessary steps if a game has the need to modify it)


Out of these options, the first one seems the nicest to me, but am open to other thoughts or corrections it I missed anything important.

ThanatosGit:
I prefer the first option but after reading the discussion again, I also need to admit that I don't understand the issues brought up here.
Dread's (non existing) `Teleportal` randomization sounds to me like all other ones. It only has fewer modes but we already have control over it and other games already doing so. You make a new class, let it inherit from `PresetTeleporterTab`, define `teleporter_mode_to_description` for the available modes and done.
The issues I see is the patcher (which is totally out of scope here + changes in the pdf) and the pools / groups of multiple teleporter randomization thing, which is exactly what this PR deals with.

Also this could make Echoes portal randomization much cleaner because it could utilize the same things, no?

The only concern I have is about (a totally to me unknown feature) `optionally randomizing shuttles and elevators separately from each other` because that sounds like a mess if we define the groups in the db.
Unless we are able to define 3 groups (1. elevator+shuttle, 2. shuttle, 3. elevator) and then just ensure that only group 1 can be shuffled if 2 and 3 are set to vanilla.

Ultimately, to me our teleporter stuff is generic and not game specific. You can add a `TeleporterConfiguration` to your game and only need to deal with very few things (where I can't see how we could eliminate that) and if you don't want / need such a thing, just don't add it to your child class from `BaseConfiguration` and you are ready to go.

Miepee:
> Dread's (non existing) Teleportal randomization sounds to me like all other ones.

You'd also need to overwrite the "display checkboxes" function so that it never displays them at all. Since, according to duncathan:
> teleportals will only ever be two-way, and will never allow picking and choosing individual pairs to randomize.

So either all teleportals are shuffled two-way, or they're vanilla. But we do have that control already, so that's easy.


---

> Unless we are able to define 3 groups (1. elevator+shuttle, 2. shuttle, 3. elevator) and then just ensure that only group 1 can be shuffled if 2 and 3 are set to vanilla.

That sounds like option 3 to me. Reason why I thought that it felt messy, is if you have a game that has more than 2 types for one pool that should get dynamically changed, e.g. `"default": ["foo", "bar", "beep"]`. Suddenly you're at 7 pools you'd need to define in the db. As opposed to just writing something like
```py
types = get_types_for_pool("group1")
...
def get_types_for_pool(pool_name):
  if pool_name in ["group1", "group2", "group3"]:
    foo = super.get_types_for_pool("foo")
    bar = super.get_types_for_pool("bar")
    beep = super.get_types_for_pool("beep")
    group1 = [foo, bar, beep]
    group2 = group3 = []
    if self.configuration.seperate_pool_for_foo:
      group1.remove(foo)
      group2.append(foo)
    if self.configuration.seperate_pool_for_beep:
      group1.remove(beep)
      group3.append(beep)
      
    if pool_name == "group1":
      return group1
    if pool_name == "group2":
      return group2
    else
     return group3
     
  else:
    return super.get_types_for_pool(pool_name)
```

henriquegemignani:
> Ultimately, to me our teleporter stuff is generic and not game specific. You can add a `TeleporterConfiguration` to your game and only need to deal with very few things (where I can't see how we could eliminate that) and if you don't want / need such a thing, just don't add it to your child class from `BaseConfiguration` and you are ready to go.

[`teleporters.py`](https://github.com/randovania/randovania/blob/800b3cab219646571d8220a51b7a7fa410444684/randovania/layout/lib/teleporters.py) has three different things specific to Prime 1 and 2, and that's not including the `static_teleporters` field that is just because these games use it for something else.

There's 3 kinds of components:
- Base, that applies to every single game
- Game specific
- Library, that is used by game specific

Teleporter code definitely fits into the library kind. But the current code is not that ready to be that imo.

ThanatosGit:
> There's 3 kinds of components:
> 
> * Base, that applies to every single game
> * Game specific
> * Library, that is used by game specific
> 
> Teleporter code definitely fits into the library kind. But the current code is not that ready to be that imo.

Yeah, the prime stuff being in there for "historical reasons" is bad.
I would say it's not ready because it definitely does not pass a certain quality gate because of that.
Don't nail me down on it but from my memory the way it is implemented in Dread, MSR and AM2R looks like it is at least "readyish".

Miepee:
Closing for now.
Not sure what'll be done with dread teleportal rando, but am2r's region rando will either wait on ER, or have some code not trying to utilize this system.

Miepee:
Closing for now.
Not sure what'll be done with dread teleportal rando, but am2r's region rando will either wait on ER, or have some code not trying to utilize this system.

PR #7048, opened 2024-08-05
AM2R: DB changes for future region rando

Miepee:
Split up from #6872


JeffGainsNGames (approved):
Split looks okay, will need play testing to really know though.

PR #7057, opened 2024-08-07
[DREAD] Freesink game modification

steven11sjf:
- added a patch to Game Modification -> Other -> Miscellaneous that disables the SubAreaManager.bKillPlayerOutsideScenario tunable
- added a note explaining freesink (or at least attempting to lol)
- added a preset migration to default to disabled

Help appreciated on the wording, i spent about 30 minutes and still don't think its very good lmao. Also, I feel like the "experimental" label isn't particularly accurate since it is well-tested to do what it says on the box lol. I'd be happy to add that requirement if that's preferred.

### DIFF
--- a/randovania/games/dread/gui/ui_files/preset_dread_chaos.ui
+++ b/randovania/games/dread/gui/ui_files/preset_dread_chaos.ui
@@ -1,0 +1,144 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>PresetDreadChaos</class>
+ <widget class="QMainWindow" name="PresetDreadChaos">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>503</width>
+    <height>387</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Other</string>
+  </property>
+  <widget class="QWidget" name="centralWidget">
+   <property name="maximumSize">
+    <size>
+     <width>16777215</width>
+     <height>16777215</height>
+    </size>
+   </property>
+   <layout class="QVBoxLayout" name="verticalLayout">
+    <property name="leftMargin">
+     <number>0</number>
+    </property>
+    <property name="topMargin">
+     <number>0</number>
+    </property>
+    <property name="rightMargin">
+     <number>0</number>
+    </property>
+    <property name="bottomMargin">
+     <number>0</number>
+    </property>
+    <item>
+     <widget class="QScrollArea" name="scroll_area">
+      <property name="sizePolicy">
+       <sizepolicy hsizetype="Expanding" vsizetype="Expanding">
+        <horstretch>0</horstretch>
+        <verstretch>0</verstretch>
+       </sizepolicy>
+      </property>
+      <property name="widgetResizable">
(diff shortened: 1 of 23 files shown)


dyceron:
Add screenshots of the UI changes

steven11sjf:
image

Miepee:
- avoid colored text
- move this into a chaos setting tab to make it less likely of being accidentally discovered/enabled. Then you also dont need your big disclaimer cause the tab name (and mini tab description) would already do that.
- probably avoid mentioning grappling without magnet? I could see people going "oh cool finally i dont die because of that anymore, lemme enable it" and then accidentally getting oob.
- had somehing to say for the last sentence but now i forgot. Will post when i get reminded.

ArcanoxDragon:
I feel like we should probably not be including the tunable in the patcher file if the setting is not **enabled** in the preset. I know that ODR at least _used_ to skip exporting a `config.ini` file if it doesn't need to, which allows for things like installing the actual Freesink mod alongside an RDV seed. If we could maintain that compatibility it'd be great.

It would also avoid having to change all of the patcher.json expected output files for the unit tests 😉

steven11sjf:
afaik `config.ini` is always touched. could just be because i always have cosmetics on though. the json contains all the config options, even the ones i have disabled.

steven11sjf:
updated ui screenshot

image

steven11sjf:
should I change floor clips to `speed AND flash shift AND (grapple OR freesink)`? or should chaos settings *never* have logical implications?

Miepee:
Chaos settings *never* have logical implications.
If you want it to be logical, then it's not a chaos setting.

henriquegemignani:
From what I understand, wouldn't you be able to start clipping almost anywhere given that you're already able to clip in most places? This seems more like a min logic or just no-logic kind of rando

steven11sjf:
Not quite. You are still required to clip through the wall, it just doesn't kill you. 

Known way to do that are
- shinesink/adc clips
- "phantom slip" (idk much about this one but it requires cloak and bomuhey skew, so a lot of requirements and usually a setup for specific locations)
- grappling walls sometimes clips you through without spider (it's a pretty awkward angle to get and is usually an accident, lol)

the behavior of OOB is basically the same as SR. you can go back into collision but require tricks to get out. it's just a lot harder to clip (requires items and advanced tech). if you have space jump and are near the edge of the scenario you can use it to break logic pretty hard but that's basically it, since it requires somewhere to store a spark and perform the tech with, and you will probably just pop out of whatever's below you. 

and as Miepee said, it wouldn't affect logic so not sure what you mean by it being min logic.

randovania/games/dread/gui/ui_files/preset_dread_chaos.ui
```diff
@@ -0,0 +1,144 @@
...
+         <number>0</number>
+        </property>
+        <property name="topMargin">
+         <number>2</number>
+        </property>
+        <property name="rightMargin">
+         <number>0</number>
+        </property>
+        <property name="bottomMargin">
+         <number>0</number>
```
Miepee, line 67: please make all of these a margin of 6

steven11sjf:
image
new ui after changes

Miepee (approved):

PR #7064, opened 2024-08-09
Dread: Revise connection to Missile Tank in Space Jump Room

Tyranisaur:
Remove possibility of using Speed Booster with Normal Bombs.

### DIFF
--- a/randovania/games/dread/logic_database/Ferenia.txt
+++ b/randovania/games/dread/logic_database/Ferenia.txt
@@ -656,21 +656,22 @@ Extra - asset_id: collision_camera_012
   > Door to Navigation Station
       Trivial
   > Pickup (Missile Tank)
-      All of the following:
-          Any of the following:
-              # Destroy the blocks
-              Lay Cross Bomb or Lay Power Bomb
-              All of the following:
-                  Lay Small Bomb
-                  Any of the following:
-                      Space Jump
-                      Movement (Beginner) and Use Spin Boost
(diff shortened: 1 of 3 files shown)


randovania/games/dread/logic_database/Ferenia.txt
```diff
@@ -656,21 +656,22 @@ Extra - asset_id: collision_camera_012
...
-                      # https://www.youtube.com/watch?v=AaSlxp8GajA
-                      Movement (Advanced)
-          Any of the following:
-              # Grab the item
-              Space Jump or Speed Booster or Movement (Advanced)
-              Movement (Intermediate) and Use Spin Boost
+      Any of the following:
+          All of the following:
+              # Destroy the blocks with Cross Bomb or Power Bomb
+              Lay Any Bomb or Lay Power Bomb
```
ArcanoxDragon, line 662: Shouldn't this be "Lay Cross Bomb" instead of "Lay Any Bomb" according to the comment? Right now, it seems like this new logic still allows getting the item with just small bombs and Speed Booster.

ArcanoxDragon (changes requested):

randovania/games/dread/logic_database/Ferenia.txt
Tyranisaur: Yes, forgot to change this when I copied everything around.

XenoWars (approved):

PR #7070, opened 2024-08-11
Put progress bar into a status bar.

Miepee:
This puts the progress bar into a statusbar, and only makes it appear when its relevant. E.g. invisible until its invoked for the first time.
This makes it take less space and look nicer than as a groupbox.

grafik
grafik
grafik
grafik


henriquegemignani:
The way the space is still reserved for the status bar before it's show is very weird to me.

Miepee:
AFAIK That is normal status bar behaviour.

Miepee:
For further elaboration on it: to my knowledge that is definitely standard.
Changing a UI layout dynamically can feel very much like the user's being pulled under the rug. This can also include changing the visibility from UI elements, which is why I'm not hiding the progress bar / stop button after they're done being useful.

Thus having the layout be constant, and only making the elements appear once they start being useful is the best compromise there for me. It's less jarring to the user than having the whole status bar magically pop up, and still avoids information overload.

Shaleblade:
I like it - I think the screen space when not used makes sense as padding, and when used highlights the importance of the progress bar

Uncle-Reggie:
This makes the UI far more attractive without a complete overhaul, imo.

Miepee:
> Maybe some sort of placeholder thing, such that it is not a total empty space? Or different color

No to both. At that point i'd rather just always have the progress bar visible.
IDK how common the sentiment is that it looks weird, but I'd rather have it merged first, get more feedback and then possibly edit after, than spend weeks discussing whether the statusbar should or should not have something in there initially.

Miepee:
I'm fine discussing it, I just don't want the PR to stall due to none of us being GUI designers and having no user feedback.

henriquegemignani (approved):

PR #7073, opened 2024-08-11
Dread/emmi zone spinner logic

Tyranisaur:


randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -1680,6 +1687,16 @@ Extra - asset_id: collision_camera_015
...
+  > Door to White EMMI Arena (Top)
+      Any of the following:
+          # https://youtu.be/RUvZaIl5vts
+          Diagonal Bomb Jump (Advanced) and Lay Normal Bomb
+          # https://youtu.be/otd7WX_dU_s
+          Speed Booster and Infinite Bomb Jump (Intermediate) and Speed Booster Conservation (Advanced) and Lay Normal Bomb
+          # https://youtu.be/OFHELQQbO4w
+          Speed Booster and Infinite Bomb Jump (Beginner) and Speed Booster Conservation (Intermediate) and Lay Cross Bomb
+          # https://youtu.be/vlhzedSfhEA
+          Morph Ball and Speed Booster and Speed Booster Conservation (Beginner) and Wall Jump (Beginner) and Use Spin Boost
```
XenoWars, line 1699: This shouldn't need the wall jump trick. I'm sure it could be done from the bottom door too and not need Morph.

randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -1581,7 +1581,6 @@ Extra - asset_id: collision_camera_015
       Any of the following:
           Space Jump
           Climb Sloped Surfaces (Beginner) and Use Spin Boost
-          Speed Booster and Speed Booster Upgrade and Speed Booster Conservation (Beginner) and Disabled Door Lock Randomizer
```
XenoWars, line 1584: This could be re-added but with adding Spin Boost, and then putting Speed Booster Upgrade and Disabled Door Lock Randomizer into an Or clause instead.

CHANGELOG.md
```diff
@@ -13,7 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ##### Artaria
 
+- Added: Various individual means of using Speed Booster to climb EMMI Zone Spinner from the Tunnel to White EMMI Area, using Speed Booster Conservation, and a choice between Normal Bombs, Cross Bombs and Spin Boost.
 - Changed: The door that is locked by both Super Missile and Charge Beam is now using lock overrides. This ensures that the requirements of these locks are both respected by logic when randomizing Doors and this door is unchanged.
+- Fixed: Logic now accounts for the need to escape through the doors after pulling open the EMMI Zone Spinner with Grapple Beam.
+- Removed: Using Speed Booster to climb the EMMI Zone Spinner Room from the Door to White EMMI Arena (Power).
```
XenoWars, line 19: To go with the other comments, this could be converted to a Changed entry where it adds Spin Boost, and needs either a Speed Booster Upgrade or Disabled Door Lock Randomizer instead of both.

XenoWars (changes requested):

randovania/games/dread/logic_database/Artaria.txt
XenoWars: Or you can use upgrades.

Tyranisaur: Also, we already have a Wall Jump requirement on jumping up to the Power door from here, so it seems consistent to also have it on this one. Though, it might make even more sense to add it to the bomb jump versions, unless we consider these to be feasible without a wall jump?

Tyranisaur:
I updated the changelog entries so this should be viable again. Can we not take 2 weeks to continue reviewing this again? I haven't done anything to address the comments so far because it's not very clear cut what tricks should be or not be on some of these moves considering consistency with what we already have. I'm also not personally invested in adding logic for the speed booster upgrades, I'm not familiar with how many it takes to charge speed for any particular distance. Also, duncathan has expressed some doubts about whether it was a good idea to have logic enabled by the upgrades in the first place. So all in all, I would prefer to consider that out of scope.

XenoWars (approved):
If Dunc has concerns with upgrade logic, then this is fine now. I do agree now that if we already have a wall jump for a spin boost method in place, it makes sense to keep it for other connections.

I can't resolve conversations, or I would.

Miepee:
@Tyranisaur can you fix the changelog conflict?

Tyranisaur:
> @Tyranisaur can you fix the changelog conflict?

Done

PR #7079, opened 2024-08-12
Redesign the customize preset window

Miepee:
This redesigns the customize preset window to remove the nested tabs, and instead use a more familiar list-view-on-the-left-and-content-on-the-right design (surely designers have come up with a name for that).
PR 1 of a few more UI redesigns PRs I'm planning while talking with a few designers.

grafik

It automatically hides headers if they're not relevant:
grafik

And also doesn't show experimental tabs (if experimental is disabled):
grafik
grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [8.6.x] - 2024-11-??
 
+- Changed: The "Customize Preset" window has been given a visual overhaul.
 - Changed: Filtering in the Generation Order tab is now case-insensitive.
 - Changed: The Area View of the Data Visualizer now always has a dark gray background to help with readability.
 
--- a/randovania/games/am2r/gui/preset_settings/am2r_chaos_tab.py
+++ b/randovania/games/am2r/gui/preset_settings/am2r_chaos_tab.py
@@ -52,8 +52,8 @@ class PresetAM2RChaos(PresetTab, Ui_PresetAM2RChaos):
         return "Chaos Settings"
 
     @classmethod
-    def uses_patches_tab(cls) -> bool:
-        return True
+    def starts_new_header(cls) -> bool:
+        return False
 
     def on_preset_changed(self, preset: Preset) -> None:
         assert isinstance(preset.configuration, AM2RConfiguration)
--- a/randovania/games/am2r/gui/preset_settings/am2r_chaos_tab.py
+++ b/randovania/games/am2r/gui/preset_settings/am2r_chaos_tab.py
@@ -52,8 +52,8 @@ class PresetAM2RChaos(PresetTab, Ui_PresetAM2RChaos):
         return "Chaos Settings"
 
     @classmethod
-    def starts_new_header(cls) -> bool:
-        return False
+    def header_name(cls) -> str | None:
+        return None
 
     def on_preset_changed(self, preset: Preset) -> None:
         assert isinstance(preset.configuration, AM2RConfiguration)
--- a/randovania/games/am2r/gui/preset_settings/am2r_gameplay_tab.py
+++ b/randovania/games/am2r/gui/preset_settings/am2r_gameplay_tab.py
@@ -44,8 +44,8 @@ class PresetAM2RGameplay(PresetTab, Ui_PresetAM2RGameplay):
         return "Gameplay Changes"
 
     @classmethod
-    def starts_new_header(cls) -> bool:
-        return True
+    def header_name(cls) -> str | None:
+        return cls.GAME_MODIFICATIONS_HEADER
 
     def on_preset_changed(self, preset: Preset):
         config = preset.configuration
--- a/randovania/games/am2r/gui/preset_settings/am2r_gameplay_tab.py
+++ b/randovania/games/am2r/gui/preset_settings/am2r_gameplay_tab.py
@@ -44,7 +44,7 @@ class PresetAM2RGameplay(PresetTab, Ui_PresetAM2RGameplay):
         return "Gameplay Changes"
 
     @classmethod
-    def uses_patches_tab(cls) -> bool:
+    def starts_new_header(cls) -> bool:
         return True
 
     def on_preset_changed(self, preset: Preset):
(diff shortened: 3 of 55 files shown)


Shaleblade:
This is such a nice change - the current version has always been a little bit of a pain to navigate, and I think sacrificing a bit of horizontal space is well worth it. I'm guessing it takes up the same amount of space as the window is resized?

Miepee:
> I think sacrificing a bit of horizontal space is well worth it

I am fairly certain that the majority of randovania users have a setup where they have double the horizontal space than vertical. So yeah, think so too. I also made the window slightly smaller in height for comparison for the default size to try to slightly balance it out.


> I'm guessing it takes up the same amount of space as the window is resized?

unsure what you're referring to with "it". The left side? If that's the case yes.
Maximized:
grafik
Smaller than default:
grafik

Uncle-Reggie:
I like this side panel a lot. Having the pieces on the side rather than in nested tabs along the top is a huge improvement.

henriquegemignani:
I don't like the bold+underline for the separators (Randomizer Logic/Game Modifications).

Though I think we could just get rid of these. I always hated how item pool is very much a game modification but fits it's logic.

JeffGainsNGames:
> Though I think we could just get rid of these. I always hated how item pool is very much a game modification but fits it's logic.

Is there a better grouping/naming convention that could be made? I feel having no separators/groups would look disorganized.

Miepee:
I think what might be looking off more is the fact that it's so big?
Looking at no-headings more, it doesn't really look bad to me. But the headings do slightly help with making the preset description entry stand out.

Fwiw I was mostly looking at plasma-settings, which has bold+slightly bigger + a line to the side (which I couldn't easily replicate here)
grafik

macos seems to use spacing for groupings instead:
grafik

gnome-settings uses a line seperator:
grafik


---

But here are more screenshot concepts. I personally prefer 2 and then 4.

1. No headings:
grafik

2. Smaller heading size:
grafik

3. No underline:
grafik

4. No bold:
grafik

5.  quick test with macos imitation:
grafik

6. quick test with gnome imitation:
grafik

Miepee:
9. Also I got recommended to do a combination of header entry + spacing, so there's that too
grafik

henriquegemignani:
I like either the macos style or the gnome style. We can also let games use multiple groups, not just two.

This is also needs some reordering and maybe renaming of titles. A tab named "Other" should be last I believe. 

Also it feels _very_ weird to have preset description last. Maybe first, but definitely not the one that is visible by default when you open.

JeffGainsNGames:
> Also it feels _very_ weird to have preset description last. Maybe first, but definitely not the one that is visible by default when you open.

I had this same discussion point too

JeffGainsNGames:
I like the gnome and not macos, the fully blank space doesnt sit well with me.

Uncle-Reggie:
The Gnome separators are very appealing. Are these mockups or are they actually being written out? The idea that RDV could do a platform check and display spacers if on macos or lines if running Linux or Windows would be a kindness, I think. (MS's own design language is inconsistent, but Win11's file explorer uses line separators between categories)

Miepee:
> We can also let games use multiple groups, not just two.

That was my thought too, I made a comment about that in the patch. Although "arbitrary groups per game" sounds nicer than "heres a set of global groups, decide what you want to be classified under"

> Also it feels very weird to have preset description last

The preset description is really weird, and I have no good idea on how to best classify it as.
Our average user, *does* use it last. They usually customize a preset first, and only when they're done, give it (maybe) a description. That's the reason why it is currently located at the very bottom, since it fits with our average user behaviour.
I will actively fight against having a first entry that is not selected by default, that's just bad UX. So either we have the preset description be the first entry, which would mean it is selected by default, or we move it somewhere else. 

I also thought about the possibility of moving the preset description functionality to somewhere outside the customize preset dialog, but it felt bad to move a specific part of the preset editing out of there.
Another thought that came to mind, was also moving the name out of the top into the same view where the preset description is edited in. But then the name isn't always accessible anymore and it felt bad to me too.
There was also the recommendation to have a description text box underneath the list, but it felt kinda messy to me:
grafik

Miepee:
>  Are these mockups or are they actually being written out?

Unsure what you mean here. I quickly hacked bad code together on this branch, ran it, and made a screenshot.

> The idea that RDV could do a platform check and display spacers if on macos or lines if running Linux or Windows would be a kindness

I don't think it would. Randovania is doing its very own themeing and design language. That sounds like behaviour that'd be more appropriate for #5845 . As long as we're forcing our own, I don't feel like we should cater to the OS design.
We also only have ~5 macos users, and I don't think extra code complexity for them is worth the effort at all.

Miepee:
> This is also needs some reordering and maybe renaming of titles.

Definitely, but I would like to do that in a seperate PR. It feels easier if this is just "this is the design change", then we can test it out and have some more user feedback, and then change more tabs to accommodate said feedback. There are a few issues that I noticed here that also affect the current design, e.g. #7081, #7080 or the aforementioned "Other" tab in am2r.

Uncle-Reggie:
> Unsure what you mean here. I quickly hacked bad code together on this branch, ran it, and made a screenshot.

You'd been doing image-based mockups for a couple things recently, was curious if that was also being done here, as a precursor to the idea that RDV could respect OS-specific conventions.

> Definitely, but I would like to do that in a seperate PR. It feels easier if this is just "this is the design change", then we can test it out and have some more user feedback, and then change more tabs to accommodate said feedback. There are a few issues that I noticed here that also affect the current design, e.g. #7081, #7080 or the aforementioned "Other" tab in am2r.

I agree with your thinking here.

Nystrata:
I like the sidebar style, since it's a lot easier to navigate than a dozen tabs

I also prefer macOS's empty space separators

Miepee:
> The double hr in the main part is weird. Is this already that way in the current state?

Yes, it is.

duncathan:
> 9. Also I got recommended to do a combination of header entry + spacing, so there's that too
>    grafik

I like this one for the most part. could we possibly center align and/or underline the headings on this one?

henriquegemignani (approved):
Changelog?

Miepee:
Marking as draft because I realized this needs quite a few changes still that I have forgotten to push.

Miepee:
Adjusted this to look like 9.

henriquegemignani (approved):

PR #7093, opened 2024-08-14
MW: hide generate without retries for non-preview

Miepee:
Unsure if i should mention in the changelog whether its locked behind preview, or just straight up say its removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [8.4.x] - 2024-08-??
 
 - Fixed: The "Last Activity" text on the Multiplayer Session window is not aligned properly.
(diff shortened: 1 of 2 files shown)


duncathan:
:-1: this is an important preset debugging tool and is invaluable for large multis. what's the point of removing it?

Miepee:
You are also a core dev and have more knowledge about the system than our average users.
So far in my personal experience, I have not heard of someone using the button intentionally. It was always "spam generation a few times, if it fails after  ~10 tries, go through preset settings and look out for potential bad options, rinse and repeat"
But my personal experience could of course be very biased.

henriquegemignani:
As I said in discord, we definitely could use some guides on how to better investigate bad presets. And this button is an excellent way of doing that, so it'd be awkward to remove it.

Definitely could use renaming, moving or whatever but it'd be bad to see it gone. Could be a context menu option when you right click a preset "investigate" or something

Miepee:
Closed in favour of #7095

PR #7124, opened 2024-08-25
Fusion: a few main deck changes

Miepee:
Reactor core now has a platform to allow non-dangerous-nettori traversal. Image in the DB is getting replaced soon:tm:
grafik

Added a new misc resource for when open transition hatches are not shuffled in DLR (currently thats not implemented, and while the plan is to implement them at one point, no clue when that will be)
Rest should be self explanantory.


JeffGainsNGames (changes requested):
Can you update the room image with the new changes.

Miepee:
I dont have a new room image :(

JeffGainsNGames (approved):
Remember to get new image.

PR #7136, opened 2024-08-27
AM2R: address preset feedback

Miepee:
grafik


JeffGainsNGames (approved):
I was thinking in my mind moving room design left, but I think this makes more sense to me now so I'm happy with this.

Miepee:
People are more likely to enable random start or enable dlr, hence why i moved the gameplay stuff to the right.

PR #7142, opened 2024-08-28
fusion: Add sector 3 trick documentation

cluekitty:
Adds trick documentation for sector 3.
Updates image for Checkpoint Crossing to show that enemies are in the room.


randovania/games/fusion/logic_database/Sector 3 PYR.txt
```diff
@@ -1040,7 +1042,12 @@ Extra - room_id: [28]
   > Pickup (Energy Tank)
       All of the following:
           Speed Booster
-          Space Jump or Can Activate Pillar
+          Any of the following:
+              Space Jump or Can Activate Pillar
+              # Wall Jump and Shinespark straight up - https://youtu.be/Pvv2dYRZAj4
+              Wall Jump (Intermediate)
+              # Wall Jump and ignore the pillar - https://youtu.be/FJxCd30N2J4
+              Wall Jump (Advanced)
```
cluekitty, line 1050: I seem to recall that including tricks in this way will cause the randomizer to always pick one over the other. Feedback on how I could improve this is appreciated. If necessary we can remove one trick or the other.

JeffGainsNGames: Yeah it'll always accept the lowest requirements. That being said I don't mind the "dual" documentation incase someone struggles maintaining the shinespark and just wants to beat their head against that other jump.

cluekitty: Would we leave the logic in place for both or remove one of the trick requirements and just have 2 comments in that case?

Miepee: Just leave both requirements. I'd rather have more of the game documented than too little.

randovania/games/fusion/logic_database/Sector 3 PYR.txt
```diff
@@ -1040,7 +1042,12 @@ Extra - room_id: [28]
   > Pickup (Energy Tank)
       All of the following:
           Speed Booster
-          Space Jump or Can Activate Pillar
+          Any of the following:
+              Space Jump or Can Activate Pillar
+              # Wall Jump and Shinespark straight up - https://youtu.be/Pvv2dYRZAj4
+              Wall Jump (Intermediate)
```
Miepee, line 1048: With 
1) this shinespark (probably) not being the intended path
2) the timer being so strict that you don't have a lot of leeway on the shinespark from the ground
3) the timer being strict-ish that you dont have too much leeway if you shinespark from the air

this should be an advanced shinespark

randovania/games/fusion/logic_database/Sector 3 PYR.txt
Miepee, line 1050: I wonder if this should be an expert walljump instead?

Miepee (commented):
can you please add new bob skip too?

Miepee:
Also since you brought it up, for this SWJ too :)
grafik

randovania/games/fusion/logic_database/Sector 3 PYR.txt
cluekitty: I had only set this as "advanced" because there's not really any negative consequence of failing it as you can retry pretty much instantly. The only negative consequence you would really have is spending too much time on a trick. I can still bump it up if you think expert better suits this trick.

Miepee: how precise do you have to be for the trick? cause it looked pretty precise for me.

Miepee: yeah after trying for a bit, it definitely feels like expert based on the other tricks we have in advanced.

cluekitty: I suppose it is fairly precise. I'll bump it up. If anything, having the other trick be advanced means this one probably should be bumped up.

Miepee (approved):

PR #7145, opened 2024-08-28
Fusion: more heat run data for most of s3

Miepee:
Draft until i figure out wtf to do with lava maze.


Miepee:
Note to self: remove the other requiement for the dumb hypermode sova processing jump

SchwartzGandhi (approved):

PR #7147, opened 2024-08-29
Restore compatibility with Python 3.11

henriquegemignani:


randovania/gui/main_window.py
```diff
@@ -753,7 +753,7 @@ def _on_menu_action_dark_mode(self):
 
     def _on_menu_action_show_multiworld_banner(self) -> None:
         banner_val = self.menu_action_show_multiworld_banner.isChecked()
-        monitoring.metrics.incr(f"gui_multiworld_banner_option_{"checked" if banner_val else "unchecked"}")
+        monitoring.metrics.incr("gui_multiworld_banner_option_" + ("checked" if banner_val else "unchecked"))
```
Miepee, line 756: does this work too?
```suggestion
        monitoring.metrics.incr(f"gui_multiworld_banner_option_{("checked" if banner_val else "unchecked")}")
```

henriquegemignani: No. Can't have a `"` inside of `f""`s, that's a 3.12 addition.

PR #7154, opened 2024-08-31
Few fixes for Planets

UltiNaruto:
## Fixed internal id for Ridley - Ridley Arena - Pickup(Tourian Key)

The internal id was set to 10085 instead of 100085.

## Fixed Tourian Key not being picked as Tourian Key

The pickup type was set to `Tourian Key #` instead of `Tourian Key`.

## Fixed Tourian Key message not including the number of the key

It would say `Tourian Key acquired` instead of `Tourian Key # acquired`

### DIFF
--- a/randovania/games/planets_zebeth/exporter/patch_data_factory.py
+++ b/randovania/games/planets_zebeth/exporter/patch_data_factory.py
@@ -27,7 +27,7 @@ class PlanetsZebethPatchDataFactory(PatchDataFactory):
     cosmetic_patches: PlanetsZebethCosmeticPatches
     configuration: PlanetsZebethConfiguration
 
-    def _create_pickups_dict(self, pickup_list: list[ExportedPickupDetails], item_info: dict, rng: Random) -> dict:
+    def _create_pickups_dict(self, pickup_list: list[ExportedPickupDetails], _rng: Random) -> dict:
         pickup_map_dict = {}
         for pickup in pickup_list:
             quantity = pickup.conditional_resources[0].resources[0][1] if not pickup.other_player else 0
@@ -50,12 +50,11 @@ class PlanetsZebethPatchDataFactory(PatchDataFactory):
                 else:
                     pickup_type = pickup.original_pickup.name
 
-            acquired_msg = f"{pickup.name} acquired"
+            acquired_msg = pickup.name
             if not self.players_config.is_multiworld:
-                acquired_msg = item_info[pickup.name]["text_header"].replace(pickup_type, pickup.original_pickup.name)
+                acquired_msg = f"{pickup.original_pickup.name} acquired"
 
             pickup_map_dict[object_id] = {
-                "index": pickup.index.index,
                 "model": pickup.model.name,
                 "type": pickup_type,
                 "quantity": quantity,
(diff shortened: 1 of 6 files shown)


randovania/games/planets_zebeth/exporter/patch_data_factory.py
```diff
@@ -43,14 +43,24 @@ def _create_pickups_dict(self, pickup_list: list[ExportedPickupDetails], item_in
...
+            pickup_type = "Nothing"
+            if not pickup.other_player:
+                if pickup.original_pickup.name.startswith("Tourian Key"):
+                    pickup_type = "Tourian Key"
+                else:
+                    pickup_type = pickup.original_pickup.name
+
+            acquired_msg = f"{pickup.name} acquired"
+            if not self.players_config.is_multiworld:
+                acquired_msg = item_info[pickup.name]["text_header"].replace(pickup_type, pickup.original_pickup.name)
```
Miepee, line 55: what is this supposed to do?

Miepee (changes requested):
Fix your tests

randovania/games/planets_zebeth/exporter/patch_data_factory.py
UltiNaruto: It fixes the name of the pickup when it's different than generic pickup
ex : Tourian Key 6 != Tourian Key

Miepee: I'm still not quite following. So your goal is to specifically make this say  `Tourian Key 6 acquired` instead of `Tourian Key acquired`, right?
Cant you change your text data to have `{item}` and then use `.format(item=pickup.original_pickup.name)`?

UltiNaruto: if there's a way to actually clone a dict I wouldn't mind since changing one of the item_data['Tourian Key `x`'] change each of them with the current way it's done

Miepee: why would .format not work for you?

UltiNaruto: because currently it uses item_data to get the acquired text unless I stop using item_data for the header which makes item_data.json only useful for description

randovania/games/planets_zebeth/exporter/patch_data_factory.py
```diff
@@ -43,14 +43,24 @@ def _create_pickups_dict(self, pickup_list: list[ExportedPickupDetails], item_in
...
+                    pickup_type = "Tourian Key"
+                else:
+                    pickup_type = pickup.original_pickup.name
+
+            acquired_msg = pickup.name
+            if not self.players_config.is_multiworld:
+                acquired_msg = f"{pickup.original_pickup.name} acquired"
+
             pickup_map_dict[object_id] = {
+                "index": pickup.index.index,
```
Miepee, line 58: don't rely on the pickup index. if you need it, add the index as extra data

UltiNaruto: the pickup index is either not referenced and null on patcher side then the patcher gets the pickup index from object id
or the pickup index is referenced and used

Miepee: When i said "dont rely on the pickup index" i didn't meant patcher side, I meant rdv side. 
If you need the pickup index on the patcher side, thats fine, but on the rdv side use extra data for it, rather than the index of the pickup node.

UltiNaruto: so do I drop the "index" key value pair?

Miepee: no, you just do something like this instead:
`"index": self.game.region_list.node_from_pickup_index(pickup.index).extra["item_index_how_its_used_in_planets"]`

UltiNaruto: unless you want me to readd "pickup_index" as extra on each pickups. as I said `"index"` is optional since the patcher can retrieve the index from object id

Miepee (approved):

PR #7158, opened 2024-09-02
MSR - Fix inverted sound sliders

dyceron:
Music and Ambient Sound were swapped in the GUI, causing the wrong field to be changed.


Miepee (commented):
i feel like either your explanation of the rpoblem is wrong, or youre not fixing hte problem youre describing.

based on this, the sliders *are* correct, but they're just mislabeled?

dyceron:
If you want to be technical, the name labels were fine, but the groups were in the wrong spot. So, the slider and % label next to music was actually for sfx, and vice versa. So, I just re-arranged them in the GUI so they matched up properly.

Miepee (approved):
Ah, ok

PR #7183, opened 2024-09-06
Fusion: correct pdf for schema, add sound settings

JeffGainsNGames:


Miepee:
needs tests fixed

JeffGainsNGames:
does that mean delete the world_1.json and re-run that one thing again

Miepee:
no need to delete it, just rerun it

Miepee (approved):
For the next PR(s), unexplored map should be relatively easy, can just copy AM2R for most things.
Missile Limit, I'm unsure what the best way to represent them is. We probably only really need 2 (vanilla), 3 (mars), 4 (zm). gotta figure out a nice understandable name for "3"

PR #7184, opened 2024-09-06
Add Stream-friendly variants for Game Art autotrackers

SchwartzGandhi:


JeffGainsNGames:
add screenshots will help reviews

SchwartzGandhi:
Prime 1
image
Echoes
image

duncathan:
are these for anything in particular? I really don't want to clutter the auto tracker layout options even further if we don't need to

SchwartzGandhi:
Currently there is just the pixel variant of the stream friendly trackers, this is just to give people the option to chose between the two.
If anything, I'd say the tracker layouts would be more cluttered without these since there wouldn't be equal formats for both game and pixel art.

duncathan:
> Currently there is just the pixel variant of the stream friendly trackers, this is just to give people the option to chose between the two.

Yes, but why? These layouts were only added to make the CGC restream easier, and afaik it was an intentional decision to only include one style so that restream would look better. Has that decision changed?

> If anything, I'd say the tracker layouts would be more cluttered without these since there wouldn't be equal formats for both game and pixel art.

...what? How does adding an entry to a list that's already way too long address the problem of the list being way too long?

SchwartzGandhi:
> These layouts were only added to make the CGC restream easier
hmm I see, I was not aware of this part. I did bring up this idea to a TO in passing and he thought it was a good idea. But if you would like me to actually ask about this, then I can.
> How does adding an entry to a list that's already way too long address the problem of the list being way too long?
Personally, I don't think the list is too long; or rather, adding this variant won't make the list any more confusing.
I know that doesn't address the core issue at hand but I do have some ideas of decluttering these layouts that I can try to work on. But that's kinda out of scope of this PR (and Miepee told me to not increase the scope of a PR lol)

Miepee:
Fwiw heres the whole discussiom we had on this
https://discord.com/channels/914291389293027329/914294505107251231/1229417307466498101

henriquegemignani:
No. I only want more variants after we change the tracker so _theme_ is separate from _layout_.

duncathan:
with dark and I both disapproving I'm gonna close this

> I only want more variants after we change the tracker so theme is separate from layout.

if TOs specifically require a new layout for tournament use again I'd be ok with adding a new one but that's just about the only exception imo

PR #7186, opened 2024-09-06
Traefik: Change maximum body size to 25MB

Miepee:
Hopefully lets larger rdvgames upload.


henriquegemignani:
Let's keep that to something more reasonable like 25MiB, please.

Miepee:
Oh i thought my 500 player multi was 55mbish, but its only 18.
turning that down then.

henriquegemignani (approved):

PR #7191, opened 2024-09-08
Readme: Mention to uncheck "copy main branch only" on fork

Miepee:
it is really unintuitive that we do not mention this anywhere, and even when it fails we only have a vague 
```
if len(self.randovania_version) != 4:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: object of type 'NoneType' has no len()
```
message.

Theoretically, we could add a mention on how to add upstream and use its tags should one have forgotten to uncheck it, but if one notices it that early they can just delete the fork and refork.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ Integration written by [Thanatos](https://github.com/ThanatosGit).
 ## Setup
 
 Getting started:
-   1. Clone this repository. (downloading the zip is *not* supported and will not work)
+   1. Clone this repository. If you want to clone your fork, make sure that during the forking process you **uncheck** the `Copy main branch only` checkbox.
    2. Open a terminal in the repository root
    3. Run the following file:
       1. Windows: `tools/prepare_virtual_env.bat`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,7 +191,7 @@ Integration written by [Thanatos](https://github.com/ThanatosGit).
 ## Setup
 
 Getting started:
-   1. Clone this repository. (downloading the zip is *not* supported and will not work)
```
henriquegemignani, line 194: Don't remove the zip mention. People press the download zip button way too often.

henriquegemignani (changes requested):
I'd rather fix any errors that come from not having tags, but fixing docs helps in the process.

README.md
Miepee: Huh. Whenever people came into the rdv server for help, I never saw them had issues due to downloading the zip.

henriquegemignani: Hopefully because both the script and the read me very explicitly call that out

henriquegemignani (approved):

PR #7192, opened 2024-09-08
Nicer error message if someone forgot tags during setup

Miepee:
Makes it clear that someone has an invalid dev environment sooner than later.

### DIFF
--- a/randovania/__init__.py
+++ b/randovania/__init__.py
@@ -172,7 +172,9 @@ _final_version = _version.version
 if is_dirty():
     _final_version += "-dirty"
 
+UNKNOWN_VERSION = "0.0.0"
+UNKNOWN_GIT_HASH = b"UNKN"
 
 __version__ = _final_version
-VERSION = _final_version
-GIT_HASH = version_hash.git_hash
+VERSION: str = _final_version or UNKNOWN_VERSION
+GIT_HASH: bytes = version_hash.git_hash or UNKNOWN_GIT_HASH
--- a/randovania/__main__.py
+++ b/randovania/__main__.py
@@ -17,6 +17,12 @@ def main() -> None:
 
     logging.debug("Starting Randovania...")
 
+    if randovania.VERSION == randovania.UNKNOWN_VERSION or randovania.GIT_HASH == randovania.UNKNOWN_GIT_HASH:
+        logging.warning(
+            "Couldn't determine the current version. If you're running from source, "
+            "do you have a git repository and tags present?"
+        )
+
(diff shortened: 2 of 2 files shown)


henriquegemignani (changes requested):
Actually I posted the message in the other pr haha

I'd rather make the situation work, even if certain features won't be perfect. Main thing is wrong version numbers but these don't really matter

Miepee:
I added dummy values. Did a quick test, didn't crash on boot, nor when genning seeds. Dunno if it'd error anywhere else.

randovania/__init__.py
```diff
@@ -172,7 +172,9 @@ def setup_logging(default_level: str, log_to_file: Path | None, quiet: bool = Fa
...
     _final_version += "-dirty"
 
+UNKNOWN_VERSION = "0.0.0"
+UNKNOWN_GIT_HASH = b"UNKN"
 
 __version__ = _final_version
-VERSION = _final_version
-GIT_HASH = version_hash.git_hash
+VERSION: str = _final_version if _final_version is not None else UNKNOWN_VERSION
+GIT_HASH: bytes = version_hash.git_hash if version_hash.git_hash is not None else UNKNOWN_GIT_HASH
```
duncathan, line 180: I believe these can just be `version_hash.git_hash or UNKNOWN_GIT_HASH`? unless you anticipate a situation in which `b""` is a valid hash

randovania/__main__.py
```diff
@@ -17,6 +17,12 @@ def main() -> None:
 
     logging.debug("Starting Randovania...")
 
+    if randovania.VERSION == randovania.UNKNOWN_VERSION or randovania.GIT_HASH == randovania.UNKNOWN_GIT_HASH:
+        logging.info(
```
duncathan, line 21: I'd say warning, since it does indicate that something is wrong with the environment even if we're trying to accomodate the error

duncathan (changes requested):

randovania/__init__.py
Miepee: Sure. Usually prefer to be more explicit, but I don't think we can even ever get `b""` for git_hash here.

henriquegemignani: `b""` is an invalid value for git_hash.

duncathan: we can do the same for the version too, right? I'm pretty sure `""` is an invalid value there as well

Miepee: It is, the version needs to b e `X.Y.Z`

duncathan (approved):

PR #7194, opened 2024-09-08
Extra video documentation for the Corruption DB, as well as a few suggestions

Belokuikuini:
While this PR is mostly about adding new strat vids or replacing the existing ones with (hopefully) more catered towards rando strats, I've also taken the liberty of modifying some existing logic which I will list below : 

# Bryyo - Fire
## Imperial Hall
 - Front of Lasso Walls -> Pickup (Missile Expansion) : 
 Added a new trick to reach the item pickup with a standable ; Beginner Standable Terrain and Movement, with video

# Bryyo - Reptilicus
## Crash Site
 - G.F.S Theseus Lower Ledge -> G.F.S Theseus Upper Ledge - Beginner Slope Jump and Terminal Fall Abuse : 
 Replaced the existing video with a fully dedicated strat vid
## Gateway Hall
 - Door to Gateway -> Bridge - Intermediate Slope Jump : 
 Added strat vid
## Jungle Generator
 - Room Center -> Event - Leviathan Shield : 
 Added either beginner knowledge or After South Jungle Court Turret to existing conditions;
 I don't think it's fair to expect completely new players to know that Jungle Court South can be skipped entirely, or how to destroy the generator after the skip. Also added video on how to do that
## Hidden Court
 - Bottom Floor -> Door to Hidden Court Hall - Beginner Slope Jump :
 Added strat vid
## Ruined Shrine
 - Door to Hidden Court -> Door to Hangar Bay - Space Jump and Morph Ball and Bomb/Spring Space Jump (Intermediate) :
 Added strat vid

# G.F.S Valhalla
## Stairwell
 - Door to Repair Bay -> Door to Weapons Cache - Beginner Slope Jump :
 Added strat vid and removed Screw Attack requirement (new vid doesn't use it)
## Aurora Chamber
 - Door to Aurora Access -> Door to Control Room Access - Beginner Terminal Fall Abuse :
 Added strat vid

I will also drop a few more general suggestions/Comments here : 
 - For quite a few strat vids I recorded, I used scan visor to gain to properties of L-Movement to make it easier and more consistent, especially when it comes to Slope Jumps. While most of the time it isn't necessary, it does make the trick much easier to perform. I'm not sure whether this should be accounted for by the logic? Also considering some rooms might not have ennemies as a substitute for a scan point, Scan Visor might also end up being an additionnal requirement in these rooms. Maybe a template like "Can use Lock-On movement" could be helpful if we decide to account for this factor?
 - Some of the added videos added are there to replace others that were already here, but those that got replaced were usually speedrun strat videos that didn't really take the time to show what exactly is being used (and often were longer than necessary), or showed a faster, but harder way to perform the same trick, which might not reflect the expected trick setting

### DIFF
--- a/randovania/games/prime3/logic_database/Bryyo - Fire.txt
+++ b/randovania/games/prime3/logic_database/Bryyo - Fire.txt
@@ -140,6 +140,8 @@ Extra - asset_id: 4215957264467390882
               Any of the following:
                   After Imperial Hall Grapple Lasso Walls or Bomb/Spring Space Jump (Advanced)
                   Boost Ball and Bomb/Spring Space Jump (Intermediate)
+                  # https://www.youtube.com/watch?v=Z-XICpMJWoE
+                  Before Imperial Hall Grapple Lasso Walls and Movement (Beginner) and Standable Terrain (Beginner)
   > Event - Grapple Lasso Walls
       Grapple Lasso
 
--- a/randovania/games/prime3/logic_database/Bryyo - Fire.txt
+++ b/randovania/games/prime3/logic_database/Bryyo - Fire.txt
@@ -140,8 +140,12 @@ Extra - asset_id: 4215957264467390882
               Any of the following:
                   After Imperial Hall Grapple Lasso Walls or Bomb/Spring Space Jump (Advanced)
                   Boost Ball and Bomb/Spring Space Jump (Intermediate)
-                  # Jump towards the Lasso Platforms https://youtu.be/Z-XICpMJWoE
-                  Movement (Beginner) and Standable Terrain (Beginner)
+                  All of the following:
+                      Standable Terrain (Beginner)
+                      Any of the following:
+                          Movement (Intermediate)
+                          # Jump towards the Lasso Platforms https://youtu.be/Z-XICpMJWoE
+                          Scan Visor and Movement (Beginner)
   > Event - Grapple Lasso Walls
       Grapple Lasso
 
--- a/randovania/games/prime3/logic_database/Bryyo - Fire.txt
+++ b/randovania/games/prime3/logic_database/Bryyo - Fire.txt
@@ -140,8 +140,8 @@ Extra - asset_id: 4215957264467390882
               Any of the following:
                   After Imperial Hall Grapple Lasso Walls or Bomb/Spring Space Jump (Advanced)
                   Boost Ball and Bomb/Spring Space Jump (Intermediate)
-                  # https://youtu.be/Z-XICpMJWoE
-                  Before Imperial Hall Grapple Lasso Walls and Movement (Beginner) and Standable Terrain (Beginner)
+                  # Jump towards the Lasso Platforms https://youtu.be/Z-XICpMJWoE
+                  Movement (Beginner) and Standable Terrain (Beginner)
   > Event - Grapple Lasso Walls
       Grapple Lasso
 
--- a/randovania/games/prime3/logic_database/Bryyo - Fire.txt
+++ b/randovania/games/prime3/logic_database/Bryyo - Fire.txt
@@ -99,7 +99,7 @@ Extra - asset_id: 4215957264467390882
   > Door to Gel Refinery Site
       Any of the following:
           After Imperial Hall Fuel Gel Walls
-          # https://www.youtube.com/watch?v=9FQFcMIH1vQ
+          # https://youtu.be/9FQFcMIH1vQ
           Space Jump Boots and Morph Ball and Bomb/Spring Space Jump (Beginner) and Standable Terrain (Beginner)
   > Event - Fuel Gel Walls
       Charge Beam or Plasma Beam
@@ -140,7 +140,7 @@ Extra - asset_id: 4215957264467390882
               Any of the following:
                   After Imperial Hall Grapple Lasso Walls or Bomb/Spring Space Jump (Advanced)
                   Boost Ball and Bomb/Spring Space Jump (Intermediate)
-                  # https://www.youtube.com/watch?v=Z-XICpMJWoE
+                  # https://youtu.be/Z-XICpMJWoE
                   Before Imperial Hall Grapple Lasso Walls and Movement (Beginner) and Standable Terrain (Beginner)
   > Event - Grapple Lasso Walls
       Grapple Lasso
@@ -201,7 +201,7 @@ Extra - asset_id: 691790544662692104
   * Normal Door to Main Lift/Door to Gel Refinery Site
   * Extra - dock_index: 3
   > Door to Gel Purification Site
-      # https://www.youtube.com/watch?v=nQRQKSyq7vU
+      # https://youtu.be/nQRQKSyq7vU
       Movement (Expert) and Use Screw Attack (Space Jump)
   > Room Center
       Trivial
@@ -237,7 +237,7 @@ Extra - asset_id: 691790544662692104
       All of the following:
           Morph Ball Bomb and Space Jump Boots and Morph Ball
           Any of the following:
-              # https://www.youtube.com/watch?v=ihLjH0TGkIg
+              # https://youtu.be/ihLjH0TGkIg
               Bomb/Spring Space Jump (Expert)
               # https://youtu.be/_sSWzSqM_0g
               Screw Attack and Bomb/Spring Space Jump (Advanced)
--- a/randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
+++ b/randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
@@ -98,7 +98,7 @@ Extra - asset_id: 15814834557214288722
   > G.F.S. Theseus Upper Ledge
       Any of the following:
           Morph Ball Bomb and Morph Ball and After G.F.S. Theseus Satellite
-          # https://www.youtube.com/watch?v=inJ7_pTqlg0
+          # https://youtu.be/inJ7_pTqlg0
           Space Jump Boots and Slope Jump (Beginner) and Terminal Fall Abuse (Beginner)
 
 > G.F.S. Theseus Upper Ledge; Heals? False
@@ -134,9 +134,9 @@ Extra - asset_id: 2768064962148517187
               Any of the following:
                   Space Jump Boots and Bomb/Spring Space Jump (Beginner)
                   Morph Ball Bomb and Bomb/Spring Space Jump (Advanced)
-          # https://www.youtube.com/watch?v=gEu-XdXFx8M
+          # https://youtu.be/gEu-XdXFx8M
           Space Jump Boots and Slope Jump (Intermediate)
-          # https://www.youtube.com/watch?v=5E81JfTtv7k
+          # https://youtu.be/5E81JfTtv7k
           Scan Visor and Movement (Advanced)
 
 > Bridge; Heals? False
@@ -478,7 +478,7 @@ Extra - asset_id: 18411620122976788705
   * Normal Door to Grand Court Path/Door to Grand Court
   * Extra - dock_index: 1
   > Door to Reliquary III
-      # https://www.youtube.com/watch?v=9A-GvQiuUSA
+      # https://youtu.be/9A-GvQiuUSA
       Boost Ball and Space Jump Boots and Morph Ball and Bomb/Spring Space Jump (Advanced)
   > Morph Tunnel Ledge
       Trivial
@@ -637,7 +637,7 @@ Extra - asset_id: 6613311822671481119
   > Event - Leviathan Shield
       All of the following:
           After Jungle Generator Ship Missile Attempt and After North Jungle Court Turret and Use Ship Missile
-          # https://www.youtube.com/watch?v=I5LtCOvlR80
+          # https://youtu.be/I5LtCOvlR80
           After South Jungle Court Turret or Knowledge (Beginner)
   > Event - Ship Missile Attempt
       Use Ship Missile
@@ -745,7 +745,7 @@ Extra - asset_id: 14006603351279167667
           All of the following:
               Space Jump Boots
               Any of the following:
-                  # https://www.youtube.com/watch?v=bFFKq-oSBJg
+                  # https://youtu.be/bFFKq-oSBJg
                   Slope Jump (Beginner)
                   Morph Ball and Bomb/Spring Space Jump (Beginner)
   > Door to Ruined Shrine
@@ -755,7 +755,7 @@ Extra - asset_id: 14006603351279167667
           After Hidden Court Burrow Gate
           Any of the following:
               Space Jump Boots or Grapple Swing
-              # https://www.youtube.com/watch?v=0h-6eP6M7tY
+              # https://youtu.be/0h-6eP6M7tY
               Boost Ball and Morph Ball and Bomb/Spring Space Jump (Advanced)
   > Door to Falls of Fire
       Trivial
@@ -926,7 +926,7 @@ Extra - asset_id: 15861154386655212106
           Morph Ball
           Any of the following:
               Boost Ball
-              # https://www.youtube.com/watch?v=oEPrZMgEPhM
+              # https://youtu.be/oEPrZMgEPhM
               Morph Ball Bomb and Bomb/Spring Space Jump (Expert) and Use Screw Attack (Space Jump)
 
 > Pickup (Missile Expansion); Heals? False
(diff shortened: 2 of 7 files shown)


randovania/games/prime3/logic_database/Bryyo - Fire.txt
```diff
@@ -140,6 +140,8 @@ Extra - asset_id: 4215957264467390882
               Any of the following:
                   After Imperial Hall Grapple Lasso Walls or Bomb/Spring Space Jump (Advanced)
                   Boost Ball and Bomb/Spring Space Jump (Intermediate)
+                  # https://www.youtube.com/watch?v=Z-XICpMJWoE
```
SchwartzGandhi, line 143: for the sake of consistency, we like to have all yt video links truncated so that they look like
`https://youtu.be/[videoid]` rather than `https://youtube.com/watch?v=[videoid]`
Small change I know but would be greatly appreciated.

randovania/games/prime3/logic_database/Bryyo - Fire.txt
```diff
@@ -140,6 +140,8 @@ Extra - asset_id: 4215957264467390882
...
+                  Before Imperial Hall Grapple Lasso Walls and Movement (Beginner) and Standable Terrain (Beginner)
```
SchwartzGandhi, line 144: I would need someone more experienced than me to remind me, but I'm not sure we need the event check here.
From what I've been told, having  `Before (event)` checks are pretty bad for generator health and should be avoided as much as possible.
Think in this scenario you should be fine to just remove it since SJ is already required in this connection

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
```diff
@@ -98,7 +98,7 @@ Extra - asset_id: 15814834557214288722
   > G.F.S. Theseus Upper Ledge
       Any of the following:
           Morph Ball Bomb and Morph Ball and After G.F.S. Theseus Satellite
-          # https://youtu.be/eHJKUTiMgO8&t=75
+          # https://www.youtube.com/watch?v=inJ7_pTqlg0
```
SchwartzGandhi, line 101: truncate

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
```diff
@@ -134,6 +134,7 @@ Extra - asset_id: 2768064962148517187
               Any of the following:
                   Space Jump Boots and Bomb/Spring Space Jump (Beginner)
                   Morph Ball Bomb and Bomb/Spring Space Jump (Advanced)
+          # https://www.youtube.com/watch?v=gEu-XdXFx8M
```
SchwartzGandhi, line 137: truncate

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
```diff
@@ -634,7 +635,10 @@ Extra - asset_id: 6613311822671481119
   > Pickup (Missile Expansion)
       Trivial
   > Event - Leviathan Shield
-      After Jungle Generator Ship Missile Attempt and After North Jungle Court Turret and Use Ship Missile
+      All of the following:
+          After Jungle Generator Ship Missile Attempt and After North Jungle Court Turret and Use Ship Missile
+          # https://www.youtube.com/watch?v=I5LtCOvlR80
```
SchwartzGandhi, line 640: truncate

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
```diff
@@ -741,6 +745,7 @@ Extra - asset_id: 14006603351279167667
           All of the following:
               Space Jump Boots
               Any of the following:
+                  # https://www.youtube.com/watch?v=bFFKq-oSBJg
```
SchwartzGandhi, line 748: truncate

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
```diff
@@ -995,6 +1000,7 @@ Extra - asset_id: 18426339925276965129
               Charge Beam or Missile ≥ 4
               Any of the following:
                   Use Screw Attack (Space Jump)
+                  # https://www.youtube.com/watch?v=LW0nbio3u78
```
SchwartzGandhi, line 1003: truncate

randovania/games/prime3/logic_database/GFS Valhalla.txt
```diff
@@ -166,6 +166,7 @@ Extra - asset_id: 6410596230384865511
           Space Jump Boots
           Any of the following:
               After Metroid Hatcher (Valhalla)
+              # https://www.youtube.com/watch?v=qY6QMWgiXoI
```
SchwartzGandhi, line 169: truncate

randovania/games/prime3/logic_database/GFS Valhalla.txt
```diff
@@ -230,8 +231,8 @@ Extra - asset_id: 7569990525320728426
                       Any of the following:
                           # https://youtu.be/p3LKN-9NaTk?t=5
                           Bomb/Spring Space Jump (Beginner)
-                          # https://youtu.be/6G8jyu4bkJ4
-                          Screw Attack and Slope Jump (Beginner)
+                          # https://www.youtube.com/watch?v=nd6HL1_2GfA
```
SchwartzGandhi, line 234: truncate

randovania/games/prime3/logic_database/Bryyo - Fire.txt
SchwartzGandhi: If you could also go and leave comments describing the trick you perform next to the video links where applicable would also be very helpful, doesn't have to be long.

SchwartzGandhi (changes requested):

randovania/games/prime3/logic_database/Bryyo - Fire.txt
Miepee: Honestly, we really should have this done automatically by rdv.

randovania/games/prime3/logic_database/Bryyo - Fire.txt
Tyranisaur: You should only but in `before whatever` if  having done or collected `whatever` makes your condition harder or impossible.  If it's not essential to have this option here, it might be worth it to forgo having it, just to avoid making `whatever` dangerous. Also worth mentioning, in Dread, we have an option in the generation section for highly dangerous logic, to cordon off this kind of dangerous logic. The point being that dangerous resources as bad for both the generator, resolver and the player experience.

For this case I was a bit confused by how video beings at the floor of the room when the connection is from Front of Lasso Walls. I think the before clause is redundant in this case, since this stuff would fit into `after walls event or here` branch. Though, the lasso event is already dangerous because of the connection from Door to Fiery Airdock to Front of Lasso Walls.

Miepee: the other question that I have: can you still do this standable after you have shoot the thingy so that it moves out of the way?

Belokuikuini: Yes, it still works, just tested it

SchwartzGandhi (approved):

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
```diff
@@ -634,7 +635,10 @@ Extra - asset_id: 6613311822671481119
   > Pickup (Missile Expansion)
       Trivial
   > Event - Leviathan Shield
-      After Jungle Generator Ship Missile Attempt and After North Jungle Court Turret and Use Ship Missile
+      All of the following:
+          After Jungle Generator Ship Missile Attempt and After North Jungle Court Turret and Use Ship Missile
+          # Get to the ledge at Field Access, then return https://youtu.be/I5LtCOvlR80
+          After South Jungle Court Turret or Knowledge (Beginner)
```
Miepee, line 641: Not sure I like this, due to 2 reasons:
1. needing to transition through a room
2. the comment of 
> if you go too far and the room unloads, this won't work.

which i'm interpreting as "one chance only, if you fuck it up,good luck"

Miepee (changes requested):

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
Belokuikuini: Not sure why transitionning to another room is a problem here? It's not OoB, and I'm fairly certain there isn't much else you can do...
Would something like this https://www.youtube.com/watch?v=03I2HiDZoCg&t=9865s be better as a video? It doesn't technically transition rooms more than once, bit it won't show that just walking into the room won't work and that it can be fixed if you do ; imo it's even worse than my previous wording when it comes to implying this only works once

Belokuikuini: Well, after looking at the room in PWE, I don't think it's about an invisible trigger at all. In fact, the ship command icon to effectively bomb the generator is activated by a counter, which itself is linked to a special function for each turret. If the corresponding turret is destroyed, then it will increment the counter by 1. The counter needs to reach 2 for the bomb run icon to appear.
My theory is that each time a destroyed turret is rendered, the special function returns true, and increments the counter. As such, if we leave without unloading the room, the counter will remain at 1, but the destroyed turret will be culled, which does not decrement the counter because of an oversight. Then, re-entering the room will render the destroyed turret again, which will increment the counter to 2 and activate the bomb run icon.
If my theory is correct, then we should also be able to perform the same manip when destroying the South Jungle Court turret. I'll test it out a later today and modify the video title and description if my theory turns out to be relevant.

Also, the only misc resource available for corruption is "Skipping Energy Cells", so I don't think it's the relevant kind of resource to put, so how should I go about adding a new kind of misc resource, and what name should I give it? Something like "Can open door" ?

Miepee: > how should I go about adding a new kind of misc resource

You click on "new..."

> what name should I give it

Entrance Rando, then have a required disabled entrance rando resource.


I would also still bump the knowledge requirement because "you can permanently lock yourself out by accidentally unloading the room and then saving" does not sound like beginner.

Belokuikuini: I may not have insisted enough, but this isn't a one shot trick at all. I've modified my description to this, if it's clearer that you can try again if you mess it up : 
> Please note that this isn't a room reload, if you go too far and the room unloads, you will still not be able to see the icon, and will have to start the manip again

Miepee: ahh ok

randovania/games/prime3/logic_database/Bryyo - Fire.txt
```diff
@@ -140,6 +140,8 @@ Extra - asset_id: 4215957264467390882
               Any of the following:
                   After Imperial Hall Grapple Lasso Walls or Bomb/Spring Space Jump (Advanced)
                   Boost Ball and Bomb/Spring Space Jump (Intermediate)
+                  # Jump towards the Lasso Platforms https://youtu.be/Z-XICpMJWoE
+                  Movement (Beginner) and Standable Terrain (Beginner)
```
dyceron, line 144: How much harder is it without a lock on (scan visor)? Worth listing with scan?

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
```diff
@@ -98,7 +98,7 @@ Extra - asset_id: 15814834557214288722
   > G.F.S. Theseus Upper Ledge
       Any of the following:
           Morph Ball Bomb and Morph Ball and After G.F.S. Theseus Satellite
-          # https://youtu.be/eHJKUTiMgO8&t=75
+          # Jump on the lip near the ledge and morph at the apex of the jump https://youtu.be/inJ7_pTqlg0
```
dyceron, line 101: Why Morph in the comment? It's not listed and you definitely don't need it here.

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
```diff
@@ -134,8 +134,9 @@ Extra - asset_id: 2768064962148517187
               Any of the following:
                   Space Jump Boots and Bomb/Spring Space Jump (Beginner)
                   Morph Ball Bomb and Bomb/Spring Space Jump (Advanced)
+          # https://youtu.be/gEu-XdXFx8M
```
dyceron, line 137: Add comment

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
```diff
@@ -634,7 +635,12 @@ Extra - asset_id: 6613311822671481119
   > Pickup (Missile Expansion)
       Trivial
   > Event - Leviathan Shield
-      After Jungle Generator Ship Missile Attempt and After North Jungle Court Turret and Use Ship Missile
+      All of the following:
+          After Jungle Generator Ship Missile Attempt and After North Jungle Court Turret and Use Ship Missile
+          Any of the following:
+              After South Jungle Court Turret
+              # Get to the ledge at Field Access, then return https://youtu.be/I5LtCOvlR80
+              Knowledge (Beginner) and Disabled Entrance Randomizer
```
dyceron, line 643: Unsure about beginner. Maybe Intermediate would be better.

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
```diff
@@ -741,6 +747,7 @@ Extra - asset_id: 14006603351279167667
           All of the following:
               Space Jump Boots
               Any of the following:
+                  # https://youtu.be/bFFKq-oSBJg
```
dyceron, line 750: Comment

randovania/games/prime3/logic_database/Bryyo - Reptilicus.txt
```diff
@@ -995,6 +1002,7 @@ Extra - asset_id: 18426339925276965129
               Charge Beam or Missile ≥ 4
               Any of the following:
                   Use Screw Attack (Space Jump)
+                  # https://youtu.be/LW0nbio3u78
```
dyceron, line 1005: Comment

randovania/games/prime3/logic_database/GFS Valhalla.txt
```diff
@@ -230,8 +231,8 @@ Extra - asset_id: 7569990525320728426
                       Any of the following:
                           # https://youtu.be/p3LKN-9NaTk?t=5
                           Bomb/Spring Space Jump (Beginner)
-                          # https://youtu.be/6G8jyu4bkJ4
-                          Screw Attack and Slope Jump (Beginner)
+                          # https://youtu.be/nd6HL1_2GfA
```
dyceron, line 234: Comment

dyceron (changes requested):

randovania/games/prime3/logic_database/Bryyo - Fire.txt
Belokuikuini: While it's possible to make the jump without lock on, you pretty much need to have a good angle, good momentum, and good timing for your jump. You can't exactly keep moving forward because there is a decorative relief blocking the direct way forward, but you still have to keep your trajectory as close to the wall as possible without sticking to it so that your SJ doesn't get eaten while you still have the shortest path from the standable to the item ledge.
In other words, it's not excruciatingly difficult, but it's definitely not free.

That being said, there is a way to have lock-on movement in this room without scan, but it involves exploiting the room layers : if you climb the room from the bottom during the shelbug layer, you can lock onto a shelbug from the standable (this probably also works with the pirate drone but I haven't tested it yet). However I'm not entirely sure what triggers the shelbugs to spawn, and if the generator is destroyed, the drone will no longer spawn. Knowing this, I don't think it's worth considering this technique in the DB.

How do these new requirements sound  ?
 - All of the following : 
   - Standable Terrain (Beginner)
   - Any of the following : 
     - Scan Visor and Movement (Beginner)
     - Movement (Intermediate)

randovania/games/prime3/logic_database/GFS Valhalla.txt
```diff
@@ -231,7 +231,7 @@ Extra - asset_id: 7569990525320728426
                       Any of the following:
                           # https://youtu.be/p3LKN-9NaTk?t=5
                           Bomb/Spring Space Jump (Beginner)
-                          # https://youtu.be/nd6HL1_2GfA
+                          # Jump on the rubble in the corner between the E-Cell slopes and Repair Bayhttps://youtu.be/nd6HL1_2GfA
```
dyceron, line 234: Missing space between comment and video

dyceron (changes requested):

randovania/games/prime3/logic_database/Bryyo - Fire.txt
dyceron: Send it

dyceron (approved):

PR #7229, opened 2024-09-15
Door Lock Rando: Find either side of door

Tyranisaur:
Instead of looking for only one of the sides of each door, look for either side. This leads to fewer impossible results, which should speed up generation. However, there's a risk that seeds could be less interesting, due to always finding the most accessible side of the door.


duncathan (changes requested):
this should be behind an experimental generator setting, since it poses the risk of being less interesting

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed: Improved server performance when importing/generating games with a huge number of worlds.
 
+### Door Lock Randomizer
+- Added: Option that changes Door Lock Rando so that when determining which types a door can be randomized into, the algorithm searches for either side of the door using the resolver. This means some doors are more likely to have locks on them. Generation will also be faster.
```
duncathan, line 13: please indicate that it's an experimental option, since otherwise users may be confused why they can't find it

duncathan (changes requested):
being that this is an experimental setting, is there any reason not to include in the upcoming release?

Miepee:
> is there any reason not to include in the upcoming release?

By the time this would get merged, it'd be like 1.5 days before the new release, which feels a little too close to comfort for my tastes

Miepee (approved):

Miepee:
@Tyranisaur merge conflicting right now.

PR #6073, opened 2024-02-26
Add basic functionality for Metroid Fusion

JeffGainsNGames:
Includes:

- Database (trickless logic has had multiple passes)
- All room images
- Starter Preset
- Basic GUI and Export/PDF components from new game generation command

PR #7237, opened 2024-09-16
Add gate bulbs for Database pictures

SchwartzGandhi:


JeffGainsNGames (changes requested):
you'll need to compress the images as they are currently >2x as large. I don't remember all the details but it's in miepees comments on https://github.com/randovania/randovania/pull/6073. Or if you know another way by all means.

Miepee (approved):
haha compression go brrrrrrrrr

PR #7239, opened 2024-09-17
Fusion: Exporter GUI

JeffGainsNGames:
image

- [x] gui
- [ ] ~~actual exporter stuff~~ (later PR)
- [x] tests

### DIFF
--- a/randovania/games/fusion/exporter/options.py
+++ b/randovania/games/fusion/exporter/options.py
@@ -10,10 +10,9 @@ from randovania.interface_common.options import PerGameOptions, decode_if_not_no
 
 @dataclasses.dataclass(frozen=True)
 class FusionPerGameOptions(PerGameOptions):
-    """ """
-
     input_path: Path | None = None
     output_path: Path | None = None
+    output_format: str = "gba"
 
     @property
     def as_json(self) -> dict:
--- a/randovania/games/fusion/exporter/options.py
+++ b/randovania/games/fusion/exporter/options.py
@@ -12,7 +12,6 @@ from randovania.interface_common.options import PerGameOptions, decode_if_not_no
 class FusionPerGameOptions(PerGameOptions):
     input_path: Path | None = None
     output_path: Path | None = None
-    output_format: str = "gba"
 
     @property
     def as_json(self) -> dict:
@@ -20,7 +19,6 @@ class FusionPerGameOptions(PerGameOptions):
             **super().as_json,
             "input_path": str(self.input_path) if self.input_path is not None else None,
             "output_path": str(self.output_path) if self.output_path is not None else None,
-            "output_format": self.output_format,
         }
 
     @classmethod
@@ -31,5 +29,4 @@ class FusionPerGameOptions(PerGameOptions):
             cosmetic_patches=cosmetic_patches,
             input_path=decode_if_not_none(value["input_path"], Path),
             output_path=decode_if_not_none(value["output_path"], Path),
-            output_format=value["output_format"],
         )
--- a/randovania/games/fusion/exporter/options.py
+++ b/randovania/games/fusion/exporter/options.py
@@ -20,6 +20,7 @@ class FusionPerGameOptions(PerGameOptions):
             **super().as_json,
             "input_path": str(self.input_path) if self.input_path is not None else None,
             "output_path": str(self.output_path) if self.output_path is not None else None,
+            "output_format": self.output_format,
         }
 
     @classmethod
@@ -30,4 +31,5 @@ class FusionPerGameOptions(PerGameOptions):
             cosmetic_patches=cosmetic_patches,
             input_path=decode_if_not_none(value["input_path"], Path),
             output_path=decode_if_not_none(value["output_path"], Path),
+            output_format=value["output_format"],
         )
--- a/randovania/games/fusion/gui/dialog/game_export_dialog.py
+++ b/randovania/games/fusion/gui/dialog/game_export_dialog.py
@@ -56,9 +56,9 @@ class FusionGameExportDialog(GameExportDialog, Ui_FusionGameExportDialog):
     def __init__(self, options: Options, patch_data: dict, word_hash: str, spoiler: bool, games: list[RandovaniaGame]):
         super().__init__(options, patch_data, word_hash, spoiler, games)
 
-        self._base_output_name = f"MARS - {word_hash}." + self.valid_file_type
-        per_game = options.options_for_game(self.game_enum())
-        assert isinstance(per_game, FusionPerGameOptions)
+        self._base_output_name = f"MARS - {word_hash}.{self.valid_file_type}"
+        fusion_options = options.options_for_game(self.game_enum())
+        assert isinstance(fusion_options, FusionPerGameOptions)
 
         # Input
         self.input_file_button.clicked.connect(self._on_input_file_button)
@@ -66,11 +66,11 @@ class FusionGameExportDialog(GameExportDialog, Ui_FusionGameExportDialog):
         # Output
         self.output_file_button.clicked.connect(self._on_output_file_button)
 
-        if per_game.input_path is not None:
-            self.input_file_edit.setText(str(per_game.input_path))
+        if fusion_options.input_path is not None:
+            self.input_file_edit.setText(str(fusion_options.input_path))
 
-        if per_game.output_path is not None:
-            output_path = per_game.output_path.joinpath(self._base_output_name)
+        if fusion_options.output_path is not None:
+            output_path = fusion_options.output_path.joinpath(self._base_output_name)
             self.output_file_edit.setText(str(output_path))
 
         add_field_validation(
@@ -111,10 +111,10 @@ class FusionGameExportDialog(GameExportDialog, Ui_FusionGameExportDialog):
         if output_file is not None:
             self.output_file_edit.setText(str(output_file))
 
-    def update_per_game_options(self, per_game: FusionPerGameOptions) -> FusionPerGameOptions:
-        assert isinstance(per_game, FusionPerGameOptions)
+    def update_per_game_options(self, fusion_options: FusionPerGameOptions) -> FusionPerGameOptions:
+        assert isinstance(fusion_options, FusionPerGameOptions)
         return dataclasses.replace(
-            per_game,
+            fusion_options,
             input_path=self.input_file,
             output_path=self.output_file,
         )
--- a/randovania/games/fusion/gui/dialog/game_export_dialog.py
+++ b/randovania/games/fusion/gui/dialog/game_export_dialog.py
@@ -58,11 +58,7 @@ class FusionGameExportDialog(GameExportDialog, Ui_FusionGameExportDialog):
         )
 
     @property
-    def valid_input_file_types(self) -> list[str]:
-        return ["gba"]
-
-    @property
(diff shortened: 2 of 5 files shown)


randovania/games/fusion/exporter/options.py
```diff
@@ -10,10 +10,9 @@
 
 @dataclasses.dataclass(frozen=True)
 class FusionPerGameOptions(PerGameOptions):
-    """ """
-
     input_path: Path | None = None
     output_path: Path | None = None
+    output_format: str = "gba"
```
Miepee, line 15: a todo option to automatically create a patch when exporting

randovania/games/fusion/gui/dialog/game_export_dialog.py
```diff
@@ -25,19 +29,69 @@ class FusionGameExportDialog(GameExportDialog, Ui_FusionGameExportDialog):
...
+
+        if per_game.output_path is not None:
+            output_path = per_game.output_path.joinpath(self.default_output_name)
+            self.output_file_edit.setText(str(output_path))
+
+        add_field_validation(
+            accept_button=self.accept_button,
+            fields={
+                self.input_file_edit: lambda: False,
+                self.output_file_edit: lambda: False,
```
Miepee, line 56: what is the point in falsing this?

randovania/games/fusion/gui/dialog/game_export_dialog.py
```diff
@@ -25,19 +29,69 @@ class FusionGameExportDialog(GameExportDialog, Ui_FusionGameExportDialog):
...
+            },
+        )
+
+    @property
+    def valid_input_file_types(self) -> list[str]:
+        return ["gba"]
+
+    @property
+    def valid_output_file_types(self) -> list[str]:
+        return ["gba"]
```
Miepee, line 66: do we ever plan to support anything *other* than gba types? if not, we could probably change these to just return strings instead of list of strings.

randovania/games/fusion/gui/dialog/game_export_dialog.py
```diff
@@ -25,19 +29,69 @@ class FusionGameExportDialog(GameExportDialog, Ui_FusionGameExportDialog):
...
+    # Output File
+    def _on_output_file_button(self):
+        output_file = prompt_for_output_file(
+            self, self.valid_output_file_types, self._base_output_name, self.output_file_edit
+        )
+        if output_file is not None:
+            self.output_file_edit.setText(str(output_file))
 
-    def update_per_game_options(self, per_game: PerGameOptions) -> FusionPerGameOptions:
+    def update_per_game_options(self, per_game: FusionPerGameOptions) -> FusionPerGameOptions:
```
Miepee, line 94: this should be PerGameOptions.

randovania/games/fusion/gui/dialog/game_export_dialog.py
JeffGainsNGames: I can't remember why I put false, I was referencing other games and couldn't make sense of the various validator functions they used since I guess it's per game to ensure the input is expected?

randovania/games/fusion/gui/dialog/game_export_dialog.py
JeffGainsNGames: I referenced AM2R and Prime and they both had their respective game options

randovania/games/fusion/gui/dialog/game_export_dialog.py
Miepee: correct. you're basically writing a condition involving each component.
e.g. from am2r:
```py
fields={
                self.input_file_edit: lambda: not (self.input_file.is_dir() and _is_valid_input_dir(self.input_file)),
                self.output_file_edit: lambda: not (self.output_file.is_dir() and self.output_file != self.input_file),
            },
```
this makes:
- the input field button will have a red border if the input is not a directory, or the directory is an invalid one (e.g. doesn't have a data.win inside of it)
- the output field will have a red border, if the output field is not a directory, or if the output field is the same as the input field

JeffGainsNGames: so I think my attempt was to say "we dont have any checks so always return false" but putting a TODO (or just make it now) would have been appropriate

randovania/games/fusion/gui/dialog/game_export_dialog.py
JeffGainsNGames: outside of my understanding of gba formats. does it hurt anything the way it is beyond could be more appropriately typed?

randovania/games/fusion/exporter/options.py
JeffGainsNGames: do we even want to support people using patch files instead of RDV games?

randovania/games/fusion/gui/dialog/game_export_dialog.py
Miepee: interesting. pretty sure that mypy should complain about that

randovania/games/fusion/exporter/options.py
Miepee: it's a common enough operation that it's probably worth having.

randovania/games/fusion/gui/dialog/game_export_dialog.py
Miepee: on second thought, no, because `prompt_for_output_file` needs a list anyway.

randovania/games/fusion/exporter/options.py
```diff
@@ -10,17 +10,17 @@
...
```
Miepee, line 15: remind me why we have the output format as an option if we cant change it?

randovania/games/fusion/gui/dialog/game_export_dialog.py
```diff
@@ -25,19 +29,69 @@ class FusionGameExportDialog(GameExportDialog, Ui_FusionGameExportDialog):
     def game_enum(cls) -> RandovaniaGame:
         return RandovaniaGame.FUSION
 
+    def __init__(self, options: Options, patch_data: dict, word_hash: str, spoiler: bool, games: list[RandovaniaGame]):
+        super().__init__(options, patch_data, word_hash, spoiler, games)
+
+        self._base_output_name = f"MARS - {word_hash}.gba"
```
Miepee, line 35: i dont like hardcoding the extension here. could you please change:
- `valid_input_file_types` to `valid_input_file_type` and make it return a string
- `valid_output_file_types` to `valid_output_file_type` and make it return a string
- use `self.valid_input_file_type` here in this line
- change the other usages in `_on_input_file_button` and `_on_output_file_button` to have a list that wrap around `self.valid_input/output_file_type`

makes more sense IMO since we dont ever really have a need to not export a gba rom file

JeffGainsNGames: if it doesn't change what's wrong with hard coding it?

Miepee: because you're now hardcoding it in two places: here and the valid input file types.

JeffGainsNGames: didn't we not want to change `list[str]` to `string` because it's used by `prompt_for_input_file`

Miepee: `prompt_for_input_file` just wants to take in a list. hence
> change the other usages in `_on_input_file_button` and `_on_output_file_button` to have a list that wrap around `self.valid_input/output_file_type`

Miepee: in practice it could look like this:
`input_file = prompt_for_input_file(self, self.input_file_edit, [self.valid_input_file_type])`

Miepee: you didnt resolve this

randovania/games/fusion/gui/dialog/game_export_dialog.py
JeffGainsNGames: I added output validation, but can't figure out what to do about input. Some games just dont have it?

Miepee: validate that its an US fusion rom by 
1. validating that its 8MB big
2. has the ascii text `METROID4USA` starting at offset 0xA0

You can look at prime1's input validator and msr's exheader validator for reference.

henriquegemignani: The rom is rather small, just has the whole thing.

duncathan: Unless you specifically intend to support modified input ROMs just hash it yeah

randovania/games/fusion/gui/dialog/game_export_dialog.py
```diff
@@ -1,21 +1,50 @@
...
-    from pathlib import Path
+    from randovania.interface_common.options import Options
 
-    from randovania.interface_common.options import PerGameOptions
+
+def is_fusion_validator(path: Path | None) -> bool:
+    """Validates if the given path is a proper input for Fusion.
+    - If input doesn't exist, returns True.
+    - If input is GBA, return False if it's Fusion otherwise True.
+    - If input is not GBA, returns False.
```
Miepee, line 30: the comment here is wrong.

randovania/games/fusion/gui/dialog/game_export_dialog.py
```diff
@@ -1,21 +1,50 @@
...
+    """
+
+    # Expected hash for vanilla "Metroid Fusion (USA).gba"
```
Miepee, line 33: please annotate taht this is an md5 has for future reference.

randovania/games/fusion/gui/dialog/game_export_dialog.py
```diff
@@ -25,19 +53,65 @@ class FusionGameExportDialog(GameExportDialog, Ui_FusionGameExportDialog):
     def game_enum(cls) -> RandovaniaGame:
         return RandovaniaGame.FUSION
 
+    def __init__(self, options: Options, patch_data: dict, word_hash: str, spoiler: bool, games: list[RandovaniaGame]):
+        super().__init__(options, patch_data, word_hash, spoiler, games)
+
+        self._base_output_name = f"MARS - {word_hash}." + self.valid_file_type
+        per_game = options.options_for_game(self.game_enum())
```
Miepee, line 60: rename this to `fusion_options`

Miepee (approved):

randovania/games/fusion/gui/dialog/game_export_dialog.py
```diff
@@ -25,19 +53,65 @@ class FusionGameExportDialog(GameExportDialog, Ui_FusionGameExportDialog):
     def game_enum(cls) -> RandovaniaGame:
         return RandovaniaGame.FUSION
 
+    def __init__(self, options: Options, patch_data: dict, word_hash: str, spoiler: bool, games: list[RandovaniaGame]):
+        super().__init__(options, patch_data, word_hash, spoiler, games)
+
+        self._base_output_name = f"MARS - {word_hash}." + self.valid_file_type
```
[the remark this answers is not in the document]
henriquegemignani: I think using the game name instead of the underlying randomizer makes more sense, but that's a different thing.

PR #7251, opened 2024-09-23
Make qsliders in UI unscrollable

quinnbbowman:
PR re: https://github.com/randovania/randovania/issues/7009

Work done: replaced qsliders in stable game UI with a new custom qslider class that cannot be scrolled with scroll wheel.

Tested by scrolling over sliders in all prime 1 UI dialogues with sliders; the new sliders work as intended and are not scrollable with the wheel, unless they are currently active.

This PR addresses the linked issue, but I personally think it would be good to replace all comboboxes and spinboxes in the UI with unscrollable ones too, seeing as they have the same issue where there are many vertical windows in rdv that are scrolled thourgh without wanting to change things, so I will be awaiting feedback on whether that seems reasonable before I address anything outside of the linked issue.  There may be objects that are desirable to remain scrollable that I do not know about.  For similar reasons, I have only replaced the sliders in stable games.

### DIFF
--- a/randovania/games/am2r/gui/ui_files/preset_am2r_chaos.ui
+++ b/randovania/games/am2r/gui/ui_files/preset_am2r_chaos.ui
@@ -101,7 +101,7 @@
               <number>6</number>
              </property>
              <item>
-              <widget class="QSlider" name="darkness_slider">
+              <widget class="ScrollProtectedSlider" name="darkness_slider">
                <property name="maximum">
                 <number>1000</number>
                </property>
@@ -265,7 +265,7 @@
               </widget>
              </item>
              <item>
-              <widget class="QSlider" name="submerged_water_slider">
+              <widget class="ScrollProtectedSlider" name="submerged_water_slider">
                <property name="maximum">
                 <number>1000</number>
                </property>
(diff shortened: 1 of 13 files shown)


Miepee:
> personally think it would be good to replace all comboboxes and spinboxes in the UI with unscrollable ones too,

IMO combo- and spinboxes are much more common to always be scrollable. The only circumstance that i know where i accidentally change the value there is in the logic db editor. Do you have any other anecdotes there?

---

Do we ever have a reason to have a scrolling qslider? If not, we might want to add a unit test that fails if theres a UI element that uses normal QSliders.

Miepee:
Idk, maybe im just more biased with which/how i use my software, but i prefer scrollable boxes.
With randovania at least i do feel that if we have a problem with accidentally changing values on default sizes on an 800x600res, that we have an issue with our ui than it being the fault of the boxes.

Miepee:
i honestly very much dislike right now how scroll-heavy a lot of the widgets in randovania are. Reworking them to have less of a need to be scrolled and changing the dropdowns to not take the full width of the window is definitely on my list of things that I plan to do.

duncathan:
This is a useful conversation, but let's call the combo boxes and such out of scope for this PR so that we can get the existing improvement through quickly. Please apply the changes to all games though, not just stable

Miepee (approved):
thanks!

PR #7261, opened 2024-09-25
Add logic node visualization for prime 1

Miepee:
This makes the nodes in the DB visualizer visible for prime 1:
grafik
As visible on the screenshot, this does *not* add visible maps for them. However, despite that it already is much more readable. Often as readable, if not better, as the discord command:
grafik

---

Some notes for reviewing:
- Check the branch out and view the rooms in the visualizer
- The door position data was already filled out, I tried to mostly keep it intact in the case that we do add maps in the future. (But in that case we probably still want to redo the positioning data to use world coordinates instead of the data in the mlvls?)
- I tried my best to have the nodes be in their respective location in the room. I'm sure though that I completely missed the mark on some nodes.


SchwartzGandhi (approved):
looked over this with miepee in vc, looks good to me but open to second opinions

henriquegemignani:
Where the locations done manually? Also, I'll leave the decision to @toasterparty

Miepee:
The dock nodes used their already existing positioning data. Some of them were slightly moved for readability's sake (e.g. rooms like great tree hall, where if you disregard the height dimension, some of the docks are exactly above each other)
Generic nodes, teleporters, event nodes and pickup nodes were all placed manually. 

I wanted to have the opinion of at least one prime db maintainer, and had planned to go through this with reggie soon. But toaster of course has the final say for (dis)approval.

JeffGainsNGames (approved):

Miepee:
we had jeff+reggie+sam looking over this in VC.

Uncle-Reggie:
Echoing support

toasterparty (approved):

Miepee:
github why :(
grafik

Miepee:
apparently doing the same thing again worked lol

PR #7278, opened 2024-09-30
Introduce Equipment Manipulation resource

SchwartzGandhi:
Misc resource for tricks that require enabling/disabling an item mid-trick

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 3 PYR.json
+++ b/randovania/games/fusion/logic_database/Sector 3 PYR.json
@@ -2283,8 +2283,42 @@
                                                     }
                                                 },
                                                 {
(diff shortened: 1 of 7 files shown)


Miepee (commented):
looks ok for me so far, 2 thoughts tho:
- *maybe* the s1 screw shaft should count under this
- depending on how many usages we find after our whole cleanup, we may want to just use movement instead. not worth a whole seperate resource for <5 usages IMO.

JeffGainsNGames:
like leaving this disabled by default then add an option to enable later. this is new tech so dont want folks having to do this without any indication stuff like this is possible.

Miepee (approved):

PR #7283, opened 2024-10-01
Revert am2r_yams back to 2.8.0

Miepee:
cant figure out how to handle the new imagemagick lib in utmt right now, so we revert


Miepee:
no. the only change between 2.8.0 and 2.9.0rc was bumping to net8 which isnt a user facing thing.

PR #7284, opened 2024-10-01
Generator: make indices/event/hint multiplier game overwritable

Miepee:
Dunno if this needs a changelog entry or not.

If an action **only** unlocks hints, we weight it less in order to avoid failing generation.
I was unsure whether i should make it have a weight of 0, instead of a reduced weight, but decided with the reduced weight for now as I assumed that Echoes probably relies on that behaviour a bit.


duncathan:
if this would affect different games to a different extent we could expose it as a variable in `GameGenerator` or a field in the db or something

Miepee:
never thought about that class. Does it make sense to just provide a function there that calculates the weight for an action (based on the available pickups/events/hints)? Dunno if we need that much flexibility, but feels nicer to do that than introduce new fields in the future for specific circumstances. I remember prime people raising the thought a few times of tweaking how actions are weighted.

duncathan:
that would add a lot of complexity and more disparity between games than I would like. if you wanna throw the other constant weights in there that'd be fine tho

Miepee:
After looking at more fusion logs; yeah, being able to tweak some of the other values would be good too.

Miepee:
Changed this PR to use dunc's suggestion of having game customizable values in GameGenerator

randovania/games/game.py
```diff
@@ -8,6 +8,11 @@
 
 import randovania
 from randovania.bitpacking.bitpacking import BitPackEnum
+from randovania.generator.filler.weights import (
```
henriquegemignani, line 11: Slightly concerned with this file importing something from the `randovania.generator` module, outside of the TYPE_CHECKING block. This might cause some import loops in weird scenarios?

Tbh looking at it we really should split this file huh? No reason for `GameData` to come from the same file as `RandovaniaGame`.

Miepee: I've been considering not having default values, but I thought that for the sake of having it easier for people to configure this'd be nicer.

But if every new game just comes from BDG, I guess that argument doesnt really hold. So there would just be the move of generator default weigths from game.py to games/[game_name]/game.py which I don't think would change much

---

I did initially have the weights in retcon.py but moved them out of there cause they caused circular dependencies. Weights being in their own file is probably fine? Cause that constants file cant really import anything else.

henriquegemignani: Oh no I like the default values. Let's use them so we can adjust the generator in a global way without having to adjust every single game.

Importing the separate file still imports every `__init__.py` in the way, which would cause issues if these had their own code.

I like to think of the different folders in the module root as it's own modules and to keep an uni-directional dependency loop. So if module A imports module B, then module B is not allowed to import anything from module A ever.

At least, not outside of TYPE_CHECKING blocks.

Miepee: I see. I don't think there's a nice way around this though. I was thinking of maybe having weights as its own class instead of just random constants, and then GameData can just reference this for type checking, but since it would still need to use the default values/class, it would precisely need to import it *not* as type checking.

Miepee:
rebased this based off of the refactor that moved game.enum out

randovania/generator/filler/weights.py
```diff
@@ -0,0 +1,7 @@
+DEFAULT_INDICES_WEIGHT_MULTIPLIER = 1
+DEFAULT_EVENTS_WEIGHT_MULTIPLIER = 0.5
+DEFAULT_HINTS_WEIGHT_MULTIPLIER = 1
+
+_DANGEROUS_ACTION_MULTIPLIER = 0.75
```
henriquegemignani, line 5: If these are going to be in a different file, they shouldn't have `_` in front.

randovania/generator/filler/retcon.py
```diff
@@ -474,21 +472,23 @@ def _calculate_weights_for(
...
 
         print(f">>> {action}")
         print(f"indices: {potential_uncollected.indices}")
-        print(f"hints: {[hint.as_string for hint in potential_uncollected.hints]}")
         print(f"events: {[event.long_name for event in potential_uncollected.events]}")
+        print(f"hints: {[hint.as_string for hint in potential_uncollected.hints]}")
         print(f"nodes: {[nodes[n].identifier.as_string for n in potential_uncollected.nodes]}")
         print()
 
+    generator = potential_reach.game.game.generator
```
henriquegemignani, line 485: I usually put things in `FillerConfiguration`, though I guess that's for preset parameters. I'm thinking maybe there should be a dataclass with all the weights, have the default values be defined directly in that class and store an instance of that class in `GameGenerator`.

randovania/generator/filler/retcon.py
```diff
@@ -474,21 +472,23 @@ def _calculate_weights_for(
         return _VICTORY_WEIGHT
 
     potential_uncollected = UncollectedState.from_reach(potential_reach) - current_uncollected
-    if debug.debug_level() > 2:
+    if debug.debug_level() > 1:
```
henriquegemignani, line 475: Aren't these prints rather verbose? Rather keep it to level 3

henriquegemignani (changes requested):

randovania/generator/filler/retcon.py
Miepee: Addressed.

output.md
henriquegemignani, line 1: stray file

henriquegemignani (approved):

PR #7292, opened 2024-10-02
start ARCold runs

SchwartzGandhi:


randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -299,17 +299,21 @@ Extra - room_id: [9]
...
   * L3 Hatch to Level 3 Security Room/Door to Security Shaft West
   * Extra - door_idx: (20,)
   > Door to Security Save Room
       All of the following:
-          # TODO:  damage reqs
-          Varia Suit
+          Any of the following:
+              # Damage Reqs
+              Varia Suit
+              Hi-Jump and Space Jump and Damage Runs (Expert) and Cold Damage ≥ 600
```
Miepee, line 316: how much damage do you take by only doing this with WJ?
also why is there HJ+SJ?

randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -400,61 +404,88 @@ Extra - room_id: [13]
...
+              # Damage Reqs
+              Varia Suit
+              Hi-Jump and Damage Runs (Advanced) and Cold Damage ≥ 238
   > Bottom of Room
-      # TODO: dmg
-      Varia Suit
+      Any of the following:
+          # Damage Reqs
+          Varia Suit
+          Damage Runs (Advanced) and Cold Damage ≥ 319
```
Miepee, line 418: what did you do to measure this, i cant replicate this at all. i'm taking 41 damage here by just falling down. are you facetanking all ripppers and blue x? the connection from the dat room to botom only takes ~150 so somethings definitely weird here

randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -400,61 +404,88 @@ Extra - room_id: [13]
   * Layers: default
   * L0 Hatch to Mimic Storage/Door to Frozen Tower
   * Extra - door_idx: (24,)
+  > Door to Data Room
```
Miepee, line 407: unsure whether we should add connections like this. i'm probably just worrying too much tho

randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -400,61 +404,88 @@ Extra - room_id: [13]
...
 > Bottom of Room; Heals? False
   * Layers: default
   > Door to Mimic Storage
       All of the following:
-          # TODO: dmg
-          Varia Suit
+          Any of the following:
+              # Damage Reqs
+              Varia Suit
+              Hi-Jump and Space Jump and Damage Runs (Expert) and Cold Damage ≥ 521
```
Miepee, line 465: seems a bit much. with HJ+SJ i can reliably get there with only ~300
without Hj i can get there with ~400

randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -400,61 +404,88 @@ Extra - room_id: [13]
...
+              Damage Runs (Advanced) and Cold Damage ≥ 48
   > Door to Data Room
-      # TODO: dmg
-      Space Jump and Varia Suit
+      All of the following:
+          Space Jump
+          Any of the following:
+              # Damage Reqs
+              Varia Suit
+              Damage Runs (Expert) and Cold Damage ≥ 692
```
Miepee, line 488: is this with or without HJ?

randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -740,36 +774,56 @@ Extra - room_id: [22, 21]
                       # https://youtu.be/a7A00q8XP9M
                       Wall Jump (Intermediate)
               Wall Jump (Intermediate) and Can Single Walljump
+          Any of the following:
+              # Damage Reqs
+              Varia Suit
+              Space Jump and Damage Runs (Expert) and Cold Damage ≥ 273
```
Miepee, line 780: might be worth having also an entry for freezing via diffusion? maybe not.

randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -1548,9 +1602,11 @@ Extra - room_id: [51]
...
       All of the following:
-          # TODO: varia / damage reqs
-          Varia Suit
           Any of the following:
               Space Jump
               Hi-Jump and Can Freeze Enemies
+          Any of the following:
+              # Damage Reqs
+              Varia Suit
+              Hi-Jump and Space Jump and Damage Runs (Expert) and Cold Damage ≥ 203
```
Miepee, line 1611: worth having an entry for freezing?
also what about SJ only.

Miepee (commented):
Other lingering thoughts
- what were your reasonings for making all difficulty advanced+?
- we maybe wanna standardize when exactly we'll "cheap out" on with damage run connections.

SchwartzGandhi:
context for a lot of these connections:
- my thought process for having these all be advanced or expert is that some people will probably want to have heat runs enabled, but not want to do specifically cold runs, tbh I kinda want to separate coldruns into their own trick category.
- I arbitraily put in a bunch of HJ+SJ checks here because the knockback you get really makes it hard to stand on rippers, climb ladders, do hardly any movement reliably, etc. I guess i could theoretically be possible to do some of these w/o SJ, but imo thats just overengineering for a trick no one really likes.
- These numbers also assume you're tanking all other damage, specifically blue X. This ties back to my previous point and that i feel like its not worth incorperating combat tricks just for something like this.

randovania/games/fusion/logic_database/Sector 5 ARC.txt
SchwartzGandhi: this one i forgot to add oops, yes this was done with hijump

Miepee:
> tbh I kinda want to separate coldruns into their own trick category.

I would do this way later, if even. Very unsure if it would be worth the effort, most cold runs seem mostly the same *and* they're limited to 10 rooms in s5 (out of like 50)

>  I guess i could theoretically be possible to do some of these w/o SJ, but imo thats just overengineering 

I'd rather use a worst case (e.g. SJ only) in those situations there tbh

> for a trick no one really likes.

If **no one** really likes those, we could probably save us the effort and just always require varia logically. I personally don't like playing with damage runs in the first place, idk about others.

> These numbers also assume you're tanking all other damage, specifically blue X

You're being *very* generous in most of your numbers then. Most of the time i was still able to cut down your numbers by a lot. Not necessarily bad (even though I think you're overdoing it in a few places), but would've been helpful to have as info earlier

SchwartzGandhi:
> You're being _very_ generous in most of your numbers then. Most of the time i was still able to cut down your numbers by a lot. Not necessarily bad (even though I think you're overdoing it in a few places), but would've been helpful to have as info earlier
Tbh I think this is mainly a consequence of how much damage you take. getting the timing wrong and dropping one space jump costs like 50 energy
> I'd rather use a worst case (e.g. SJ only) in those situations there tbh
fair enough, this would at least give an excuse to try and get better numbers lol

Miepee:
> Tbh I think this is mainly a consequence of how much damage you take. getting the timing wrong and dropping one space jump costs like 50 energy

I don't think the timing is *that* bad. On my attempts I was purposefully trying to be bad and still ended up taking much less health.

randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -404,61 +412,88 @@ Extra - room_id: [13]
...
   > Door to Mimic Storage
       All of the following:
-          # TODO: dmg
-          Varia Suit
+          Any of the following:
+              # Damage Reqs
+              Varia Suit
+              Space Jump and Damage Runs (Advanced) and Cold Damage ≥ 59
           # Drop down from ledge above https://youtu.be/lUc4sT4UyQY
           Jump Extends (Expert) and Movement (Hypermode)
```
Miepee, line 459: this doesnt make much sense. it will still require the jump extend despite you having SJ.

randovania/games/fusion/logic_database/Sector 5 ARC.txt
Miepee: Didnt you say this was done with hijump?

randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -404,61 +412,88 @@ Extra - room_id: [13]
   * Layers: default
   * L0 Hatch to Mimic Storage/Door to Frozen Tower
   * Extra - door_idx: (24,)
+  > Door to Data Room
+      All of the following:
+          Space Jump
+          Any of the following:
+              # Damage Reqs
+              Varia Suit
```
Miepee, line 420: i would prefer if youd remove varia suit from this for duplications sake.

randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -744,36 +782,56 @@ Extra - room_id: [22, 21]
...
   * Extra - door_idx: (109,)
   > Pickup (Power Bomb Tank)
-      # TODO: varia / damage reqs
-      Varia Suit and Can Use Any Bombs
+      All of the following:
+          Can Use Any Bombs
+          Any of the following:
+              # Damage Reqs
+              Varia Suit
+              Damage Runs (Advanced) and Cold Damage ≥ 140
```
Miepee, line 805: why does this take so much more damage tahn falling down from l3=

randovania/games/fusion/logic_database/Sector 5 ARC.txt
Miepee: I think an SJ only entry would still be helpful here.

Miepee:
Whatever you pushed is mismatching between json and txt

Miepee:
on a second look, seems like just whitespace changes. just open it in the editor and resave it.

Miepee (approved):

Miepee:
@SchwartzGandhi conflicts :(

Miepee:
nvm fixed it myself

PR #7296, opened 2024-10-03
Fix possibility of obfuscator secrets being supplied incorrectly

### LINKED ISSUE #7268: Make lib.obfuscator raise missingsecret exception when the secret is an empty/whitespace-only string
turns out github actions will happily use secrets not declared in the project but set them to an empty string which passes the "is not none" check in create_release

Miepee:
Fixes #7268

### DIFF
--- a/tools/create_release.py
+++ b/tools/create_release.py
@@ -183,7 +183,7 @@ async def main():
     icon_path = randovania.get_icon_path()
     shutil.copyfile(icon_path, icon_path.with_name("executable_icon.ico"))
 
-    if (secret := os.environ.get("OBFUSCATOR_SECRET")) is not None:
+    if (secret := os.environ.get("OBFUSCATOR_SECRET")) is not None and len(secret) >= 1 and not secret.isspace():
         write_obfuscator_secret(
             _ROOT_FOLDER.joinpath("randovania", "lib", "obfuscator_secret.py"),
             secret.encode("ascii"),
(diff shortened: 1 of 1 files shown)


tools/create_release.py
```diff
@@ -183,7 +183,7 @@ async def main():
     icon_path = randovania.get_icon_path()
     shutil.copyfile(icon_path, icon_path.with_name("executable_icon.ico"))
 
-    if (secret := os.environ.get("OBFUSCATOR_SECRET")) is not None:
+    if (secret := os.environ.get("OBFUSCATOR_SECRET")) is not None and len(secret) >= 1 and not secret.isspace():
```
henriquegemignani, line 186: maybe this?
```python
secret = os.environ.get("OBFUSCATOR_SECRET", "")
if secret.trim() and not secret.isspace():
```

henriquegemignani (changes requested):

tools/create_release.py
Miepee: What is `trim` supposed to be for a function? Do you mean strip?

henriquegemignani: probably, too used to different names in different languages

Miepee: doesn't work then.
Stripping an empty string, results in an empty string. and isspace only works on strings with length bigger than 0.
```py
>>> s = ""
>>> s.strip().isspace()
False
>>> 
```

henriquegemignani: I did suggest both conditions?

Miepee: oh, i got switched up by it. forget sometimes that a string itself can be truthy.

henriquegemignani (approved):

PR #7298, opened 2024-10-04
Fusion: use better templates for freezing enemies

Miepee:


JeffGainsNGames:
How is the s4.json a -200 line diff only?

Miepee:
Github just displays it weird. if you look at the actual diff you'll see that there are additions too
grafik

SchwartzGandhi (approved):

PR #7304, opened 2024-10-04
Make useless pickup category not refer to etm

Miepee:
stops cave story confusion by not talking about ETMs anymore.

### DIFF
--- a/randovania/game_description/pickup/pickup_category.py
+++ b/randovania/game_description/pickup/pickup_category.py
@@ -54,7 +54,7 @@ class PickupCategory:
 
 
 USELESS_PICKUP_CATEGORY = PickupCategory(
-    name="useless", long_name="Useless", hint_details=("an ", "Energy Transfer Module"), hinted_as_major=False
+    name="useless", long_name="Useless", hint_details=("a ", "useless item"), hinted_as_major=False
 )
 
 GENERIC_KEY_CATEGORY = PickupCategory(name="key", long_name="Key", hint_details=("a ", "key"), hinted_as_major=False)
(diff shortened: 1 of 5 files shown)


henriquegemignani (changes requested):
Changing the text for Echoes in the other hand is not interesting.

Miepee:
Darn, I thought I could get away with being lazy.

randovania/game_description/pickup/pickup_category.py
```diff
@@ -54,7 +54,7 @@ def is_expansion(self) -> bool:
 
 
 USELESS_PICKUP_CATEGORY = PickupCategory(
-    name="useless", long_name="Useless", hint_details=("an ", "Energy Transfer Module"), hinted_as_major=False
+    name="useless", long_name="Useless", hint_details=("a ", "useless item"), hinted_as_major=False
```
duncathan, line 57: this loses a lot of specificity for CS because CS already has a category with these details which was treated as separate from the Nothing hint category

Miepee: oh, wasn't aware of that. So CS should instead have another text for the hint details instead, so that "a useless item" can keep being used for what its used for in CS right now, and hten the other one for the nothing/empty items?

henriquegemignani:
Proper solution would be to make pickups for the nothings.

Miepee:
That would also allow saving non-etms into the rdvgames, right?

Miepee:
Gonna close this then, as the proper solution would be completely different anyway and able to fix more things.

PR #7309, opened 2024-10-06
Prime: add combat visor to essence

JeffGainsNGames:


Samuel6710 (approved):

JeffGainsNGames:
While I'm here, any interest in removing the "kill metroids requirement" since you have phazon beam?
image

toasterparty (approved):

CHANGELOG.md
```diff
@@ -53,6 +53,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Logic Database
 
+##### Impact Crater
+
+- Added: In Metroid Prime Lair: Require combat visor for Essence fight.
```
Miepee, line 58:
```suggestion
- Added: In Metroid Prime Lair: Require Combat Visor for Essence fight.
```

Miepee:
> While I'm here, any interest in removing the "kill metroids requirement" since you have phazon beam?

I'm not a prime maintainer, but i would say go for it. My only concern would be: what happens if you somehow get caught by a metroid before reaching the pool? Does it automatically die when you later reach it? Or do you need to get it off with bombs?

Uncle-Reggie:
> > While I'm here, any interest in removing the "kill metroids requirement" since you have phazon beam?
> 
> I'm not a prime maintainer, but i would say go for it. My only concern would be: what happens if you somehow get caught by a metroid before reaching the pool? Does it automatically die when you later reach it? Or do you need to get it off with bombs?

The Metroids must be bombed off or Power Bombed to death. There needs to be some kind of requirement for killing them.

Miepee:
Let's leave the requrements as they are then right now and readjust them later.

Samuel6710:
> The Metroids must be bombed off or Power Bombed to death. There needs to be some kind of requirement for killing them.

Not exactly, metroids will latch off of you after enough time has passed. You can have health requirements as an alternative to bombs or power bombs.

PR #7310, opened 2024-10-06
Fusion: misc logic tweaks

Miepee:
Went through a few generation logs and adjusted things based on what I was seeing.

### DIFF
--- a/randovania/games/fusion/logic_database/Main Deck.txt
+++ b/randovania/games/fusion/logic_database/Main Deck.txt
@@ -1080,12 +1080,14 @@ Extra - room_id: [38]
   * Extra - blockx: 53
   * Extra - blocky: 10
   > Beside Door
-      Trivial
+      Wall Jump (Beginner) or Enabled Generator Workaround
 
 > Door to Crew Quarters East; Heals? False
   * Layers: default
   * L0 Hatch to Crew Quarters East/Door to Arachnus Arena
   * Extra - door_idx: (88,)
+  > Arena
+      Enabled Generator Workaround
   > Beside Door
       Trivial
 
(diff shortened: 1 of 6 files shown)


randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -332,7 +332,7 @@ Extra - room_id: [6]
               Any of the following:
                   After Sector 4 Water Level Lowered
                   # TODO: damage values
```
JeffGainsNGames, line 334: remove todo

randovania/games/fusion/logic_database/Main Deck.txt
```diff
@@ -1109,7 +1111,9 @@ Extra - room_id: [38]
   * Open Passage to The Attic/Other to Arachnus Arena
   * Extra - door_idx: (200,)
   > Beside Door
-      Missiles
```
JeffGainsNGames, line 1112: can you briefly explain what this is all doing for my awareness.

Miepee: Going to arachnus: makes this not a double unsafe action (first generic node you loose access to the door, going down makes you loose access to the generic node). it boosts arachnnus' weight from ~0.3 to1.2 or something
e + attic to generic node: nothing really. just makes it clear that the reason these are currently trivial and not WJ beginner is because of the generator seeing these as unsafe actions otherwise.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
Miepee: removed

JeffGainsNGames (approved):

PR #7312, opened 2024-10-06
Fusion: tweak etank weight

Miepee:
Causes etanks to be spread out more and not dumped all at the end when the generator realized it needs access to dna/bosses.

### DIFF
--- a/randovania/games/fusion/pickup_database/pickup-database.json
+++ b/randovania/games/fusion/pickup_database/pickup-database.json
@@ -519,7 +519,9 @@
             "custom_count_for_shuffled_case": 20,
             "extra": {
                 "StartingItemCategory": "Energy"
-            }
+            },
+            "probability_offset": 2.0,
+            "probability_multiplier": 2.0
         },
         "Level 0 Locks": {
             "pickup_category": "misc",


JeffGainsNGames:
why 2.0 for each setting? I understand a bit how it works, offset being an absolute increase in weight and multiplier being relative, but curious why those numbers were chosen and if any checks were done to understand the change?

Miepee:
Mostly just cause the results looked ok.

There are a lot of cases in the current situation, where the default weight for etanks is 0. 0 being multiplied by 2 still results in 0. and then the offset makes it back up to 2.
Most of the times, the etanks are combined to something like:
[E, E, E, E, Missile, Missile, Missile ,Missile], which in the current formula would be something along base_weight * [sum(each individual multipler) / length] + [sum(each individual offset) / length]. Which in most cases just is 0 (because boss is unsafe) * doesnt_matter + [(2 + 2 + 2 + 2 + 0 + 0 + 0 + 0) /8] = 0.5

If there ends up being an etank being placed individually, then its because for some individual required expected damage path we have (e.g. getting past silo scaffoling) in which case the weight is just gonna be at 5 or so, which I'm fine with. P much all our required expected damage are lowish, and giving E to players soonish is not a bad thing for fusion IMO.

JeffGainsNGames:
thanks for the rundown, let's see how it plays out. can always tweak it later during testing.

JeffGainsNGames (approved):

Miepee:
The main important thing for tweaking is, that you want to look at the generation log to see how it affects things.

PR #7313, opened 2024-10-06
Allow tweaking multipliers for ammo

Miepee:
Fusion (and other games maybe too?) would benefit from changing missile tank weights.
I don't know if this needs a migration


henriquegemignani (approved):
Migration is technically needed to properly indicate that the file format changed schema and older versions of the app aren't fully compatible.

But we only care about migration for these formats to easily support people who have games on branches or whatever.

PR #7316, opened 2024-10-06
Move GameData and similar classes into a `randovania.game` module.

henriquegemignani:


henriquegemignani:
This actually tells me `GameWebInfo` should just be a required argument for `GameData` with no default.

Miepee (approved):

PR #7317, opened 2024-10-06
finish ARCold runs

SchwartzGandhi:


randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -283,16 +309,24 @@ Extra - room_id: [8]
...
   * Extra - door_idx: (64,)
   > Door to Arctic Containment
-      # TODO: varia / damage runs
-      Varia Suit and Can Kill PB Geron
+      All of the following:
+          Can Kill PB Geron
+          Any of the following:
+              Varia Suit
+              # Damage Reqs
+              Damage Runs (Advanced) and Cold Damage ≥ 180
```
Miepee, line 329: how come t he 80 damage discrepancy between the directions?

randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -888,8 +929,14 @@ Extra - room_id: [26]
...
               Movement (Expert)
-          # Damage Runs
-          Varia Suit
+          Any of the following:
+              # Damage Reqs
+              Varia Suit
+              All of the following:
+                  # No springball: https://youtu.be/a4urCWRPTNM
+                  Damage Runs (Hypermode) and Cold Damage ≥ 733
+                  Mid-Air Morph (Beginner) or Can Use Bombs
```
Miepee, line 938: this is both mid air morph and bombs, no?
Also wonder whether it should really be hypermode. conceptually its not *that* much different from doing it with spring.

randovania/games/fusion/logic_database/Sector 5 ARC.txt
SchwartzGandhi: this is due to you you tanking more hits from the zeela

randovania/games/fusion/logic_database/Sector 5 ARC.txt
SchwartzGandhi: > this is both mid air morph and bombs, no?

oops i didnt think about that i guess ill just remove mid air morph then since that since you need bombs to do this.

> Also wonder whether it should really be hypermode. conceptually its not that much different from doing it with spring.

bombs take significantly longer to detonate than to use springball, and in that time difference is a knockback from the cold damage, which makes it worth the separatation imo

JeffGainsNGames: yeah I'd likely agree with timing bombs between cold hits makes it harder rather than the instantaneous springball (not that I have/will try it)

randovania/games/fusion/logic_database/Sector 5 ARC.txt
Miepee: why not just kill the zeelas via shooting? does that not save health over running into them?

SchwartzGandhi: ok trying it out i do get lower numbers, but this connection is a little weird to test with no geron

Miepee (approved):

PR #7327, opened 2024-10-07
Fusion: Serris balances

JeffGainsNGames:


randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1971,17 +1971,16 @@ Extra - room_id: [42]
   > Event - Serris
       All of the following:
           Any of the following:
-              # Serris Weapon Requirements
+              # Serris Weapon Requirements (50HP)
               Charge Beam or Missiles ≥ 10
               Missiles ≥ 5 and 20 DMG Missiles
```
Miepee, line 1976: not missing missiles should definitely not be trickless

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1971,17 +1971,16 @@ Extra - room_id: [42]
...
           # Core-X Requirements
           Missiles
           Any of the following:
               # Health Requirements
-              Combat (Expert) or Normal Damage ≥ 400
-              Combat (Beginner) and Normal Damage ≥ 300
+              Combat (Advanced) or Normal Damage ≥ 300
```
Miepee, line 1981: link the vid: https://youtu.be/1M1GCCa4jc0

Advanced is fine considering that a blind spot exists where you literally cant get hit.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
JeffGainsNGames: bumped the missiles a bit, I don't think it's worth getting more granular than this?

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1971,17 +1971,18 @@ Extra - room_id: [42]
   > Event - Serris
       All of the following:
           Any of the following:
-              # Serris Weapon Requirements
-              Charge Beam or Missiles ≥ 10
+              # Serris Weapon Requirements (50HP)
+              Charge Beam or Missiles ≥ 15
               Missiles ≥ 5 and 20 DMG Missiles
```
Miepee, line 1976: IMO:
- 10 dmg + 20
- 20 dmg + 10
- combat advanced + 
  - 10 dmg + 10
  - 20 dmg + 5
  
 basically what i did for box.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1971,17 +1971,23 @@ Extra - room_id: [42]
   > Event - Serris
       All of the following:
           Any of the following:
-              # Serris Weapon Requirements
-              Charge Beam or Missiles ≥ 10
-              Missiles ≥ 5 and 20 DMG Missiles
+              # Serris Weapon Requirements (50HP)
+              Charge Beam or Missiles ≥ 20
```
Miepee, line 1975: this missile should have 10 dmg missile just in case

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1971,17 +1971,23 @@ Extra - room_id: [42]
...
+              Missiles ≥ 10 and 20 DMG Missiles
+              All of the following:
+                  Combat (Advanced)
+                  Any of the following:
+                      Missiles ≥ 10
+                      Missiles ≥ 5 and 10 DMG Missiles
```
Miepee, line 1981: missing 10 dmg missile, the second one should be 20 dmg missile

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1971,17 +1971,23 @@ Extra - room_id: [42]
...
           # Core-X Requirements
           Missiles
           Any of the following:
               # Health Requirements
-              Combat (Expert) or Normal Damage ≥ 400
-              Combat (Beginner) and Normal Damage ≥ 300
+              Normal Damage ≥ 300
+              Combat (Beginner) and Normal Damage ≥ 200
               Combat (Intermediate) and Normal Damage ≥ 150
```
Miepee, line 1988: since serris appears before box, and you have after serris more chances to find more etanks in vanilla, thoughts on bumping these down to 250, 180, 110 instead?
(after multiplier these would be 375, 270, 165)

randovania/games/fusion/logic_database/Sector 4 AQA.txt
JeffGainsNGames: this is for advanced logic, not missing any shots. Don't see why we we need it to be 10x10DMG and 5x20DMG >> 50HP.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
JeffGainsNGames: 20x5DMG missiles is 100DMG

Miepee: because of the lack of dmg specifier i didnt even know that you implicitly referenced 5dmg here, especially because we use this nowhere else in the logic db. I don't like that. Please change this in some way to make this either more obvious, or hard require 10 dmg missiles

randovania/games/fusion/logic_database/Sector 4 AQA.txt
Miepee: see above comment.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
JeffGainsNGames: We seem to use it in a few places?
image
image

JeffGainsNGames: and it doesn't even need to specifically be 5dmg, it's just the minimum possible. If you have any other missile upgrades it would and X number of missiles it also satisfies the requirement. just 5DMG is what needs to be considered for the calculation

Miepee: It just feels very inconsistent with how we've been doing all the other boss/combat logic (outside of templates)

JeffGainsNGames: so what.. create a template for 5DMG missile and just make it =diffusion?? it seems so extra an unnecessary

JeffGainsNGames: or would you rather just say no boss is expected to be killed with 5DMG/diffusion missiles and only ever use 10DMG and above templates for them?

Miepee: I prefer seeing an explicit "5 DMG Missiles" over "okay, how much damage did diffusion-only do again?"

We currently have no plans to have the possibility to use anything lower than 10 dmg missiles. Even with the future "any missile upgrade gives ability to fire missiles" option. Hence why basically no other combat logic bothers with it, and my suggestion of "we can just hard require 10 dmg missiles here"
And if we do change the plans, then we need to change quite a few places in the DB anyway, one more to the pile doesn't hurt.

Miepee: So yes, id prefer for now if we only use 10dmg+ for bosses.

JeffGainsNGames: > I prefer seeing an explicit "5 DMG Missiles" over "okay, how much damage did diffusion-only do again?"
you don't need to do this though at all, you just need to say "hey I need 20 missiles regardless of dmg" which is pretty simple to understand...

I'm done arguing I'll just concede.

Miepee (approved):

PR #7350, opened 2024-10-11
Fusion: Varia core x rework

Miepee:


randovania/games/fusion/logic_database/Sector 6 NOC.txt
```diff
@@ -552,14 +552,17 @@ Extra - room_id: [13]
...
+              Normal Damage ≥ 315
               # Fight Boss Damageless: https://youtu.be/CfgKTcobo0s
               Combat (Hypermode)
-              Combat (Beginner) and Normal Damage ≥ 400
-              Combat (Intermediate) and Normal Damage ≥ 200
-              Combat (Advanced) and Normal Damage ≥ 100
+              Combat (Beginner) and Normal Damage ≥ 250
+              Combat (Intermediate) and Normal Damage ≥ 180
+              Combat (Advanced) and Normal Damage ≥ 120
+              Combat (Expert) and Normal Damage ≥ 60
```
JeffGainsNGames, line 565: I feel like damageless could just be expert honestly

randovania/games/fusion/logic_database/Sector 6 NOC.txt
```diff
@@ -552,14 +552,17 @@ Extra - room_id: [13]
               # Core-X Requirements
               Missiles
               Combat (Beginner) or 20 DMG Missiles
+              # A way to dodge
+              Combat (Intermediate) or Have Any Jump Upgrade
```
JeffGainsNGames, line 556: I guess we could have something like this, charge beam does do an insane knockback though

SchwartzGandhi, line 556: is this a thing for every core-x? or is this just for mega core?

randovania/games/fusion/logic_database/Sector 6 NOC.txt
Miepee: I thought so too, but wanted to stay on the safe side here, because
1. AFAIK there's no consistent way on getting the big varia-mass damageless
2. If you mess up once anywhere during this whole 4 minute fight on min reqs, then you will have to redo the whole thing from wherever your last save was (not necessarily the save right before/after varia core)

randovania/games/fusion/logic_database/Sector 6 NOC.txt
Miepee: This is for the core-x. Charge does not do anything to it.

Miepee: this is only for the core-x of mega-core. It's very fast and if people don't know how to dodge it (which isnt exactly obvious), will get hit.

randovania/games/fusion/logic_database/Sector 6 NOC.txt
JeffGainsNGames: seems like you should save if you are going to fight this boss at 0 Etanks just saying
I dont care enough to block because of this just seems extra

Miepee: the last comment was less implying "forgot to save" and more "saves could be blocked and backtrack is huge"

SchwartzGandhi: imo distance from a save station shouldn't factor into how difficult a trick should be, rather just the fact you can die/softlock is a good enough indicator (in general).
as for this connection, I agree with Jeff that this is fine as is

SchwartzGandhi (approved):

PR #7356, opened 2024-10-12
AM2R: slight gui alignment adjustments

Miepee:


Miepee:
> at least say what has been changed.

I did, it's in the PR title. 
But now thinking of it, it's actually *spacing* adjustments and not alignment. Which is now in the changelog which I forgot to initially add, thanks.

before:
grafik

after:
grafik

This is an example of another entry:
grafik

PR #7359, opened 2024-10-12
Remove citra requirement for multiworld

ThanatosGit:
i leave it here for later.
lot of things needs to be done:
- ~~remove exheader selection~~
- ~~remove region selection~~
- ~~adjust input selection to check for correct file~~
- osrr bump, which requires meds bump
- ~~migration~~
- ~~fix Tests~~
- fix Guides
- ~~fix ui~~
image
image
image


randovania/games/samus_returns/gui/dialog/msr_connector_prompt_dialog.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class MSRConnectorPromptDialog(ConnectorPromptDialog):
+    def __init__(self, **kwargs: dict[str, Any]) -> None:
+        super().__init__(**kwargs)
+
+        self.top_radio.setText("Citra")
+        self.top_label.setText("Connects to a Citra running on this computer.")
+
+        self.ip_radio.setText("Luma3DS")
```
Miepee, line 15: Probably not worth specifying, but theoretically this would also allow connecting to a remote citra, yes?

Miepee (approved):
Looks good to me.

How do you plan to merge these together with the ODR bump? Do you want to make merge May's commits into this branch?
Otherwise we could also probably abuse owner privileges and merge them directly, ignoring the required tests, but I'd prefer to not do that.

henriquegemignani (approved):

PR #7362, opened 2024-10-13
Fusion: slight template cleanup

Miepee:


randovania/games/fusion/logic_database/header.txt
```diff
@@ -86,7 +86,7 @@ Templates
 * 10 DMG Missiles:
       All of the following:
           Missiles
-          Ice Missile Data or Missile Data
+          Ice Missile Data or Missile Data or Super Missile Data
```
SchwartzGandhi, line 89: super missiles do 20 damage and shouldn't be in here

randovania/games/fusion/logic_database/header.txt
```diff
@@ -111,8 +111,9 @@ Templates
 
 * Can Kill Tough Beam-Resistant Enemy:
       Any of the following:
-          Missiles ≥ 10 or Screw Attack or Can Use Power Bombs
+          Missiles ≥ 10 or Screw Attack
           Charge Beam and Have Any Beam Upgrade
+          Power Bombs ≥ 4 and Can Use Power Bombs
```
SchwartzGandhi, line 116: do we need this much cushioning for PBs?

SchwartzGandhi (changes requested):

randovania/games/fusion/logic_database/header.txt
Miepee: 20 > 10

randovania/games/fusion/logic_database/header.txt
Miepee: I would prefer to have more than less cushioning. The amount of how many PBs enemies take can vary a lot, and having a little cushioning avoids situations where youre expected to go through multiple strong enemies in a row with only 1 pb.

randovania/games/fusion/logic_database/header.txt
Miepee: to clarify: yes, they absolutely should be in here. we have in multiple places requirements with something like "10 missiles & 10 dmg or 1 missile & 45 dmg". A super missile would, because 20 is greater than 10, fullfill the first condition. but if we never list supers, then they never will be logically considered.
in other words, our templates are "deal at least x dmg" not "deal exactly x dmg"

randovania/games/fusion/logic_database/header.txt
SchwartzGandhi: Fair enough, and to tell the truth, using PBs everywhere to kill tough enemies isnt really fun imo, so I'm fine with this.

randovania/games/fusion/logic_database/header.txt
SchwartzGandhi: I would like the template to be renamed to clarify this, something like `10+ DMG Missiles` perhaps?

Miepee: sure

Miepee: only renamed this one, as 20, 30 and 45 are always exact

SchwartzGandhi (approved):

PR #7366, opened 2024-10-13
CLI benachmark: print minimums and maximums

Miepee:
```
      Results | Fixes | Failures |     Mean |    Stdev |   Median |  Minimum |  Maximum
Metroid Prime |     0 |        0 |    4.533 |    0.693 |    4.564 |    3.250 |    6.381

```


henriquegemignani:
The minimum and maximum generation time are not useful information when benchmarking and optimisation generation times. Having these values are purely a curiosity, but are likely distractions from what you should be looking at. I believe having these values be printed is a net negative to the usability of this tool.

PR #7370, opened 2024-10-13
Support seed count for benchmarks

Miepee:


henriquegemignani:
IMO no. This command is intended for the CI job and nothing else. Focus improvements in the batch generate one.

henriquegemignani:
Actually it's a different command. But I remember there being something else wonky with this command 🤔

Miepee:
If you remember what was wonky with this, that'd be nice to know.

henriquegemignani:
Afaik, having to decode/encode permalinks to the json file quite so often is ridiculously slow.
The json file is meant to be preserved to re-run in the future, but it's not quite flexible/future proof enough.

Miepee:
I realistically don't see usage above 1k at least. For most games that's already way plenty to get a good measurement.

henriquegemignani (approved):

PR #7371, opened 2024-10-14
Fixed "QSlider" capitalization in UI custom widget definition

quinnbbowman:
Fixed custom scroll protected slider "QSlider" import in UI files from the previous PR.  I don't know enough to say what it broke to have the wrong uncapitalized class extended, or why it functions anyway, but this fixes it :)


Miepee (approved):
ty

PR #7382, opened 2024-10-15
Fusion: fix door id in s2 central shaft

JeffGainsNGames:
image

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.txt
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.txt
@@ -460,7 +460,7 @@ Extra - room_id: [13, 46]
 > Door to Ripper Tower; Heals? False
   * Layers: default
   * Open Hatch to Ripper Tower/Door to Central Shaft
-  * Extra - door_idx: (31,)
+  * Extra - door_idx: (113,)
   > Beside Drop
       Morph Ball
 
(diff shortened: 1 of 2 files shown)


randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -460,7 +460,7 @@ Extra - room_id: [13, 46]
 > Door to Ripper Tower; Heals? False
   * Layers: default
   * Open Hatch to Ripper Tower/Door to Central Shaft
-  * Extra - door_idx: (31,)
+  * Extra - door_idx: (113,)
```
Miepee, line 463: history says that at one point, it was
grafik
 this:

JeffGainsNGames: that's from combining the room states. Since the other room isn't used is there any point to add it back or just raising the comment for history sake

Miepee: forgot that we didnt use the other room anymore.

Miepee (approved):

randovania/games/fusion/logic_database/Sector 2 TRO.txt
JeffGainsNGames: image
the door doesn't even exist in this state

PR #7388, opened 2024-10-16
More SRX stuff

SchwartzGandhi:


randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -830,6 +830,9 @@ Extra - room_id: [21]
   > Door to Ripper Maze Access
       Trivial
 
+> Event- Break Missile Blocks; Heals? False
+  * Layers: default
```
Miepee, line 834: this node seems useless? nothing connects to it, it connects to nothing else. 
it also isn't really necessary.


What *would* be useful, is an event for the shoot blocks that lead to animorphs.

Miepee (approved):

PR #7404, opened 2024-10-18
Mypy: override typing for generic generated gui module

Miepee:
We don't want these typed either.


henriquegemignani:
Ideally we could generate some `.pyi` files for these.

henriquegemignani (approved):

Miepee:
Didn't you look into tooling for that at one point?

henriquegemignani:
> Didn't you look into tooling for that at one point?

Yeah and someone shared one they've made, but I didn't go further.

PR #7405, opened 2024-10-18
DB: Add method to allow game specific db integrity checks

### LINKED ISSUE #4495: Ensure that certain resources always are used with a requirement
Some resources like Bombs, Shinesparking etc., all have certain requirements that one needs to fullfill before using them. In the case of Bombs, it's Morph, shinesparking requires speedbooster and so on.

I implemented a small test for am2r that checks some of am2r's resources for that, but it would be a good idea to have that for other games as well.
One approach would be to simply abstract that test I've written a bit more, and add the other games' data to it.
Another approach would be to have an `implied_requirements` field in the resource DB. This would make the connections much cleaner, as one doesn't have to have create a dozen templates for all the condition prerequisites.

Miepee:
Fixes #4495


randovania/db_integrity/integrity.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def items_to_be_replaced_by_templates(
+    game: GameDescription, context: NodeContext, items_to_templates: dict[str, str]
+) -> Iterator[str]:
+    for source_node in game.region_list.iterate_nodes():
+        for destination_node, req in game.region_list.potential_nodes_from(source_node, context):
+            for resource, template in items_to_templates.items():
+                if does_requirement_contain_resource(req, resource):
+                    yield (
```
henriquegemignani, line 49: Please add tests for this file, and docstrings. And add to the mypy list

randovania/db_integrity/integrity.py
```diff
@@ -0,0 +1,67 @@
+from collections.abc import Iterator
```
henriquegemignani, line 1: Does this file need to be in a new module?

henriquegemignani (changes requested):

randovania/db_integrity/integrity.py
Miepee: probably not, didnt know where else to put it.
logic_database sounds wrong cause thats not a code file
 layout?

henriquegemignani: game_description? There's literally `integrity_check.py` there.

Miepee: and also create game_description for within `games` then too?

henriquegemignani: does it need to be a new folder?

Miepee: IDK putting it right into games felt kinda weird to me, but I guess that worsk.

randovania/db_integrity/integrity.py
Miepee: Done.

henriquegemignani (approved):

PR #7406, opened 2024-10-18
Chore: remove code from pickup db folder

Miepee:


randovania/game_connection/connector/am2r_remote_connector.py
```diff
@@ -3,6 +3,7 @@
 from collections import defaultdict
 from typing import TYPE_CHECKING
 
+from games.am2r.layout import progressive_items
```
henriquegemignani, line 6: that's just wrong :D

Miepee: my pycharm setup is weird woops

henriquegemignani (approved):

PR #7410, opened 2024-10-20
Dread/ferenia cold room requirements

Tyranisaur:


Tyranisaur:
In order to avoid merge conflicts, this is draft until #7380 gets merged.

randovania/games/dread/logic_database/Ferenia.txt
```diff
@@ -2049,6 +2047,24 @@ Extra - asset_id: collision_camera_030
               Gravity Suit
               Heat/Cold Runs (Advanced) and Cold Damage ≥ 50
 
+> Room Center; Heals? False
+  * Layers: default
+  > Dock to Wave Beam Tutorial
+      All of the following:
+          Screw Attack and After Ferenia - Storm Missile Gate Cold and Lay Any Bomb
```
XenoWars, line 2054: This does not need bombs.

randovania/games/dread/logic_database/Ferenia.txt
```diff
@@ -2049,6 +2047,24 @@ Extra - asset_id: collision_camera_030
...
+          Any of the following:
+              Use Spin Boost
+              # https://youtu.be/RMmofIi1N3k
+              Wall Jump (Beginner)
+  > Door to Cold Room (Energy Recharge Station) (Charge)
+      All of the following:
+          Screw Attack
+          Any of the following:
+              Lay Small Bomb
+              Power Bombs ≥ 2 and Lay Power Bomb
```
XenoWars, line 2064: Commenting here since it is related to something I can't place a comment with. If you want this to be effective, you should remove the connection to it from Dock to Wave Beam Tutorial and connect that to Room Center instead.

XenoWars (changes requested):

randovania/games/dread/logic_database/Ferenia.txt
Tyranisaur: How so? The point is that to break the gate and then go back to the energy recharge station with power bombs, you need to break the bomb blocks twice without an ammo refill. But any time you get across the whole room you only need one PB ammo. If I removed the connection directly from the dock to wave beam tutorial to energy recharge station in favor of going through the room center node, then it would take 2 ammo to go right to left.

XenoWars: I'm thinking of a situation where you can get to the dock and back, but not do anything beyond the dock. There are no pb refills beyond the dock, so it could very well see the logic of going to the dock then going back with one PB but you are trapped because you used the one PB already.

Tyranisaur: Oh but there are Autools in the Wave Beam Tutorial. Sure we should ideally make sure you actually have enough weaponry to destroy them. I suppose there's also an Autsharp in the Energy Recharge Station that you would need to get a pb drop from when going the other way with minimal ammo. Either way, this problem isn't any worse than it already is.

XenoWars: That autsharp you speak of is the one at the Plasma door right? That makes me think, that you need to use 2 PBs to even get to the storm gate, so shouldn't it require 3 to get out? I don't know the stance on using Autsharps and Autools to refill on consumables especially since logic only knows of ammo and total recharge station currently. I did think the Wave Beam Tutorial did not have any, though, so that's my mistake. Getting back with the drops might be a bit difficult though due to the EMMI being close behind.

Miepee: > logic only knows of ammo and total recharge station currently.

Not sure if thats what you meant, but logic assumes that ammo is *never* spend. In other words, it will never assume you need to refill ammo, regardless whether you can do so at enemies, recharge stations, a magic trigger that refills you or whatever else.

Tyranisaur: We don't handle ammo in logic the same way as health (yet), so ammo logic is just a capacity check, and we have to make some sort of best effort to handle ammo requirements based on that, which leads to these complicated considerations for "there and back again" and multi room situations. I don't know that we have any guidelines or existing practice for what to expect for refilling ammo, or health for that matter, from defeating enemies.

Though, shortly after I wrote the above message, I realized that in this situation you have screw attack, so I think for this case I think it should be a safe assumption that the enemies can be defeated.

XenoWars: > Though, shortly after I wrote the above message, I realized that in this situation you have screw attack, so I think for this case I think it should be a safe assumption that the enemies can be defeated.

There is still the issue of getting back with the drops with the EMMI essentially right behind you, and the point of needing 3 to get out instead of 2.

Tyranisaur: As far as I can tell, you can reset to checkpoint at the EMMI door, which appears to cause the Autsnipers in to reappear, which also is a source of ammo on top of the Autools. I'll acknowledge that if the EMMI has been defeated then there isn't a checkoint at those doors, but unless we do some wile shenanigans in the future, then you need to have used a bomb launcher, so ammo for the bomb blocks shouldn't be a concern at that point.

But also, in terms of the bomb blocks in the Energy Recharge room, the Autsharp spawns to the right of the block, so it seems fair that we don't need to account for extra ammo for those blocks.

Tyranisaur: For reference, see this map image:
image
Green circles are bomb blocks and red circles as enemies. The top left enemy respawns indefitively. The other ones respawn after a reset to checkpoint to the EMMI door.

Before, all of the relevant connections required 1 pb ammo (when using pbs instead of other bombs). What I did was to change it so that moving from the room center to the left now requires 2 ammo, but I've left all the other connections as is, including keeping the connection directly from the EMMI door to the to the Charge door, which is the highlighted connection.

XenoWars: The Autools are on the other side of the wall.

XenoWars: If the Autsnipers and Autsharps guarantee PB drops, though, it is probably fine. You wouldn't even have to reload the area to get the drop from the snipers after opening the gate.

Tyranisaur: Well you need to reload the room because they get destroyed when you break  the blocks with power bombs coming through the first time.

I could go either way on this, but I still think it's at least not worse than it already was. Given that power bombs normally give you 2 ammo to start with, I think it's fair that we can make some assumptions that people who choose to play with only one would be some levels of resourceful over the bare minimum.

XenoWars (approved):

PR #7437, opened 2024-10-23
Visualizer: Save door overrides when editing node

Miepee:
Makes it not annoying anymore to change descriptions of door nodes that have overrides.


henriquegemignani (commented):
Won't this break when changing a node into a dock node?

Miepee:
it indeed breaks.

henriquegemignani (approved):

PR #7443, opened 2024-10-23
Randomprime v1.28.1

toasterparty:
Technically out of spec with the JSON schema by omitting the new `startingItems`, but the defaults are fine and it'll be way more convenient to fix in the custom items PR

### DIFF
--- a/randovania/games/prime1/exporter/patch_data_factory.py
+++ b/randovania/games/prime1/exporter/patch_data_factory.py
@@ -955,6 +955,11 @@ class PrimePatchDataFactory(PatchDataFactory):
         else:
             boss_sizes = {}
 
+        if self.configuration.spring_ball:
+            spring_ball_item = "Morph Ball Bomb"
+        else:
+            spring_ball_item = "Spring Ball"
+
         data: dict = {
             "$schema": "https://randovania.github.io/randomprime/randomprime.schema.json",
             "seed": self.description.get_seed_for_player(self.players_config.player_index),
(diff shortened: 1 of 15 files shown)


randovania/games/prime1/exporter/patch_data_factory.py
```diff
@@ -955,6 +955,11 @@ def get_random_size(minimum, maximum):
         else:
             boss_sizes = {}
 
+        if self.configuration.spring_ball:
+            spring_ball_item = "Morph Ball Bomb"
+        else:
+            spring_ball_item = "Spring Ball"
+
```
Miepee, line 962: 1. Spring ball is a custom/independant item that can be collected now, yes?
2. If the item is set to bombs, does it mean that collecting bombs automatically grants you the ability to use spring, or do you still need the custom spring item in addition?

toasterparty: Read this code as:

> By default, you must collect the `Spring Ball` item in order to use Spring Ball, unless the user has enabled the Spring Ball option, in which case Spring Ball functionality will be granted upon acquisition of `Morph Ball Bomb`

Miepee: Gotcha. My main confusion was from the doc where it wasn't quite clear to me whether it meant "Being able to jump in ball means collecting spring ball and morph ball bombs" or "once you collected morph ball bombs, you get the ability to jump in ball"

Miepee (approved):

PR #7448, opened 2024-10-24
Fusion: Sector 6 Documentation

Miepee:


JeffGainsNGames (approved):
I feel like the reason we left blue X as normal dmg was so if cold environmental dmg was modifiable it would need to be a different dmg type. Otherwise fine with it being changed to cold dmg since both are negated by varia.

Miepee:
I thought internally they're both cold damage?
But if that causes issues in the future ig we can always adjust.

PR #7450, opened 2024-10-24
Better Multiworld heuristic

Miepee:
Someone can have a Solo MW going, with only one world, which should be considered an MW.

### DIFF
--- a/randovania/interface_common/players_configuration.py
+++ b/randovania/interface_common/players_configuration.py
@@ -22,4 +22,4 @@ class PlayersConfiguration:
 
     @property
     def is_multiworld(self) -> int:
-        return len(self.player_names) > 1
+        return self.get_own_uuid() != INVALID_UUID


henriquegemignani (changes requested):
It seems every single HintDetailsTab uses this property, but the `GameDetailsWindow` does not have access to the UUIDs, nor is it possible to always have it since you can open that windows from just an rdvgame.

Also, main thing this would do is cause hints to have your own world name and your pickups to have the `Your` prefix. Not sure this is your intention.

Miepee:
Maybe testing locally a little instead of assuming everything will be fine, would've been a good idea xP

duncathan:
> which should be considered an MW.

mind explaining why?

Miepee:
If it was generated from the Multiworld tab/browser/session, then the user expectation is that whatever is generated is considered an MW, regardless of how many worlds.
Currently this can have weird affects on dread for example, because it has an option to export without connector capabilities. A user would expect that if generated/exported from there, that the connector would be forced on.

MayberryZoom:
What would be the point in forcing the connector on in that case? It doesn't actually do anything necessary that I can think of.

Miepee:
Usually people use it for testing that their MW setup works. So that if they do have issues for some reason, they find it during their 1-world testing session, rather than in a scenario where everyone is waiting on them.
Which usually involves seeing that item/location tracking works.

MayberryZoom:
I thought local items don't get logged though? I think for Dread they show up in the items collected, but not in the log of items sent if I remember correctly.

It's definitely still useful so you can see the In-Game/Title Screen indicator which should show that the connection is working though.

Miepee:
The session shows the % of items collected in your game, and for Dread that includes items regardless whether local or offworld. But only offworld show up in the session history.

MayberryZoom:
Yeah, that's what I was trying to say, just worded poorly haha

duncathan:
> Currently this can have weird affects on dread for example, because it has an option to export without connector capabilities. A user would expect that if generated/exported from there, that the connector would be forced on.

if this is the only reason to do this, we should just force the connector no matter what. there's not really any reason for it to be optional any more

curious whether co-op affects this PR in any way tho

Miepee:
> curious whether co-op affects this PR in any way tho

Only by creating a conflict. Co-op always needs to force multis on, so that single-world coop sessions can properly function.

> we should just force the connector no matter what

Was talking with Dyceron about this a while ago too and agreed there. Citra/Ryujinx don't properly handle stopping/restarting emulation with it being active but closing the emulator first and reopening the emu trivial and IMO outweighs the confusion we have over people having the option accidentally turned off. Especially since on certain OS/games/gpu's, the emu already can't handle stopping/restarting properly anyway at times.


There are a few other things that would still change if a single-world in an MW session would be considered MW, mainly hints/credits/pickup descriptions would always have the player name. I would personally prefer that solely for consistency reasons but also wouldn't loose sleep over it if we decide against it.

duncathan:
> There are a few other things that would still change if a single-world in an MW session would be considered MW, mainly hints/credits/pickup descriptions would always have the player name. I would personally prefer that solely for consistency reasons but also wouldn't loose sleep over it if we decide against it.

I prefer the consistency of single player looking like single player than your suggested consistency of single player sometimes looking like multiplayer :P

> Forcing the connector to be always on is not an option for MSR. If you export with the connector on Luma and have wifi disabled or simply currently no connection to an acces point, it will not even boot and run straight into a crash screen. Maybe this can be fixed in the remote connector, but given the randomness of the errors with the current injection technique, I don't know if I'm ever willing to waste my time on it.

Fair enough. The current setup hasn't really caused problems tho so I'd rather just leave it be if need be haha

Miepee:
> your suggested consistency of single player sometimes looking like multiplayer

My suggestion is more "have the **multiworld** session window always generate multiworld seeds, rather than it sometimes generating single player seeds"

Miepee:
Closing for now.
The issue doesn't exist for Samus Returns anymore, and Dread has plans to fix it as well.
Avoids having the weirdness with hints/opening spoiler logs

PR #7459, opened 2024-10-27
Prime: Fix mismatching screen brightness

Miepee:
Technically a bug fix, but since it has existed for ages, I don't really care if it gets moved into the next release.


Miepee:
rebased

PR #7460, opened 2024-10-27
Better handle incompatible ISOs for reading Prime 1/2 ISOs

henriquegemignani:


Miepee (commented):
This feels like something that should be done via the "throw user error exception when inputs are known to have bad hash"

henriquegemignani:
Oh I misread the code. When a `UnableToExportError` it _doesn't_ hash the inputs. Ugh

PR #7465, opened 2024-10-28
Visually and audibly alert user when a game has been generated

Nystrata:
Addresses https://github.com/randovania/randovania/issues/7008

Use OS features to visually alert (using [QApplication::alert](https://doc.qt.io/qt-5/qapplication.html#alert)) and audibly alert (using [QApplication::beep](https://doc.qt.io/qt-5/qapplication.html#beep)) users when a game has been generated for both solo and multiworld games.    
Audible system beep alert can be toggled in the Preferences menu.  

(Edit, new image!) 
image


randovania/gui/lib/common_qt_lib.py
```diff
@@ -294,3 +296,9 @@ def set_icon_data_paths(label: QtWidgets.QLabel):
     repl = rf'<img src="{get_data_path().as_posix()}/\g<1>"/>'
     new_text = image_pattern.sub(repl, label.text())
     label.setText(new_text)
+
+
+def alert_user(parent: QtWidgets, options: Options):
+    QtWidgets.QApplication.alert(parent)
+    if not parent._options.disable_sound_alerts:
```
henriquegemignani, line 303: The intent of the `options` argument is that you don't need to find it in an undocumented place in the widget

randovania/gui/multiplayer_session_window.py
```diff
@@ -750,13 +753,19 @@ def generate_layout(progress_update: ProgressUpdateCallable):
...
+                # Alert the person who generates on errors since update_game_tab
+                # doesn't generation errors
                 raise
 
             except Exception as e:
                 await self.failure_handler.handle_exception(
                     e,
                     self.update_progress,
                 )
+                alert_user(self, self._options)
```
henriquegemignani, line 766: If the intent is to notify when it finished, error or not, you can use the `finally` block

henriquegemignani (dismissed):
Maybe also add an option for the visual alert?

randovania/gui/multiplayer_session_window.py
Miepee: We don't want to alert on cancel error or on success

Miepee:
> Maybe also add an option for the visual alert?

IMO the visual alert is harmless enough where it doesn't need an option. 
At least in other programs that I used, I've never seen an option for flashing-on-taskbar.

randovania/gui/ui_files/main_window.ui
```diff
@@ -653,6 +651,17 @@ For more information, check [Randovania Help](help://tab_multiworld)!</string>
     <string>Verify installation</string>
    </property>
   </action>
+  <action name="menu_action_disable_sound_alerts">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>Disable Sound Alerts</string>
```
Miepee, line 659: This text should show the opposite: `Enable Sound Alerts`

henriquegemignani:
> > Maybe also add an option for the visual alert?
> 
> IMO the visual alert is harmless enough where it doesn't need an option. At least in other programs that I used, I've never seen an option for flashing-on-taskbar.

Discord
image

randovania/gui/multiplayer_session_window.py
henriquegemignani: On cancel, yeah.
On success, not sure? You can definitely leave it generating and alt-tab while it runs. Especially for big multis.

Miepee:
Huh, never knew that existed.

randovania/gui/multiplayer_session_window.py
Miepee: the host will still be notified on success, just not via this function. It's from [`update_game_tab`](https://github.com/randovania/randovania/pull/7465/files#diff-53f4482d0e5d76127403793fa68c960d5c93cd78fd4646fb7fea40d2989241aaR452) instead.

randovania/gui/ui_files/main_window.ui
Nystrata: Do y'all want it to be Sound Alerts to be enabled by default?
And if so, how would I do so?

Miepee: Not sure whether they should be enabled by default. On one hand, people are probably never finding the setting otherwise, on the other, I feel like more people find a sound annoying and would just turn it off. Personally leaning more towards disabled by default (due to bias)

But this was more just for phrasing to make it more sensible.

Nystrata:
> > > Maybe also add an option for the visual alert?
> > 
> > 
> > IMO the visual alert is harmless enough where it doesn't need an option. At least in other programs that I used, I've never seen an option for flashing-on-taskbar.
> 
> Discord image

I can look into that.

Couple things -   
Should it be coupled together as one option, so "Sound and Visual Alerts" is one option.  
  
or
  
Should it be two separate options for the sound and visual alerts?
and if so,
On Windows, it flashes in the taskbar.
On macOS, it bounces in the dock.
Is there a preferred name for the option if it toggles both option?
The best I can think of is "Taskbar Flashing/Dock Bouncing", but that sounds cumbersome.

randovania/gui/lib/common_qt_lib.py
Nystrata: I'm unsure what that means

Miepee:
> Should it be two separate options for the sound and visual alerts?

I personally would turn off the sound alerts but have the visual alerts on, so I would prefer two.
If you want phrasing, you can just have "Enable visual/audio alerts for when Generation is done" or something.

Nystrata:
> > Should it be two separate options for the sound and visual alerts?
> 
> I personally would turn off the sound alerts but have the visual alerts on, so I would prefer two. If you want phrasing, you can just have "Enable visual/audio alerts for when Generation is done" or something.

If separate options are preferred, how does this look?
image

randovania/gui/lib/common_qt_lib.py
henriquegemignani: a `QtWidget` having a `_options` field is not a rule, so you shouldn't be doing `parent._options`. But this is exactly why I suggest for this function to receive an `Options` as argument, so you can just use that directly

Nystrata: Oh, I'm silly - I didn't realize that I still was calling on parent._options instead of the passed in options

randovania/gui/ui_files/main_window.ui
Nystrata: Added to "Alert When Generation Completes" submenu, which should make the option understandable without renaming to "Enable Sound Alerts"
382442773-09367ee5-f8bc-48c9-a4f4-fc2a756d5c20

randovania/gui/multiplayer_session_window.py
Nystrata: Yeah, on testing, doing it here, it only alerts the host. 
Doing it with update_game_tab alerts everyone.

randovania/gui/lib/common_qt_lib.py
Nystrata: Fixed!

randovania/gui/multiplayer_session_window.py
```diff
@@ -750,13 +753,19 @@ def generate_layout(progress_update: ProgressUpdateCallable):
             except AnyNetworkError:
                 # We're interested in catching generation failures.
                 # Let network errors be handled by who called us, which will be captured by handle_network_errors
+                alert_user(self, self._options)
+                # Alert the person who generates on errors since update_game_tab
```
Miepee, line 757: I dont like putting comments *underneath* the statement thats relevant. Would prefer something more akin to this (hopefully this fits into 1 line?)
Also text should say more akin to "Alert the person who generates on errors, since 'update_game_tab' doesn't show generation errors to other clients

randovania/gui/multiplayer_session_window.py
```diff
@@ -750,13 +753,19 @@ def generate_layout(progress_update: ProgressUpdateCallable):
...
+                # doesn't generation errors
                 raise
 
             except Exception as e:
                 await self.failure_handler.handle_exception(
                     e,
                     self.update_progress,
                 )
+                alert_user(self, self._options)
+                # Alert the person who generates on errors, since update_game_tab
```
Miepee, line 767: ditto

randovania/gui/ui_files/main_window.ui
```diff
@@ -653,6 +661,28 @@ For more information, check [Randovania Help](help://tab_multiworld)!</string>
...
+  </action>
+  <action name="menu_action_visual_taskbar_dock_alerts">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="checked">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>Visual Taskbar/Dock Alerts</string>
```
Miepee, line 683: I would just say `Visual Alerts`

randovania/interface_common/options.py
```diff
@@ -131,6 +131,8 @@ def from_json(cls, value: dict) -> PerGameOptions:
     "dark_mode": Serializer(identity, bool),
     "show_multiworld_banner": Serializer(identity, bool),
     "experimental_settings": Serializer(identity, bool),
+    "sound_alerts": Serializer(identity, bool),
+    "visual_taskbar_dock_alerts": Serializer(identity, bool),
```
Miepee, line 135: I would prefer to call these everywhere as `sound_generation_alert` (maybe audio instead?) and `visual_generation_alert`

randovania/gui/ui_files/main_window.ui
Nystrata: My concern with just "Visual Alerts" is that user can't really tell what it's supposed to affect. Saying Taskbar/Dock would at least give them an indication of that.

Miepee: The issue is that this is a *very* OS specific thing. So technically this would be more "Taskbar/Dock/Panel/topbar/whatever-android-has/whatever-ios-has" which is just a complete mess
IMO if the user doesnt need to know what exact audio alerts they are, they dont need to know the exact visual alerts.

randovania/interface_common/options.py
Nystrata: > sound_generation_alert (maybe audio instead?)  

Gonna go for "audible_generation_alert" then

randovania/gui/ui_files/main_window.ui
Nystrata: Changed it to
image

randovania/gui/multiplayer_session_window.py
Nystrata: Had to really wrangle the comment to make it fit pre-commit's constraints

```py
alert_user(
    self, self._options
)  # Alert the user who gens on errors, since 'update_game_tab' doesn't show gen errors to other clients
```

henriquegemignani: Why not just above? I'd put a blank line between the comment block explaining why we ignore network errors and the block explaining the alert

Miepee (approved):

PR #7470, opened 2024-10-29
Fusion: Fix 1st action giving items to unlock hints only

Miepee:
```
[P: Missile Launcher Data] - 1.75
[P: Morph Ball] - 1.75
[P: Level 2 Locks] - 0.0
[P: Level 4 Locks] - 0.0
```

### DIFF
--- a/randovania/games/fusion/logic_database/Main Deck.json
+++ b/randovania/games/fusion/logic_database/Main Deck.json
@@ -1145,6 +1145,15 @@
                                         "amount": 1,
                                         "negate": false
                                     }
+                                },
+                                {
+                                    "type": "resource",
+                                    "data": {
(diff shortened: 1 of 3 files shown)


JeffGainsNGames:
Won't this not help if you say, start in sector 1 entrance save? since the event would be trivial to reach?

Miepee:
If you start in e.g. sector 1 entrance, then this isn't really an issue. The generator cant place anything just to reach those hints. It would need to place morph, or speedbooster. If it does that, not only does it unlock more locations for you but also getting l2/l4 *after* allows you to get checks (e.g. go to s3/s4, go to reactor)

JeffGainsNGames (approved):

PR #7490, opened 2024-11-05
Implement Co-op support for Multiworlds

### LINKED ISSUE #6932: Multiworld: Support Co-op
## Goals
Co-op should allow multiple people to play the same world. Any player that collects an item, will be sent to every player (excluding the the who picked it up) who's also playing that world.

## Non-Goals
- Seeing other players in your own world.
- Syncing game state between players. E.g. Destroyed locks, in-game events, picked up locations, defeated bosses etc.

## Required changes
- GUI work
  - to allow a world to be assigned to multiple players
  - that changes MW help guide to clear this case slightly more, as it currently assumes 1 world 1 player
- Server should [persist local checks](https://github.com/randovania/randovania/blob/b2eefaaf0be3ee539a2205514e8c7a80c277aa2a/randovania/server/multiplayer/world_api.py#L125-L127)
- Remote pickups should only include items for yourself if the world is a co-op world. (Wouldn't this get a bit weird, since the world assignment can change at any time?)
- Maybe change how the [pickup exporter](https://github.com/randovania/randovania/blob/8b5bb6f9f34ab3c6e29e0d35b7603709b24cc352/randovania/exporter/pickup_exporter.py#L285) is displaying text for local worlds?
- Probably some more I missed.

Miepee:
Opening to get a little more feedback on this.

### Usage

Session admins now have the ability to toggle coop for the session if they havent generated yet:
grafik

If a session is in coop, then an "all worlds" category shows up. This is to make it easier to assign worlds:
grafik

### Game stuff

Some things with games I found:
#### AM2R
- This feature exposed a flaw in the MW implementation where it apparently sometimes eats items when they're received. Only way to get the items is to start a new save file and let all the items load in. Not sure what's up with that, can probably happen in normal multis too though. Could also have been a fluke though, who knows. EDIT: seems to have been a fluke, we can't replicate it.
#### Dread
- In the one playthrough session we did, we encountered a thing, where once we received wide beam, we both had a really weird pseudo-charge effect hovering around samus until resetting to checkpoint/travelling to another area. 
- Progressive items for your own world will always show the first progression model. Not quite sure how easy it is to fix yet, the conditional resources always are an empty tuple. 
- Nothing items are using the Multiworld icon on the minimap. Should be fairly simple to fix.
#### Echoes
- ~~Item percentage is never given to the player, so they always finish the game with 0%~~ EDIT: considered non-issue by Henrique.
#### MSR
- The feature that shows you how many dna are in the current scenario/region is always set to 0.
- Progressive items models have the same thing as in dread

For most games, I don't know whether the things are considered blocking or not, so I'd appreciate advice on them. Would probably fix them in a future PR though.

### Other notes

The feature is currently disabled in dev builds cause I don't want complete public testing yet. 
Currently collecting an item means "sent item to world" and then followed by "received item from world". There was small discussion to remove this, but the more I played it, the more I appreciated both messages being shown, because it helps knowing whether you have connector issues or if the item/location was collected already. So far haven't heard other people complain about it either.

### Missing stuff
- Changelog for when this feature is not locked to running from source
- Tests. Probably want to have an rdvgame with all games in coop to test the PDFs, and then a few to test server/ui.

Fixes #6932


randovania/network_common/game_connection_status.py
```diff
@@ -11,6 +11,8 @@ class GameConnectionStatus(Enum):
     Disconnected = "disconnected"
     TitleScreen = "title-screen"
     InGame = "in-game"
+    Unclaimed = "unclaimed"
+    Empty = "empty"
```
henriquegemignani, line 15: Where is this used?

henriquegemignani (changes requested):

randovania/network_common/game_connection_status.py
Miepee: https://github.com/randovania/randovania/pull/7490/files#diff-4660f7803bfa7c1d3aa65e7e92d48cfb09a701fc3afda26fffb30a89261a4c93R83-R86

henriquegemignani: And what sets something to empty?

Miepee: anything in the "any worlds" group

randovania/exporter/pickup_exporter.py
```diff
@@ -278,7 +278,7 @@ def create_details(
         If for yourself, use the solo creator but adjust the name to mention it's yours.
         For offworld, create a custom details.
         """
-        if pickup_target.player == self.players_config.player_index:
+        if pickup_target.player == self.players_config.player_index and not self.players_config.is_coop:
```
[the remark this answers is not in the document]
Miepee: thought about something similar too, could also be used for the CS PDF

randovania/gui/widgets/multiplayer_session_users_widget.py
```diff
@@ -95,7 +100,7 @@ class MultiplayerSessionUsersWidget(QtWidgets.QTreeWidget):
     _last_session: MultiplayerSessionEntry | None = None
     _session: MultiplayerSessionEntry
     _user_widgets: dict[int, UserWidgetEntry]
-    _world_widgets: dict[uuid.UUID, WorldWidgetEntry]
+    _world_widgets: dict[tuple[int | None, uuid.UUID], WorldWidgetEntry]  # (user id, world UUID): worldWidgetEntry
```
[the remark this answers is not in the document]
Miepee: don't know what you mean here. as an extra attribute for MultiplayerSessionUsersWidget? or do you mean to just have it as a type alias, so that you can do `dict[tuple[UserID, uuid.UUID], ...]`?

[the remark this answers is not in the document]
henriquegemignani: The later

randovania/exporter/pickup_exporter.py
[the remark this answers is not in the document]
Miepee: Not quite happy with the function name, but after sitting there for a while I couldn't think of a better one.

randovania/server/multiplayer/session_admin.py
```diff
@@ -583,7 +594,7 @@ def _create_patcher_file(sa: ServerApp, session: MultiplayerSession, world_uid:
 def admin_player(sa: ServerApp, session_id: int, user_id: int, action: str, *args):
     monitoring.set_tag("action", action)
 
-    verify_has_admin(sa, session_id, user_id)
+    verify_has_admin(sa, session_id, sa.get_current_user().id)
```
henriquegemignani, line 597: This needs a test

randovania/server/multiplayer/session_admin.py
```diff
@@ -542,6 +546,16 @@ def _set_allow_everyone_claim(sa: ServerApp, session: MultiplayerSession, new_st
         session.save()
 
 
+def _set_allow_coop(sa: ServerApp, session: MultiplayerSession, new_state: bool):
```
henriquegemignani, line 549: Let's do better with new code. Return typing and docstring

randovania/server/multiplayer/session_admin.py
```diff
@@ -542,6 +546,16 @@ def _set_allow_everyone_claim(sa: ServerApp, session: MultiplayerSession, new_st
...
+    verify_has_admin(sa, session.id, None)
+
+    with database.db.atomic():
+        session.allow_coop = new_state
```
henriquegemignani, line 553: When disabling co-op, it should either fail if there's world claimed by multiple players or delete these extra claims.

randovania/server/multiplayer/session_admin.py
```diff
@@ -560,10 +574,7 @@ def _create_patcher_file(sa: ServerApp, session: MultiplayerSession, world_uid:
 
     layout_description = session.layout_description
     players_config = PlayersConfiguration(
-        player_index=player_index,
-        player_names=player_names,
-        uuids=uuids,
-        session_name=session.name,
+        player_index=player_index, player_names=player_names, uuids=uuids, session_name=session.name, is_coop=True
```
henriquegemignani, line 577: Maybe keep the trailing comma so it's still multiple lines?

randovania/gui/widgets/multiplayer_session_users_widget.py
```diff
@@ -363,19 +369,35 @@ def _create_world_item(
...
-                        connect_to(claim_menu.addAction(p.name), self._world_claim_with, world_id, p.id)
+                    create_claim_for_each_player_entry()
 
-            else:
+            elif owner is not None and (
+                self.is_admin() or self._session.allow_everyone_claim_world or owner == self.your_id
+            ):
                 world_menu.addSeparator()
-                connect_to(world_menu.addAction("Unclaim"), self._world_unclaim, world_id, owner)
+                create_unclaim_entry()
```
henriquegemignani, line 400: The Unclaim should show up for your claimed worlds in the "All worlds" section

henriquegemignani (changes requested):

randovania/gui/widgets/multiplayer_session_users_widget.py
Miepee: Not so sure about that. If you have a scenario like:
```
Alice (admin)
Bob
|- Prime
Caesar
|- Prime
All Worlds
|- Prime
```
And alice clicks on the actions for prime under the all worlds section, what should happen for her?
If "unclaim" shows, its not quite clear from *who* alice is unclaiming it from. With the world being directly under the user its clear with "ok, i'm unclaiming the world from the user above" which isn't present here.

So IMO either don't make it show at all there (lazy implementation lol), or we show it as "Unclaim from yourself" with an admin option of "unclaim from > [dropdown of every user in the session that has it claimed]" (and maybe an admin option of "unclaim from all users"?)

henriquegemignani: If "Claim for yourself" shows, then "Unclaim for yourself" should also show

randovania/server/multiplayer/session_admin.py
Miepee: should ideally have some gui handling/feedback too.

Miepee (commented):
Some recent commit mustve broken somehing, where an admin that has claimed a world is unable to claim the world for other people now.

Miepee (commented):
Renaming a world/preset should cause a meta update so that it renames properly for everyone.
Also double check permissions/handling for renaming worlds.

Same goes for replacing a preset

randovania/server/multiplayer/session_admin.py
Miepee: On second thought, I'm not quite sure if I want to have this change
It was used so that on "everyone can claim worlds", that a user can unclaim a world from other people. But we probably shouldn't allow that.

randovania/gui/multiplayer_session_window.py
```diff
@@ -887,6 +896,50 @@ async def export_game_button_clicked(self):
...
+                QtWidgets.QMessageBox.Icon.Information,
+                "Information",
+                (
+                    "Co-op is still *very* experimental and may have issues.\nFor Prime 1 and Echoes in particular, "
+                    "please ensure that Randovania is always connected to the game before you collect items, as "
+                    "otherwise they will be lost permanently!"
+                ),
+            )
+        else:
+            world_count = collections.defaultdict(int)
```
henriquegemignani, line 913: Maybe first iterate to create a `world_users: dict[WorldId, list[UserId]]`, so the unclaim loop at the end is simpler and `if is_world_assigned_to_multiple` could just be `if any(len(users) > 1 for users in world_users.values):`?

test/test_files/patcher_data/prime1/prime1_and_2_multi/world_1.json
```diff
@@ -27,11 +27,7 @@
         "automaticCrashScreen": true,
         "quickplay": false,
         "quiet": false,
-        "suitColors": {
-            "variaDeg": 40,
-            "gravityDeg": 350,
-            "phazonDeg": 12
-        },
+        "suitColors": {},
```
henriquegemignani, line 30: Suspicious! You seem to have just deleted the cosmetic_1.json for this file

henriquegemignani (changes requested):
A lot of these test files seems to have been recreated. Seems very wrong.

Also your PR touches a bunch of important things. Please make sure `codecov/patch` hits it's target.

randovania/gui/multiplayer_session_window.py
Miepee: yeah didn't occur to me. is indeed cleaner.

Miepee:
> A lot of these test files seems to have been recreated. Seems very wrong.

They're not recreated. We currently have in upstream/main leftover world files for whatever reason, which I deleted in this PR.
E.g. for `multi-am2r+cs+dread+prime1+prime2` you can see two prime jsons here: https://github.com/randovania/randovania/tree/dd582d5/test/test_files/patcher_data/prime1/multi-am2r%2Bcs%2Bdread%2Bprime1%2Bprime2

Even though the rdvgame doesn't specify any prime/echoes games whatsoever!

henriquegemignani:
> > A lot of these test files seems to have been recreated. Seems very wrong.
> 
> They're not recreated. We currently have in upstream/main leftover world files for whatever reason, which I deleted in this PR. E.g. for `multi-am2r+cs+dread+prime1+prime2` you can see two prime jsons here: https://github.com/randovania/randovania/tree/dd582d5/test/test_files/patcher_data/prime1/multi-am2r%2Bcs%2Bdread%2Bprime1%2Bprime2
> 
> Even though the rdvgame doesn't specify any prime/echoes games whatsoever!

Can you do these cleanup and the file moving in a different PR, so it's clear what exactly changed in these files by this PR directly?

Miepee:
Sure

Miepee:
> The idea, at least mine, was to add an additional field in the ExportedPickupDetails not modify the existing one

The reason I went with modifying an existing one, is because conceptually, there is no difference between sending a remote item to another world and sending a remote item to your world. Both are just sending an item to the server and not giving anything locally. If a game really needs to have a distinction between that, then they can still do that via `pickup_target.player == self.players_config.player_index`. For normal multis, the resources will just be empty.

> With an additional field there should be no required (additonal) change in the pdfs

The PDFs already don't need additional changes, because they coded "if item is going to server, dont give any additional resources locally". The only thing that does is dread and MSR because of how they have it coded with progressive models. Every other game (am2r, cs, prime, echoes) does not do anything different in the coop test compared to normal multis.

> It now says "World 5's" instead of "Your" 

This is intentional. "Your" singular is wrong when multiple players are playing a world as it will be sent to all. And "Your" plural sounds very weird cause its hard to tell while reading who this encompasses (all people from the session? all people from the world). "World 5's" is much clearer there and accurate.
Also accidentally makes it a good identifier if a world is coop lol. During testing we had a few sessions where we forgot to enable coop and noticed it because the text said "your" instead of "world bla's"

> which also means we need to change all games to respect the quantity properly.

No, see above. Every other game (other than maybe double checking msr and dread) doesnt need any patcher changes.

> No idea about prime but it looks like it doesn't even has a quantity, so you will get it locally + received remote?

Incorrect, Prime does have a quantity, it's "currincrease". Prime and Echoes are kinda weird there in general cause they give a local MW item if an item goes to the server with a specific amount, so that the connector later can then check which location has been collected.

Miepee:
> Lastly, I think it's very weird to talk about the world you are playing right now in third person and the decision to use the world's name instead of the plural "Your" is very technical driven

IDK, I've seen waaaay too many people confused by plural "you" that I try to avoid it as much as possible and if I do need to use it, just use "y'all" instead. But obviously "sent progressive jump to y'all" sounds very weird as well.
Maybe compromise on "your world" instead?

> That's more up to the game, not to the pdf.

Right, and only Dread and MSR need changes, regardless if it's a new field or just the existing field repurposed.

Nystrata:
As someone who's played with coop a bit, I think I very much prefer just having a the world name rather a form of "you", "your world", or the like.  

I expect there will be quite a few games where people will swap between different coop worlds withn the same multi, so having an in-game reminder of what coop world they're in is useful.  
It's already been useful for me in a similar case.

randovania/exporter/pickup_exporter.py
```diff
@@ -275,21 +279,53 @@ def create_details(
...
+                description,
             )
-            return dataclasses.replace(details, name=f"Your {details.name}")
+            if (
+                self.players_config.should_target_local_player(pickup_target.player)
+                or pickup_target.pickup == useless_pickup
+            ):
+                return dataclasses.replace(details, name=f"Your {details.name}")
+
+            new_resources = [
```
henriquegemignani, line 307: maybe add a comment here leave it explicit it's the section regarding coop?

henriquegemignani (approved):

henriquegemignani (approved):

PR #7510, opened 2024-11-12
Prompt for confirmation on closing window while having background processes

### LINKED ISSUE #7023: WARN the user if attempting to close randovania while generation is in progress

Nystrata:
* resolves #7023

Prompt for confirmation for closing the window while having background processes:
* During solo game generation in the main window
* During multiworld game generation in the multiplayer session window
  * **This will now also stop multiworld game generation when the multiplayer session window is closed**  
  by calling `self.stop_background_process()`.

***
I'm unsure if we should add a confirmation for the multiworld session window, since it seems that, without this change, multiworld game generation will happen even after the multiworld session window is closed but will stop if all windows are closed.

I'm choosing to confirm on any background tasks rather than specifically for game generation due to this discussion in [#randovania-dev](https://discord.com/channels/914291389293027329/914294272579227668/1305725050737004565)

***
Screenshots:
* Main Window:
  * image
* Multiplayer Session Window
  * image


randovania/gui/main_window.py
```diff
@@ -321,6 +322,20 @@ def highlight_logo(
         self.main_tab_widget.setCurrentIndex(0)
 
     def closeEvent(self, event):
+        if self.has_background_process:
+            event.ignore()
+            dialog = QtWidgets.QMessageBox(
```
Miepee, line 327: use async_dialog.warning instead

Miepee (commented):
> I'm unsure if we should add a confirmation for the multiworld session window, since it seems that, without this change, multiworld game generation will happen even after the multiworld session window is closed but will stop if all windows are closed.

I consider this a bug because its very non-intuituve and does whacky things when you reopen the session im fairly certain

---

- changelog

randovania/gui/main_window.py
Nystrata: I was avoiding trying to mix async with non async code, since I'm not sure how to call async code from non async funciton

using something like `await async_dialog.warning` in closeEvent gives me 
"SyntaxError: 'await' outside async function" errors

Miepee: I wanted to avoid code copy paste but looks like we're already doing this sort of thing a few times in the codebase.
avoiding this means doing some hacky shenanigans with asyncio so probably not worth it.

Nystrata:
>     * changelog

Woops, added

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed: Added options to alert the user via flashing the taskbar/playing a sound when generation completes in the Preferences menu.
 - Changed: Add confirmation dialog when deleting a world in Multiworld.
+- Changed: Add confirmation dialog when closing the main window or multiplayer session window while a task is in progress.
```
Miepee, line 12: would phrase this as "while generation is in progress"
task is really generic and not really helping an end user

randovania/gui/main_window.py
```diff
@@ -321,6 +322,20 @@ def highlight_logo(
...
+                QtWidgets.QMessageBox.Icon.Warning,
+                "Confirm close window",
+                "Are you sure you want to close this window?\nClosing this window will abort current tasks.",
+                (StandardButton.Yes | StandardButton.No),
+            )
+            dialog.setDefaultButton(StandardButton.No)
+            dialog.setWindowIcon(QtGui.QIcon(os.fspath(randovania.get_icon_path())))
+            result = dialog.exec()
+            if not result == StandardButton.Yes:
```
Miepee, line 336:
```suggestion
            if result != StandardButton.Yes:
```

randovania/gui/multiplayer_session_window.py
```diff
@@ -308,6 +309,19 @@ async def create_and_update(
...
+        if self.has_background_process:
+            event.ignore()
+            result = await async_dialog.warning(
+                self,
+                "Confirm close window",
+                "Are you sure you want to close this window?\nClosing this window will abort current tasks.",
+                buttons=async_dialog.StandardButton.Yes | async_dialog.StandardButton.No,
+                default_button=async_dialog.StandardButton.No,
+            )
+            if not result == StandardButton.Yes:
```
Miepee, line 321: ditto

randovania/gui/main_window.py
henriquegemignani: `closeEvent` in particular must not be async, because we need to blockingly handle the event

CHANGELOG.md
```diff
@@ -9,8 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added: Drag and dropping a preset file into the main window now imports it.
 - Changed: Added options to alert the user via flashing the taskbar/playing a sound when generation completes in the Preferences menu.
-- Changed: Add confirmation dialog when deleting a world in Multiworld.
+- Changed: Added confirmation dialog when deleting a world in Multiworld.
 - Fixed: When editing the preset description after another change in a preset, the cursor position is now retained.
+- Changed: Added confirmation dialog when closing the main window or multiplayer session window while generation is in progress.
```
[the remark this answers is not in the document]
Nystrata: oh what the heck, I crossed the streams lmao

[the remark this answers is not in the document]
Miepee: Ok you missed what i was trying to point at

Miepee (approved):

PR #7526, opened 2024-11-16
Dread: Restructure and adjust Early Grapple Room

Tyranisaur:
The Ledge before Slide Jump node is deemed unnecessary and removed. The doors are now connected directly to each other.

The difficulty of the slide jumps are switched. Using Cross Bomb to cross the right gap is now behind Beginner Movement.


randovania/games/dread/logic_database/Dairon.txt
```diff
@@ -946,15 +946,23 @@ Extra - asset_id: collision_camera_012
...
-              Slide and Slide Jump (Intermediate)
+              All of the following:
+                  Slide
+                  Any of the following:
+                      Slide Jump (Intermediate)
+                      All of the following:
+                          Slide Jump (Beginner)
+                          Any of the following:
+                              Flash Shift
+                              Cross Bomb and Movement (Beginner)
```
XenoWars, line 965: If you're removing the ledge node (which I do admit is unnecessary), then Intermediate Slide Jump is the only thing that works without spin boost. Cross Bomb and Flash won't help you get into the tunnel above the lava. Disregarding that the second jump is actually more lenient due to being able to save yourself if you fail and is why it was rated lower, you need to do two slide jumps for the whole sequence and failing either will plummet you into the lava.

randovania/games/dread/logic_database/Dairon.txt
```diff
@@ -991,6 +999,15 @@ Extra - asset_id: collision_camera_012
...
   * Extra - right_shield_def: None
+  > Door to Early Grapple Access
+      All of the following:
+          Morph Ball
+          Any of the following:
+              Varia Suit
+              Heat/Cold Runs (Intermediate) and Heat Damage ≥ 450
+          Any of the following:
+              Flash Shift or Use Spin Boost
+              Cross Bomb and Movement (Beginner)
```
XenoWars, line 1010: Not sure why we never had this as just trivial, since you just fall down onto the slope and jump up to the tunnel. Cross Bomb can be removed for the movement requirement.

XenoWars (changes requested):

randovania/games/dread/logic_database/Dairon.txt
Tyranisaur: I'm going to insist that the second one should be intermediate even though it can be retried by jumping back from the slope. The timing is relatively tight, and you have to jump surprisingly late. I'm convinced that the first one has more leniency, and is fine as beginner. Though, it's possible to make a case for adding a beginner wall jump. It's not necessarily an important point, but there's a solid number of options for the second jump that are just completely irrelevant if the first jump is not rated lower than the second one. I'm also convinced that other people agree with me on this rating.

randovania/games/dread/logic_database/Dairon.txt
Tyranisaur: Good point, I'll remove this one. We already had beginner movement for making the left jump in this this direction, so it doesn't make a specific impact if the right one is also considered  beginner movement or would be trickless.

randovania/games/dread/logic_database/Dairon.txt
XenoWars: Based on discussion in Discord, perhaps we make it intermediate and advanced, then call it a day.

Tyranisaur: Done.

CHANGELOG.md
```diff
@@ -35,7 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ##### Dairon
 
-- Changed: In Early Grapple Room: The Slide Jumps have had their difficulty switched. The first one is now Beginner and the second is Intermediate.
+- Changed: In Early Grapple Room: The Slide Jumps have had their difficulties change. The first one is now Intermediate and the second is Advanced.
```
XenoWars, line 38: The first one is no longer being changed, and the "d" is missing on the word "change"

XenoWars (changes requested):

CHANGELOG.md
Tyranisaur: Right of course.

Miepee (approved):
confirmed good by xeno in discord

PR #7527, opened 2024-11-16
Fix multiworld generation complete alert

Nystrata:
Previously, alerts being called in update_game_tab would cause alerts on opening or making any changes to the the multiplayer session window, rather than solely on generation complete.

This works around that by only calling alerts in game_session_api.prepare_to_upload_layout
This change means that only the person generating will have the alerts, rather than all users as previously intended.

> I'm marking this one as draft since I do actually want to properly fix it to alert all users rather than this workaround


randovania/gui/multiplayer_session_window.py
```diff
@@ -753,21 +751,17 @@ def generate_layout(progress_update: ProgressUpdateCallable):
...
                 await self.failure_handler.handle_exception(
                     e,
                     self.update_progress,
                 )
-                # Alert the user who gens on errors, since 'update_game_tab' doesn't show gen errors to other clients
-                alert_user_on_generation(self, self._options)
 
             finally:
                 self._generating_game = False
+                alert_user_on_generation(self, self._options)
```
Miepee, line 765: revert this. You dont want to alert when its being cancelled

Miepee (commented):
I'd rather merge the workaround for now and then find on doing this properly in the meantime than have the risk of this not hitting the feature freeze

Miepee (commented):
probably needs changelog changed? unless we just say that "multiworld alerts" is vague enough where it doesnt need the clarification on whether its the generatee or everyone.

Nystrata:
> probably needs changelog changed? unless we just say that "multiworld alerts" is vague enough where it doesnt need the clarification on whether its the generatee or everyone.

Eh, maybe
Maybe not
I don't think anyone's built an expectation of how it should work yet - I don't think anyone outside of devs came across it at all, since I didn't hear anything about it in beta-testing since it's been merged

henriquegemignani (approved):

PR #7528, opened 2024-11-16
Retain cursor position for preset description even if preset "_is_dirty"

ThanatosGit:


henriquegemignani (approved):

henriquegemignani:
Feels we could have a helper function for this "set text while preserving cursor" or outright avoid setting the text if it's the same, but this works too

PR #7548, opened 2024-11-22
Prime 1: Be able to open morph doors with just morph

Miepee:
[Way back then](https://github.com/randovania/randovania/commit/f882f54ec7d6c9371e6bf33bfc2f7ead8cb58afa#diff-aa6d2234af85c964b8945afcab912a8317b4ee261cbb7d31c3377261a8a29cf0) @toasterparty changed this to require a beam. I'm not quite sure why, an explanation would be welcome. Was it just out precaution to rather put a player in logical BK, than let them being able to logically go places?
Since Morph Ball bombs are able to open normal blue doors, it means that locking morph ball doors behind a beam means that you cannot e.g. start in burn dome access with morph+bombs and get progression in lower furnace.

Even if we want to logically require being able to shoot to go through a morph ball door, the fact that the current condition gets fullfilled with Morph+Wave+Scan only is definitely flawed.


duncathan:
should be changed to `Morph Ball AND Open Normal Door` instead. afaik the point is to prevent the player from being beamless or visorless for any significant duration

Miepee:
There are very few morph ball doors in the game. Most, if not all are basically immediately followed by normal doors. So in the state of this PR, you are locked for any significant duration by bombs/beams.
However, this PR does allow for a few interesting "morph tunnel" starts that aren't "get beams immediately" - just like starting in normal rooms with morph and bomb does.

E.g: On current main, if you start in upper furnace with morph+boost you can go checkout lower furnace and get progression there.
But now if you start in crossway access west with morph+boost, the seed is suddenly logically impossible because of the requirement that you need bombs/beam to go through the tunnel you are able to just morph in-game through.

Samuel6710 (approved):
I agree with Miepee and like these changes, but @toasterparty might have intentionally designed it this way so I would run it by him as well.

toasterparty (approved):

PR #7556, opened 2024-11-25
Properly introduce PickupIndex age v2

Miepee:
Recontinuation of #6607 

Pickups have an impact in the age when they're assigned. The weight of an index depends on it's age, where younger age = more likely

This PR does introduce an option, but it is currently off for all games.
In future PRs, I want to do a bit more analysis on nicer values (especially for Prime)

Fixes https://github.com/randovania/randovania/issues/4496

### DIFF
--- a/randovania/generator/filler/retcon.py
+++ b/randovania/generator/filler/retcon.py
@@ -182,10 +182,10 @@ def increment_index_age(locations_weighted: WeightedLocations, increment: float)
     Increments the pickup index's age for every already collected index.
     """
     for player, location, _ in locations_weighted.all_items():
-        was_present = location in player.pickup_index_age
-        player.pickup_index_age[location] += increment
+        was_present = location in player.pickup_index_ages
+        player.pickup_index_ages[location] += increment
         if not was_present:
-            # if it wasn't present, thenwe get to log!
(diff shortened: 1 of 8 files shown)


randovania/generator/filler/retcon.py
```diff
@@ -32,16 +32,24 @@
...
+    In practice, one group for each region respecting preset exclusions
+    """
     result = {}
 
     for indices in indices_groups:
         weight_from_collected_indices = math.sqrt(len(indices) / ((1 + len(assigned_indices & indices)) ** 2))
 
         for index in sorted(uncollected_indices & indices):
-            weight_from_considered_count = min(10, considered_counts[index] + 1) ** -2
+            weight_from_considered_count = min(10.0, index_age[index] + 1.0) ** -2
```
henriquegemignani, line 52: Rename `weight_from_considered_count`

henriquegemignani (changes requested):

randovania/generator/filler/retcon.py
Miepee: good catch, fixed.

henriquegemignani (approved):

PR #7569, opened 2024-12-01
Call door rando setting "mode" in UI

Miepee:
No screenshot cause new machine and haven't bothered setting up running from source yet. But shouldn't be too hard to review anyway.

Reasoning: Saying "I set my Door randomization type to 'Type' " sounds very wrong.

### DIFF
--- a/randovania/gui/ui_files/preset_dock_rando.ui
+++ b/randovania/gui/ui_files/preset_dock_rando.ui
@@ -99,7 +99,7 @@
            <item row="0" column="0">
             <widget class="QLabel" name="mode_label">
              <property name="text">
-              <string>Randomization Type</string>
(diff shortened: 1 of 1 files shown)


dyceron:
Honestly, should even rename the modes themselves. "Doors" is too vague. Yes I get that the description is there, but to me, "Types" is more obvious. We should change the modes to be: "Original Door Locks", "Shuffle Individual Door Locks" and "Shuffle Door Locks by Type."

Miepee:
Maybe for a next release
At least, as long as I still haven't setup my machine fully, I don't want to do it, especially because those are mentioned in more places than just here.

Miepee:
Will close and redo with dcr's suggestion later

PR #7589, opened 2024-12-05
Prime 1 Remove Superheat Option in Twin Fires Tunnel

M4RKERSM4SK:
Idea from Amojini, Baja also thought it could be a _cool_ idea.

> Ideas for opening up prime progression for cgc was removing superheated from twin fires tunnel. Allows phen and maybe mines without Varia, maybe makes gravity more important logically

Since this change seemed like the remove bars feature, I thought I could implement it in a similar fashion. 

- Added option "Remove Superheat in Twin Fires Tunnel" and it is disabled by default
- Removed out HUD Heat Notifications and similar Items in Transport to Tallon Overworld West to remove confusion while entering the room

Tested

- Verified removed objects in Transport to Tallon Overworld West and no crashes/weirdness
- Forced Starting Location to Twin Fires tunnel to test Logic and Generator
- Made sure room was not superheated and tried the spider ball track etc

I believe I did the logic edits correctly, but would like a triple check since I am new at it.

If I need to edit one of the testing rdv files like the remove bars feature implementation, please let me know. I can do the refills one again or whatever you all decide is best.

### DIFF
--- a/randovania/games/prime1/exporter/patch_data_factory.py
+++ b/randovania/games/prime1/exporter/patch_data_factory.py
@@ -740,6 +740,19 @@ class PrimePatchDataFactory(PatchDataFactory):
             ]
 
         # serialize room modifications
+        if self.configuration.remove_superheat_twin_fires_tunnel:
+            magmoor_region = level_data["Magmoor Caverns"]["rooms"]
+            magmoor_region["Twin Fires Tunnel"]["superheated"] = False
+            magmoor_region["Transport to Tallon Overworld West"]["deleteIds"] = [
+                1048683,  # 0x0010006B - Timer for Varia Check
+                1048684,  # 0x0010006C - Function for Varia Check
+                1048682,  # 0x0010006A - Hud Memo Trigger for Heat
+                1048686,  # 0x0010006E - Hud Memo for Heat
+                1048685,  # 0x0010006D - Hud Memo Timer for Heat
+                1048681,  # 0x00100069 - Steam Effect
+                1048687,  # 0x0010006F - Lava that was meant to trigger "Warning" and Heat Glow Effect
+            ]
+
         if self.configuration.superheated_probability != 0:
             probability = self.configuration.superheated_probability / 1000.0
             for region in regions:
--- a/randovania/games/prime1/generator/bootstrap.py
+++ b/randovania/games/prime1/generator/bootstrap.py
@@ -40,6 +40,7 @@ class PrimeBootstrap(MetroidBootstrap):
             "superheated_probability": "superheated_probability",
             "submerged_probability": "submerged_probability",
             "remove_bars_great_tree_hall": "remove_bars_great_tree_hall",
+            "remove_superheat_twin_fires_tunnel": "remove_superheat_twin_fires_tunnel",
         }
         for name, index in logical_patches.items():
             if getattr(configuration, name):
--- a/randovania/games/prime1/gui/preset_settings/prime_patches_qol.py
+++ b/randovania/games/prime1/gui/preset_settings/prime_patches_qol.py
@@ -26,6 +26,7 @@ _FIELDS = [
     "phazon_elite_without_dynamo",
     "spring_ball",
     "remove_bars_great_tree_hall",
+    "remove_superheat_twin_fires_tunnel",
 ]
 
 
--- a/randovania/games/prime1/gui/ui_files/preset_prime_qol.ui
+++ b/randovania/games/prime1/gui/ui_files/preset_prime_qol.ui
@@ -42,9 +42,9 @@
        <property name="geometry">
         <rect>
          <x>0</x>
-         <y>-358</y>
+         <y>-143</y>
          <width>487</width>
-         <height>1200</height>
+         <height>1272</height>
         </rect>
        </property>
        <layout class="QVBoxLayout" name="scroll_layout">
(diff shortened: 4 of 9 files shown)


Miepee:
IMO, this change is very different from gth bars and I'm not sure i like it.
Bar gone is a change that affects logic on the trick levels most people play as, and has pretty significant logic side effects. Partly due to gth being essentially a hub.

What significance does this change bring? For trickless this has literally no changes, which is the majority of our userbase. For knowledge beginner+, it still will have minimal impact, but feel free to prove me wrong with data otherwise.

duncathan:
I thought the first paragraph of the PR was pretty clear that this is intended for CGC to adjust how progression plays out in that format, and was endorsed by Baja (a TO)

Miepee:
I'm still not entirely convinced without some data to back this up as a good change.

From what I understand, this is supposed to devalue varia suit by making it required less early.
The issue but by making giving the generator an easy pass to go between tallon/phen/mines, is that it undervalues varia suit by a lot, because there is no logical use for varia outside of magmoor/magma pool.
So I worry that by giving the player access to every area easily, it will lead to a lot of seeds where varia either isnt needed, or can easily bypassed with heat runs due to the amount of etanks you have gotten until then.
So i'd like to see something that shows
- how does this change general progression
- how often were varia locked locations completely dead
- how often did they have macguffins that were able to be easily bypassed
Because so far these were all just theorized in small discussions

Miepee:
Would also be neat to see how this logically changes things when its *combined* with the removed bars.

M4RKERSM4SK:
>  For trickless this has literally no changes, which is the majority of our userbase.

On trickless, Morph and Spider are used to traverse the room without Varia with this change. Tested the generation a few times on trickless and this is what it gave me to traverse the room.

> I'm still not entirely convinced without some data to back this up as a good change.

This PR was something I thought I could help with as the racing part of the community is searching for more options to explore on improving CGC. I don't know if this classifies as a "good change", but was going more towards having a change that could help CGC. The problem is, I don't know if the analyzing tools really work with multi-world. IIRC, I believe that you told me it was kinda broken for multi-world so I am not sure I can get good data like that to back this up in that specific case for CGC.

Cestrion at the time that kicked off this discussion:

> As much as I love CGC format how it currently is with it's preset, it should evolve. Looking for more preset variants that are actually fun and also not BK inducing is the goal for me.

Since Varia was one contender that came up on consistently placing the prime 1 player in logical BK early/mid game, I think the intention is that this preset change could give the prime player more options with wave/spider/grapple/thermal or maybe PPC with spider/pbs/wave/ice etc. 

> The issue but by making giving the generator an easy pass to go between tallon/phen/mines, is that it undervalues varia suit by a lot, because there is no logical use for varia outside of magmoor/magma pool.

I had the same thoughts and Baja did mention another concern that required Thardus Hallways could happen a ton more

Doing this change in addition with GTH Bar skip was something I brought up as well, could be really interesting

I will see what I can do about getting some data to address the concerns and see if this change is worth it.

I will open a thread in the discord prime dev chat if I cant figure out the batch generation, but I am going to try to make one of the spreadsheets you all do that can show:

- Trickless
- Trickless + Remove Superheat
- Trickless + Remove Superheat + Remove GTH Bars
- Prime 1 Beginner Trick Racing Preset
- Prime 1 Beginner Trick Racing Preset+ Remove Superheat
- Prime 1 Beginner Trick Racing Preset+ Remove Superheat + Remove GTH Bars

Would this work in what you are thinking of for data to backup this change?

Miepee:
> On trickless, Morph and Spider are used to traverse the room without Varia with this change. Tested the generation a few times on trickless and this is what it gave me to traverse the room.

This is a logic bug! Every non-superheated magmoor room should have "knowledge (beginner) or heat resistant suit", see the connections in magmoor workstation for example. However, for some reason, this is missing for for all the rooms between workstation and twin fires tunnel. (Which also means right now, you could e.g. spawn in twin fires with grapple+sj+ice+boost+spider+bombs and be expected go collect your progression in plasma processing trickless.)
In other words, no, this would not have any effect on trickless. Sure, you could go through twin fires tunnel, but then you'd immediately need varia at twin fires.

>  IIRC, I believe that you told me it was kinda broken for multi-world 

They are, but shouldnt be tooooo hard to fix, i just never bothered.

> Since Varia was one contender that came up on consistently placing the prime 1 player in logical BK early/mid game,

Trying to balance to never BK is a bad choice IMO. The more you try to take BK possibilities away from prime, the more you will cause BK for echoes (bad!) and/or place prime into "full-clear as fast as possible". IMO the possibility of a BK should exist, if only to force the other players to route differently than they would do in solo - which is what no BK would do.


> Would this work in what you are thinking of for data to backup this change?

yes

Samuel6710:
The option is something I'm cool with including, but Miepee's concerns are valid. I'd like to see some batch generated seeds with this setting enabled before making any decisions.

Regardless of that, the logic needs to be adapted. Besides the logic bug where several rooms don't require Knowledge (Beginner) that Miepee mentioned, the option to remove Superheated effects should not logically apply to subsequent rooms past Twin Fires Tunnel. Those should still require Knowledge (Beginner) or a superheated suit.

M4RKERSM4SK:
Here is some data. I don't have a google account so I apologize I cannot share this that way.

"Base" is CGC Prime Weekly (August 2024)
"RemoveHeatTFT" is CGC Prime Weekly (August 2024) + Remove Super Heat in Twin Fires Tunnel
"RemoveHeatTFT+RemoveBars" is CGC Prime Weekly (August 2024) + Remove Super Heat in Twin Fires Tunnel + Remove Bars

1k Seeds per each of the 3 presets and I only did the analyzer on the majors and progression for now so we all can get an idea

[RemoveSuperheatData.ods](https://github.com/user-attachments/files/18025898/RemoveSuperheatData.ods)

JeffGainsNGames:
https://docs.google.com/spreadsheets/d/14p5beTWl9zw26Ojgo2LILVwGfKWN8odNzL2RNllgIH8/edit?usp=sharing

M4RKERSM4SK:
`I don't think this change has a place in RDV in general when there are preset options that achieve the same goal. `

`once multiworld logic rework that has proper energy requirements, this problem will be solved by that.`

`patching RDV specifically for a CGC hallway isn't something I think we should support`

This change essentially would circumvent the core issues needing to be fixed. Closing

PR #7591, opened 2024-12-05
bump caver

duncathan:
(note: this is a PR to stable)

see also https://github.com/cave-story-randomizer/cave-story-randomizer/releases/tag/2.4.1

- Fixed: Curly's item in Plantation is now always sent in multiworld
- Changed: Loading a save in Jail no. 1 will now remove the block if the chest has been opened
- Changed: The helper block in Grasstown to return from east to west has been lowered by one block to make the jump easier
- Fixed: The helper block in Grasstown always spawns when it is supposed to
- Fixed: Freeware no longer crashes when alt-tabbing in fullscreen mode
- Fixed: Teleporter icons in Arthur's House now always appear in the correct order
- Fixed: Music randomization in Outer Wall now works regardless of which way you entered the room
- Added: Ikuyo and kl3cks7r have been added to the ingame credits


Miepee (approved):
Apart from the fact that I can't read TSC, looks good to me

Miepee:
several tests seem to be failing though

PR #7601, opened 2024-12-08
Use a dedicated field for last displayed version for dev builds

henriquegemignani:


Miepee (commented):
what happens when old clients, which already had a dev build saved in their self._last_changelog_displayed, will try to connect to a new server after this PR?

randovania/interface_common/options.py
```diff
@@ -385,14 +387,25 @@ def user_dir(self) -> Path:
...
     @property
-    def last_changelog_displayed(self) -> StrictVersion:
-        return StrictVersion(self._last_changelog_displayed)
+    def last_changelog_displayed(self) -> version_lib.Version:
+        if randovania.is_dev_version():
+            version_str = self._last_changelog_displayed_dev
+        else:
+            version_str = self._last_changelog_displayed
+
+        return version_lib.parse_string(version_str)
```
Miepee, line 396: IMO, if version_str is an invalid version string, we should reset the field.
Just in the case that someone somehow gets an invalid field in there, that they aren't completely blocked from using the application.

henriquegemignani: We really should be using `version_lib.parse_string` as the method for it in `_SERIALIZER_FOR_FIELD` because then it falls back to the popup saying "invalid field found. delete and continue?".
This error only happens when we cause an error, so I don't want to silently delete user data.

Miepee (approved):

PR #7614, opened 2024-12-11
Fusion: change name locks to keycards

JeffGainsNGames:
you collect `a thing` to open `a lock`, you do not collect a lock. Keycard seems fitting given the in-game sprite.

image


duncathan:
thank god

randovania/games/fusion/logic_database/header.json
```diff
@@ -220,7 +220,7 @@
                 }
             },
             "L0Locks": {
```
Miepee, line 222: we might wanna rename the short key at one point too

PR #7621, opened 2024-12-16
Add pickup index and provider uuid to remote pickups

ThanatosGit:
Need this to fix MSR DNA counter in coop.
Provider UUID to compare to remote connector UUID. If it's the same, it's a coop item.
Pickup index to find the region of the pickup for the coop world.

### DIFF
--- a/randovania/game_connection/connector/mercury_remote_connector.py
+++ b/randovania/game_connection/connector/mercury_remote_connector.py
@@ -142,12 +142,12 @@ class MercuryConnector(RemoteConnector):
 
         self.in_cooldown = True
 
-        provider_name, pickup, pickup_index, is_coop = remote_pickups[num_pickups]
+        provider_name, pickup, coop_location = remote_pickups[num_pickups]
         item_name, items_list = self.resources_to_give_for_pickup(self.game.resource_database, pickup, inventory)
 
         scenario_id = ""
-        if is_coop:
-            location_node = self.game.region_list.node_from_pickup_index(pickup_index)
+        if coop_location is not None:
+            location_node = self.game.region_list.node_from_pickup_index(coop_location)
             scenario_id = self.game.region_list.nodes_to_region(location_node).extra["scenario_id"]
 
         self.logger.debug("Resource changes for %s from %s", pickup.name, provider_name)
--- a/randovania/game_connection/connector/mercury_remote_connector.py
+++ b/randovania/game_connection/connector/mercury_remote_connector.py
@@ -151,7 +151,7 @@ class MercuryConnector(RemoteConnector):
         await self.game_specific_execute(item_name, items_list, provider_name, region_name, provider_uuid)
 
     async def game_specific_execute(
-        self, item_name: str, items_list: list, provider_name: str, region_name: str, provider_uuid: str
+        self, item_name: str, items_list: list, provider_name: str, region_name: str, provider_uuid: uuid.UUID
     ) -> None:
         raise NotImplementedError
 
--- a/randovania/game_connection/connector/mercury_remote_connector.py
+++ b/randovania/game_connection/connector/mercury_remote_connector.py
@@ -10,9 +10,9 @@ from qasync import asyncSlot  # type: ignore
 
 from randovania.exporter.pickup_exporter import _conditional_resources_for_pickup
 from randovania.game_connection.connector.remote_connector import (
-    PickupEntryWithOwner,
     PlayerLocationEvent,
     RemoteConnector,
+    RemotePickupTuple,
 )
 from randovania.game_description import default_database
 from randovania.game_description.resources.inventory import Inventory, InventoryItem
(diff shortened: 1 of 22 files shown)


randovania/game_connection/connector/remote_connector.py
```diff
@@ -17,7 +17,8 @@
     from randovania.game_description.db.area import Area
     from randovania.game_description.db.region import Region
 
-PickupEntryWithOwner = tuple[str, PickupEntry]
+# provider_name, PickupEntry, PickupIndex, provider_uuid
+RemotePickupTuple = tuple[str, PickupEntry, PickupIndex, str]
```
Miepee, line 21: why is the uuid a string instead of uuid.UUID?

Miepee (commented):
will give full review later, just quickly wanted to throw this out

henriquegemignani (changes requested):
I think in some kind of "network common" file has a bunch of api version numbers. There's one for at least server api, would have to raise it because this is changing is changing the data structures that are sent along

randovania/game_connection/connector/remote_connector.py
Miepee: Yeah I can't see any other usecase either. Would be fine with a boolean.

randovania/game_connection/connector/msr_remote_connector.py
```diff
@@ -25,7 +27,9 @@ def get_resources_for_details(
     ) -> list:
         return get_resources_for_details(pickup, conditional_resources, other_player)
 
-    async def game_specific_execute(self, item_name: str, items_list: list, provider_name: str) -> None:
+    async def game_specific_execute(
```
Miepee, line 30: PR to interact with the game fix dna hints n stuff will come later?

randovania/game_connection/connector/remote_connector.py
```diff
@@ -11,13 +12,12 @@
...
 if typing.TYPE_CHECKING:
-    import uuid
-
     from randovania.game.game_enum import RandovaniaGame
     from randovania.game_description.db.area import Area
     from randovania.game_description.db.region import Region
 
-PickupEntryWithOwner = tuple[str, PickupEntry]
+# provider_name, PickupEntry, PickupIndex, provider_uuid
+RemotePickupTuple = tuple[str, PickupEntry, PickupIndex, uuid.UUID]
```
Miepee, line 20: at this point we should just make it a class inheriting from NamedTuple.

And somewhere in network common so that it can also be used by multiplayer_session.py

randovania/server/multiplayer/world_api.py
```diff
@@ -350,6 +351,8 @@ def emit_world_pickups_update(sa: ServerApp, world: World):
                 {
                     "provider_name": action.provider.name,
                     "pickup": _base64_encode_pickup(pickup_target.pickup, resource_database),
+                    "provider_uuid": str(action.provider.uuid),
+                    "location": action.location,
```
Miepee, line 355: would be neat for these to have the same ordering as everywhere else (e.g. location first, uuid after)

randovania/game_connection/connector/remote_connector.py
henriquegemignani: > No hard reason.
> uuid.UUID is not json serializable. Therefore the server sends it as string and I just didn't converted it to uuid again on client side because I don't really have to.

Our automatic serialization of server/client should be handling UUID, which means it's no effort. Also a good chunk of the client-server api is actually binary, not json and so sending the UUID class is more efficient since it's just 16 bytes.

Of course it could be this part of the API uses none of that tooling, but I sure hope this rather high volume part is using it...

Miepee (approved):
Changes look fine on my end.
@henriquegemignani probably knows better than I if another arbitrary number would need to be adjusted somewhere else.

randovania/game_connection/connector/remote_connector.py
duncathan: I'd rather send the `uuid.UUID` and just fix the serialization tbh. should be pretty trivial to do and it's easy to believe that having the uuid might come in handy later

randovania/game_connection/connector/mercury_remote_connector.py
```diff
@@ -141,14 +141,18 @@ async def receive_remote_pickups(self) -> None:
 
         self.in_cooldown = True
 
-        provider_name, pickup = remote_pickups[num_pickups]
+        provider_name, pickup, pickup_index, is_coop = remote_pickups[num_pickups]
         item_name, items_list = self.resources_to_give_for_pickup(self.game.resource_database, pickup, inventory)
+        location_node = self.game.region_list.node_from_pickup_index(pickup_index)
+        region_name = self.game.region_list.nodes_to_region(location_node).extra.get("scenario_id", "")
```
henriquegemignani, line 147: I'd call this variable scenario_id

Is it ever missing?

henriquegemignani (changes requested):

randovania/server/multiplayer/world_api.py
```diff
@@ -350,6 +351,8 @@ def emit_world_pickups_update(sa: ServerApp, world: World):
                 {
                     "provider_name": action.provider.name,
                     "pickup": _base64_encode_pickup(pickup_target.pickup, resource_database),
+                    "location": action.location,
```
henriquegemignani, line 354: Does providing the index when it's not co-op ever make sense? The world it's from is not included so you don't have access to the correct GameDescription.

Could instead merge both is_coop with location, making it say `coop_location` which is an `Pickup Index | None`

henriquegemignani (changes requested):

henriquegemignani (approved):

randovania/server/multiplayer/world_api.py
duncathan: I'm happier with this

PR #7626, opened 2024-12-16
replace Prime 1 flavor scans with advice

Shaleblade:
first draft of the scan tips PR, which replaces many flavor scans in Prime with advice, including information on all Knowledge tricks

### DIFF
--- a/randovania/games/prime1/exporter/scan_tips.py
+++ b/randovania/games/prime1/exporter/scan_tips.py
@@ -1,0 +1,696 @@
+from __future__ import annotations
+
+SCAN_TIPS = {
+    "3671417093": [
+        # Stone Toad Standable Tip
+        # Replaces Stone Toad scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Stone Toad",
+        (
+            "&main-color=#89D6FF;It is possible to stand on Stone Toads.\n\n\n"
+            "By slope jumping off of the spine on its back, "
+            "one can reach the top of Reflecting Pool without draining the water."
+        ),
+    ],
+    "2124053457": [
+        # Ruined Shrine Boost Skip Tip
+        # Replaces Ruined Shrine halfpipe scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;One can reach the top with the"
+            "&main-color=#FF3333;Space Jump Boots&main-color=#89D6FF;"
+            ' by jumping on the pointed root protruding from the wall. \n The name "Mr. Peg" is etched into the root.'
+        ),
+    ],
+    "3745923772": [
+        # Sheegoth Standable Tip
+        # Replaces Sheegoth scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Sheegoth",
+        (
+            "&main-color=#89D6FF;It is possible to stand on Sheegoths.\n\n\n"
+            "By jumping off of them with &main-color=#FF3333;Space Jump Boots&main-color=#89D6FF;, "
+            "one can climb Ice Ruins East without missiles."
+        ),
+    ],
+    "3162940989": [
+        # Baby Sheegoth Standable Tip
+        # Replaces Baby Sheegoth scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Baby Sheegoth",
+        (
+            "&main-color=#89D6FF;It is possible to stand on Baby Sheegoths."
+            "\n\n\nBy jumping off of them with &main-color=#FF3333;Space Jump Boots&main-color=#89D6FF;, "
+            "one can climb Ice Ruins East without missiles."
+        ),
+    ],
+    "3508298443": [
+        # Bloodflower Standable Tip
+        # Replaces Bloodflower scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Bloodflower",
+        (
+            "&main-color=#89D6FF;Using a double bomb jump, one can roll on top of a Bloodflower.\n\n"
+            "By using another double bomb jump on top of it, "
+            "Great Tree Chamber is reachable from Great Tree Hall without Space Jump Boots."
+        ),
+    ],
+    "2541494929": [
+        # Vault Jump Tip
+        # Replaces Main Plaza fallen stones scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;One can reach the ledge leading to Vault with "
+            "&main-color=#FF3333;Space Jump Boots&main-color=#89D6FF;.\n\nIt requires holding "
+            "&image=SA,3.0,1.0,1.0,46434ED3,34E79314; while jumping from a brick platform under a twisting tree."
+        ),
+    ],
+    "2903166635": [
+        # Metroid Prime Flamethrower Exo Tip
+        # Replaced Metroid Prime Exo scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Metroid Prime",
+        (
+            "&main-color=#89D6FF;When Metroid Prime's exoskeleton is weak to the Power Beam, "
+            "it can also be damaged by the &main-color=#FF3333;Flamethrower&main-color=#89D6FF;."
+        ),
+    ],
+    "2271263048": [
+        # Hive Mecha Bridge Tip
+        # Replaces Hive Mecha bridge scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        "&main-color=#89D6FF;To cross the bridge without fighting Hive Mecha, carefully walk along its right edge.",
+    ],
+    "2053272936": [
+        # Boost Pipe Speed Tip
+        # Replaces Tallon Canyon halfpipe scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;When using the &main-color=#FF3333;Boost Ball&main-color=#89D6FF; in a half pipe, "
+            "hold only &image=8B0C22A7; or &image=778CCD6A; and boost at the lowest point to gain the most height."
+        ),
+    ],
+    "399696251": [
+        # Boost Speed Limit Tip
+        # Replaces Ruined Gallery morph tunnel scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;When using the &main-color=#FF3333;Boost Ball&main-color=#89D6FF;, "
+            "if the control stick is kept in a neutral position, the Boost Ball has no speed limit."
+        ),
+    ],
+    "4175379874": [
+        # Flaahgra Wavebuster Tip
+        # Replaced Flaahgra scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Flaahgra",
+        (
+            "&main-color=#89D6FF;Flaahgra can be defeated with the &main-color=#FF3333;Wavebuster&main-color=#89D6FF;. "
+            "On Normal difficulty, doing so without any other sources of damage requires 230 missiles."
+        ),
+    ],
+    "889553182": [
+        # Frozen Metroid Hop Tip
+        # Replaces Metroid scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Metroid",
+        (
+            "&main-color=#89D6FF;Jumping on the center of a frozen Metroid causes one to bounce on top of it.\n\n"
+            "When it breaks out of the ice, anyone on top is launched high in the air\n\nThis can be used to traverse "
+            "Metroid Quarantine A without Space Jump Boots, Morph Ball Bombs, or Spider Ball."
+        ),
+    ],
+    "4176439895": [
+        # Frozen Hunter Metroid Hop Tip
+        # Replaces Hunter Metroid scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Hunter Metroid",
+        (
+            "&main-color=#89D6FF;Jumping on the center of a frozen Hunter Metroid causes one to bounce on top of it."
+            "\n\nWhen it breaks out of the ice, anyone on top is launched high in the air.\n\n"
+            "This can be used to acquire Frost Cave's item without the Grapple Beam."
+        ),
+    ],
+    "425486343": [
+        # Arboretum door DBJ tip
+        # Replaces Arboretum big branch scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;It is possible to surmount the gate "
+            "at the top of this room with a carefully-timed double bomb jump."
+        ),
+    ],
+    "1102487982": [
+        # FCS climb tip
+        # Replaces Frigate Crash Site frigate scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;It is possible to climb up to this room's top door from the bottom. "
+            "There are multiple methods, the easiest of which only needs the "
+            "&main-color=#FF3333;Space Jump Boots&main-color=#89D6FF; to climb the crashed frigate's exterior."
+        ),
+    ],
+    "3737672425": [
+        # Monitor Station climb tip
+        # Replaces Monitor Station 2f left computer scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;By jumping along the rectangular extrusions on this building's exterior, "
+            "it is possible to reach the roof without Space Jump Boots."
+        ),
+    ],
+    "3893588753": [
+        # Monitor Station climb tip
+        # Replaces Monitor Station 2f middle computer scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;While on this building's roof, it is possible to reach Warrior Shrine without "
+            "lowering the bridge. To do so, stand on top of the roof computer's tallest monitor, "
+            "lock onto a Puffer to the right or left, and perform a lock-on dash towards the ledge."
+        ),
+    ],
+    "1839087853": [
+        # Tree Weakness Tip
+        # Replaces Main Plaza tree scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Opening the tree requires a &main-color=#FF3333;Super Missile&main-color=#89D6FF;."
+            "\n\n\nAdditionally, it is possible to peek the item inside the tree without opening it. "
+            "To do so, stand at a specific angle and location on the right side of the tree's lip, then jump in place."
+        ),
+    ],
+    "1530752134": [
+        # Phendrana Canyon L-Jump Tip
+        # Replaces Phendrana Canyon platform scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;The item in this room can be acquired without the Scan Visor "
+            "by using the &main-color=#FF3333;Space Jump Boots&main-color=#89D6FF; and L-Jumping."
+        ),
+    ],
+    "3740523398": [
+        # Phendrana Canyon Escape Tip
+        # Replaces Phendrana Canyon half-pipe fish scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;As long as the boxes in this room remain intact, "
+            "they can be climbed to exit this room without any items."
+        ),
+    ],
+    "442524052": [
+        # Rapid Fire Missiles Tip
+        # Replaces Main Plaza holobanner
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Hold&image=5D9F9796;and rapidly tap "
+            "&image=EA2A1C5C; to fire missiles as fast as possible."
+        ),
+    ],
+    "807982837": [
+        # Underwater Slope Jump Tip
+        # Replaces Watery Hall plant flavor text
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Underwater slope jumps performed without the Gravity Suit allow access "
+            "to great heights.\n\nYou can use an underwater slope jump to reach the sunken item in this "
+            "room without the Gravity Suit. This is significantly easier with the "
+            "&main-color=#FF3333;Space Jump Boots&main-color=#89D6FF;."
+        ),
+    ],
+    "2077227745": [
+        # Meta Ridley Boost Tip
+        # Replaces Meta Ridley scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Meta Ridley",
+        (
+            "&main-color=#89D6FF;During the first phase of combat against Meta Ridley, "
+            "hitting it with a &main-color=#FF3333;Boost Ball&main-color=#89D6FF; "
+            "charge causes enough damage to immediately move to the fight's second phase."
+        ),
+    ],
+    "2967215000": [
+        # Non-Superheated Back Magmoor Tip
+        # Replaces Puddle Spore scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Puddle Spore",
+        (
+            "&main-color=#89D6FF;Everything south of Twin Fires Tunnel in Magmoor Caverns is not superheated,\n\n"
+            "allowing for safe travel without the Varia Suit."
+        ),
+    ],
+    "1735564172": [
+        # Arboretum Sandstone PBs Tip
+        # Replaces Arboretum sandstone scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;The &main-color=#FF3333;Sandstone&main-color=#89D6FF; structures in this room are "
+            "some of the few that cannot be destroyed by Power Bombs.\nNote that some game versions change that."
+        ),
+    ],
+    "1618265307": [
+        # Burn Dome PB Sandstone Tip
+        # Replaces Burn Dome sandstone scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Like most &main-color=#FF3333;Sandstone&main-color=#89D6FF; structures, "
+            "this can also be destroyed by &main-color=#FF3333;Power Bombs&main-color=#89D6FF;."
+        ),
+    ],
+    "745991660": [
+        # Gathering Hall PB talloric alloy Tip
+        # Replaces Gathering Hall lamp scans
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Like most &main-color=#FF3333;Talloric Alloy&main-color=#89D6FF; structures, "
+            "the grate above can also be destroyed by &main-color=#FF3333;Power Bombs&main-color=#89D6FF;."
+        ),
+    ],
+    "3344095433": [
+        # Ruined Shrine PB Sandstone Tip
+        # Replaces Ruined Shrine sandstone scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Like most &main-color=#FF3333;Sandstone&main-color=#89D6FF; structures, "
+            "this can also be destroyed by &main-color=#FF3333;Power Bombs&main-color=#89D6FF;."
+        ),
+    ],
+    "3485565452": [
+        # Vault PB Slots Tip
+        # Replaces Vault bomb slot seal scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;These Bomb Slot covers are able to be destroyed by &main-color=#FF3333;Power Bombs"
+            "&main-color=#89D6FF;.\n\nAdditionally, these Bomb Slots are also able to be activated by Power Bombs."
+        ),
+    ],
+    "1354393071": [
+        # Watery Hall PB Sandstone Tip
+        # Replaces Watery Hall sandstone scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Like most &main-color=#FF3333;Sandstone&main-color=#89D6FF; structures, "
+            "this can also be destroyed by &main-color=#FF3333;Power Bombs&main-color=#89D6FF;."
+        ),
+    ],
+    "3507547160": [
+        # Lava Lake PB Sandstone + No-Respawn Tip
+        # Replaces Lava Lake sandstone scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Like most &main-color=#FF3333;Sandstone&main-color=#89D6FF; structures, "
+            "this can also be destroyed by &main-color=#FF3333;Power Bombs&main-color=#89D6FF;.\n\n"
+            "Additionally, in most game versions, these rocks will not reappear after being destroyed."
+        ),
+    ],
+    "3269646811": [
+        # Reflecting Pool PB Talloric Alloy Tip
+        # Replaces Reflecting Pool lid scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Like most &main-color=#FF3333;Talloric Alloy&main-color=#89D6FF; structures, "
+            "this can also be destroyed by &main-color=#FF3333;Power Bombs&main-color=#89D6FF;."
+        ),
+    ],
+    "1876916293": [
+        # Fungal Hall B PB Tip
+        # Replaces MQB lower right computer screen scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;The item within Fungal Hall B's mushroom ring can be uncovered with "
+            "&main-color=#FF3333;Morph Ball Bombs&main-color=#89D6FF; or "
+            "&main-color=#FF3333;Power Bombs&main-color=#89D6FF;."
+        ),
+    ],
+    "4186358462": [
+        # Life Grove Tunnel PB Tip
+        # Replaces Life Grove & LGT Bendezium rock scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;This &main-color=#FF3333;Bendezium&main-color=#89D6FF; surface has a note:\n\n\n "
+            "Life Grove and Life Grove Tunnel each contain a &main-color=#FF3333;Sandstone&main-color=#89D6FF; "
+            "structure.\n\nLike most &main-color=#FF3333;Sandstone&main-color=#89D6FF; structures, "
+            "they can also be destroyed by &main-color=#FF3333;Power Bombs&main-color=#89D6FF;."
+        ),
+    ],
+    "323204925": [
+        # Watery Hall Access Wall Tip
+        # Replaces Watery Hall Access holobanner scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;In addition to missiles, the weak wall section in this room is able to be destoyed "
+            "by the &main-color=#FF3333;Charge Beam&main-color=#89D6FF;."
+        ),
+    ],
+    "56434148": [
+        # Central Dynamo Unlocked Door Tip
+        # Replaces Central Dynamo lower right computer scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        "&main-color=#89D6FF;Fighting the Cloaked Drone locks all doors except the door to Quarantine Access A.",
+    ],
+    "3355561537": [
+        # Central Dynamo IS Tip
+        # Replaces Central Dynamo lower left computer screen scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;A corner of this room can be used to acquire &main-color=#FF3333;Infinite Speed"
+            "&main-color=#89D6FF;. Using Infinite Speed before defeating the Cloaked Drone "
+            "results in taking significantly less damage."
+        ),
+    ],
+    "1755484832": [
+        # Central Dynamo leaving tip
+        # Replaces Central Dynamo upper left computer screen scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;When the Cloaked Drone is defeated, "
+            "it is possible to leave the room without collecting the item."
+        ),
+    ],
+    "2271012115": [
+        # Spinner PB Tip
+        # Replaces Spinner scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Spinner",
+        (
+            "&main-color=#89D6FF;In lieu of Morph Ball Bombs, one can also exit Spinners "
+            "with &main-color=#FF3333;Power Bombs&main-color=#89D6FF;."
+        ),
+    ],
+    "263442559": [
+        # Flaahgra Reverse Kill Tip
+        # Replaces Sunchamber mirror scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;If coming from Sun Tower Access, Flaahgra can be defeated without shooting the first "
+            "mirror.\nScan dashing or &main-color=#FF3333;Space Jump Boots&main-color=#89D6FF; "
+            "allows one to reach the center and bomb the bomb slots."
+        ),
+    ],
+    "345113941": [
+        # Metroid Tube Break Tip
+        # Replaces Research Lab Aether top left computer scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;In addition to a missile, a &main-color=#FF3333;Power Bomb&main-color=#89D6FF; or "
+            "&main-color=#FF3333;Plasma Beam&main-color=#89D6FF; shot will break the Metroid's tank."
+        ),
+    ],
+    "437035956": [
+        # HotE Ghost Skip Tip
+        # Replaces HotE statue-side holobanner scans
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;If entering from Reflecting Pool Access before the Chozo Ghosts have been awakened,"
+            "\n\none can move through the room along a specific path while avoiding awakening the ghosts.\n"
+            "This can either be used to activate the Bomb Slots or to reach Crossway Access South."
+        ),
+    ],
+    "213585783": [
+        # Main Quarry Bombless Tip
+        # Replaces Main Quarry left computer scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;The nearby item is acquirable without Morph Ball Bombs by unmorphing on the crane "
+            "platform's edge.\nAlternatively, one can slope jump off of this "
+            "cabin's roof to scale the crane platform and reach the item."
+        ),
+    ],
+    "130269422": [
+        # Main Quarry Disposal Dash Tip
+        # Replaces Main Quarry right computer scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;It is possible to reach the door to Waste Disposal with a scan dash "
+            "and &main-color=#FF3333;Space Jump Boots&main-color=#89D6FF;."
+        ),
+    ],
+    "3474604570": [
+        # Control Tower Fight Skip Tip
+        # Replaces Control Tower ice scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;The Space Pirates in this room are alerted when one steps in front of the entrance "
+            "to West Tower. One can avoid this by jumping around the space directly in front of the entrance."
+        ),
+    ],
+    "3573990646": [
+        # Thardus Skip Tip
+        # Replaces Quarantine Monitor left screen scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;One can cross Quarantine Cave without fighting "
+            "Thardus by using aerial movement around the room's edges."
+        ),
+    ],
+    "3820360882": [
+        # Sun Tower Early Wild Tip
+        # Replaces Sun Tower gate scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;When entering from Sun Tower Access, one can awaken the Chozo Ghosts in Sunchamber "
+            "without falling to the bottom of the room. This requires either using the &main-color=#FF3333;Spider Ball"
+            "&main-color=#89D6FF; to roll to the bottom of the magnetic track before rolling back up, or stopping the "
+            "descent with an Oculus and using complex bomb jumps to rise back to the top."
+        ),
+    ],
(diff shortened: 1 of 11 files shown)


Nystrata:
This is my personal opinion:
When referring to the player/player character in third person, consider substituting "one" with "Samus" to make it slightly less ambiguous.

> Jumping on the center of a frozen Hunter Metroid causes one to bounce on top of it.   

vs

> Jumping on the center of a frozen Hunter Metroid causes Samus to bounce on top of it.

and

> One can reach the ledge leading to Vault with...

vs

> Samus can reach the ledge leading to Vault with...

Nystrata:
Actually, you're right, it's not ambiguous as I first thought.  
Though, I actually find "one" more awkward than "Samus", though that's likely because that's what I'm used to reading from strategy guides/tutorials.   
There's also the third option of using "You" - this actually falls in line with how the in-game Inventory entries in the Prime games go:
> The Gravity Suit eliminates the effects of liquid on movement. Samus's Notes: This modification improves your defensive shielding...

> Timing is important when using the Space Jump Boots. Experiment to discover ways to increase the height and length of your jumps...

So, "You" would match the cadence of the in-game text.

I'd prefer "You" or "Samus" - I still feel "one" is weird.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "399696251": [
+        # Boost Speed Limit Tip
+        # Replaces Ruined Gallery morph tunnel scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;When using the &main-color=#FF3333;Boost Ball&main-color=#89D6FF;, "
+            "if the control stick is kept in a neutral position, the Boost Ball has no speed limit."
```
Samuel6710, line 104: - This scan node seems better suited to explaining the wallboost rather than the mechanics of boost ball.
- Boost Ball's speed cap is also removed if you are holding left or right, not just neutral.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+            "On Normal difficulty, doing so without any other sources of damage requires 230 missiles."
+        ),
+    ],
+    "889553182": [
+        # Frozen Metroid Hop Tip
+        # Replaces Metroid scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Metroid",
+        (
+            "&main-color=#89D6FF;Jumping on the center of a frozen Metroid causes one to bounce on top of it.\n\n"
```
Samuel6710, line 123: This seems unnecessary. This information isn't useful for players who don't have this trick in logic; those who do have it in logic very likely already know how it works. Most players will also be focused on killing the metroids instead of scanning them, so the likelihood they would see this scan and find it useful is unlikely.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+    ],
+    "4176439895": [
+        # Frozen Hunter Metroid Hop Tip
+        # Replaces Hunter Metroid scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Hunter Metroid",
+        (
+            "&main-color=#89D6FF;Jumping on the center of a frozen Hunter Metroid causes one to bounce on top of it."
+            "\n\nWhen it breaks out of the ice, anyone on top is launched high in the air.\n\n"
+            "This can be used to acquire Frost Cave's item without the Grapple Beam."
```
Samuel6710, line 136: See MQA Metroid hop comment.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "425486343": [
+        # Arboretum door DBJ tip
+        # Replaces Arboretum big branch scan
```
Samuel6710, line 141: This scan point would be better suited for the gate node rather than the branch.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;It is possible to surmount the gate "
+            "at the top of this room with a carefully-timed double bomb jump."
```
Samuel6710, line 146: It would be worth noting that the bomb jump must be directed at the top right of the gate.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+    ],
+    "3893588753": [
+        # Monitor Station climb tip
+        # Replaces Monitor Station 2f middle computer scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;While on this building's roof, it is possible to reach Warrior Shrine without "
+            "lowering the bridge. To do so, stand on top of the roof computer's tallest monitor, "
+            "lock onto a Puffer to the right or left, and perform a lock-on dash towards the ledge."
```
Samuel6710, line 178: - Calling this a `lock-on dash` is a little confusing. I'd change it to `combat dash` to make it consistent with the trick name.
- It would also be worth mentioning that the gap can be traversed with an R-Jump.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+    ],
+    "1839087853": [
+        # Tree Weakness Tip
+        # Replaces Main Plaza tree scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Opening the tree requires a &main-color=#FF3333;Super Missile&main-color=#89D6FF;."
+            "\n\n\nAdditionally, it is possible to peek the item inside the tree without opening it. "
+            "To do so, stand at a specific angle and location on the right side of the tree's lip, then jump in place."
```
Samuel6710, line 189: This isn't a trick and it replaces scan text that conveys information about the vulnerability in the tree, so I'd either remove this scan or find a different node to use.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;The item in this room can be acquired without the Scan Visor "
+            "by using the &main-color=#FF3333;Space Jump Boots&main-color=#89D6FF; and L-Jumping."
+        ),
+    ],
+    "3740523398": [
+        # Phendrana Canyon Escape Tip
+        # Replaces Phendrana Canyon half-pipe fish scan
```
Samuel6710, line 204: `fish scan`?

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "442524052": [
+        # Rapid Fire Missiles Tip
+        # Replaces Main Plaza holobanner
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Hold&image=5D9F9796;and rapidly tap "
+            "&image=EA2A1C5C; to fire missiles as fast as possible."
```
Samuel6710, line 219: This isn't a trick and is a slightly arbitrary location to have the hint so I would remove it.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "Meta Ridley",
+        (
+            "&main-color=#89D6FF;During the first phase of combat against Meta Ridley, "
+            "hitting it with a &main-color=#FF3333;Boost Ball&main-color=#89D6FF; "
+            "charge causes enough damage to immediately move to the fight's second phase."
+        ),
+    ],
+    "2967215000": [
+        # Non-Superheated Back Magmoor Tip
+        # Replaces Puddle Spore scan
```
Samuel6710, line 247: I think this is good information to convey but choosing a good scan node is difficult. While I believe the puddle spore is not a good spot for it since it is not something players are likely to scan, there are not many options. Maybe the Magmoor Workstation pump monitor?

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "1735564172": [
+        # Arboretum Sandstone PBs Tip
+        # Replaces Arboretum sandstone scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;The &main-color=#FF3333;Sandstone&main-color=#89D6FF; structures in this room are "
+            "some of the few that cannot be destroyed by Power Bombs.\nNote that some game versions change that."
```
Samuel6710, line 262: `Note that some game versions change that.` What versions change this behavior and for which blocks? I cannot find them in [version differences.](https://tcrf.net/Metroid_Prime/Version_Differences)

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Like most &main-color=#FF3333;Sandstone&main-color=#89D6FF; structures, "
+            "this can also be destroyed by &main-color=#FF3333;Power Bombs&main-color=#89D6FF;."
+        ),
+    ],
+    "745991660": [
+        # Gathering Hall PB talloric alloy Tip
+        # Replaces Gathering Hall lamp scans
```
Samuel6710, line 277: This should be the grate above, not the lamps.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "3485565452": [
+        # Vault PB Slots Tip
+        # Replaces Vault bomb slot seal scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;These Bomb Slot covers are able to be destroyed by &main-color=#FF3333;Power Bombs"
+            "&main-color=#89D6FF;.\n\nAdditionally, these Bomb Slots are also able to be activated by Power Bombs."
```
Samuel6710, line 302: `...can be destroyed`/`...can also be activated` flows a bit better than `...are able to be destroyed`/`...are also able to be activated`.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+    ],
+    "3507547160": [
+        # Lava Lake PB Sandstone + No-Respawn Tip
+        # Replaces Lava Lake sandstone scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Like most &main-color=#FF3333;Sandstone&main-color=#89D6FF; structures, "
+            "this can also be destroyed by &main-color=#FF3333;Power Bombs&main-color=#89D6FF;.\n\n"
+            "Additionally, in most game versions, these rocks will not reappear after being destroyed."
```
Samuel6710, line 323: I do not recall any game versions having this change. Randomprime does have a QOL change that makes it so that they do not reappear, which may be what you are thinking of.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+    ],
+    "1876916293": [
+        # Fungal Hall B PB Tip
+        # Replaces MQB lower right computer screen scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;The item within Fungal Hall B's mushroom ring can be uncovered with "
+            "&main-color=#FF3333;Morph Ball Bombs&main-color=#89D6FF; or "
+            "&main-color=#FF3333;Power Bombs&main-color=#89D6FF;."
```
Samuel6710, line 344: This isn't a trick, and players likely already know that this can be destroyed with power bombs and morph ball bombs. Additionally, this scan node provides information for a different room. This is superfluous.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        (
+            "&main-color=#89D6FF;This &main-color=#FF3333;Bendezium&main-color=#89D6FF; surface has a note:\n\n\n "
+            "Life Grove and Life Grove Tunnel each contain a &main-color=#FF3333;Sandstone&main-color=#89D6FF; "
+            "structure.\n\nLike most &main-color=#FF3333;Sandstone&main-color=#89D6FF; structures, "
+            "they can also be destroyed by &main-color=#FF3333;Power Bombs&main-color=#89D6FF;."
+        ),
+    ],
+    "323204925": [
+        # Watery Hall Access Wall Tip
+        # Replaces Watery Hall Access holobanner scan
```
Samuel6710, line 361: Does the debris not already have a scan node? That would be a more suitable node.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+            "&main-color=#89D6FF;In addition to missiles, the weak wall section in this room is able to be destoyed "
+            "by the &main-color=#FF3333;Charge Beam&main-color=#89D6FF;."
+        ),
+    ],
+    "56434148": [
+        # Central Dynamo Unlocked Door Tip
+        # Replaces Central Dynamo lower right computer scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        "&main-color=#89D6FF;Fighting the Cloaked Drone locks all doors except the door to Quarantine Access A.",
```
Samuel6710, line 374: There are different triggers that lock the doors. This can be triggered without triggering the drone fight. I would reword to indicate that the doors lock when approached after entering the room and will only unlock after defeating the drone and doing the maze.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+    ],
+    "3355561537": [
+        # Central Dynamo IS Tip
+        # Replaces Central Dynamo lower left computer screen scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;A corner of this room can be used to acquire &main-color=#FF3333;Infinite Speed"
```
Samuel6710, line 382: It would be worth explaining that it takes advantage of the crates. You may also consider explaining how infinite speed is actually obtained.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "1755484832": [
+        # Central Dynamo leaving tip
+        # Replaces Central Dynamo upper left computer screen scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;When the Cloaked Drone is defeated, "
+            "it is possible to leave the room without collecting the item."
```
Samuel6710, line 394: This is true regardless of whether the drone has been defeated or not.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "2271012115": [
+        # Spinner PB Tip
+        # Replaces Spinner scan
```
Samuel6710, line 399: I'd adjust this comment to indicate which spinner(s) this is used for.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "",
+        (
+            "&main-color=#89D6FF;If coming from Sun Tower Access, Flaahgra can be defeated without shooting the first "
+            "mirror.\nScan dashing or &main-color=#FF3333;Space Jump Boots&main-color=#89D6FF; "
+            "allows one to reach the center and bomb the bomb slots."
+        ),
+    ],
+    "345113941": [
+        # Metroid Tube Break Tip
+        # Replaces Research Lab Aether top left computer scan
```
Samuel6710, line 420: The metroid inside the container may be a more suitable node for this information.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;In addition to a missile, a &main-color=#FF3333;Power Bomb&main-color=#89D6FF; or "
+            "&main-color=#FF3333;Plasma Beam&main-color=#89D6FF; shot will break the Metroid's tank."
+        ),
+    ],
+    "437035956": [
+        # HotE Ghost Skip Tip
+        # Replaces HotE statue-side holobanner scans
```
Samuel6710, line 430: This scan node seems too out of the way. I would replace it with the peg scan point and explain that it is used to reach the bomb slot.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "213585783": [
+        # Main Quarry Bombless Tip
+        # Replaces Main Quarry left computer scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;The nearby item is acquirable without Morph Ball Bombs by unmorphing on the crane "
+            "platform's edge.\nAlternatively, one can slope jump off of this "
```
Samuel6710, line 446: I would also mention the wallboost.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+            "cabin's roof to scale the crane platform and reach the item."
+        ),
+    ],
+    "130269422": [
+        # Main Quarry Disposal Dash Tip
+        # Replaces Main Quarry right computer scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;It is possible to reach the door to Waste Disposal with a scan dash "
```
Samuel6710, line 456: I would also mention the R-Jump.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "3573990646": [
+        # Thardus Skip Tip
+        # Replaces Quarantine Monitor left screen scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;One can cross Quarantine Cave without fighting "
+            "Thardus by using aerial movement around the room's edges."
```
Samuel6710, line 477: I would say `...room's pillars` instead.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+    "3820360882": [
+        # Sun Tower Early Wild Tip
+        # Replaces Sun Tower gate scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;When entering from Sun Tower Access, one can awaken the Chozo Ghosts in Sunchamber "
+            "without falling to the bottom of the room. This requires either using the &main-color=#FF3333;Spider Ball"
+            "&main-color=#89D6FF; to roll to the bottom of the magnetic track before rolling back up, or stopping the "
+            "descent with an Oculus and using complex bomb jumps to rise back to the top."
```
Samuel6710, line 489: I would say `...using an infinite bomb jump to...` instead.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "1530945290": [
+        # Training Chamber Ghost Skip Tip
+        # Replaces Meditation Fountain left sculpture scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;In Training Chamber, one can avoid awakening the Chozo Ghosts "
+            "by climbing architecture near the entrance to Training Chamber Access."
```
Samuel6710, line 499: This scan seems like too much. The trick in question is hypermode and the scan node is quite out of the way and inconspicuous. I think this should be removed.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;There are several ways to cross Magma Pool without the Grapple Beam.\n\n"
+            "The simplest uses the &main-color=#FF3333;Space Jump Boots&main-color=#89D6FF; and &main-color=#FF3333;"
+            "Gravity Suit&main-color=#89D6FF; to jump off of a raised stone in the lava near the far ledge.\n"
+            "A similar method uses a double bomb jump from the same location, though it requires not having the "
+            "Gravity Suit.\nA harder method uses &main-color=#FF3333;Space Jump Boots&main-color=#89D6FF; to scan-dash"
+            " off of pickups left by the crates.\n\nFinally, it is possible to get across with "
+            "&main-color=#FF3333;Space Jump Boots&main-color=#89D6FF; by L-jumping between the lips of the "
```
Samuel6710, line 514: R-Jumping*

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;When performing a scan dash or lock-on dash, jump by tapping "
+            "&image=SI,0.70,0.68,833BEE04; for as little time as possible to gain extra height."
+        ),
+    ],
+    "3668253850": [
+        # Item Drop Tip
+        # Replaces Waterfall Cavern hole scan
```
Samuel6710, line 550: This seems like an arbitrary location to put this scan.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+    "3673769160": [
+        # L-Jump Frames Tip
+        # Replaces Elite Research top floor computer scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Holding &image=SA,3.0,1.0,1.0,46434ED3,34E79314; for the first 13 frames of a jump "
+            "and then letting go\nslightly extends jumps. This is called an L-Jump. When using the &main-color=#FF3333;"
+            "Space Jump Boots&main-color=#89D6FF;, holding &image=SA,3.0,1.0,1.0,46434ED3,34E79314; "
+            "for 7 frames after the second jump grants an additional, smaller extension."
```
Samuel6710, line 567: This is an arbitrary location to put this scan. It would be much more suitable to put information about doing this spinner without boost instead.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "1807281950": [
+        # RDV FAQ Tip
+        # Replaces North Atrium right sculpture scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        "&main-color=#89D6FF;Each game in Randovania has a Frequently Asked Questions section.",
```
Samuel6710, line 575: This scan gives information outside the scope of the game. The FAQ is already visible and available on the Randovania application, so I do not see the point of it.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+    ],
+    "2367804388": [
+        # Shiny Missile Tip
+        # Replaces North Atrium disc scans
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;There is a 1/1024 chance for Missile Expansions to be &main-color=#FF3333;shiny"
+            "&main-color=#89D6FF;.\n\n\nShiny Missile Expansions are "
+            "&main-color=#3333FF;blue&main-color=#89D6FF;.\n\n\n\nThat is all."
```
Samuel6710, line 595: Same as above

Shaleblade:
> Actually, you're right, it's not ambiguous as I first thought. Though, I actually find "one" more awkward than "Samus", though that's likely because that's what I'm used to reading from strategy guides/tutorials. There's also the third option of using "You" - this actually falls in line with how the in-game Inventory entries in the Prime games go:
> 
> > The Gravity Suit eliminates the effects of liquid on movement. Samus's Notes: This modification improves your defensive shielding...
> 
> > Timing is important when using the Space Jump Boots. Experiment to discover ways to increase the height and length of your jumps...
> 
> So, "You" would match the cadence of the in-game text.
> 
> I'd prefer "You" or "Samus" - I still feel "one" is weird.
> 
> It's just my small opinion, I'm not hung up on it.

As stated in Discord, I ran a poll in January to decide the tone of these entries. "Logbook voice" indeed won. Logbook voice is not consistent in Metroid Prime - you can look at the Parasite Queen's "_Use your auto-targeting to acquire this new target!_" as an example of that. Inventory entries also break convention a lot too. Let me break down the rules I'm working with:

1. No instruction. The scan visor's entries prints descriptions, not directions. You can say "X can do Y" and leave it up to the reader to decide whether they should do Y, but you do not command.
2. Do not address Samus by name. The scan visor's entries are written in an encyclopedic fashion, not engaging with the reader. "You" is acceptable if it's the only way to convey the information, but an unspecific "one" is best where possible.
3. Avoid breaking "kayfabe" as much as possible. Yes, it will often not be possible, but part of sticking with Logbook Voice means trying to keep things in-universe as much as possible. (This one wasn't really in question, but I'm trying to list all of the rules I can think of.)

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "4213747411": [
+        # Sideways speed limit tip
+        # Replaces Landing Site lichen scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        "&main-color=#89D6FF;While unmorphed, there is no sideways limit to your speed.",
```
Samuel6710, line 603: This is not a trick, the scan node location is arbitrary, and it is highly unlikely that a player would encounter this scan.

Samuel6710 (changes requested):

Shaleblade:
I want to address some comments on Discord, but first, I should be clear about the goals of these tip scans. They may not be obvious, but they are important to me:

1. Mention every single Knowledge trick without exception. A complete explanation is ideal, but the reader must, at minimum, get enough of an idea of the trick's existence that they are able to do their own research to learn exactly what it is & how to do it.
2. Provide beginner-friendly advice to make playing the game easier and more fun. This includes basic mechanics advice that the vanilla game does not reinforce (e.g. the Don't Hold B tip; the tip on holding L for improved air control), RDV usage advice (e.g. the FAQ tip), mentioning community resources (e.g. the discord & video database links), and guidance for tricks that are particularly load-bearing or open up play a lot (e.g. the scan dash to Waste Disposal; the SJB jump to the Vault ledge).
3. Explain unintuitive mechanics, regardless of difficulty level. Hopping off of frozen Metroid is not a Knowledge trick, but the "Metroid hop" when they unfreeze is not obvious/intuitive. Ideally, after reading all tips, players will have all the knowledge they need to tackle whatever challenges any difficulty of seed can throw at them. (They will, of course, still need executional skill & the wisdom/experience to know *where* to apply that knowledge.) 
4. When space permits, otherwise explain individual tricks that are not obvious. Many, many tricks in Metroid Prime are not obvious even if you know the mechanics of the game. This is due to a mix of deceptive collision geometry, unintuitive ways individual mechanics can interact, and simply the ways an untrained eye would reasonably assume some things aren't possible. The Bloodflower tip is an example of this one - there's no reason to think that its standable surface extends as far as it does, and that warrants a mention regardless of difficulty or niche application. I want to stress that this goal is an ideal - I know I will not have the space in the game (or time on this earth) to mention every single unintuitive trick. However, I do not believe there's any value in a flavor scan left vanilla when it could impart something practical. Please do not suggest removing a tip if the information is accurate and there's nothing to replace it with - I do not think this is constructive, as the target audience is not exclusively beginners. (I'm open to being told "this needs to note the difficulty," though.)

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "Bloodflower",
+        (
+            "&main-color=#89D6FF;Using a double bomb jump, one can roll on top of a Bloodflower.\n\n"
+            "By using another double bomb jump on top of it, "
+            "Great Tree Chamber is reachable from Great Tree Hall without Space Jump Boots."
+        ),
+    ],
+    "2541494929": [
+        # Vault Jump Tip
+        # Replaces Main Plaza fallen stones scan
```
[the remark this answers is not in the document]
Shaleblade: That's the scan node in front of the door leading to Nursery Access, noting the stones that fell from the bridge above.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+            "&main-color=#89D6FF;When Metroid Prime's exoskeleton is weak to the Power Beam, "
+            "it can also be damaged by the &main-color=#FF3333;Flamethrower&main-color=#89D6FF;."
+        ),
+    ],
+    "2271263048": [
+        # Hive Mecha Bridge Tip
+        # Replaces Hive Mecha bridge scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        "&main-color=#89D6FF;To cross the bridge without fighting Hive Mecha, carefully walk along its right edge.",
```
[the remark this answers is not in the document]
Shaleblade: 1. Some players might find walking along the bridge easier than scan-dashing - walking at the right angle doesn't require any fast movements
2. Even if it's not useful for that player on that run, now that they know, they can use that on future runs (or tell their MW pals about it, who may find it immediately useful).

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+            "one can climb Ice Ruins East without missiles."
+        ),
+    ],
+    "3508298443": [
+        # Bloodflower Standable Tip
+        # Replaces Bloodflower scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Bloodflower",
+        (
+            "&main-color=#89D6FF;Using a double bomb jump, one can roll on top of a Bloodflower.\n\n"
```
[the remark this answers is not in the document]
Shaleblade: This falls within my fourth goal, explaining unintuitive tricks (even difficult ones). If you would need to see someone else do it or read about it to know it exists, I want to have it mentioned ingame.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "&main-color=#89D6FF;To cross the bridge without fighting Hive Mecha, carefully walk along its right edge.",
+    ],
+    "2053272936": [
+        # Boost Pipe Speed Tip
+        # Replaces Tallon Canyon halfpipe scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;When using the &main-color=#FF3333;Boost Ball&main-color=#89D6FF; in a half pipe, "
+            "hold only &image=8B0C22A7; or &image=778CCD6A; and boost at the lowest point to gain the most height."
```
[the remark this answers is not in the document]
Shaleblade: This falls underneath goals 2 (add tips that make the game easier & more fun, and I think moving faster fits both) and 3 (explain unintuitive mechanics, which this obviously fits).

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: That's fair, though I'm not sure where else to put the speed limit explanation. I'd love to find a scan near a long tunnel (like this horseshoe), but that was the only one I could find. Any ideas?

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: This falls within my third goal of explaining unintuitive mechanics, and I also just disagree that players who set high difficulties will know about all the tricks that fall within them. There's not much reason to assume there's something special about standing on frozen Metroids as a trick - I would have just assumed the trick difficulty was so high because manipulating them into a good position is hard. Plus, a player might read this, find the idea interesting, and then want to try it. I don't think every tip needs to be for the purpose of helping a player follow the logic of their seed - educating players on niche mechanics and tacitly inviting them to try weird stuff like the Metroid hop out is enough.

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: This replacement provides more specific information about how to open the tree, so I'm surprised you think it's worth removing. Not saying you're wrong, I'm just confused.

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: One of the scan points in the room is just "hey check out this frozen fish & half-pipe"

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: This fits in goals 2 & 3, and I think Main Plaza is a pretty good place for beginner-focused, low-inventory-requirement tips. I don't want to remove it, but I'm open to alternate suggestions as to where to put it.

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: I'm definitely open to moving it (I also think the Puddle Spores aren't great), but I think the Magmoor Workstation pump monitor text changes as you activate it - that, plus the fact that it's a functional hint for players, has me wary about using it instead.

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: This was added at [Miepee's request](https://ptb.discord.com/channels/914291389293027329/1305369147881816076/1307433452114677823) - the version in question is the Archipelago version of Prime, if I understand correctly

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: Like the other Game Versions comment, this was added because (IIRC) it doesn't work that way in AP.

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: I've experienced beginners assuming it had to be PBs and being surprised when normal bombs work. I wouldn't have added this otherwise.

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: Agreed, but if I'm reading PWE right, I don't think it actually does

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: Just so I understand, the doors lock when you hit invisible triggers separate from the drone fight?

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: I think explaining IS will take more space that I'm comfortable using on that trick, but I'd like to dedicate some time in the future to writing up an IS guide to spread across a couple terminals (maybe in this same room if there're scan points available.) The crate mention is good, though.

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: This replaces the logbook entry for spinners, so it's all of them

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: I don't know how to replace scans for specific entities (e.g. one Metroid, not all of them), unfortunately.

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: I thought it was just a CBJ, but I trust you'd know the trick much better than me

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: This is part of my first goal, mentioning every single Knowledge trick. The scan node is in a room with several other scan nodes containing tips - if you find one, you'll probably read them all. I think it would be reasonable to note that doing so is highly difficult, though.

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: The database says it's L-jumping?

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: I wanted to place it early in the game (given a vanilla start), but it's true that it's otherwise arbitrary. I'm open to alternate suggestions.

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: Goal 2, and for this specifically, I've seen too many people obviously not know the FAQ exist to assume people notice it frequently. It's not like there's something else competing for this spot, so I think a reminder like this is useful to cut down on the number of times people ask something in #prime-help and get answered by the FAQ bot.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "&main-color=#89D6FF;Each game in Randovania has a Frequently Asked Questions section.",
+    ],
+    "3979124144": [
+        # Spoiler Log Tip
+        # Replaces North Atrium left sculpture scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;By default, seeds in Randovania generate with spoiler logs.\n\n"
+            "Spoiler logs can be used to check what items are at specific locations, among other information."
```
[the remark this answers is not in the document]
Shaleblade: Again, Goal 2, and this is another case of "many people are incurious about programs even when it would help them."

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: Do you have a suggestion for where better to put L-jump advice? (Also, I don't know much about boostless spinner use, so I'm not sure what advice would be best there.)

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: Goal 2, and again, I think preempting "what is this?" questions + a fun fact that isn't competing for real estate is worth keeping.

randovania/games/prime1/exporter/scan_tips.py
Shaleblade: Goal 2, plus this is Landing Site - a place with a bunch of scan tips, and with one of the most foundational scan-dashes, which builds off of this information.

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        "&main-color=#89D6FF;While unmorphed, there is no sideways limit to your speed.",
+    ],
+    "3234096437": [
+        # Discord link tip
+        # Replaces Landing Site glowing spidervine scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Those seeking guidance can find "
+            "community at &main-color=#3333FF;discord.gg/mprando&main-color=#89D6FF;."
```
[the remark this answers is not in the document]
Shaleblade: Goal 2, and the lack of clickability is why I'm keeping them to short, readable URLs (nothing like youtube links)

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+    ],
+    "1430090441": [
+        # RDV MW window tip
+        # Replaces Landing Site Tallon fern scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;When playing multiworld games in Randovania, one does not need to have the multiworld "
+            "browser navigated to the related session. So long as Randovania itself "
+            "is open and connected to the game, items will be sent and received properly."
```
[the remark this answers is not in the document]
Shaleblade: Goal 2, and this is definitely something some beginners don't realize - RDV not needing you to navigate to the right MW session window is nice, but not intuitive

randovania/games/prime1/exporter/scan_tips.py
Miepee: might be worth to say "different randomizers" instead

randovania/games/prime1/exporter/scan_tips.py
Miepee: this doesn't work in vanilla / legacy mode. because this is as mentioned by sam, a randomprime QoL change.

randovania/games/prime1/exporter/scan_tips.py
toasterparty: recommended for observatory

randovania/games/prime1/exporter/scan_tips.py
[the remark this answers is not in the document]
toasterparty: recommended for observatory

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "577744447": [
+        # Video database link tip
+        # Replaces Canyon Cavern saturnine scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#89D6FF;Every video in Randovania's Metroid Prime database can be found at "
+            "&main-color=#3333FF;randovania.org/\ngames/prime1/videos&main-color=#89D6FF;."
```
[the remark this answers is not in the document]
toasterparty: recommended for observatory

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,696 @@
...
+        ),
+    ],
+    "4140247203": [
+        # SJB speed limit tip
+        # Replaces Landing Site red starburst scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "",
+        (
+            "&main-color=#FF3333;Space Jump Boots&main-color=#89D6FF; "
+            "raise the maximum vertical speed at which one can travel."
```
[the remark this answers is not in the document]
toasterparty: Technically the maximum vertical speed limit is unchanged. What changes is the magnitude of the first jump's impulse - in effect the first jump goes a tiny bit higher, but it almost never matters. I personally wouldn't have included this, but I'm not entirely opposed either.

toasterparty (changes requested):
Thank you so much for your patience shale. It's not fair you have to wait almost half a year for your hard work to be seriously considered, but it's the situation we're in - sorry for that.

I kind of get where Sam is coming from in his conservative suggestions. You end up replacing the text on a greater % of the game's scannable than I was imagining - but I also think all of the content you came up with is good. I think a good path forward is to make this a toggleable option to satisfy trickless or experienced players who want to reserve the ability to look at vanilla scan text. I like the setting enough to recommend it as enabled by default, however. I can help making the setting, migration, gui etc, if you'd like the help.

Also, I don't like representing patcher-introduced changes as "in some game versions". You don't have to worry about archipelago in this pull request, as the text in this file will never be used for archipelago. If you want to add a similar change to archipelago, then you're free to add/remove whatever you want for their implementation.

randovania/games/prime1/exporter/scan_tips.py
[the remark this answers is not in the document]
toasterparty: Beginners should be encourage to learn the dash to skip the fight. I suggest mentioning both or just the dash.

You can also mention that the fight can be escaped if accidentally started by dashing left off of a wasp to leave the way you come in. I see a lot of people reset because they didn't know this

randovania/games/prime1/exporter/scan_tips.py
toasterparty: `limit to your sideways speed` reads better than `sideways limit to your speed`

randovania/games/prime1/exporter/scan_tips.py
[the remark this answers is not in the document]
toasterparty: recommended for observatory

Miepee:
> Also, I don't like representing patcher-introduced changes as "in some game versions". 

My suggestion behind it was three-fold:
1) behaviour where randomprime is different from vanilla
2) behaviour where randomprime has had different defaults/behaviour over time
3) behaviour where rdv prime rando is different from other prime randos

magmoor lava lake for example is a case where the destroyable bomb walls were only introduced quite late - they are not a thing in vanilla, nor a thing in earlier randovania versions.
it was essentially a thing to make "hey, whatever you're reading only holds true for this specific rando version you're playing right now. everything else, be it vanilla, earlier or future randos, or different randos may have different behaviour" clear.
But maybe there's a better way to communicate that.

randovania/games/prime1/exporter/scan_tips.py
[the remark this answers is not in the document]
Miepee: any reason why specifically observatory? does that have better discoverability than landing site or....?

toasterparty:
I only care that the text the player is currently reading is accurate for the randovania version that they are currently using to read that text. Caring about past and future randomizer versions is only something that weighs on your mind because you are a developer.

Think about our FAQ - We don't require that it covers behavior in past randovania versions

Miepee:
> Caring about past and future randomizer versions is only something that weighs on your mind because you are a developer.

I don't think it's *fully* because of that.
The FAQ is very easy to see that its bound to the randovania version due to it being 1) in-app 2) easily findable and small.

The in-game tips are mentally seperated from the randovania version (you don't have any way to check during gameplay), are too spread out and difficult to easily check upon (either scan everything or read source code) and are more akin to a read-once-and-remember deal. If someone reads "you don't need thermal if you know where the thermal-pads are", and we change that behaviour, then they can't double check the scan message and are gonna be confused about it since they remember reading/being told otherwise. 
If they're being told upfront about it, that they shouldn't expect the behaviour everywhere, the there'll be less confusion about it.

But that also makes me realize that putting that into the scan messages is not a good idea, since we theoretically could change almost *anything* and that mentioning it in the GUI is the better solution.

Uncle-Reggie:
I find the opposition to helpful information being made available to players in any fashion confusing. As long as the information is factual and presented in a relevant location there should be no issue in my mind.

If thermal conduits later get patched to hard require thermal visor then we can edit the scan text. It's not like there won't be a major note in the hypothetical future RDV release where that happens.

The goal of this being provided in-game is that people already don't read the FAQ. If it gets longer, this will be even more true.

SeBiTaS908 (approved):

randovania/games/prime1/exporter/scan_tips.py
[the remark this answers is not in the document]
Miepee: resolving as per knowledge goal

randovania/games/prime1/exporter/scan_tips.py
Miepee: resolving as per knowldege goal

randovania/games/prime1/exporter/scan_tips.py
Miepee: see mqa hop comment

randovania/games/prime1/exporter/scan_tips.py
```diff
@@ -0,0 +1,694 @@
...
+SCAN_TIPS = {
+    "3671417093": [
+        # Stone Toad Standable Tip
+        # Replaces Stone Toad scan
+        "&main-color=#FF3333;Hidden message discovered.",
+        "Stone Toad",
+        (
+            "&main-color=#89D6FF;It is possible to stand on Stone Toads.\n\n\n"
+            "By slope jumping off of the spine on its back, "
+            "one can reach the top of Reflecting Pool without draining the water."
```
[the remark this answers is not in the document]
Shaleblade: That's fair. I'm inclined to avoid passive voice, in part for length. This one should now read as follows:

"It is possible to stand on Stone Toads and slope jump off their back spines.
This makes it possible to reach the top of Reflecting Pool without draining the water."

PR #7639, opened 2024-12-18
Prime: Magmoor coordinates

JeffGainsNGames:
updates all coordinates for docks, pickups, logic. items in every room nodes may have weird Z coords since I just moved them in the visualizer.

only part of #7636


Miepee (changes requested):
Please do not change the x/y/z coords. Add these as extra data.

See this comment: https://github.com/randovania/randovania/issues/7636#issuecomment-2550731786

JeffGainsNGames:
why? is this not what these values are for? Toasters thoughts?

Miepee:
> They're approximations based on how they look in rooms. For visualization purposes, the db maintainers explicitly did *not* want to have them matching with their internal coordinates because it would mean that there are several rooms where nodes completely overlap each other, thus completely nullifying the visuals.
We would need to provide the node coordinates as extra data.

As examples:
grafik
grafik
grafik
grafik
grafik

toasterparty:
Did you do this by hand or by automated ROM dump?

Miepee:
By hand.

PR #7645, opened 2024-12-19
Add new "quantity" parameter to "get_parent_for" from ODR

ThanatosGit:


henriquegemignani:
Is this missing a PR to ODR or just updating ODR?

henriquegemignani:
Biggest cost is just having some discipline that nothing is broken, mainly with meds.
I'll make some tags later

Miepee:
If we rebase rds onto latest meds, then we also need to rebase osrr, no?

henriquegemignani (approved):

PR #7670, opened 2024-12-29
Prime: Tallon Overworld world coordinates

JustinDMS:
Added new data for each node under `world_coordinates` that contains the in-game coordinate values. Coordinates were pulled from PWE and entered by hand.

Generic and Event node coordinates were made in good faith and used existing in-game object coordinates with few exceptions.

Pickup nodes used for Items Every Room have a `null` value.

Part of #7636 with considerations to requests made in #7639


Miepee:
@JustinDMS you need to run randovania, open the data visualizer and then save to update the txt representation and thus make tests pass

MrMiguel211:
Did you use the bottom X,Y,Z values in PWE or are only 4 decimals allowed in the json? Because using those values will yield incorrect results as they will not correctly be 1:1 to in-game coordinates, using the Position values from the object properties is the way to go, as they will have all 6 of their decimal values instead of just 4.

Screenshot 2024-12-29 162335

Door to gully PWE values

vs.

Object properties values


Notice how the X value is actually `-407.892761` and not `-407.8928`

Miepee:
randovania limits the xyz values iirc

MrMiguel211:
Well in that case you can ignore my previous comment, however it still won't be 1:1 to in-game coordinates

JustinDMS:
Yes, I did use the bottom values and noticed about halfway through that they were rounded.

I'll go back through and fix that and the .txt file

JustinDMS:
@Miepee How do I properly update the text file? Following the instructions to save the database reverts all the changes I made

JustinDMS:
Closing this pr. Will open a new one later that has transforms for each node in every region

PR #7689, opened 2025-01-04
Add typing to game and game_connection modules

duncathan:


randovania/game/game_enum.py
```diff
@@ -1,3 +1,6 @@
+# mypy: disable-error-code="no-redef"
+# we specifically abuse redefinition in this class to import game_module
```
Miepee, line 2: Huh, I assumed there was some way to do what we have here without redefining game_module.

randovania/game_connection/connector/am2r_remote_connector.py
```diff
@@ -170,14 +170,15 @@ def new_inventory_received(self, new_inventory_response: str) -> None:
                     "Response should contain a '|', but it doesn't. Original response: %s", new_inventory_response
                 )
                 continue
-            (item_name, quantity) = position.split("|", 1)
-            if not quantity.isdigit():
+            (item_name, q) = position.split("|", 1)
+            if not q.isdigit():
```
Miepee, line 174: does python not have something akin to `bool TryParseInt(string, out int)`?

randovania/game/game_enum.py
duncathan: there is, but we do it this way to avoid importing until runtime. it's necessary

randovania/game_connection/connector/am2r_remote_connector.py
Miepee: Apparently nothing from the standard library. Would require writing a custom function akin to 
```py
def try_parse_int(s: str): -> int | None:
  if s.isdigit():
    return int(s)
  return None
```

Which afaik we don't need right now and also doesn't solve the issue of needing to explicitly cast after

randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -7,6 +7,7 @@
 from typing import TYPE_CHECKING
 
 from open_prime_rando.dol_patching import all_prime_dol_patches
+from ppc_asm import assembler  # type: ignore[import-untyped]
```
henriquegemignani, line 10: Why did this get moved out of the type checking block?

randovania/game_connection/executor/dolphin_executor.py
```diff
@@ -2,9 +2,9 @@
 
 import platform
 
-import dolphin_memory_engine
-import pid
-from dolphin_memory_engine._dolphin_memory_engine import DolphinStatus
+import dolphin_memory_engine  # type: ignore[import-untyped]
+import pid  # type: ignore[import-untyped]
```
henriquegemignani, line 6: This is not one of our packages

henriquegemignani (changes requested):

randovania/game_connection/connector/prime_remote_connector.py
duncathan: line 39, `PatchInstructions = list[assembler.BaseInstruction]`

randovania/game_connection/executor/dolphin_executor.py
duncathan: lol yes i wrote that before my other pr

randovania/game_connection/connector/prime_remote_connector.py
duncathan: ah, turns out I can put those definitions inside the type checking block! neat. hopefully nobody ever has to `cast()` to these types tho

test/gui/widgets/test_game_connection_window.py
```diff
@@ -142,10 +142,10 @@ async def test_add_connector_builder_dread(window: GameConnectionWindow, abort,
         assert window.game_connection.add_connection_builder.call_args[0][0].ip == "my_ip"
 
 
-@pytest.mark.parametrize("system", ["Darwin", "Windows"])
+@pytest.mark.parametrize("system", ["darwin", "win32"])
 def test_setup_builder_ui_all_builders(skip_qtbot, system, mocker: MockerFixture, is_dev_version, is_frozen):
     # Setup
-    mocker.patch("platform.system", return_value=system)
+    mocker.patch("sys.platform", system)
```
henriquegemignani, line 148: Why this change?

randovania/game_connection/connector_builder_choice.py
```diff
@@ -27,13 +27,12 @@ def is_usable(self) -> bool:
...
         if self is ConnectorBuilderChoice.DOLPHIN:
-            match platform.system():
-                case "Darwin":
-                    return False
-                case "Linux" if randovania.is_frozen():
-                    return os.getuid() == 0 and not randovania.is_flatpak()
-                case _:
-                    return True
+            # using sys.platform here instead of platform.system() due to a mypy limitation
+            # see: https://github.com/python/mypy/issues/8166
```
duncathan, line 31: @henriquegemignani

henriquegemignani (approved):

PR #7692, opened 2025-01-04
minimize unnecessary use of type: ignore

duncathan:
mypy now errors on redundant ignores, as well as ignores without a specific error code declared. this makes sure we avoid ignoring unless we absolutely have to

also, external libraries without type stubs are now ignored in pyproject.toml instead of inline ignore comments. this helps make it clear that they're not our responsibility to fix and again keeps ignore comments to the absolute minimum necessary

### DIFF
--- a/pyuic.json
+++ b/pyuic.json
@@ -57,5 +57,5 @@
     "pyrcc": "pyside6-rcc",
     "pyrcc_options": "",
     "pyuic": "pyside6-uic",
-    "pyuic_options": "--star-imports"
+    "pyuic_options": "--absolute-imports"
 }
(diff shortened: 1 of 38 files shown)


pyuic.json
```diff
@@ -57,5 +57,5 @@
     "pyrcc": "pyside6-rcc",
     "pyrcc_options": "",
     "pyuic": "pyside6-uic",
-    "pyuic_options": "--star-imports"
-}
\ No newline at end of file
+    "pyuic_options": "--absolute-imports"
```
duncathan, line 60: users running mypy locally will need to run `python setup.py build_ui --force` once to rebuild UI files after this change

henriquegemignani (approved):

PR #7701, opened 2025-01-06
Spider method of crossing mining facility gamma nest access

Mr-Ledger:
Added spider method along the ceiling to both horizontal docks in mining facility gamma nest access making sure you can reach the ceiling as a ball first

Intermediate mid-air morph since the short height means you have to do the jump > morph > aim inputs in fairly quick succession

A single bomb jump inside the doorways lets you reach the ceiling

Plus disabled entrance rando since you can roll in from the adjacent rooms


JeffGainsNGames (approved):
likely will have changelog conflicts with 7700

PR #7708, opened 2025-01-08
Fix typo in Dread logic database

tom-overton:
I noticed a small typo in Dread's logic database yesterday involving this event:
image

It didn't seem intentional (as literally everywhere else, it's written as "speedboost"), so I fixed it. I have two questions about this change:
- Do I need to do a migration for this? I saw that there was `dairon_typo` migration data for Dread, but that seems to be because those strings are used for `dock_weakness`. I see no such thing for this event, which makes me think a migration is unnecessary.
- Do I need to update the changelog for this? The [contributing guide](https://github.com/randovania/randovania/blob/6012fa9dc176789bd38709182ea3febbb50cb428/CONTRIBUTING.md#example-of-changes-that-should-have-changelog) says it should be updated for "any database change for a stable game," which is what this fix is. I also see such a change in the [6.2.0 changelog](https://github.com/randovania/randovania/releases/tag/v6.2.0) where the aforementioned Dairon typo was fixed. However, I wasn't sure if this change was necessary to call out since, as I said above, it doesn't seem like it needs a migration.

Sorry for my newbie questions! I'm happy to change things if it's necessary to get this in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added: Mining Facility Alpha Nest Access now has a Spider Ball method of crossing the room from either side.
 
+### Metroid Dread
+
+#### Logic Database
+
+- Fixed: A typo in the event name Artaria - Prepare Speedboost in Map Station has been changed from Prepare Speeboost in Map Station.
+
 ## [8.8.0] - 2025-01-02
 
 - Added: Experimental preset option to place all majors or pickups logically. This makes sure pickups that are not required can be collected.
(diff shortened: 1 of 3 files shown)


Miepee:
> Do I need to do a migration for this?

No


> Do I need to update the changelog for this?

Yes

tom-overton:
I updated the changelog using the Dairon typo fix as a template; let me know if you'd like to see it look a different way. I also didn't know if the games should be in any particular order in the changelog, so I just stuck Dread at the end. Sidenote: is it intended for there to be two different AM2R sections?

Miepee:
> I also didn't know if the games should be in any particular order in the changelog

Alphabetical. E.g. am2r -> dread -> metroid primes

> Sidenote: is it intended for there to be two different AM2R sections?

no, seems like something got messed up there

tom-overton:
Thanks. I also noticed a duplicate Samus Returns section too, so I fixed both that and the AM2R one. For AM2R, when I combined the two Logic Database sections, I just put the room names in alphabetic order.

PR #7717, opened 2025-01-11
Make Dairon Power Switch 2 non dangerous and keep various other event nodes accessible after the fact

Tyranisaur:
Primary goal is to make the Freezer generator be seen as not dangerous by the logic machinery, considering that the only reason it is dangerous is that the path into the generator itself gets cut off until Gravity Suit is acquired. However, for the purposes of the resolver, having any negated requirements visible in logic, even when there's no path to them, is also enough to cause problems. Ideally it should be acceptable to hand wave away the fact that this means it's technically possible to traverse through parts of a cold room without Gravity Suit, but only because doing so doesn't lead to anything interesting.

While understanding the impact of this change, other related changes were identified, where making sure that event nodes for dangerous events can still be accessed after collecting the event itself will lead to better better results in the generator and resolver as well.


randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -1669,7 +1669,9 @@ Extra - asset_id: collision_camera_024
       Lay Small Bomb or Shoot Beam
   > Dock to Kraid Arena
       All of the following:
-          Before Kraid and Can Slide
+          Any of the following:
+              Before Kraid and Can Slide
+              Morph Ball and After Kraid and After Artaria - ArbEnky Blob East and After Cataris - Diffusion Tutorial Blob 2 and After Cataris - Blob above Kraid and After Cataris - Kraid Eyedoor Ceiling Blob and After Cataris - Kraid Eyedoor Blob
```
XenoWars, line 1674: Artaria's probably isn't needed here I'm sure, but why are the Above Kraid and Eyedoor ceiling blobs needed if they don't have any influence on the backdoor path?

randovania/games/dread/logic_database/Dairon.txt
```diff
@@ -1924,53 +1912,37 @@ Extra - asset_id: collision_camera_025
...
-              Morph Ball and Single-wall Wall Jump (Intermediate)
+      Any of the following:
+          Space Jump or Speed Booster or Simple IBJ
+          All of the following:
+              Wall Jump (Beginner)
+              Flash Shift or Use Spin Boost
+          Morph Ball and Single-wall Wall Jump (Intermediate)
   > Door to Power Switch 2
       Any of the following:
-          Gravity Suit
```
XenoWars, line 1939: Gravity Suit is still a valid option even with the power out since you can wall jump out of the water.

XenoWars (changes requested):

randovania/games/dread/logic_database/Cataris.txt
Tyranisaur: My thought process is that I want it to take this path: 
image
Using this path instead of jumping over the lava pool seemed reasonable, because there seems to be many different options to do that part. Comparatively, requiring the extra blobs seems like a reasonable shorthand here. It's a coincidennce that this path doesn't go through any doors. The Kraid Eyedoor is excluded from randomization, so I don't think there would be anything wrong with doing that.

XenoWars (approved):

PR #7728, opened 2025-01-16
small migration_lib refactor

duncathan:
something I needed for the hint rework

removes the redundant return from the migration signature, and adds an alternative function that provides a RandovaniaGame to the migrations


randovania/lib/migration_lib.py
```diff
@@ -1,36 +1,48 @@
...
     pass
 
 
-Migrations = typing.Sequence[Callable[[dict], dict] | None]
+class Migration(typing.Protocol):
+    def __call__(self, __data: dict) -> None: ...
+
+
+class GameMigration(typing.Protocol):
+    def __call__(self, __data: dict, game: RandovaniaGame) -> None: ...
```
duncathan, line 20: this is a cool thing with protocols. this protocol describes a callable (which returns `None`) with exactly two arguments, where the first argument is a `dict` and can have any name (due to starting with a double underscore), while the second argument is a `RandovaniaGame` and must be named `game`

randovania/games/dread/assets/migration_data.json
```diff
@@ -867,5 +867,32 @@
     "dairon_typo": {
         "Ferenia/East Transport to Darion/Door to Total Recharge Station": "Ferenia/East Transport to Dairon/Door to Total Recharge Station",
         "Ferenia/Total Recharge Station/Door to East Transport to Darion": "Ferenia/Total Recharge Station/Door to East Transport to Dairon"
+    },
+    "generated_pickups": {
```
henriquegemignani, line 871: This shouldn't be in this pr

henriquegemignani (changes requested):

randovania/games/dread/assets/migration_data.json
duncathan: lol ty

henriquegemignani (approved):

PR #7730, opened 2025-01-16
Include generated pickups (DNA, STKs, etc.) and their categories in the pickup database

duncathan:
another necessary refactor for my hint rework. I'm specifically interested in thoughts on:
1) the use of `{i}` as the name of the variable for string formatting
2) whether `create_generated_pickup()` should automatically increase the identifier by 1 if it's an int. every game that passes an int currently passed `i + 1` and the old functions incremented it automatically

### DIFF
--- a/randovania/cli/commands/new_game.py
+++ b/randovania/cli/commands/new_game.py
@@ -269,6 +269,7 @@ def create_pickup_database(game_enum: RandovaniaGame) -> PickupDatabase:
     }
     pickup_db = PickupDatabase(
         pickup_categories=pickup_categories,
+        generated_pickups={},
         standard_pickups={
             "Powerful Weapon": StandardPickupDefinition(
                 game=game_enum,
--- a/randovania/cli/commands/new_game.py
+++ b/randovania/cli/commands/new_game.py
@@ -22,7 +22,7 @@ from randovania.game_description.db.pickup_node import PickupNode
 from randovania.game_description.db.region import Region
 from randovania.game_description.db.region_list import RegionList
 from randovania.game_description.game_description import GameDescription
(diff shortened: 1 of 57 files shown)


duncathan:
oh and as for reasoning: currently any pickup DB migrations need to migrate the hardcoded generated pickups manually. this PR means that migrations can be automatically applied to all pickups and categories, including generated ones (the only exception being GENERIC_KEY_CATEGORY which i may be getting rid of altogether later anyway)

henriquegemignani:
>     1. the use of `{i}` as the name of the variable for string formatting
The templating stuff seems fine. It being `{i}` in particular I dislike. It should be something more explicit, maybe `template`?

>     2. whether `create_generated_pickup()` should automatically increase the identifier by 1 if it's an int. every game that passes an int currently passed `i + 1` and the old functions incremented it automatically
Hard no for me. At least it's very explicit that what you're passing is the template argument, without any magic behind it.

duncathan:
> > ```
> > 1. the use of `{i}` as the name of the variable for string formatting
> > ```
> 
> The templating stuff seems fine. It being `{i}` in particular I dislike. It should be something more explicit, maybe `template`?

What about `id`? Since it's the `identifier` argument that gets passed to it. Could also do `identifier` written in full but I kinda prefer keeping it short for ease of reading/writing

> > ```
> > 2. whether `create_generated_pickup()` should automatically increase the identifier by 1 if it's an int. every game that passes an int currently passed `i + 1` and the old functions incremented it automatically
> > ```
> 
> Hard no for me. At least it's very explicit that what you're passing is the template argument, without any magic behind it.

I felt similar, yeah. Would rather be explicit even if it means you have to remember the +1

henriquegemignani (approved):

PR #7735, opened 2025-01-18
MSR: Add Database Integrity Checks

dyceron:


randovania/games/samus_returns/db_integrity.py
```diff
@@ -0,0 +1,43 @@
...
+)
+
+# Short name resource -> recommended template name
+use_templates_over_items = {
+    "Spring": "Use Spring Ball",
+    "Spider": "Use Spider Ball",
+    "Bomb": "Lay Bomb or Lay Any Bomb",
+    "PBAmmo": "Lay Power Bomb or Lay Any Bomb",
+    "Wave": "Shoot Wave Beam",
+    "Spazer": "Shoot Spazer Beam",
```
Miepee, line 16: Plasma too or nah? Feels a but weird to have templates for anybeam except last, but i could live with it.

dyceron: The templates are there because of how beam progression works. Plasma is last so it has no special case. But, with split beams coming soon there's no real need for the templates anymore.

Miepee (approved):

PR #7737, opened 2025-01-18
Fusion: Revise Omega Metroid

Miepee:
Health:

| db | 1.5x | 
|---|---|
|500 | 750 |
|300 |450 |
|200 |300 |
|100 |150 |

### DIFF
--- a/randovania/games/fusion/logic_database/Main Deck.txt
+++ b/randovania/games/fusion/logic_database/Main Deck.txt
@@ -12,15 +12,24 @@ Extra - unlocked_save_recharge_station: True
   > Event - Omega Metroid
       All of the following:
           After Navigation Controls Activated
-          # Weapon Requirements - TODO: more beams
-          Wide Beam or Combat (Intermediate)
+          Any of the following:
+              # Weapon Requirements
+              # Trickless
+              Plasma Beam and Wide Beam
+              All of the following:
+                  # Easy 1 Beam methods
+                  Combat (Beginner)
+                  Plasma Beam or Wide Beam
+              All of the following:
+                  # Harder 1 Beam methods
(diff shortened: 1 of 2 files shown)


randovania/games/fusion/logic_database/Main Deck.txt
```diff
@@ -12,15 +12,24 @@ Extra - unlocked_save_recharge_station: True
...
               # Energy requirements
-              Combat (Hypermode) or Normal Damage ≥ 1000
-              Combat (Beginner) and Normal Damage ≥ 800
-              Combat (Intermediate) and Normal Damage ≥ 600
-              Combat (Advanced) and Normal Damage ≥ 300
-              Combat (Expert) and Normal Damage ≥ 100
+              Combat (Expert) or Normal Damage ≥ 500
+              Combat (Beginner) and Normal Damage ≥ 300
+              Combat (Intermediate) and Normal Damage ≥ 200
+              Combat (Advanced) and Normal Damage ≥ 100
```
JeffGainsNGames, line 32: Does the final fight not matter if we have higher health reqs to guarantee access to E-tanks in logic?
I'm imagining a scenario where all your upgrades are as early as possible with the least amount of game-opening progression (bombs, SJ, speed), and all the E being locked in sectors.

Miepee: SA-X has higher health reqs than the omega. We could have the whole omega fight as trivial right now and basically nothing would change.


That said, the omega is *way* easier to dodge and deal with than the omega, since its a very predicatable movement pattern. With trickless you can also essentially just spam beams really fast and not get hit at all.

JeffGainsNGames (approved):

PR #7738, opened 2025-01-18
Fix option to check if base patches is beatable with all pickups

### LINKED ISSUE #7096: Fix check_if_beatable_after_base_patches
This is a feature that's meant to be a quick way to check if random base patches, like starting location and transporters, lead to a seed being impossible regardless of how items are randomized. When this check works correctly, it enables the generator to quickly retry generating possible base patches without needing to attempt generating an item placement. This would be extra helpful for multiworlds, where the item placement would fail very late in the generation.

It's currently broken, and the prevailing theory is that something is leaking into global state what makes something fail later for the generator/resolver.

Tyranisaur:
Closes #7096 

There's a funky ownership model where the connection cache lives in the GamePatches object that is shared with later stages of generation, while the underlying connection data is not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [8.9.x] - 2025-02-??
 
 - Added: It is now possible to focus the game images in the "Games" tab via a Keyboard.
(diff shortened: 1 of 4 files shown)


henriquegemignani (approved):

henriquegemignani:
Maybe changelog? 🤔

Tyranisaur:
Maybe, I was wondering about what to put in the changelog given that there weren't anything in the changelog when the option was introduced or hidden.

PR #7745, opened 2025-01-20
Add support for launching Dolphin from within RDV

ToxicFrog:
This is a mild convenience, but more importantly, it fixes PTRACE permission issues on linux by making dolphin a child process of RDV.

Not entirely happy with the UI, but it works.

image
image
image

### DIFF
--- a/randovania/game_connection/builder/dolphin_connector_builder.py
+++ b/randovania/game_connection/builder/dolphin_connector_builder.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from subprocess import Popen, TimeoutExpired
 from typing import TYPE_CHECKING
 
 from randovania.game_connection.builder.prime_connector_builder import PrimeConnectorBuilder
@@ -11,12 +12,69 @@ if TYPE_CHECKING:
 
 
 class DolphinConnectorBuilder(PrimeConnectorBuilder):
+    dolphin_cmd: str
+    dolphin_comm: str
+    dolphin_child_process: Popen | None
+
+    def __init__(self, dolphin_cmd: str = "", dolphin_comm: str = ""):
+        super().__init__()
+        self.dolphin_cmd = dolphin_cmd
+        self.dolphin_comm = dolphin_comm
+        self.dolphin_child_process = None
+
+    @property
+    def pretty_text(self) -> str:
+        name = super().pretty_text
+
+        # If dolphin_comm is set, display both the program being run (if applicable)
+        # and the commname we will search for in the connector description.
+        if self.dolphin_comm:
+            if self.dolphin_cmd:
+                return f"{name}: run '{self.dolphin_cmd}', attach to '{self.dolphin_comm}'"
+            else:
+                return f"{name}: attach to already-running '{self.dolphin_comm}'"
+
+        # Otherwise just display the program we're running (again, if any) and
+        # rely on pyDME's compiled-in defaults for autodetection.
+        elif self.dolphin_cmd:
+            return f"{name}: run '{self.dolphin_cmd}', then autodetect"
+        else:
+            return f"{name}: autodetect using defaults"
+
     @property
     def connector_builder_choice(self) -> ConnectorBuilderChoice:
         return ConnectorBuilderChoice.DOLPHIN
 
+    def start_dolphin(self) -> None:
+        if not self.dolphin_cmd:
+            return
+
+        if self.dolphin_child_process is None:
+            self.logger.info(f"Dolphin not running, attempting to start it: {self.dolphin_cmd}")
+            self.dolphin_child_process = Popen(self.dolphin_cmd, shell=True)
+
+        elif self.dolphin_child_process.poll() is not None:
+            self.logger.info(f"Dolphin has exited, restarting it: {self.dolphin_cmd}")
+            self.dolphin_child_process = Popen(self.dolphin_cmd, shell=True)
+
+    def stop_dolphin(self) -> None:
+        if self.dolphin_child_process is not None:
+            self.logger.info("Shutting down child Dolphin process.")
+            # Send TERM twice, since the first one just opens a
+            # "are you sure you want to exit?" popup.
+            self.dolphin_child_process.terminate()
+            self.dolphin_child_process.terminate()
+            try:
+                self.dolphin_child_process.wait(timeout=1)
+            except TimeoutExpired:
+                self.dolphin_child_process.kill()
+            self.dolphin_child_process = None
+
     def create_executor(self) -> MemoryOperationExecutor:
-        return DolphinExecutor()
+        return DolphinExecutor(self.dolphin_comm)
 
     def configuration_params(self) -> dict:
-        return {}
+        return {
+            "dolphin_cmd": self.dolphin_cmd,
+            "dolphin_comm": self.dolphin_comm,
+        }
--- a/randovania/game_connection/connector_builder_choice.py
+++ b/randovania/game_connection/connector_builder_choice.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import os
 import sys
 from enum import Enum
 from typing import Self
@@ -32,7 +31,7 @@ class ConnectorBuilderChoice(Enum):
             if sys.platform == "darwin":
                 return False
             if sys.platform == "linux" and randovania.is_frozen():
-                return os.getuid() == 0 and not randovania.is_flatpak()
+                return not randovania.is_flatpak()
 
         return True
 
--- a/randovania/game_connection/executor/dolphin_executor.py
+++ b/randovania/game_connection/executor/dolphin_executor.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import platform
 
 import dolphin_memory_engine  # type: ignore[import-untyped]
@@ -24,10 +25,24 @@ def _validate_range(address: int, size: int) -> None:
 
 
 class DolphinExecutor(MemoryOperationExecutor):
-    def __init__(self) -> None:
+    dolphin_comm: str
+
+    def __init__(self, dolphin_comm: str) -> None:
         super().__init__()
         self.dolphin = dolphin_memory_engine
+        self.dolphin_comm = dolphin_comm
         self._pid = pid.PidFile("randovania-dolphin-backend")
+
+        # Do we have a custom process name to search for? If so, use this envar to
+        # override pyDME's internal defaults.
+        if self.dolphin_comm:
+            os.environ["DME_DOLPHIN_PROCESS_NAME"] = self.dolphin_comm
+
+        # Otherwise, clear it -- this is stateful, so if the user has been experimenting
+        # with Dolphin settings, it may be set from an earlier DolphinExecutor, which
+        # would result in pyDME looking for the wrong thing.
+        elif "DME_DOLPHIN_PROCESS_NAME" in os.environ:
+            del os.environ["DME_DOLPHIN_PROCESS_NAME"]
 
     @property
     def lock_identifier(self) -> str | None:
--- a/randovania/games/common/prime_family/gui/dolphin_connector_prompt_dialog.py
+++ b/randovania/games/common/prime_family/gui/dolphin_connector_prompt_dialog.py
@@ -1,0 +1,111 @@
+from __future__ import annotations
+
+from typing import Any
+
+from PySide6 import QtWidgets
+
+from randovania.gui.dialog.text_prompt_dialog import TextPromptDialog
+from randovania.gui.lib import async_dialog
+
+
+class DolphinConnectorPromptDialog(TextPromptDialog):
+    """Similar to ConnectorPromptDialog, but with more options."""
+
+    def __init__(self, **kwargs: Any):
+        super().__init__(**kwargs)
+
+        # Top line: Launch & Autodetect radio buttons
+        self.launch = QtWidgets.QRadioButton("Launch", self)
+        self.launch.setToolTip(
+            "Run a custom command under Randovania's control, then connect to it. "
+            "This may be needed to work on Linux systems with default security settings."
+        )
+        self.autodetect = QtWidgets.QRadioButton("Autodetect", self)
+        self.autodetect.setToolTip("Attempt to find and connect to an already running Dolphin.")
+        self.gridLayout.addWidget(self.autodetect, 1, 0, 1, 1)
+        self.gridLayout.addWidget(self.launch, 1, 1, 1, 2)
+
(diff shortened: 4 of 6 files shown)


randovania/game_connection/builder/dolphin_connector_builder.py
```diff
@@ -11,12 +12,62 @@
 
 
 class DolphinConnectorBuilder(PrimeConnectorBuilder):
+    dolphin_cmd: str
+    dolphin_comm: str
+    dolphin_child: Popen
+
+    def __init__(self, dolphin_cmd: str = "", dolphin_comm: str = ""):
```
ThanatosGit, line 19: We usually don't use defaults here. I assume it's for not breaking existing connectors.
Add a migration to: https://github.com/randovania/randovania/blob/main/randovania/interface_common/persisted_options.py

randovania/game_connection/builder/dolphin_connector_builder.py
```diff
@@ -11,12 +12,62 @@
...
+        self.dolphin_child = None
+
+    @property
+    def pretty_text(self) -> str:
+        if self.dolphin_comm:
+            if self.dolphin_cmd:
+                return f"{super().pretty_text}: run '{self.dolphin_cmd}', attach to '{self.dolphin_comm}'"
+            else:
+                return f"{super().pretty_text}: autodetect as '{self.dolphin_comm}'"
+        elif self.dolphin_comm:
```
ThanatosGit, line 32: Smells like unreachable code: 
```
if self.dolphin_comm:
...
elif self.dolphin_comm:
```

ThanatosGit (commented):
I'm just passing by.
When it's not in draft state anymore, make sure to add type hints where it's missing.
(methods: `start_dolphin`, `stop_dolphin`, param: `dolphin_comm`)

Eventually add a screenshot of the ui to the PR.

tools/randovania.nix
```diff
@@ -0,0 +1,58 @@
+{ stdenv
```
henriquegemignani, line 1: These are unrelated and should be added in their own pull request

henriquegemignani (changes requested):

randovania/game_connection/builder/dolphin_connector_builder.py
ToxicFrog: I based this on the DebugConnectorBuilder, which does use defaults (but looks like it's the only one that does?)

What's the best way to do this with persisted_options? It looks like changes that add new settings (rather than removing/changing existing ones) just add a new `_only_new_fields` entry, but if I understand how it works correctly, that won't work here because if someone has an existing DolphinConnector those arguments will be absent when they update, with tragic results.

So I think I need an actual migrator that sets those settings to sensible defaults (in this case `""`, i.e. the old behaviour of autodetect-default) here?

tools/randovania.nix
duncathan: highly recommend using individual branches for unrelated changes rather than doing everything on `main`

ToxicFrog: Done as #7749

randovania/game_connection/builder/dolphin_connector_builder.py
ToxicFrog: Yep, that second one was meant to be `dolphin_cmd`. Thanks.

ToxicFrog:
> I'm just passing by. When it's not in draft state anymore, make sure to add type hints where it's missing. (methods: `start_dolphin`, `stop_dolphin`, param: `dolphin_comm`)
> 
> Eventually add a screenshot of the ui to the PR.

Done type hints, added some screenshots :)

randovania/game_connection/connector_builder_choice.py
```diff
@@ -31,8 +30,6 @@ def is_usable(self) -> bool:
             # see: https://github.com/python/mypy/issues/8166
             if sys.platform == "darwin":
                 return False
-            if sys.platform == "linux" and randovania.is_frozen():
-                return os.getuid() == 0 and not randovania.is_flatpak()
```
ToxicFrog, line 35: I don't have any way of testing the flatpak version here, so I'm not 100% confident in this. I *suspect* that if dolphin is installed in a way that lets it be launched from in the flatpak, this will work, but I don't know (a) if that's possible or (b) how to do it, although passing comments on other flatpak-based software imply that it might be possible.

randovania/game_connection/builder/dolphin_connector_builder.py
ThanatosGit: > I based this on the DebugConnectorBuilder, which does use defaults (but looks like it's the only one that does?)

`DebugConnectorBuilder` is only available from source but if something changes for the frozen builds, it's better to use a migration which also bumps the version of the json config file.

> What's the best way to do this with persisted_options? It looks like changes that add new settings (rather than removing/changing existing ones) just add a new `_only_new_fields` entry, but if I understand how it works correctly, that won't work here because if someone has an existing DolphinConnector those arguments will be absent when they update, with tragic results.

Tbh, I think the connector builder options are a little bit weird compared to the other options because the field is just an arbitrary dictionary `params: dict`, which makes it impossible to just use `_only_new_fields` because there is no new field in the dataclass. (If you ask me that has a good potential of a future refactor and make it more type safe).

The migration should look like this:
```python
def _dolphin_configuration_fields(options: dict) -> None:
    for connector_builder in options["connector_builders"]:
            if connector_builder["choice"] == "dolphin":
                connector_builder["params"] = {
                    "dolphin_cmd": "",
                    "dolphin_comm": "",
                }
```

ThanatosGit (changes requested):
@Miepee Can you, as a linux user, please look at this? Is this good to go besides the two following things? 

In the meantime things to fix after ci:
`test/game_connection/executor/test_dolphin_executor` needs to be fixed because `TypeError: DolphinExecutor.__init__() missing 1 required positional argument: 'dolphin_comm'`

And `dolphin_child: Popen` needs to be `dolphin_child: Popen | None` because `None` is assigned in the constructor.

randovania/game_connection/connector_builder_choice.py
Miepee: This won't work. Unless you make your own fork of randovania-flatpak that also includes dolphin-flatpak this doesn't work because the only way for flatpak applications to launch other applications is by signalling it to the host system. At which point it does not become a child process anymore.

There is a small discussion I once had with some gnome terminal devs in #flatpak:matrix.org, from what I recall their solution was to essentially have a daemon on the host system.

randovania/game_connection/builder/dolphin_connector_builder.py
```diff
@@ -11,12 +12,62 @@
...
+        super().__init__()
+        self.dolphin_cmd = dolphin_cmd
+        self.dolphin_comm = dolphin_comm
+        self.dolphin_child = None
+
+    @property
+    def pretty_text(self) -> str:
+        if self.dolphin_comm:
+            if self.dolphin_cmd:
+                return f"{super().pretty_text}: run '{self.dolphin_cmd}', attach to '{self.dolphin_comm}'"
```
Miepee, line 29: make a variable for super().pretty_text

randovania/game_connection/builder/dolphin_connector_builder.py
```diff
@@ -11,12 +12,62 @@
...
+            else:
+                return f"{super().pretty_text}: autodetect as '{self.dolphin_comm}'"
```
Miepee, line 31: the whole function seems a bit messy to read, especially with how its all phrased, but this one case confuses me in particular.

randovania/game_connection/builder/dolphin_connector_builder.py
```diff
@@ -11,12 +12,62 @@
...
+
+        elif self.dolphin_child.poll() is not None:
+            self.logger.info(f"Dolphin has exited, restarting it: {self.dolphin_cmd}")
+            self.dolphin_child = Popen(self.dolphin_cmd, shell=True)
+
+    def stop_dolphin(self) -> None:
+        if self.dolphin_child is not None:
+            self.logger.info("Shutting down child Dolphin process.")
+            # Send TERM twice, since the first one just opens a "are you sure
+            # you want to exit?" popup.
```
Miepee, line 57: small style nit, but i'd split the comment so so that the quote is all on one line

randovania/game_connection/builder/dolphin_connector_builder.py
```diff
@@ -11,12 +12,62 @@
 
 
 class DolphinConnectorBuilder(PrimeConnectorBuilder):
+    dolphin_cmd: str
+    dolphin_comm: str
+    dolphin_child: Popen
```
Miepee, line 17: feel like there could be some better ways for dolphin_cmd and dolphin_comm, will revisit that later

dolphin_child i'd rename to dolphin_child_process for clarity.

randovania/game_connection/executor/dolphin_executor.py
```diff
@@ -24,10 +25,17 @@ def _validate_range(address: int, size: int) -> None:
...
+
+    def __init__(self, dolphin_comm: str) -> None:
         super().__init__()
         self.dolphin = dolphin_memory_engine
+        self.dolphin_comm = dolphin_comm
         self._pid = pid.PidFile("randovania-dolphin-backend")
+        if self.dolphin_comm:
+            os.environ["DME_DOLPHIN_PROCESS_NAME"] = self.dolphin_comm
+        elif "DME_DOLPHIN_PROCESS_NAME" in os.environ:
+            del os.environ["DME_DOLPHIN_PROCESS_NAME"]
```
Miepee, line 38: why is this done?

randovania/games/common/prime_family/gui/dolphin_connector_prompt_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+        self.cmd_label = QtWidgets.QLabel("Command", self)
+        self.prompt_edit.setToolTip(
+            "The command to run. This can either be an absolute path (e.g. /usr/bin/dolphin-emu) "
+            "or a plain command name (e.g. dolphin-emu)."
+        )
+        self.gridLayout.addWidget(self.cmd_label, 2, 0, 1, 1)
+        self.gridLayout.addWidget(self.prompt_edit, 2, 1, 1, 2)
+
+        # Third line, custom commname search
+        self.comm_label = QtWidgets.QLabel("Search", self)
```
Miepee, line 38: note for me to research better names than `Command` and `Search`.

randovania/games/common/prime_family/gui/dolphin_connector_prompt_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+            "This may be needed to work on Linux systems with default security settings."
+        )
+        self.autodetect = QtWidgets.QRadioButton("Autodetect", self)
+        self.autodetect.setToolTip("Attempt to find and connect to an already running Dolphin.")
+        self.gridLayout.addWidget(self.autodetect, 1, 0, 1, 1)
+        self.gridLayout.addWidget(self.launch, 1, 1, 1, 2)
+
+        # Next line, Run Command radio button, text field
+        self.cmd_label = QtWidgets.QLabel("Command", self)
+        self.prompt_edit.setToolTip(
```
Miepee, line 30: do not use tooltips for labels. Absolutely not intuitive.
Use a question-mark/info-icon button instead. See plasma-settings' Display KCM as well as our preset-settings-item-tab for reference
grafik

randovania/games/common/prime_family/gui/dolphin_connector_prompt_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self.description_label.setText(
+            "On Windows, the default of 'Autodetect' should work for most users. "
+            "If you are running Linux, you may need to experiment with these settings somewhat -- "
+            "mouse over the options for more information."
+        )
+
+        self.description_label.setWordWrap(True)
+        self.gridLayout.setColumnStretch(1, 1)
+        # Size in the .ui file isn't quite large enough for that chunky multiline
```
Miepee, line 64: confused by this comment. What .ui file?

randovania/games/common/prime_family/gui/dolphin_connector_prompt_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+        # description above.
+        self.resize(self.width(), self.sizeHint().height())
+
+    def _select_autodetect(self) -> None:
```
Miepee, line 68: I would make it so that when autodetect is enabled, it will hide the layout that contains the two labels+two lineedits

Miepee (commented):
gave this a brief look

randovania/game_connection/builder/dolphin_connector_builder.py
ToxicFrog: Ideally we'd do away with `dolphin_comm` entirely, but DME is really constraining here: it looks for a process with the right memory map *and* a commname exactly matching DME_DOLPHIN_PROCESS_NAME (or `dolphin-emu` if unset).

My preference would be to simply pass it the PID to attach to. There's also room for improving the autodetection -- look at argv0 as well as commname, or even ignore the program name entirely and just look for the memory map, which is highly distinctive.

These would not be terribly complicated features, but they'd need to be added to upstream DME (or added as a patch series on the C++ parts of pyDME and rebased whenever upstream changes are pulled in).

henriquegemignani: DME has code for supporting multiple hooked Dolphins at once and I do wish to have it exposed in pyDME. In this case you'd have the global API that works as is right now, then we could also have an API for communicating with an specific PID.

ToxicFrog: My understanding, though, is that pyDME includes (a subset of) upstream DME to handle the actual interface with dolphin, and that not diverging from upstream too much is a priority.

duncathan: pyDME is our (specifically Henrique's) library, so I trust him when he suggests we modify its API

randovania/games/common/prime_family/gui/dolphin_connector_prompt_dialog.py
henriquegemignani: I was going to actually ask that this dialog is made with an UI file. There's way too much widget creation here that isn't dynamic.

henriquegemignani (changes requested):

Miepee:
@ToxicFrog With the existance of [DME-daemon](https://github.com/randovania/dolphin-memory-engine-daemon) is this still something you're interested in?

PR #7746, opened 2025-01-20
Fusion: Ridley revision

Miepee:
A few things to hopefully explain the thought process a little better:
- i removed the space jump requirement to fight ridley, since its barely useful for casual play anyway
- wide+plasma only has been bumped to intermediate combat because due to the lack of dps it can be fairly easy to get a ton of damage accumulated by ridley
- space jump cheese strat is currently charge only. Supposedly possible with missiles too, but since I haven't heard of anyone actually doing this yet, so will probably be added later when we have a video
- to my knowledge the only way to do it hitless is with vanilla+ beams. may be possible with lower beams (or missiles only) but for me it seemed too unpredictable for now.
- i ran 1k seeds on this, ridley now had progression of ~6% (where "progression" means "appeared in the gen order")

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 1 SRX.json
+++ b/randovania/games/fusion/logic_database/Sector 1 SRX.json
@@ -6848,7 +6848,7 @@
                                                 {
                                                     "type": "and",
                                                     "data": {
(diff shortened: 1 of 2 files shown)


SchwartzGandhi (approved):
Everything looks in order, nothing jumps out at me immediately.
I haven't done much boss logic so if Jeff says its fine, I'll believe him.

PR #7749, opened 2025-01-20
Add nix/nixos support files.

ToxicFrog:
These will only be of use to people running nixos (or running not-nixos but with a nix-centric development environment) who want to use and/or hack on RDV, but since that includes me and I've done the work anyways, I might as well publish.

It might be possible to upstream `randovania.nix` into nixpkgs so that people can install it via the package manager, but it's not ready for that yet -- at minimum it needs additional metadata marking it not cross-compileable and whatnot.

### DIFF
--- a/tools/shell.nix
+++ b/tools/shell.nix
@@ -1,0 +1,70 @@
+# Development shell environment for Randovania on Nix.
+#
+# This will:
+# - create the venv and install all requirements, if it doesn't already exist
+# - set up $LD_LIBRARY_PATH with the system libraries needed for RDV
+# - define $RANDOVANIA_PATH as the root of the RDV source checkout
+# - define `rdv` as a shell function that runs RDV from the source root
+#
+# So, basic usage is:
+# $ nix-shell tools/shell.nix
+# [...long wait while it installs requirements, if this is the first time]
+# [...do some coding...]
+# $ rdv
+#
+# Or, if you just want to run rdv as a one-off:
+# $ nix-shell tools/shell.nix --run rdv
+
+{ pkgs ? import <nixpkgs> {} }:
+
+let
+  libs = with pkgs; [
+    dbus
+    fontconfig
+    freetype
+    glib
+    libgcc.lib
+    libGL
+    libxkbcommon
+    libz
+    wayland
+    xcb-util-cursor
+    xorg.libxcb
+    xorg.xcbutilimage
+    xorg.xcbutilkeysyms
+    xorg.xcbutilrenderutil
+    xorg.xcbutilwm
+    xorg.libX11
+    zstd
+  ];
+in pkgs.mkShell {
+  name = "randovania-dev-shell";
+
+  packages = with pkgs; [
+    fontconfig
+    python312
+    python312Packages.pip
+  ];
+
+  shellHook = ''
+    cd "${toString ./..}" || exit 1
+    [[ -f randovania.spec ]] || exit 1
+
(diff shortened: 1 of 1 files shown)


henriquegemignani (dismissed):

tools/randovania.nix
```diff
@@ -0,0 +1,58 @@
...
+, libxkbcommon
+, lttng-ust_2_12
+, SDL2
+, qt6
+, xorg
+}:
+
+stdenv.mkDerivation rec {
+  pname = "randovania";
+  version = "8.8.0";
```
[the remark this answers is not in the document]
ToxicFrog: The correct way to do this is to make a package definition that just builds at HEAD, yeah. The problem is that this requires making the build hermetic and reproduceable, and that's work that hasn't been done for most of the packages in RDV's huge dependency closure, and that I don't have the time (or, in many cases, the expertise) to do.

In cases like this we usually just do the above and package the binary release, which works fine for the package repository where it's ok if it there's a bit of lag between official releases and updates int he repo, but is, as you point out, a bit of a land mine here where people will naturally expect it to always build at HEAD.

I would like to revisit this at some point, but for now I'll pull it from this PR and perhaps try to get it into nixpkgs separately.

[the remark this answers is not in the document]
Miepee: How bad would it be to always point to the latest release and ignore the sha (or fetch it from gh if they provide it?) while the file is in our repo?

tools/shell.nix
```diff
@@ -0,0 +1,70 @@
...
+    libxkbcommon
+    libz
+    wayland
+    xcb-util-cursor
+    xorg.libxcb
+    xorg.xcbutilimage
+    xorg.xcbutilkeysyms
+    xorg.xcbutilrenderutil
+    xorg.xcbutilwm
+    xorg.libX11
```
Miepee, line 37: curios, why are both wayland and x11 deps specified? Just as a fallback if someone runs this on an x11 system?

ToxicFrog: Yeah. RDV uses both (since it supports both wayland and X11), so both need to be in the closure.

tools/randovania.nix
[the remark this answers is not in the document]
ToxicFrog: Building without an input hash is only practically supported for local builds (i.e. where the source is a file or directory on disk already) -- turning it off for fetchurl won't work unless reproduceable builds are turned off entirely, which no-one does.

Similarly, there's no URL in this project that's always "the latest release tarball" because the version number is part of the tarball name.

If there's some kind of release automation machinery this could be wired into, which could automatically update the the version number and hash once the tarball is generated, that could work.

tools/shell.nix
```diff
@@ -0,0 +1,70 @@
+# Development shell environment for Randovania on Nix.
+#
+# This will:
+# - create the venv and install all requirements, if it doesn't already exist
+# - set up $LD_LIBRARY_PATH with the system libraries needed for RDV
+# - define $RANDOVANIA_PATH as the root of the RDV source checkout
+# - define `randovania` as a shell function that runs RDV from the source root
```
Miepee, line 7: Comment seem wrong? It doesnt define `randovania` as a shell function, it defines `rdv`. Unless theres an alias somewhere im not seeing

ToxicFrog: Whoops, you're right! That went through a few variations before I settled on `rdv`.

Miepee (approved):

PR #7751, opened 2025-01-21
Add DLR exclusions to Prime 3

SchwartzGandhi:
Excludes all hangar doors, teleporters, and cutscene transitions from DLR


duncathan (changes requested):
this is all wrong. there is no reason to be changing the morph ball doors, open passages, or teleporters. from what I understand the problem with the docks currently marked as normal doors is that they're... not. if that's the case, do it properly and make a new dock type for them

PR #7496, opened 2024-11-07
Update starter_preset.rdvpreset

Swaraj2259:
Added description.
Fix of the issue #7493

PR #7762, opened 2025-01-23
Update starter_preset.rdvpreset

Swaraj2259:
#7496


dyceron:
Why did you make a new pull request for this? You could have just updated the one you already have open.

PR #7764, opened 2025-01-23
Fusion: Sector 4 Logic part 2

Miepee:


JeffGainsNGames:
Why no 'has any suit on this one`?

Miepee:
Specify please
grafik

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -322,7 +325,7 @@ Extra - room_id: [6]
                       Stand On Frozen Enemies (Beginner) and Can Freeze Enemies With Any Weapon and Can Use Bombs
                       Mid-Air Morph (Beginner) or Can Use Springball
                   All of the following:
-                      # Reach blocks via PBs
+                      # Reach blocks via PBs. Let crab crawl to top of bomb blocks, freeze there, lay PB, refreeze when its falling down.
```
SchwartzGandhi, line 328: sounds like this needs either a vid or a TODO for one

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -929,11 +929,11 @@ Extra - room_id: [20]
       Any of the following:
           After Sector 4 Water Level Lowered
           All of the following:
-              # Damage Reqs
+              # Damage Run
               Damage Runs (Beginner)
               Any of the following:
                   Electrified Water Damage ≥ 31
-                  Gravity Suit and Electrified Water Damage ≥ 8
+                  Electrified Water Damage ≥ 8 and Have Any Suit Upgrade
```
SchwartzGandhi, line 936: This applied to all of these:
I dont think lumping in varia suit with these numbers via the template is a good idea.
Gravity suit give you extra mobility, which for certain rooms, lowers the amount of damage you take substantially.
Unless you're implying damage logic accounts for this somehow

SchwartzGandhi (changes requested):

randovania/games/fusion/logic_database/Sector 4 AQA.txt
Miepee: Jeff also raised the concern to me
Turns out that no, Gravity doesn't substantially lower the amount of damage you take compared to Varia

I tried this out in a couple of the rooms that require you to take 40+ damage as well as 90 damage. They were always close to each other in terms of how much damage i took. Definitely feasible within the 1.5x multiplier we have, The Drain Pipe IIRC needs either either very tight movement for 1x multiplier, or we increase the required damage by 5 or something.


Lumping these together just makes things way more readable than having different damage reqs for varia and gravity. Especially with how neglible the electric water damage is.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1999,10 +1999,13 @@ Extra - room_id: [40]
...
-              Missiles ≥ 2 and Morph Ball
+              All of the following:
+                  Missiles ≥ 2
+                  # Either move to the Scizers or precisely shoot from the outside
+                  Morph Ball or Movement (Intermediate)
           Any of the following:
               # Kill the Golden Scisers
-              Charge Beam or Missiles
+              Charge Beam or Missiles ≥ 5
               Gravity Suit and Screw Attack
```
shark20061, line 2009: Only works with the Morph Ball + Missiles route

Wave Beam + Gravity Suit + Screw Attack would pass this logic, but would not be enough to allow you through the room
(Missing Morph Ball + Missiles, or Charge Beam, as Charge Beam + Wave Beam alone is sufficent)

(It seems to have been wrong originally too)

shark20061 (changes requested):
Some logic breaks with these changes, though some of it was broken before the changes

randovania/games/fusion/logic_database/Sector 4 AQA.txt
Miepee: we have a vid by this from someone on the fusion server but it doesn't meet our quality standards. I can add in a todo to redo it.

Miepee:
Addressed reviews

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1999,11 +2003,22 @@ Extra - room_id: [40]
...
+              # Kill from outside with Wave
+              Charge Beam and Wave Beam
+              All of the following:
+                  # Get into the cage
+                  Missiles ≥ 2 and Morph Ball
+                  Any of the following:
+                      Charge Beam or Missiles ≥ 5
+                      Gravity Suit and Screw Attack
+                  # Kill from outside with Missiles
+                  Missiles ≥ 10 and Movement (Intermediate)
```
shark20061, line 2021: This is one level too deep, it's an alternative to charge + wave or going inside

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1531,12 +1533,14 @@ Extra - room_id: [33]
...
+              Damage Runs (Beginner) and Electrified Water Damage ≥ 11 and Have Any Suit Upgrade
           Any of the following:
               # Get Up
               Space Jump
-              # https://youtu.be/5cASpPGdiig
-              Hi-Jump and Wall Jump (Intermediate)
+              All of the following:
+                  # WJ+HJ: https://youtu.be/5cASpPGdiig
+                  Hi-Jump and Wall Jump (Intermediate)
+                  Gravity Suit or After Sector 4 Water Level Lowered
```
shark20061, line 1543: The comment also applied to space jump, as it doesn't work underwater without gravity either

shark20061 (changes requested):
Couple tweaks needed

shark20061 (approved):
Looks good

PR #7766, opened 2025-01-24
Fusion: more small main deck fixes

Miepee:


randovania/games/fusion/logic_database/Main Deck.txt
```diff
@@ -727,8 +727,8 @@ Extra - room_id: [22]
   > Other to Habitation Deck (Upper)
       Any of the following:
           Space Jump or Can Freeze Enemies With Any Weapon
-          # TODO: check if you can do this with jump extneds
-          Wall Jump (Beginner) and Can Single Walljump
+          # Kill Namihe's and then SWJ up. TODO: check if you can do this with jump extneds
+          Wall Jump (Beginner) and Can Kill Tough Beam-Resistant Enemy and Can Single Walljump
```
SchwartzGandhi, line 731: would screw attack eventually be implied to kill the Namihes?

randovania/games/fusion/logic_database/Main Deck.txt
```diff
@@ -1343,8 +1343,9 @@ Extra - room_id: [47]
           All of the following:
               # Destroy Blocks
               Can Use Power Bombs
-              # PBs only requires two because one doesn't destroy all of them.
+              # PBs-only requires two because one doesn't destroy all of them.
```
SchwartzGandhi, line 1346: why did you add a hyphen?

randovania/games/fusion/logic_database/Main Deck.txt
```diff
@@ -571,7 +571,7 @@ Extra - room_id: [18]
       After Main Deck Power Bomb Geron Destroyed
   > Other to Western Hub
       Any of the following:
-          Morph Ball and Can Break Single Bomb Blocks
```
SchwartzGandhi, line 574: I know this is technically handled by the dock node internally, but is there any other good reason to remove it? Im fairly certain he have a tonne of other connections like this elsewhere.

randovania/games/fusion/logic_database/Main Deck.txt
SchwartzGandhi, line 1346:
```suggestion
              # These blocks requires two PBs because one doesn't destroy all of them.
```

randovania/games/fusion/logic_database/Main Deck.txt
Miepee: Not necessarily because Can Single Walljump may be an item and not backed by Screw.

randovania/games/fusion/logic_database/Main Deck.txt
Miepee: To make it clear that this is applying to if you "only have PBs". Not that if you have PBs, you "only require two".

Miepee: That's better phrasing then what I did with the hyphen lol

randovania/games/fusion/logic_database/Main Deck.txt
Miepee: The reason here was that the shinesparking case didn't have it.
And so my solution was either:
- Explicitly add it to the speedbooster case, which makes it even harder to read
- Make a `ALL: Morph and (ANY: Can Break Single Bomb Blocks or (Speedbooster shenanies))` which adds further nesting and is less readable
- Remove the morph from the break single bomb block case and make it implied that you need morph due to the dock type.

SchwartzGandhi: fwiw I dont think it would be that unreadable, we definitely have worse connections lol.
Won't debate on it further tho, works for me.

randovania/games/fusion/logic_database/Main Deck.txt
zarakava: In this case, drop the s on requires

Miepee: addresed

SchwartzGandhi (approved):

PR #7767, opened 2025-01-24
Generator: Move check if beatable check to after assigning pool results

Tyranisaur:
Games with pre placed key items such as Dread need these items to be accessible in order to beat the game. Considering that these are placed without checking if the locations are accessible, it stands to reason that the check_if_beatable logic should run after these are placed, to reject base patches that make these key items inaccessible even when having everything else in the item pool.

### DIFF
--- a/randovania/generator/generator.py
+++ b/randovania/generator/generator.py
@@ -62,9 +62,7 @@ def _validate_pickup_pool_size(
 
 
 async def check_if_beatable(patches: GamePatches, pool: PoolResults) -> bool:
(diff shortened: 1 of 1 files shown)


Tyranisaur:
> Something tells me there was a reason for doing it the old way, but I can't figure out why nor find any possible issues with the change so let's go.

🤷 I figured that I had to remove the assign own pickups because otherwise there's an assert if you try to assign something that was assigned previously. Because it's in the other order now, it doesn't need to do that the second time.

PR #7770, opened 2025-01-24
Dread: Add preset option to remove all light sources in a region

MayberryZoom:
As titled, when this option is enabled all light actors within a scenario will be deleted. This is a preset option rather than a cosmetic option, as it provides a substantially different gameplay experience. Enabling it makes many elements of the game very difficult to see, primarily items.

This also includes an open-dread-rando bump from 2.15.0 -> 2.16.0.

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,7 +82,7 @@ exporters = [
     "factorio-randovania-mod>=0.5.0",
 
     # Metroid Dread
-    "open-dread-rando>=2.15.0",
+    "open-dread-rando>=2.16.0",
 
     # Metroid Prime 1
     "py_randomprime>=1.27.0",
--- a/randovania/games/dread/exporter/patch_data_factory.py
+++ b/randovania/games/dread/exporter/patch_data_factory.py
@@ -44,18 +44,6 @@ _ALTERNATIVE_MODELS = {
     PickupModel(RandovaniaGame.METROID_DREAD, "PROGRESSIVE_SPIN"): ["powerup_doublejump", "powerup_spacejump"],
 }
 
-_SCENARIO_IDS = {
-    "artaria": "s010_cave",
-    "cataris": "s020_magma",
-    "dairon": "s030_baselab",
-    "burenia": "s040_aqua",
-    "ghavoran": "s050_forest",
-    "elun": "s060_quarantine",
-    "ferenia": "s070_basesanc",
-    "hanubia": "s080_shipyard",
-    "itorash": "s090_skybase",
-}
-
 
 def get_item_id_for_item(item: ItemResourceInfo) -> str:
     if "item_capacity_id" in item.extra:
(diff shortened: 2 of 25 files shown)


randovania/games/dread/exporter/patch_data_factory.py
```diff
@@ -44,6 +44,18 @@
...
+    "cataris": "s020_magma",
+    "dairon": "s030_baselab",
+    "burenia": "s040_aqua",
+    "ghavoran": "s050_forest",
+    "elun": "s060_quarantine",
+    "ferenia": "s070_basesanc",
+    "hanubia": "s080_shipyard",
+    "itorash": "s090_skybase",
+}
+
```
ThanatosGit, line 58: That's just the region name to scenario id. I don't like having data doubled.
Replace it with the suggestion in the other comment

randovania/games/dread/exporter/patch_data_factory.py
```diff
@@ -486,6 +498,16 @@ def _tilegroup_patches(self):
             }
         ]
 
+    def _light_patches(self):
+        config = self.configuration.disabled_lights.as_json
+        patches = []
+
+        for region, is_disabled in config.items():
+            if is_disabled:
+                patches.append({"scenario": _SCENARIO_IDS[region], "actor_layer": "rLightsLayer", "method": "all"})
```
ThanatosGit, line 507:
```suggestion
        for region_name, is_disabled in config.items():
            if is_disabled:
                scenario_id = self.game.region_list.region_with_name(region_name.capitalize()).extra.get("scenario_id")
                patches.append({"scenario": scenario_id, "actor_layer": "rLightsLayer", "method": "all"})
```

pyproject.toml
```diff
@@ -82,7 +82,7 @@ exporters = [
     "factorio-randovania-mod>=0.5.0",
 
     # Metroid Dread
-    "open-dread-rando>=2.15.0",
+    "open-dread-rando>=2.16.0",
```
ThanatosGit, line 85: That's hard required for the changes.
Add it to the `requirements.txt`

ThanatosGit (changes requested):
Would be nice to also add tests for it.
IMO it is sufficient to just modify an existing test to use this new option.

Tests needs to be fixed. No idea what's the issue because GH is slow in showing `133865` lines.

randovania/games/dread/exporter/patch_data_factory.py
MayberryZoom: Thanks, I was looking for something like this method but couldn't find it.

pyproject.toml
MayberryZoom: Well I did, but apparently per #7740 it's already in main. It's part of my commit but it isn't showing up in this PR because of that, lol

MayberryZoom:
Yeah, I totally forgot to update the test files. Good catch

pyproject.toml
ThanatosGit: Oops, my fault. Haha, I merged that. Now I remember that my thought was "it only adds a (currently) unused feature and I can just merge it without a changelog"

ThanatosGit (approved):

PR #7781, opened 2025-01-29
Fusion: Rework sector 2 eastern shaft

Miepee:
Reason for WJ intermediate on most is due to you being on a time frame until enemy unfreezes
Reason for freezing intermediate on most is that you more or less have one attempt at freezing the first kihunter in the correct position and if that fails you need to reload the room because its then too low. Since its initial descent happens relatively quickly, if you're not prepared for it it will catch you off-guard.

Due to the room change most if not all of the old videos are now outdated. I'd need to search which ones still apply and which ones don't.

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.json
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.json
@@ -7783,6 +7783,18 @@
                                     {
                                         "type": "or",
                                         "data": {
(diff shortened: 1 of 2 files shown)


shark20061 (commented):
Do you want to leave TODO notes for the videos that were removed for now?

Miepee:
Wasn't planning to. I have a script that gets all places that need a vid, and these show up there.

shark20061 (approved):

PR #7796, opened 2025-02-01
add CLI for features-per-node

JeffGainsNGames:
spits out 
1) num of features per pickup node
2) num of pickup nodes with certain amounts of pickups
3) all pickup nodes with 0 features

image

### DIFF
--- a/randovania/cli/database.py
+++ b/randovania/cli/database.py
@@ -9,6 +9,7 @@ from typing import Any
 
 from randovania.game.game_enum import RandovaniaGame
 from randovania.game_description import default_database, trick_documentation
+from randovania.game_description.db.pickup_node import PickupNode
 from randovania.game_description.resources.search import MissingResource, find_resource_info_with_long_name
 from randovania.game_description.trick_documentation import TrickUsageState
 from randovania.games import binary_data, default_data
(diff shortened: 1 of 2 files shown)


randovania/cli/database.py
```diff
@@ -438,6 +439,54 @@ def trick_usage_documentation_command(sub_parsers: _SubParsersAction) -> None:
...
+def features_per_node_command_logic(args: Namespace) -> None:
+    """Reports features for given pickup nodes, taking into account area and node specific features"""
+    db = load_game_description(args)
+    pickup_node_count = 0
+    featureless_nodes = []
+    no_features = 0
+    one_features = 0
+    two_features = 0
+    threeplus_features = 0
+    for region in db.region_list.regions:
```
Miepee, line 451: since you never use this, you could also just do `for area ion db.region_list.all_areas:`

randovania/cli/database.py
```diff
@@ -438,6 +439,54 @@ def trick_usage_documentation_command(sub_parsers: _SubParsersAction) -> None:
...
+                    else:
+                        threeplus_features += 1
+                    print(area.name + "/" + node.name + ": " + str(len(all_features)))
+    print("no features:     " + str(no_features))
+    print("1 feature :      " + str(one_features))
+    print("2 features:      " + str(two_features))
+    print("3+ features:     " + str(threeplus_features))
+    print("total pickups: /" + str(pickup_node_count))
+    print("-----------------------------")
+    print("featureless nodes: " + str(featureless_nodes))
```
Miepee, line 478: use fstrings

randovania/cli/database.py
```diff
@@ -438,6 +439,54 @@ def trick_usage_documentation_command(sub_parsers: _SubParsersAction) -> None:
...
+                            all_features.append(hint)
+                    if len(all_features) == 0:
+                        no_features += 1
+                        featureless_nodes.append(str(area.name) + "/" + str(node.name))
+                    elif len(all_features) == 1:
+                        one_features += 1
+                    elif len(all_features) == 2:
+                        two_features += 1
+                    else:
+                        threeplus_features += 1
```
Miepee, line 470: seems better as a [match statement](https://docs.python.org/3/reference/compound_stmts.html#overview)

Miepee (commented):
Having one test would be nice to ensure it doesn't break

JeffGainsNGames:
not sure what I am supposed to assert for the proposed test

Miepee:
Either just make a test that only runs this to ensure this function itself doesnt break weirdly

Or if you wanna go very fancy, use a test DB (BDG or making one yourself) that makes sure that everything gets printed out correctly.

duncathan (approved):

PR #7802, opened 2025-02-02
Fusion: Sector 2 cleanup

Miepee:
There is one shinespark that I discovered:
grafik
Modelling this one accurately is bad.
We cant have the connection in any node in Central shaft, because that means the generator will be able to take the connection despite you not being able to in-game.
Adding an event for this in puyo palace is meh.
And adding a seperate helper dock that warps you to the beside drop node is also meh.

So I decided to not put it into logic.
If anyone knows on how to include that one somewhat nicely, lemme know.


randovania/games/fusion/logic_database/Sector 2 TRO.txt
```diff
@@ -1645,7 +1670,13 @@ Extra - room_id: [50]
   * Extra - door_idx: (114,)
   > Pickup (Power Bomb Tank)
       Any of the following:
-          Morph Ball and Can Freeze Enemies With Any Weapon
+          All of the following:
+              # Vanilla way: freeze rippes and get across
```
JeffGainsNGames, line 1674: rippers

JeffGainsNGames (approved):

PR #7677, opened 2025-01-02
MP3 Add handling for Phaaze skip

SchwartzGandhi:
This was a feature I added for the Schwork, but figured it would be nice to have support for it in RDV as well.

PR #7820, opened 2025-02-05
MP3 port more CRV stuff to RDV

SchwartzGandhi:
relies on #7677 

This adds an additional configuration option, `MP3Update`, that cannot be altered by players since that is what the Schwork is for. This should also allow Caretaker/Staging to read CRV rdvgames again.

Also ports the function that prevents starting with ship grapple


dyceron:
Closing as a result of #7991

PR #7824, opened 2025-02-05
display correct region name in LocationHintFeatureTab

duncathan:
image


henriquegemignani (approved):
I don't really like showing nodes with `/` separators in end-user UI, especially without spaces.

duncathan:
> I don't really like showing nodes with `/` separators in end-user UI, especially without spaces.

is there an existing function for displaying them nicely?

henriquegemignani:
> > I don't really like showing nodes with `/` separators in end-user UI, especially without spaces.
> 
> is there an existing function for displaying them nicely?

`RegionList.area_name` does it, but doesn't include the node.
Goes hand-hand that I also find that listing region/area/node all at once to be very overwhelming, especially because our pickup nodes are rather... hard to read.

I think the tab itself would work great if it excluded pickups when it's just area nodes at least.

PR #7831, opened 2025-02-06
Reorder fusion goal bosses

Miepee:
No screenshot because i did this from phone


Miepee:
Yeah they already were (minus ridley) and I think it makes things slightly easier to track down if someone uses it as a resource.
E.g. we had arachnus+yakuza grouped together and zazabi+nettori grouped together

(Do need to fix the pr though because idk why tests are failing. Prboably the beginning newline?)

JeffGainsNGames (approved):
Happy with change once tests fixed

PR #7839, opened 2025-02-06
Rebalances keybearer hints

duncathan:
the plan I executed on:

> 1) add PB/missile expansions to chozo tech, and beam ammo expansions to luminoth tech
> 2) remove the old "missile-related", "beam-related", etc. categories
> 3) change keybearers to use a low-precision feature. this should mean almost all keybearer hints say "a key", "some chozo tech", or "some luminoth tech". this is a slight nerf but honestly not by very much at all. you'll still have a pretty clear picture of "that's probably X expansion, but has a small chance of being Y major"

also updates the hints tab:
image


test/server/multiplayer/test_world_api.py
```diff
@@ -97,7 +97,7 @@ def test_emit_world_pickups_update_one_action(
     # Run
     world_api.emit_world_pickups_update(sa, w1)
 
-    # Uncomment this to encode the data once again and get the new bytefield if it changed for some reason
+    # # Uncomment this to encode the data once again and get the new bytefield if it changed for some reason
```
JeffGainsNGames, line 100: Superfluous #

duncathan: this is actually more consistent with other instances of this pattern across the test suite lol

randovania/generator/hint_distributor.py
```diff
@@ -468,6 +473,11 @@ def get_hint_precision(
             if self.use_detailed_location_precision:
                 additional_loc_precisions.append(HintLocationPrecision.DETAILED)
 
+            if isinstance(precision.location, SpecificHintPrecision):
```
henriquegemignani, line 476: I don't like these 4 lines being duplicated in the two blocks

henriquegemignani (changes requested):

randovania/generator/hint_distributor.py
```diff
@@ -510,21 +522,27 @@ def get_hint_precision(
...
         player_pools: Sequence[PlayerPool],
+        hint_kinds: Container[HintNodeKind] = set(HintNodeKind),
     ) -> GamePatches:
         """
         Adds precision to all hints that are missing one.
         :param player_state:
         :param patches:
         :param rng:
+        :param player_pools:
+        :param hint_kinds: Which `HintNodeKind`s should be replaced?
```
henriquegemignani, line 536: This is still weird. The function is all about touching the `Hint` objects, but these don't have a `HintNodeKind`.

duncathan: not really? the function is all about touching `patches.hints`, which is a `dict[NodeIdentifier, Hint]`. both key and value are important here

henriquegemignani (approved):

PR #7851, opened 2025-02-07
Fusion: Support custom room names

Miepee:


randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -306,6 +306,20 @@ def create_useless_pickup(self) -> PickupEntry:
             model_name="Empty",
         )
 
+    def _create_room_names(self):
```
JeffGainsNGames, line 309: does this need a `-> dict`

Miepee: it does, thanks for catching.

PR #7852, opened 2025-02-07
Add HintSuitability to merge interesting/less_interesting_to_hint

henriquegemignani:


randovania/generator/hint_distributor.py
```diff
@@ -139,6 +140,16 @@ def choose_feature(
...
 
 
+class HintSuitability(enum.Enum):
+    """
+    How interesting a pickup is for hinting.
+    """
+
+    INTERESTING = enum.auto()
+    LESS_INTERESTING = enum.auto()
+    NOT_INTERESTING = enum.auto()
```
duncathan, line 150: I'd make this an `IntEnum` with explicit values, so that `INTERESTING > LESS_INTERESTING > NOT_INTERESTING`

randovania/games/prime2/generator/hint_distributor.py
```diff
@@ -36,8 +36,13 @@ def num_joke_hints(self) -> int:
         return 2
 
     @override
-    def less_interesting_pickup_to_hint(self, pickup: PickupEntry) -> bool:
-        return not pickup.has_hint_feature("key")
+    def hint_suitability_for_pickup(self, pickup: PickupEntry) -> HintSuitability:
+        if pickup.show_in_credits_spoiler:
+            return HintSuitability.INTERESTING
+        elif not pickup.has_hint_feature("key"):
+            return HintSuitability.LESS_INTERESTING
```
duncathan, line 43: you've introduced a bug here. keys are shown in the credits spoiler, so now they'll always be considered INTERESTING. that's why I designed these functions the way I did, so that interesting was a strict subset of less interesting

duncathan (changes requested):
I don't like how much easier it is to make mistakes with this setup (as evidenced by you making them). I agree the names could be improved but I'm not sure this is the right way

duncathan:
alternative: combine the approaches
```py
@final
def hint_suitability_for_pickup(self, pickup: PickupEntry) -> HintSuitability:
    if not self.interesting_pickup_to_hint(pickup):
        return HintSuitability.UNINTERESTING
    if not self.more_interesting_pickup_to_hint(pickup):
        return HintSuitability.INTERESTING
    return HintSuitability.MORE_INTERESTING
```
(where less_interesting is now interesting, and interesting is now more_interesting, for clarity)

duncathan:
this way subclasses can define the criteria for interesting and more interesting, but they *cannot* override the rule that MORE_INTERESTING >= INTERESTING >= NOT_INTERESTING

henriquegemignani:
> alternative: combine the approaches
> 
> ```python
> @final
> def hint_suitability_for_pickup(self, pickup: PickupEntry) -> HintSuitability:
>     if not self.interesting_pickup_to_hint(pickup):
>         return HintSuitability.UNINTERESTING
>     if not self.more_interesting_pickup_to_hint(pickup):
>         return HintSuitability.INTERESTING
>     return HintSuitability.MORE_INTERESTING
> ```
> 
> (where less_interesting is now interesting, and interesting is now more_interesting, for clarity)

I actually thought of doing this, but I don't like the `interesting_pickup_to_hint` and `more_interesting_pickup_to_hint` as method names.

duncathan:
> I actually thought of doing this, but I don't like the interesting_pickup_to_hint and more_interesting_pickup_to_hint as method names.

I'm ok with changing the names, it's just important to mii that the ordering can't be violated

henriquegemignani:
#7864 is a better implementation

PR #7856, opened 2025-02-07
Improve the appearance of the hint feature tabs

duncathan:
Location features now use database links, and omit the pickup name if all pickups in the area have the feature.

Also, features can define a description which gets displayed at the top of their foldable in the feature tab.

image


henriquegemignani (approved):

Miepee:
I still think that having each entry as a bullet point makes them easier to pass.
But I won't block merging over it.

PR #7861, opened 2025-02-08
Fusion: Sector 5 revamp

Miepee:


randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -916,14 +923,18 @@ Extra - room_id: [24]
   > Door to Kago Speedway
       Space Jump
   > Door to Groznyj Grad
+      # Kago disapears once you have PBs
```
JeffGainsNGames, line 926: kago never appears ?

Miepee: No, the kago is there but small.
Once you collect PBs, it dissapears.

The kago will never grow to its purple variant.

Miepee: Here is the small variant for example on mostly-current main
grafik

JeffGainsNGames (approved):

PR #7863, opened 2025-02-08
A couple quick fixes to hint distribution

### LINKED ISSUE #7858: Generation error if an ETM is on a flying ing cache
hint_distributor.py, line 508: `item = patches.pickup_assignment[hint.target]`
this raises a KeyError if there's an ETM, because it's still not included in the assignment

### LINKED ISSUE #7860: Hints aren't all being placed in some circumstances
happened in a multi. will grab the rdvgame later

Image

duncathan:
fixes #7860 
fixes #7858

### DIFF
--- a/test/generator/filler/test_runner.py
+++ b/test/generator/filler/test_runner.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 from random import Random
 from unittest.mock import MagicMock
 
+import pytest
+
 from randovania.game_description.db.hint_node import HintNode, HintNodeKind
 from randovania.game_description.db.node_identifier import NodeIdentifier
 from randovania.game_description.hint import (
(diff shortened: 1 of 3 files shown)


henriquegemignani (changes requested):
Can you add a test case to check for these?

test/generator/filler/test_runner.py
```diff
@@ -85,8 +87,6 @@ def test_fill_unassigned_hints_empty_assignment(echoes_game_description, echoes_
     filler_config = MagicMock()
     filler_config.minimum_available_locations_for_hint_placement = 0
     hint_state = HintState(filler_config, echoes_game_description)
-    empty_set: frozenset[PickupIndex] = frozenset()
-    hint_state.hint_initial_pickups = {node.identifier: empty_set for node in hint_nodes}
```
duncathan, line 89: turns out we literally already had a test for this, and I was just... intentionally skipping over the part where we test what needed to be tested, lmfao

henriquegemignani (approved):

PR #7870, opened 2025-02-09
Dread: Shinespark logic in Screw Attack Room

Tyranisaur:


randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -4829,18 +4829,36 @@ Extra - asset_id: collision_camera_081
...
+                  # https://youtu.be/QJuxx0SSQMI
+                  Gravity Suit and Morph Ball
   > Event - Screw Attack Rotatable
       Trivial
   > Under left side blocks
       All of the following:
           After Artaria - Screw Attack Rotatable
           Morph Ball or Screw Attack
+  > Event - Screw Attack Rotatable to left of water
+      Speed Booster and Before Artaria - Screw Attack Rotatable and Speed Booster Conservation (Beginner) and Disabled Door Lock Randomizer and Enabled Highly Dangerous Logic and Shoot Missile
```
XenoWars, line 4861: With Screw Attack, this would not actually be dangerous. So it could be either Highly Dangerous Logic or Screw Attack.

CHANGELOG.md
```diff
@@ -22,9 +22,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ##### Artaria
 
-- Changed: Flipping the spinner. in Screw Attack Room when by going from the door to the transport is now logically possible also after having flipped the spinner.
+- Added: Accessing the top pickup in Screw Attack Room with a Shinespark by going down through the water basin. Requires either Screw Attack or Morph Ball and Gravity Suit.
+- Added: Logically flipping the Screw Attack Room Spinner and using Shinespark to immediately reach the left side platform. Requires Highly Dangerous Logic.
```
XenoWars, line 26: Coupled with the other comment, this could include it requires Screw Attack as an alternative.

XenoWars (changes requested):

CHANGELOG.md
Tyranisaur: I was confused about what this is referring to, given that the comment is attached to two lines, one of which refers to the same piece of the logic that the other comment is referring to, and the other line refers to a piece of logic that has a branch for using screw attack. But do you mean the connection that uses speed booster to get across the flipper before it's flipped, and implying that that maneuver is not dangerous based on the possibility of doing a shinespark up from the water basin? If so, isn't there an implicit assumption that you've already escaped the water basin before? Is this a way to enable the two items over here to be what lets you escape the water basin? I suspect that the logic becomes harder to reason about if we try to make exceptions to dangerous logic being locked by highly dangerous logic.

XenoWars: it is for the one immediately going to the left side platform, since you can then just screw attack back up to the door. It would, however, apply to the item as well as you mention, since you can go to the left first then do it again for the item; it would just need to also include the trick and would no longer need to check for the flipper event. At that point, whether you use the flipper first or not, it would just be a normal dangerous action since either way you are logically forced to go down without highly dangerous logic and you would just be softlocked without some other way back up.

randovania/games/dread/logic_database/Artaria.txt
Tyranisaur: After careful consideration, I added Screw Attack here, and I ended up restricting the hand waving on flipping the spinner to when highly dangerous logic is disabled. I think this should cover most of the bases, as I think it's still effectively in logic to get the top item with screw attack and some tricks outside of highly dangerous logic this way.

XenoWars (approved):

PR #7903, opened 2025-02-17
prime: make artifacts less interesting

### LINKED ISSUE #7900: Metroid Prime considers Artifacts as valid targets for hints

JeffGainsNGames:
closes #7900

### DIFF
--- a/randovania/games/prime1/hint_distributor.py
+++ b/randovania/games/prime1/hint_distributor.py
@@ -10,9 +10,10 @@ if TYPE_CHECKING:
 
 
 class PrimeHintDistributor(HintDistributor):
+    # Makes Artifacts less likely to be hinted since gaurunteed hints already exist in Artifact Temple
     @override
     def is_pickup_interesting(self, pickup: PickupEntry) -> bool:
-        return not pickup.has_hint_feature("key")
+        return not pickup.has_hint_feature("artifact")
 
     @override
     @property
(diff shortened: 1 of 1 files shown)


randovania/games/prime1/hint_distributor.py
```diff
@@ -1,12 +1,19 @@
...
 from randovania.generator.hint_distributor import HintDistributor
 
+if TYPE_CHECKING:
+    from randovania.game_description.pickup.pickup_entry import PickupEntry
+
 
 class PrimeHintDistributor(HintDistributor):
+    @override
+    def is_pickup_interesting(self, pickup: PickupEntry) -> bool:
+        return not pickup.has_hint_feature("key")
```
duncathan, line 15: using the `"artifact"` feature here might make it more obvious what's being excluded

Miepee: Something that may be worth discussing in the future is whether we want the key feature to still exist in the first place.

AFAIK the previous justification was for it to introduce ambiguitiy between whether that's the echoes' STK/DTK or the prime's Artifact. But with how early prime gets artifact hints in like 90% of cases one could also argue that it's not really doing anything

JeffGainsNGames:
Would we also want to include Phazon suit?

Miepee:
Can see the case made either way.
Would refer to your discretion or if in doubt consult some CGC peeps
(or toasterparty if he's available, but afaik he currently isn't)

Uncle-Reggie:
> Would we also want to include Phazon suit?

No. It's an actual upgrade with tangible use and the hint provided with a vague location is at the very end of Prime's logic chain.

Miepee (approved):

PR #7915, opened 2025-02-18
Use UV

henriquegemignani:
Actually merging is blocked by https://github.com/dependabot/dependabot-core/issues/10478


README.md
```diff
@@ -208,28 +207,25 @@ In order to update your repository:
    1. Update the git repository. (With `git pull` or anything else)
    2. Make sure that Randovania is closed.
    3. Re-run the steps from "Getting Started", starting at step 2.
-      1. In case of unexpected errors, delete the `venv` in the root of the repository and start again.
+      1. In case of unexpected errors, delete the `.venv` in the root of the repository and start again.
```
Miepee, line 210: does windows hide dotfiles by default? I know most linux file managers do. so we may want to add an aditional sentence of something like "enable the setting to show hidden files in your file manager"

henriquegemignani: I don't even like this mention here, and hopefully uv can actually handle things better. I also don't like the venv being .venv, maybe I can configure it.

Miepee: i know that .venv is *technically* the standard nowadays but i do prefer venv for visibility purposes.
Especially with how often our peeps have issues with some weird dependency where the easiest thing to do is just delete the venv and reinstall deps....

duncathan:
I'm used to using commands like `python -m pytest -n auto`, `python -m mypy`, `pip install -e some/local/patcher/repo`, etc in the venv - will this change the syntax for doing those?

henriquegemignani:
> I'm used to using commands like `python -m pytest -n auto`, `python -m mypy`, `pip install -e some/local/patcher/repo`, etc in the venv - will this change the syntax for doing those?

If you use the uv command for entering the venv, it doesn't. But I think using the proper UV apis for it will be better.

duncathan:
> > I'm used to using commands like `python -m pytest -n auto`, `python -m mypy`, `pip install -e some/local/patcher/repo`, etc in the venv - will this change the syntax for doing those?
> 
> If you use the uv command for entering the venv, it doesn't. But I think using the proper UV apis for it will be better.

oh it'll absolutely be better and is what we should encourage ppl to do, but it's a relief to know my muscle memory won't screw mii up as I get to used to the new APIs

ThanatosGit (approved):

PR #7917, opened 2025-02-19
EditableTableModel improvements

henriquegemignani:
- Move new element support into AppendableEditableTableModel
- Add FieldDefinition.read_only
- Add FieldDefinition.default_factory
- Add DateFieldDefinition


ThanatosGit (approved):
Lgtm.
Though given how many return paths + if- and else-clauses are there, it might be worth to address the uncovered lines with some tests. I leave it to you.

PR #7918, opened 2025-02-19
Don't use copy.copy in EnergyTankDamageState

henriquegemignani:
```
> python -m randovania layout validate '.\Dread Randomizer - Shinespark Charge Autclast.
Took 3.938 seconds. Game is possible.
> python -m randovania layout validate '.\Dread Randomizer - Shinespark Charge Autclast.rdvgame'
Took 3.084 seconds. Game is possible.
```
🙃

### DIFF
--- a/randovania/resolver/energy_tank_damage_state.py
+++ b/randovania/resolver/energy_tank_damage_state.py
@@ -31,6 +31,9 @@ class EnergyTankDamageState(DamageState):
         self._resource_database = resource_database
         self._region_list = region_list
 
+    def __copy__(self) -> Self:
+        return self._duplicate()
+
     @override
     def resource_database(self) -> ResourceDatabase:
         return self._resource_database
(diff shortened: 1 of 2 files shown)


randovania/resolver/energy_tank_damage_state.py
```diff
@@ -44,8 +43,18 @@ def _maximum_energy(self, resources: ResourceCollection) -> int:
         num_tanks = resources[self._energy_tank]
         return self._starting_energy + (self._energy_per_tank * num_tanks)
 
+    def _duplicate(self) -> Self:
```
duncathan, line 46: would it be worth adding this?
```py
def __copy__(self) -> Self:
    return self._duplicate()
```

henriquegemignani: Wouldn't be the first class to have that

duncathan (approved):

PR #7921, opened 2025-02-19
Update Database Format.md node image to match description

### LINKED ISSUE #7720: Discrepancy in Database Format.md
The image and text are conflicting in the Node Management and Redundant Edges section.

The image should show arrows between A and B, as well as B and C, and not between A and C.

The text states that a connection does not explicitly need to be made between A and C, but the image shows that connection (and not the B and C connection which would make A and C unnecessary)

image

dyceron:
Fixes #7720 

image


dyceron:
Why does this matter? I'm not following.

henriquegemignani:
When you clone a git repository, you're not downloading the current state but every single commit ever. So a commit that reduces the size of a file actually increases the repository size.

It's worth it when it's a file that ends in the distributed builds, but this one isn't.

ThanatosGit (approved):

PR #7927, opened 2025-02-20
Video Integrity Check

### LINKED ISSUE #7367: Database test: do *not* have youtube playlist links.
It breaks embeds

JeffGainsNGames:
fixes #7367 

upon saving database, now checks if any youtube links are to a playlist (which breaks embeds)

ex below, wording slightly modified:
image

### DIFF
--- a/randovania/game_description/integrity_check.py
+++ b/randovania/game_description/integrity_check.py
@@ -392,7 +392,7 @@ def find_incompatible_video_links(game: GameDescription) -> Iterator[str]:
                     yield from get_videos(requirement, node, target)
 
 
-def get_videos(req: Requirement, node: Node, target: Node):
+def get_videos(req: Requirement, node: Node, target: Node) -> Iterator[str]:
     if isinstance(req, RequirementArrayBase):
         if req.comment is not None:
             for word in req.comment.split(" "):
--- a/randovania/game_description/integrity_check.py
+++ b/randovania/game_description/integrity_check.py
@@ -397,7 +397,7 @@ def find_incompatible_video_links(game: GameDescription) -> Iterator[str]:
                             if "youtu" not in word:
                                 continue
                             if "&list" in word:
-                                yield f"Playlist linked in {node.identifier.as_string} -> {target.name}."
+                                yield f"Youtube Playlist linked in {node.identifier.as_string} -> {target.name}."
 
 
 def find_database_errors(game: GameDescription) -> list[str]:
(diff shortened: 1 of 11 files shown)


randovania/game_description/integrity_check.py
```diff
@@ -384,6 +384,22 @@ def check_for_resources_to_use_together(
             continue
 
 
+def find_incompatible_video_links(game: GameDescription) -> Iterator[str]:
+    db = game
```
henriquegemignani, line 388: This extra variable seems pointless

randovania/game_description/integrity_check.py
```diff
@@ -384,6 +384,22 @@ def check_for_resources_to_use_together(
...
+    for region in db.region_list.regions:
+        for area in region.areas:
+            for node in area.nodes:
+                for target, requirement in area.connections.get(node, {}).items():
+                    if not isinstance(requirement, RequirementArrayBase):
```
henriquegemignani, line 393: This only finds comments if they're in the "root" of a connection.

You could put the if `if requirement.comment is not None` block all in a function, which then calls itself recursively for every item of the array.

henriquegemignani (changes requested):

randovania/game_description/integrity_check.py
JeffGainsNGames: really? I followed `randovania\cli\commands\export_db_videos.py` for most of this. So I must be misunderstanding something between the two

henriquegemignani: You missed these lines I think: https://github.com/randovania/randovania/blob/90868b013291c043b7079ed604f77c8616d2f052/randovania/cli/commands/export_db_videos.py#L177-L178

JeffGainsNGames: ahh, admittedly didn't understand the whole thing at first
have also never came across `yield` so will need to read up on what that actually does..

randovania/game_description/integrity_check.py
```diff
@@ -384,6 +384,22 @@ def check_for_resources_to_use_together(
...
+                        continue
+                    if requirement.comment is not None:
+                        for word in requirement.comment.split(" "):
+                            if "youtu" not in word:
+                                continue
+                            if "&list" in word:
+                                yield f"Youtube Playlist linked in {node.identifier.as_string} -> {target.name}."
```
Miepee, line 400:
```suggestion
                                yield f"YouTube Playlist linked in {node.identifier.as_string} -> {target.name}."
```

henriquegemignani (approved):

PR #7939, opened 2025-02-21
Don't send the patch data to the exporter dialog

henriquegemignani:


henriquegemignani:
> looks fine to me, but you forgot to adjust CS tests

... I ran all tests and went over all failures and somehow missed some. amazing

henriquegemignani:
> > looks fine to me, but you forgot to adjust CS tests
> 
> ... I ran all tests and went over all failures and somehow missed some. amazing

Ah they're mypy failures not actual test failures

PR #7947, opened 2025-02-21
Change PickupIndexCollected to be emitted as soon as possible.

henriquegemignani:
This removes the chance of it being lost by the following execute_remote_patches failing.

### DIFF
--- a/randovania/game_connection/connector/prime_remote_connector.py
+++ b/randovania/game_connection/connector/prime_remote_connector.py
@@ -164,14 +164,14 @@ class PrimeRemoteConnector(RemoteConnector):
 
         return Inventory(inventory)
 
-    async def known_collected_locations(self) -> set[PickupIndex]:
-        """Fetches pickup indices that have been collected.
-        The list may return less than all collected locations, depending on implementation details.
-        This function also returns a list of remote patches that must be performed via `execute_remote_patches`.
+    async def check_for_collected_location(self) -> bool:
+        """
+        Queries the game for the multiworld magic item and checks if a new location was collected.
+        :return: True, if a new location was found and remote patches were executed. False otherwise.
         """
         multiworld_magic_item = self.multiworld_magic_item
         if multiworld_magic_item is None:
-            return set()
+            return False
 
         memory_ops = await self._memory_op_for_items([multiworld_magic_item])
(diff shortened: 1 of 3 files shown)


randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -194,9 +194,9 @@ async def known_collected_locations(self) -> set[PickupIndex]:
                 )
             ]
             await self.execute_remote_patches(patches)
-            return locations
-        else:
-            return set()
+            return True
```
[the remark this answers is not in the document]
henriquegemignani: It'd fail by throwing an exception meaning this line is also not executed.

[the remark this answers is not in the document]
henriquegemignani: The only situation where this PR changes anything is that if during `execute_remote_patches`, we send the code to Nintendont to remove the magic number from the inventory, that _actually_ gets sent but we get a network error while waiting for the response, leading to an exception which stops us from storing rdv side that the location was found.

With the PR, what happens is we store the location anyway and would find it again next time.

randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -165,14 +165,14 @@ async def get_inventory(self) -> Inventory:
...
-        This function also returns a list of remote patches that must be performed via `execute_remote_patches`.
+    async def check_for_collected_location(self) -> bool:
+        """
+        Queries the game for the multiworld magic item and checks if a new location was collected.
+        :return: True, if a new location was found and remote patches were executed. False otherwise.
         """
         multiworld_magic_item = self.multiworld_magic_item
         if multiworld_magic_item is None:
-            return set()
+            return False
```
duncathan, line 175: out of curiosity, when is this condition ever satisfied?

henriquegemignani: Corruption used to not have this be None. Guess code got refactored around to being an exception?

henriquegemignani: Gonna remove this so coverage is happy

henriquegemignani:
We got some testing that indicates no regression, so I'm comfortable merging this to main and getting further testing from dev builds.

PR #7948, opened 2025-02-21
Only track object count for Echoes if the debug tracker is enabled

henriquegemignani:


randovania/game_connection/connector/remote_connector.py
```diff
@@ -94,3 +94,6 @@ async def force_finish(self) -> None:
     def is_disconnected(self) -> bool:
         """When True, this connector has lost connection with the game and must be discarded."""
         raise NotImplementedError
+
+    def inform_connected_tracker(self, tracker_details: dict | None) -> None:
+        """Called when an AutoTracker is created using this connector."""
```
duncathan, line 99: I know this is valid but I really dislike having a function with only a docstring as its contents. Would prefer an explicit `pass` statement

randovania/data/gui_assets/tracker/prime2-debug.json
```diff
@@ -343,5 +343,6 @@
                 "Object Count"
             ]
         }
-    ]
-}
\ No newline at end of file
+    ],
+    "read_object_count": true
```
duncathan, line 347: I don't like this being top level. Would rather enclose it in an `extra` field

duncathan (changes requested):

duncathan:
why is this a thousand line pr

henriquegemignani:
> why is this a thousand line pr

I rebased on top of the other PR that was in the merge queue, but gh didn't like this I guess

PR #7950, opened 2025-02-22
Echoes: make e-tanks less interesting to hint

JeffGainsNGames:


duncathan:
I actually want to try an alternative approach to fixing this

duncathan:
but we can try this for now in the meantime

duncathan (approved):

PR #7951, opened 2025-02-22
Document expectations of work done to games in development.

henriquegemignani:


docs/New Game.md
```diff
@@ -152,3 +152,28 @@ misconfiguration in Randovania.
...
 
 **TODO**
+
+# Expectations of a Development Game
+
+After a game has been merged and is available in development builds, the responsibility of maintaining it is now shared between you and the Randovania team:
+- Randovania: We make sure your game is included in all Randovania changes.
+- You: Work towards the game reaching stable.
+
+We're happy to support new games by doing our part, but since that has a cost we don't want to be doing it forever there's some expectations regarding time frames:
```
Miepee, line 162: This sentence right now sounds weird
```suggestion
We're happy to support new games by doing our part, but since that has a cost, that we don't want to be doing forever, there are some expectations regarding time frames:
```

docs/New Game.md
```diff
@@ -152,3 +152,28 @@ misconfiguration in Randovania.
 Before a game can work in Multiworld, it must first work properly as a solo game. It is highly recommended to finish the basic integration first.
 
 **TODO**
+
+# Expectations of a Development Game
```
Miepee, line 156: I would move this section further up.
Either above Multiworld, or above Adding a new game entry.

henriquegemignani: When reading the document, it feels weird to go over all this text about after the game is added, before I can read how to add a game.

Miepee: it's why i proposed "above MW" as the other moving location. (Realistically, games will probably never launch with MW for stable)

For me what is also a bit weird is that all this info is very much helpful for deciding on when/if you want to upstream/add a game at all.

docs/New Game.md
```diff
@@ -37,7 +37,7 @@ Replace the values of the command with appropriate values for your game. The mea
 - short-name: Used to prefix all classes used by your game, and in user-facing elements that needs to be brief.
 - long-name: Used in user-facing elements to mention your game.
 
-The command check if the provided values fit the naming rules for each of these. 
+The command check if the provided values fit the naming rules for each of these.
```
duncathan, line 40: since this line is being changed anyway: should say "the command checks"

docs/New Game.md
duncathan: agreed with moving it above MW

docs/New Game.md
```diff
@@ -152,3 +152,28 @@ misconfiguration in Randovania.
...
+How good the experience is when playing a seed is up to your discretion.
+We do however recommend integrating it as early as it is possible, for a better developer experience.
+
+### Becoming Stable
+
+After roughly two years, we expect a game to be close to being considered stable.
+
+### Active Development
+
+Please don't go for too long without visible development of either patcher or non-logic Randovania changes.
```
duncathan, line 176: this should be a bit more specific and mention that this is primarily for *before* going stable. once a game is stable, by nature further active development isn't really needed.

duncathan: I'd also add a line explicitly mentioning that games which aren't in active development may be subject to removal if they don't meet the time frames mentioned above

duncathan (changes requested):

docs/New Game.md
henriquegemignani: Good point on stable. Any suggestions on the later?

docs/New Game.md
```diff
@@ -147,6 +147,35 @@ Define a [JSON Schema](https://json-schema.org/understanding-json-schema/) that
...
+
+### Becoming Stable
+
+After roughly two years, we expect a game to be close to being considered stable.
+
+## Remarks
+
+All of these only apply to games in development. After a game reaches stable, further development is not needed.
+
+Please keep us in the loop of what's going on - if you'll need to step away for some time, if you're having trouble with anything and so on.
```
Miepee, line 176: This is probably a good spot to ammend the "potential to get removed" part.

```suggestion
Please keep us in the loop of what's going on - if you'll need to step away for some time, if you're having trouble with anything and so on.
We may remove games from the codebase after they have been merged, if we feel that there no work done from your side towards making the game stable. This does not mean that we reject the game! It can always come back, but we don't want to maintain something that does not feel worth being maintained.
```

docs/New Game.md
```diff
@@ -82,7 +82,7 @@ After all of the above is done, you are able to generate games.
 
 ### Exporting games
 
-Up to now, everything is just theoretical values inside Randovania. The next step is applying these modifications to the game. 
+Up to now, everything is just theoretical values inside Randovania. The next step is applying these modifications to the game.
```
Uncle-Reggie, line 85:
```suggestion
Up to now, everything has just been theoretical values inside Randovania. The next step is applying these modifications to the game.
```
Tense edit for continuity

docs/New Game.md
```diff
@@ -152,3 +152,28 @@ misconfiguration in Randovania.
...
 
 **TODO**
+
+# Expectations of a Development Game
+
+After a game has been merged and is available in development builds, the responsibility of maintaining it is now shared between you and the Randovania team:
+- Randovania: We make sure your game is included in all Randovania changes.
+- You: Work towards the game reaching stable.
+
+We're happy to support new games by doing our part, but since that has a cost, that we don't want to be doing forever, there are some expectations regarding time frames:
```
Uncle-Reggie, line 156:
```suggestion
We're happy to support the addition of new games by sharing what we know,  but too much investment pulls our time from core projects so we have some expectations regarding time frames:
```
The language originally used seems a bit cold and needed some polish. Open to tweaking this further.

docs/New Game.md
henriquegemignani: Don't like "worth being maintained". let me think

Miepee (approved):
Would like another core dev's opinion as well

PR #7953, opened 2025-02-22
MSR - Make tanks less interesting to hint

dyceron:


Miepee (approved):
Wonder if at one point we should maybe have some helper function for it.

PR #7964, opened 2025-02-24
Fusion: Provide acquiration message to patcher

Miepee:


randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -320,6 +330,29 @@ def _create_room_names(self) -> list[dict]:
                     )
         return names
 
+    def create_memo_data(self) -> dict:
```
JeffGainsNGames, line 333: where is this called?

Miepee: By randovania.exporter.patch_data_factory somewhere. 
It's an overwritten method

Miepee: The original definition is here:
https://github.com/randovania/randovania/blob/2a11aa86739f66c9d9ddfe65a380af5a7fddcfc8/randovania/exporter/patch_data_factory.py#L87

The call here:
https://github.com/randovania/randovania/blob/2a11aa86739f66c9d9ddfe65a380af5a7fddcfc8/randovania/exporter/patch_data_factory.py#L67

JeffGainsNGames (approved):

PR #7972, opened 2025-02-24
Fusion: fix more gui issues

JeffGainsNGames:
part of #7206 

image
image


JeffGainsNGames:
they're resizable and have scrolls when made real small

Miepee (approved):

PR #7987, opened 2025-02-26
echoes: increase hint action weight

duncathan:
should make translators significantly more common in echoes, which helps make up for hints being less powerful overall


JeffGainsNGames (approved):

duncathan:
misclick lol

PR #7988, opened 2025-02-26
Renamed placeholder ETM to Nothing

joaobenedetmachado:
### Changes Made:
- In the file `randovania/randovania/layout/game_patches_serializer.py`
- Changed the variable `_ETM_NAME` from `"Energy Transfer Module"` to `"Nothing"`

- This change was requested in issue #7983
- The placeholder name "Energy Transfer Module" is no longer needed and should be replaced with "Nothing"

- **Let me know if any additional changes are required**


randovania/layout/game_patches_serializer.py
```diff
@@ -27,7 +27,7 @@
     from randovania.game_description.pickup.pickup_entry import PickupEntry
     from randovania.layout.base.base_configuration import BaseConfiguration
 
-_ETM_NAME = "Energy Transfer Module"
+_ETM_NAME = "Nothing"
```
Miepee, line 31: Please renane the variable (and its references) now that its no longer referencing an etm

Miepee (commented):
You forgot step 2 of the issue: write a migration

randovania/layout/game_patches_serializer.py
```diff
@@ -48,7 +48,7 @@ def _pickup_assignment_to_item_locations(
             else:
                 item_name = f"{target.pickup.name}"
         else:
-            item_name = _ETM_NAME
+            item_name = "Nothing"
```
Miepee, line 51: This is not renaming the variable.

This is inlining the variable and thus making the variable useless.

joaobenedetmachado: could you please check it now?

Miepee: Oh, and you missed an occurence in line 252

joaobenedetmachado: So sorry, i corrected it now

Miepee:
Your last commit was not a migration, it was a rename.

For the migration, you want to look into randovania/game_description/game_migration.py and your failing tests.

Miepee:
If you need help adressing the comments, please do let us know!

joaobenedetmachado:
> If you need help adressing the comments, please do let us know!

yeah, i would like!

Miepee:
You do need to state what exactly you need help with. We aren't mind readers after all :sweat_smile:

Also as a side note, if you prefer more synchronous communication, we have a discord server that we use for development as well: https://discord.gg/M23gCxj6fw

joaobenedetmachado:
> You do need to state what exactly you need help with. We aren't mind readers after all :sweat_smile:
> 
> Also as a side note, if you prefer more synchronous communication, we have a discord server that we use for development as well: https://discord.gg/M23gCxj6fw

So, I would like to understand the relationship with the changes I made to game_patches_serializer.py and game_migration, what type of migration is this? what will it be used for?

duncathan:
the game patches serializer is used to write data to an `.rdvgame` file, which is where we store all data about a randomized game in a JSON format. an important feature of `.rdvgame` files is that they're forwards compatible - every version of randovania needs to support loading any old version of an `.rdvgame`. to achieve this, we write a migration. the migration converts the JSON data for the previous version of `.rdvgame` into a valid format for the new version. 

since your changes here affect the way we read and write `.rdvgame` files, you'll need to add a migration that converts pickup entries with the `"Energy Transfer Module"` value to use the `"Nothing"` value instead.

duncathan:
> ok, so i make a new function, like `def _migrate_28`, and and put it in the _MIGRATIONS?

yes, that's correct.

randovania/game_description/game_migration.py
```diff
@@ -299,6 +299,14 @@ def _migrate_v27(data: dict, game: RandovaniaGame) -> None:
                     node["hint_features"] = []
 
 
+def _migrate_v28(data: dict, game: RandovaniaGame) -> None:
+    for world in data["worlds"]:
+        for area in world["areas"].values():
+            for node in area["nodes"].values():
+                if node["node_type"] == "pickup" and node.get("pickup_type") == "Energy Transfer Module":
+                    node["pickup_type"] = "Nothing"
+
```
Miepee, line 308: Can you explain what your thought process was when implementing this function?
I'm very confused on where you got all of these from :sweat_smile: . They're also wrong as they're making the tests fail.

Miepee (commented):
Considering that we're having a rule that every PR must pass all the tests, it is really beneficial for you if you look at the test output.

You can run tests locally by activating the virtual environment and then doing `python -m pytest test`. If needed for further reruns you can narrow it down to a directory/file by editing the last path, e.g. `python -m pytest test/generator/filler`.

If for some reason you are unable to run tests locally, then you should at the very least look at the test output from the CI. You can click on a failing test at the bottom of a PR: 
grafik
That will redirect you to a page where you can see what CI did. Thus also letting you see small stacktraces on why CI failed, and which tests have failed.

Miepee:
@joaobenedetmachado Any updates from your side?
If this feels too overwhelming/difficult to do, then that's fine. There is no shame in that.

I'm not sure if you have used our application at all (does not sound like it at least? I could just be getting the wrong impression though), but those kinds of changes are just very hard to explain and guide someone through unfamiliar with them.
I don't mind helping you guide through it, but it would be a better use of both our time to drop it in that case. Because for us, it's easier and takes less time to explain it to people who do use our app, and for you the knowledge would be mostly worthless if you're never planning to use the app.

Miepee:
obsoleted by #8242

PR #7996, opened 2025-02-26
Dlr/fix locked door condition

Tyranisaur:


randovania/gui/main_window.py
```diff
@@ -473,7 +473,7 @@ def work(progress_update: ProgressUpdateCallable):
             response = await async_dialog.warning(
                 self,
                 "Unexpected hash",
-                f"Expected has to be {expected}. got {new_layout.shareable_hash}. Do you wish to continue?",
+                f"Expected hash to be {expected}. got {new_layout.shareable_hash}. Do you wish to continue?",
```
Miepee, line 476: gui change thats missing from the changelog

randovania/generator/dock_weakness_distributor.py
```diff
@@ -279,7 +279,9 @@ def _determine_valid_weaknesses(
 
     if state is not None:
         reach = ResolverReach.calculate_reach(logic, state)
-        if state.node == target:
+        if state.node == target or isinstance(state.node, DockLockType) and state.node.dock == target:
```
Miepee, line 282: is it really possible that state.node can be either a docklocktype or a docknode?

henriquegemignani, line 282:
```suggestion
        if state.node == target or (isinstance(state.node, DockLockType) and state.node.dock == target):
```

henriquegemignani (dismissed):

randovania/generator/dock_weakness_distributor.py
duncathan: never trust operator precedence in conditionals, lmao

duncathan: my immediate intuition is that it would certainly be possible. if anything, it's more likely to be a docklocknode than a docknode I think

Tyranisaur: I'm convinced that `DockLockType` was a slip, but I found through experience that they can be `DockLockNode`.

PR #8000, opened 2025-02-27
filter for uncollected resource nodes in advance with unsafe

duncathan:
this is extremely significant when combined with the new experimental setting. it fixes vanilla missiles in echoes starter preset!!!!
image

also adds the experimental setting to the preset describer, which I had missed earlier


randovania/generator/generator_reach.py
```diff
@@ -77,6 +79,17 @@ def nodes(self) -> Iterator[Node]:
     def safe_nodes(self) -> Iterator[Node]:
         raise NotImplementedError
 
+    @property
+    def safe_uncollected_resource_nodes(self) -> Iterator[Node]:
+        context = self.node_context()
+        for node in self.safe_nodes:
+            if node.is_resource_node:
+                res = typing.cast(ResourceNode, node)
+                if res.should_collect(self.node_context()) and res.requirement_to_collect().satisfied(
```
henriquegemignani, line 88: Re-use the `context` variable from above

henriquegemignani (changes requested):

test/generator/filler/test_player_state.py
```diff
@@ -35,13 +35,16 @@ def state_for_blank(
 def test_current_state_report(state_for_blank):
     result = state_for_blank.current_state_report()
     assert result == (
-        "At Intro/Starting Area/Event - Post Weapon after 0 actions and 0 pickups, "
-        "with 4 collected locations, 24 safe nodes.\n\n"
```
henriquegemignani, line 39: Would be nice to have the count of safe uncollected resource nodes I guess?

test/generator/filler/test_player_state.py
```diff
@@ -50,7 +53,7 @@ def test_current_state_report(state_for_blank):
         "None\n"
         "\n"
         "Reachable nodes:\n"
-        "32 nodes total"
+        "Intro/Back-Only Lock Room/Pickup (Extra Key)"
```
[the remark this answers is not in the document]
duncathan: I'm a little unsure myself, but it's important to point out that it's also collected an additional resource node now. Could be that collecting that additional node does not remove access to any other resource nodes, so now it goes and collects it which removes access to the trivial nodes?

test/generator/filler/test_player_state.py
duncathan: good call

randovania/generator/generator_reach.py
duncathan: LMAO that's literally why i added it whoops

henriquegemignani (approved):

PR #8008, opened 2025-02-28
Echoes: nsj sa logic adjustments

JeffGainsNGames:
bump mining plaza, trial tunnel
lower temple assembly site (also change the video to use easier method)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Keybearer hints may refer to pickups using different categories than before.
 - Fixed: When using the new experimental generator setting to consider possible unsafe resources, the Starter Preset can once again see Missile Launcher placed in the GFMC Compound crate.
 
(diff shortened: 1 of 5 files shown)


JeffGainsNGames:
pinging @BajaBlood @CMos75 as tourney admins

BajaBlood:
I do appreciate being looped into changes that add/remove beginner tricks.

Am personally pretty indifferent on plaza/TAS, both feel similarly difficult and borderline beginner/intermediate to me. I'd be fine with any of both beginner, both intermediate, or (as proposed) only plaza intermediate due to more punishing failure.

duncathan:
yeah, I don't think plaza is *that* much more precise than TAS, but the fact it takes like 10 seconds to retry and then requires another bit of NSJ SA movement immediately after just makes it feel so much worse. it's very easy to imagine someone losing a minute or more to it at the beginning of a race, which is pretty extreme for a single trick so early on IMO. and as someone who generally only has the skill for beginner tricks, it just feels really bad lmao

duncathan (approved):

PR #8017, opened 2025-03-02
in individual DLR, use the resolver for hint placement

duncathan:
depends on #8036 and #8037 

also adds a hidden preset option to enable this in non-DLR seeds. i don't want to expose it in the GUI atm

### DIFF
--- a/randovania/generator/hint_distributor.py
+++ b/randovania/generator/hint_distributor.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 import functools
+import logging
 import math
 from abc import ABC, abstractmethod
 from collections import defaultdict
@@ -27,7 +28,8 @@ from randovania.game_description.hint_features import HintFeature
 from randovania.game_description.resources.pickup_index import PickupIndex
 from randovania.generator.filler.filler_library import UnableToGenerate
 from randovania.generator.filler.player_state import HintState, PlayerState
-from randovania.resolver import debug
+from randovania.layout.base.dock_rando_configuration import DockRandoMode
+from randovania.resolver import debug, resolver
 
 if TYPE_CHECKING:
     from randovania.game_description.assignment import PickupTarget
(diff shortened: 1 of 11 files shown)


duncathan:
so, this has some problems, most notably that the results suck:
image

@Tyranisaur you're the resolver expert, could you take a look and see if i'm missing something dumb?

Tyranisaur:
It could be that the resolver doesn't pick up hint nodes aggressively enough compared to when it picks up items. Maybe it would work better if hint nodes were given the same treatment as major items when they are considered safe. Though, the current idea is to pick up safe major items on the first pass over the satisfied actions and if the resolution fails when picking up that action, it immediately rolls back out. So if you really want hint nodes to have higher priority than that, you would probably need to do another pass over the actions before picking the first safe major item.

Tyranisaur:
Also, it appears that you're skipping EventPickupNodes.

duncathan:
Thanks for the pointers, Ty. After some changes I was able to get results like this:
image

As far as I can tell, most of the majors being hinted are in the progression chain, and their hints were available before the majors themselves were. Debug output during hint placement indicated that these were in fact logical hints, which is another good sign

henriquegemignani (dismissed):
I think if you split all the code-refactoring necessary for this pull request into a separate one, it'll both be easier to review this and the other PR can be merged without being worried about the quality of the results.

henriquegemignani:
And unblock #8035.

duncathan:
ok, now depends on #8036 and #8037 

also adds a hidden preset option to enable this in non-DLR seeds. i don't want to expose it in the GUI atm

duncathan:
played an echoes DLR seed on this branch and I've gotta say it's night and day. nearly every hint was genuinely helpful, compared to old DLR seeds where if you were lucky you might find one or two helpful hints. this is a massive improvement

randovania/generator/hint_distributor.py
```diff
@@ -685,6 +692,24 @@ async def assign_precision_to_hints(
...
+        new_state = await resolver.resolve(patches.configuration, patches, collect_hint_data=True)
+
+    if new_state is None:
+        logger.warning(
+            f"Unable to solve game for player {player + 1} after placing doors. Hints will be based on generator state."
+        )
+        return None
+    else:
+        debug.debug_print(f">> Player {player + 1} is solve-able after door placement. Beginning hint placement.")
+        return new_state.hint_state
```
Tyranisaur, line 710: Is it just me or is the (optional) check after generation that the seed can be solved simply duplicated work when you do this? Is there any way to skip the duplicated work?

Tyranisaur (approved):

randovania/generator/hint_distributor.py
duncathan: it is technically duplicated work, but it would be *really* ugly to try and skip it imo. frankly i'm not too concerned about adding one extra resolver run at the end of like 300 of them, lol

PR #8020, opened 2025-03-02
Move select_weighted_action to random_lib

henriquegemignani:


randovania/lib/random_lib.py
```diff
@@ -48,12 +44,24 @@ def iterate_with_weights(
         yield pickup_node
```
duncathan, line 44: while you're in here, wanna rename `pickup_node` and `action` into generic names?

duncathan (changes requested):
tests/mypy

randovania/lib/random_lib.py
henriquegemignani: oh yeah oops

duncathan (approved):

PR #8023, opened 2025-03-02
Resolver: Fix additional requirements when rolling back progressive items

Tyranisaur:


duncathan:
was the problem being fixed here an off-by-one error with the index?

Tyranisaur:
Effectively yes, but in the case where this was originally written, it was offset by how it found the progression chain by looking at the wrong node. So it was a case of this https://media.licdn.com/dms/image/v2/D5622AQEWPufj0I7lkA/feedshare-shrink_1280/B56ZPy.5KaHIAk-/0/1734948395751?e=2147483647&v=beta&t=HBbui2VI0hjoVOn4fv4P7uE-5k9H5tshiDiWN7Iuovo

duncathan (approved):

PR #8030, opened 2025-03-03
place hints with no valid targets at the end instead of the start

duncathan:
this prevents completely non-logical hints from taking potential logical targets from other hints


henriquegemignani (approved):
I feel like this is a wonderful thing to have tests for.

duncathan:
what do you want to see tested, exactly?

duncathan:
kk, I'll write a test later

PR #8033, opened 2025-03-04
add test for hint node order

duncathan:


henriquegemignani (approved):
Was thinking of testing fill_unsassigned_hints but this works too lmao

PR #8039, opened 2025-03-04
increase DLR attempts for echoes

duncathan:
on my preset, most doors are solved in under 300 attempts, and all are solved in under 500 attempts. the 600 is just a little bit of leeway in case of weird presets. critically, this means the doors in Sky Temple can actually be non-blue, which is an important part of my intended design for echoes DLR and hasn't been working for a while


Tyranisaur (approved):
Love it

PR #8044, opened 2025-03-05
calculate precision for HintLocationPrecision.DETAILED

### LINKED ISSUE #7891: Calculate HintLocationPrecision.DETAILED precision
it's currently always 100% precision, but I realized that a hint like "in Plantation" in cave story is much less precise than "with Curly", with both being applicable to one of the pickups in plantation. shouldn't affect most cases in other games but CS just has rooms with so many pickups in em

duncathan:
fixes #7891


henriquegemignani (changes requested):
Can you add a test for this?

henriquegemignani (approved):

PR #8057, opened 2025-03-06
fusion: update underwater walljump logic in sector 4

cluekitty:
Community feedback suggests that long underwater walljump tricks should be hypermode/ludicrous.
Additionally, I think they should be hypermode/ludicrous so we do not give players who chose expert repetitive strain injury.

Changed:
Sector 4
* Security Access
  * Door to Security -> Beside Hole
  * Beside Hole -> Door to Cheddar Bay

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 4 AQA.json
+++ b/randovania/games/fusion/logic_database/Sector 4 AQA.json
@@ -8984,7 +8984,7 @@
                                                     "data": {
                                                         "type": "tricks",
                                                         "name": "UnderwaterWallJump",
-                                                        "amount": 4,
+                                                        "amount": 5,
                                                         "negate": false
                                                     }
                                                 },
@@ -9117,7 +9117,7 @@
                                                     "data": {
                                                         "type": "tricks",
                                                         "name": "UnderwaterWallJump",
-                                                        "amount": 4,
+                                                        "amount": 5,
                                                         "negate": false
                                                     }
                                                 }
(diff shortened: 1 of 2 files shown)


Miepee (commented):
1) needs a rebase
2) I'm personally too biased to to always playing on a keyboard, where these WJs don't feel ludicrous to me, but rather expert.

Would like to have @SchwartzGandhi or @JeffGainsNGames thoughts

herculesbenchpress:
Setting these to hypermode is the correct choice, especially the ones that require a lot of climbing. It's pretty easy to drop and have to start over as well, even for experienced players. Plus, avoiding any injuries for anyone except those that are willing to do the dumbest things in the logic.

duncathan:
> The injuries part is not a valid reason IMO for ludicrous.

I don't think there's a single more compelling reason for something to be categorized as ludicrous

Miepee:
Then we should make UWJs a ludicrous-only trick

herculesbenchpress:
Some, not all. Not every UWJ requires 30+ wall jumps like the one from level 4 security. The one going to Serris and Serris tank is only 2 or 3. The one to Oasis storage is only about 5 and isn't as intensive. I think the intensive ones should be switch to hypermode/ludicrous.

duncathan:
> Is it possible to add a warning to the slider page for UWJ? Something to let the player know that underwater wall jumping can put some strain on the hands and to include only at their own risk?

that can be added to the trick description

herculesbenchpress:
I think we should do that, first and foremost. Then look at what the trick levels are at and see how difficulty they really are and go from there?

cluekitty:
It seems there is not yet videos for the UWJ for Security Access. I'll spend some time later today seeing if I can record these while I have this PR open. Or if someone else records them sooner I can add the videos to the PR.

JeffGainsNGames (approved):

PR #8061, opened 2025-03-06
Add global HintConfiguration

### LINKED ISSUE #3444: Configurable Random Hints
Currently, Echoes only has preset configuration GUI for changing how key hints are presented. There should be additional GUI options to control how the random hints behave.

At minimum, there should be an option to disable hints entirely, but other options such as selecting what types of hints are available is desirable.

Ideally this would be implemented game agnostic such that Prime1 can utilize these configuration options for it's forthcoming hint system.

duncathan:
Fixes #3444 

adds standard configuration settings and preset tab for hints. this consolidates all the existing per-game hints settings into one global system, and adds a couple new settings to enable/disable random hints and specific location hints

I also exposed the json-only hint settings, using a new feature where preset tabs can have settings that display only in preview mode. I'm sure some other preset tabs will be able to take advantage of this as well

finally, i renamed the HintNodeKind enum values because holy moly those were bad lmao

echoes:
image

cave story:
image

samus returns:
image

prime: 
image

echoes in preview mode:
image


henriquegemignani (changes requested):
Add hint_tab.py to the mypy required list

randovania/layout/preset_describer.py
```diff
@@ -108,6 +108,26 @@ def _calculate_pickup_pool(self, configuration: BaseConfiguration) -> list[str]:
 
         return result
 
+    def _hints_info(self, configuration: BaseConfiguration) -> list[str]:
+        strings: list[str] = []
+        game = default_database.game_description_for(configuration.game)
+
+        if game.has_random_hints:
+            if not configuration.hints.enable_random_hints:
+                strings.append("Random hints disabled")
```
henriquegemignani, line 117: Add a test for this and specific location

henriquegemignani (changes requested):

henriquegemignani:
I'm not entirely sure of the wording of "specific location hints". It's not exactly obvious at first glance.

duncathan:
> I'm not entirely sure of the wording of "specific location hints". It's not exactly obvious at first glance.

I think this is a problem with our terminology in general rather than with this name. We've been getting more consistent at using "pickup" rather than "item" for pickups, but I think we need to make more conscious efforts to move towards calling PickupNode/PickupIndex "pickup locations" or just "locations" in our user-facing stuff. We already have this to some extent with the Location Pool and Pickup Pool preset tabs.

This is also how the Hint Locations tab already refers to this kind of hint (although that tab needs some serious visual improvements too lol)

duncathan:
image

henriquegemignani:
I do know of that tab you just posted, but I had actually checked it before posting and it feels very disconnected. It takes really lot of effort to understand what it means.
Even right now I'm not completely sure on it. I think for Echoes it's the Keybearer hints?
And why should that disable just these and not the U-Mos 2 hint?

duncathan:
> I do know of that tab you just posted, but I had actually checked it before posting and it feels very disconnected. It takes really lot of effort to understand what it means. Even right now I'm not completely sure on it. I think for Echoes it's the Keybearer hints? And why should that disable just these and not the U-Mos 2 hint?

I agree that the definitions and that tab are really unclear ATM, I just don't think it's in scope for this PR to hammer all those issues out. This PR uses the existing user-facing terminolgy. I have some ideas to improve the terminology but I'd prefer to leave it for a follow-up PR

duncathan:
> Disagree. You're exposing a setting with this terminology. I'd rather then you not include the box for disabling these hints.

image
I've added this to the descriptions. I don't think very much more than this is warranted (for this PR, anyway)

henriquegemignani (approved):

PR #8076, opened 2025-03-09
Give the playthrough tab a long-overdue makeover

duncathan:
this should be infinitely easier for users to actually read and follow along with, lmao

screenshots:
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Some joke hints are now shared between all games that use them.
 - Changed: Hints now prioritize non-progression major items rather than progression minor items.
 - Changed: The "Spoiler: Playthrough" tab is significantly easier to read and follow along with.
+- Fixed: The "Spoiler: Playthrough" tab is much faster when calculating the playthrough.
 - Fixed: It is now properly impossible for multiple hints to point to the same location.
 - Fixed: A recent regression in the resolver that made it significantly slower has been fixed.
 - Fixed: Generator action weights were being incorrectly calculated.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: The Hints Spoiler tab sorts its entries in a more helpful manner.
 - Changed: Some joke hints are now shared between all games that use them.
 - Changed: Hints now prioritize non-progression major items rather than progression minor items.
+- Changed: The "Spoiler: Playthrough" tab is significantly easier to read and follow along with.
 - Fixed: It is now properly impossible for multiple hints to point to the same location.
 - Fixed: A recent regression in the resolver that made it significantly slower has been fixed.
 - Fixed: Generator action weights were being incorrectly calculated.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -271,6 +271,7 @@ files = [
     "randovania/gui/preset_settings/hints_tab.py",
     "randovania/gui/lib/signal_handling.py",
     "randovania/gui/widgets/game_connection_window.py",
+    "randovania/gui/widgets/game_validator_widget.py",
     "randovania/gui/widgets/node_selector_widget.py",
     "randovania/lib/",
     "test/bitpacking/",
--- a/randovania/cli/commands/validate.py
+++ b/randovania/cli/commands/validate.py
@@ -12,7 +12,6 @@ from randovania.resolver import debug, resolver
 
 if TYPE_CHECKING:
     from argparse import ArgumentParser, Namespace, _SubParsersAction
-    from typing import Any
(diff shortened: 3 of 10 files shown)


henriquegemignani:
How does it look like when there's backtracking?

duncathan:
> How does it look like when there's backtracking?

image

duncathan:
oh btw, as a fun bonus this makes the high verbosities *way* faster to calculate from within this window

henriquegemignani:
I would say it needs some indicator that the action was undone, without having to open the folding. Just something to indicate that row is different from the others.

henriquegemignani:
Maybe the resolver should always be ran with higher verbosity and we just don't display the extra columns by default?

duncathan:
> Maybe the resolver should always be ran with higher verbosity and we just don't display the extra columns by default?

In a random echoes DLR seed I'm trying it with, it's still a fairly significant difference in time. 14.2s on normal vs 120s on extreme, or 64s on extreme with door locks hidden from the list. I think the time savings of normal verbosity are still pretty relevant.

(That's compared to 20s on normal and 253s on extreme on `main`, incidentally.)

duncathan:
> I would say it needs some indicator that the action was undone, without having to open the folding. Just something to indicate that row is different from the others.

image

henriquegemignani:
> In a random echoes DLR seed I'm trying it with, it's still a fairly significant difference in time. 14.2s on normal vs 120s on extreme, or 64s on extreme with door locks hidden from the list. I think the time savings of normal verbosity are still pretty relevant.
> 
> (That's compared to 20s on normal and 253s on extreme on `main`, incidentally.)

Very good reasons indeed. In this case, the wording of that needs improving - logging is programming jargon, let's avoid it here. Plus the levels don't indicate what additional information we get.

Miepee:
Would there be a possibility to not show actions that were rolled back at all?
From a casual user PoV, the concept of a rollback is foreign. And trying to follow on what they need to do to beat a seed, only to then realize they need to mentally reverse what they have just read/noted down isn't that great.

duncathan:
> Very good reasons indeed. In this case, the wording of that needs improving - logging is programming jargon, let's avoid it here. Plus the levels don't indicate what additional information we get.

I can drop the "log" and just call it Verbosity. I'm not sure how much explaining we need to do about the different levels - 99% of the time the average user will just want Normal, the default.

duncathan:
> Would there be a possibility to not show actions that were rolled back at all? At least for Normal Verbosity
> From a casual user PoV, the concept of a rollback is foreign. And trying to follow on what they need to do to beat a seed, only to then realize they need to mentally reverse what they have just read/noted down isn't that great.

Easy enough to hide rollbacks on normal verbosity. I'll leave the decision up to @henriquegemignani whether we do so

Miepee:
Also 'cause I wasn't sure whether my phrasing was too vague, when I said "hide rollbacks" i meant "rollbacks and all relevant things that were rolled back". Not just entries with the type "rollback".
So in the image above that you shared where the resolver collected a lock, a pickup and then both were rolled back, it should then show nothing instead.

duncathan:
> Also 'cause I wasn't sure whether my phrasing was too vague, when I said "hide rollbacks" i meant "rollbacks and all relevant things that were rolled back". Not just entries with the type "rollback". So in the image above that you shared where the resolver collected a lock, a pickup and then both were rolled back, it should then show nothing instead.

oh. that's hard to do

henriquegemignani (commented):
Since you're touching this file, can you add `game_validator_widget.py` to the mypy list?

duncathan:
made it scale better when the window is resized:
image

also figured I'd actually show the label asking for a refresh. it currently shows when the filters are changed, should I make it show when verbosity is changed too?

henriquegemignani:
Since the label shows up when the check boxes and changed and not when you when the verbosity, IMO it implies that doesn't need re-running. So I think should show up in that case too.

duncathan:
image

duncathan:
oh yeah here's light mode. it's extremely ugly but the colored text is readable. I don't want to change the colors because they currently match the node colors in the data visualizer
image

duncathan:
image
this is light mode without the dark background in the type column. i also have no idea how to theme these colors according to dark/light mode being active so if we really want that to be done then I'll need to be told how to do it

henriquegemignani (approved):

henriquegemignani:
I think I prefer without the black background?

duncathan:
ah I forgot to include hints, where it's particularly bad on the white bg. honestly though, maybe we should look into making the light theme use a light grey bg instead of white to maybe alleviate some of this? lol

duncathan:
image
could still remove the dark bg if preferred

henriquegemignani:
Yeah that Hints text is unreadable.

Clearly what we need is text shadows but I dunno how to do that

duncathan:
> Clearly what we need is text shadows but I dunno how to do that

yeah I looked into adding shadows or borders but it seems to be *really* involved, making a custom QPainter and all that nonsense. something for another day

PR #8085, opened 2025-03-11
Further improvements to playthrough tab

duncathan:
at this point I think I actually find the playthrough tab *more* readable than the generation order tab, which is a massive success imo

- adds region name
- splits major and minor pickups and allows them to be filtered separately
- filters rollbacks based on the type of action being rolled back

the big remaining issue is that EventPickup names are soooo long but idk what can be done about that lmao

image
image

### DIFF
--- a/randovania/gui/widgets/game_validator_widget.py
+++ b/randovania/gui/widgets/game_validator_widget.py
@@ -240,7 +240,7 @@ class GameValidatorWidget(QtWidgets.QWidget, Ui_GameValidatorWidget):
                     player = int(pickup_match.group("world_num"))
                     pickup = pickup_match.group("pickup_name")
 
-                    if self.layout_description.generator_parameters.world_count == 1:
+                    if self.layout_description.world_count == 1:
                         action_text = pickup
                     else:
                         action_text = f"{self.players.player_names[player]}'s {pickup}"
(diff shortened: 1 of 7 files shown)


randovania/gui/widgets/game_validator_widget.py
```diff
@@ -200,61 +222,96 @@ def write_to_log(*a: Any) -> None:
...
+
+                if (action_type in {"Major", "Minor", "Pickup"}) and action_text != "Nothing":
+                    pickup_match = pickup_action_re.match(action_text)
+                    assert pickup_match is not None
 
-            item = QtWidgets.QTreeWidgetItem(parent)
+                    player = int(pickup_match.group("world_num"))
+                    pickup = pickup_match.group("pickup_name")
+
+                    if self.layout_description.generator_parameters.world_count == 1:
```
henriquegemignani, line 243: Could be just `self.layout_description.world_count`

duncathan: intentionally future-proofed for multiworld

henriquegemignani (approved):

randovania/gui/widgets/game_validator_widget.py
henriquegemignani: Sure, but that's not what I said :P
`layout_description.world_count` is an alias for `layout_description.generator_parameters.world_count`

duncathan: oh, sure. I thought you meant the whole condition could just be that

PR #8091, opened 2025-03-12
remove redundant and incorrect connections in NOC geron crossing

duncathan:
brought to you by my seed being impossible without a shinespark trick i can't do :(

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 6 NOC.json
+++ b/randovania/games/fusion/logic_database/Sector 6 NOC.json
@@ -1190,27 +1190,6 @@
                                 ]
                             }
                         },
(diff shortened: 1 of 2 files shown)


duncathan:
forgot label

SchwartzGandhi (changes requested):
Honestly this whole room needs refactoring. Without this connection, the game would expect you to reach the upper left hatch to enter the morph tunnel. What it should be is to have an extra generic node in the lower left and put the shinespark there instead.

duncathan:
someone else can feel free to do that to fix the false negative, I was just concerned with fixing the false positive

SchwartzGandhi:
I'll get a new pr going right now

PR #8092, opened 2025-03-12
Fusion Updated Videos and Trick Logic for Reservoir Vault

M4RKERSM4SK:
Added Wall Jump Video

Added Stand on Frozen Enemies Video

Changed from Freeze with any Weapon to only Ice Beam or Ice Missiles. After 40 min of attempts I could not get the diffusion missile to time right due to the radius of diffusion freezing both Owtch. Either its impossible due to the enemey timings or much harder then beginner.

Edit: I have also found and added a new trick for the upper part of the room for review.

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 4 AQA.txt
+++ b/randovania/games/fusion/logic_database/Sector 4 AQA.txt
@@ -642,7 +642,7 @@ Extra - room_id: [13]
                   # Mid-Air Morph: https://youtu.be/F-XptNF7OOg
                   Mid-Air Morph (Advanced)
                   # Freeze the Owtch: https://youtu.be/U1LHQA0NoFM
-                  Stand On Frozen Enemies (Advanced) and Can Freeze Enemies With Any Weapon
+                  Stand On Frozen Enemies (Expert) and Can Freeze Enemies With Any Weapon
 
 ----------------
 Aquarium Shaft
--- a/randovania/games/fusion/logic_database/Sector 4 AQA.txt
+++ b/randovania/games/fusion/logic_database/Sector 4 AQA.txt
@@ -603,8 +603,12 @@ Extra - room_id: [13]
       Trivial
   > Next to Chain Blocks
       Any of the following:
+          # Wall Jump: https://youtu.be/c2ojj7S_-5c
           Space Jump or Wall Jump (Beginner)
-          Stand On Frozen Enemies (Beginner) and Can Freeze Enemies With Any Weapon
+          All of the following:
+              # Freeze the Owtch: https://youtu.be/iztxJUZuEjA
+              Stand On Frozen Enemies (Beginner)
+              Can Freeze Enemies With Ice Beam or Can Freeze Enemies With Ice Missiles
 
 > Beside Upper Pickup; Heals? False
   * Layers: default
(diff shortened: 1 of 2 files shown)


herculesbenchpress:
I can assume the diffusion missile strat is possible. We cannot ignore this in the requirements since this is a way to freeze an enemy.

Also, for the future, try not to add to the PR with additions that don't encompass the original changes. It can make it confusing to review. If something new needs to be added, open it as a separate PR.

M4RKERSM4SK:
Since the diffusion strat its not beginner, I think it would be added as a seperate requirement correct and then added later if someone can do? Recording or doing the diffusion is beyond my ability at this time.

Would it work if I would have added a comment to the PR with the new trick and details for review? Trying to follow what Jeff stated for reworking a single room, but what you stated makes sense too. Since its the same branch with the same room, I would have had to make a seperate branch but named simular. Either way no problem, but if a comment works after the addition, it would save a ton on time on my end if that also works for the DB team.

biosp4rk:
https://github.com/user-attachments/assets/cd6d8f1b-95f1-4279-a550-3ecdd525f491

M4RKERSM4SK:
> owtch.mp4

What difficulty would this be considered at?

M4RKERSM4SK:
Ok, I took the Diffusion method that Biospark found here and got a recording with Biospark's permission.

I upgraded the Ice Missile and Ice Beam method to intermedate. Reason is that if Wall jumps are not considered after freezing the enemy (since the wall jumps trick is already beginner) then knowing you need to have the owtches in a specific position and direction in order for the trick to work seems more than beginner.

The DIffusion method is hard. You need to enter the room with the correct timing, precharged Diffusion, the individual owtches on the right direction, freeze with perfect timing and spots to not freeze both, and freeze at the right time at the end or take damage and fall. I placed it at expert for review.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -635,6 +641,8 @@ Extra - room_id: [13]
                   Can Use Springball
                   # Mid-Air Morph: https://youtu.be/F-XptNF7OOg
                   Mid-Air Morph (Advanced)
+                  # Freeze the Owtch: https://youtu.be/U1LHQA0NoFM
+                  Stand On Frozen Enemies (Advanced) and Can Freeze Enemies With Any Weapon
```
JeffGainsNGames, line 645: I'm on the edge of saying to bump this to expert for:
1) how different it is from most other tricks in this category (having to wait so long and the enemy roaming offscreen)
2) resetting is likely slow and annoying
Would appreciate another DB maintainers input

herculesbenchpress: I agree. This feels, and looks, like an expert level trick with all of the waiting and needing to get things right. The devil's advocate inside of me wonders if this will ever be relevant in a randomizer since most people that would set something to expert and likely just going to wall jump up to the platform anyway.

M4RKERSM4SK: If we are talking the trick that Jeff as highlighted for review, I couldnt mid air morph (advanced) so thats how I found this new trick to get past the chain blocks without springball or bombs. It scored me varia out of logic extreamly early in an actual seed, so I think this trick is extreamly relevant for Rando and for Stand on Frozen Enemies logic

I think expert is fair though given the points Jeff made, but would also like to state this could be a nice "advance" alternative for folks that cant mid air morph yet logically. It is a bit obscure though. 

Please let me know offically what you all want this new trick at and I can get it pushed.

JeffGainsNGames: after our discussion in your stream, please bump to expert

M4RKERSM4SK: bumped to expert

JeffGainsNGames (approved):

PR #8106, opened 2025-03-12
Run mypy in entire Fusion, delay mars_patcher import

henriquegemignani:


pyproject.toml
```diff
@@ -254,9 +254,7 @@ files = [
     "randovania/games/cave_story/",
     "randovania/games/dread/exporter/hint_namer.py",
     "randovania/games/dread/gui/",
-    "randovania/games/fusion/gui/",
-    "randovania/games/fusion/generator/",
-    "randovania/games/fusion/layout/",
```
JeffGainsNGames, line 259: can I get an explanation just for my own knowledge?

henriquegemignani: The added line below is the entire folder, so now mypy runs for every file in the Fusion.

JeffGainsNGames (approved):

PR #8107, opened 2025-03-12
Split HintNode into multiple subclasses

duncathan:
First step on my quest to make specific-pickup hints real.
image
image
image
image
image
image


randovania/game_description/db/hint_node.py
```diff
@@ -54,3 +52,92 @@ def is_collected(self, context: NodeContext) -> bool:
...
+
+    @property
+    def target_index(self) -> PickupIndex:
+        """Which PickupIndex this hint points to."""
+        return PickupIndex(self._target)
+
+    def data_editor_message(self, region_list: RegionList, db: ResourceDatabase) -> str:
+        message = "Specific Location Hint"
+        target = region_list.node_from_pickup_index(self.target_index)
+        fmt_message = '\n<br />Target: <a href="node://{}">{}</a>'
```
henriquegemignani, line 87: I don't like having effectively GUI code right up into the core classes

duncathan: could do it as a series of `isinstance()` checks where it's actually used, if you prefer

randovania/game_description/db/hint_node.py
```diff
@@ -54,3 +52,92 @@ def is_collected(self, context: NodeContext) -> bool:
...
+    HintNodeKind,
+    {
+        HintNodeKind.GENERIC: "Generic",
+        HintNodeKind.SPECIFIC_LOCATION: "Specific Location",
+        HintNodeKind.SPECIFIC_PICKUP: "Specific Pickup",
+    },
+)
+
+HINT_KIND_TO_CLASS = {
+    cls.hint_kind(): cls for cls in (GenericHintNode, SpecificLocationHintNode, SpecificPickupHintNode)
```
henriquegemignani, line 136: Maybe `HintNode.__subclasses__()`?

randovania/games/prime2/generator/hint_distributor.py
```diff
@@ -38,7 +38,7 @@ def is_pickup_interesting(self, target: PickupTarget, player_id: int, hint_node:
                 return False
         if target.player == player_id and ("Translator" in target.pickup.name):
             # don't place a translator hint on its color of lore scan
-            return hint_node.extra.get("translator") not in target.pickup.name
+            return hint_node.requirement_display_name not in target.pickup.name
```
henriquegemignani, line 41: I don't like either that you're using something you called display name for core logic.

duncathan: I'll add a FIXME; I never should've used that attribute for this in the first place really

henriquegemignani (changes requested):

henriquegemignani:
Oh I identified why I'm very unsure about this PR. I'm unsure about everything about `HintNodeKind.SPECIFIC_PICKUP`, especially because we have 0 use cases. That kind of hint node hasn't been proven to work.

duncathan:
> Oh I identified why I'm very unsure about this PR. I'm unsure about everything about HintNodeKind.SPECIFIC_PICKUP, especially because we have 0 use cases. That kind of hint node hasn't been proven to work.

I have a WIP branch that makes specific pickup nodes actually real. I just wanted to get this part of the change through first since it significantly improves the UX for the other two in the database editor/viewer

henriquegemignani:
This PR might've died to bitrot?

PR #8117, opened 2025-03-14
Fusion Puyo Palace Trick and Video Updates

M4RKERSM4SK:
### Pickup (Power Bomb Tank) -> Door to Nettori Save Room

Made and added new video that was requested from trick spreadsheet

### Door to Nettori Save Room -> Pickup (Power Bomb Tank)

I increased the initial movement trick with no upgrades from intermediate to advanced

Reasons:
- If you fail any of the below you are punished heavily esp in alternate start
- Hanging, jumping, holding L, aiming down, shooting and then holding right was pretty tough to do quickly
- If you pressed "b" or did the sequence too late, you shoot the block below ruining your chances of getting out with no upgrades (which that movement trick is set to advanced. IMO this trick should match that trick since it ties to and impacts the ability to do that advanced trick at all)

In addition I also found and added in some easier methods

- You can shoot the blocks with Wave Beam or Diffusion from above making the movement much easier to get in
- Or you can use SA and just jump through
- I left this at intermedate since if you still fail the jump you are heavily punished

I dont know if I did the logic right for what items requires Diffusion Missiles in this case, please double check me there

Edit: I forgot last night to add in the SA method too, that is what the second commit is there.

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.json
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.json
@@ -10225,7 +10225,7 @@
                                                 {
                                                     "type": "or",
                                                     "data": {
-                                                        "comment": "Use Wave Beam: https://youtu.be/xd5-5I8YFwM",
+                                                        "comment": "Screw Attack: https://youtu.be/CEmsUhfUjCc or Use Wave Beam: https://youtu.be/xd5-5I8YFwM",
                                                         "items": [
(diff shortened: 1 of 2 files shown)


M4RKERSM4SK:
Added in powerbombs and changed wording on video to just be "Destroy blocks from Above"

JeffGainsNGames (approved):

PR #8119, opened 2025-03-14
Dread: Increase difficulty of getting past fans with cross bombs

Tyranisaur:


MayberryZoom (approved):
lgtm, I agree with this difficulty bump

henriquegemignani (approved):
since may ok'd

PR #8121, opened 2025-03-14
fusion(sector6): Add alternate room id for Clogged Cavern

cluekitty:
Fixes a missing room name in sector 6 after you defeat Varia Core boss.


JeffGainsNGames (approved):

JeffGainsNGames:
tests are failing, since we have a test for the expected output to the patcher (which depends on room IDs) we have to update those files

if you go to `test\games\test_log_file_export.py` and uncomment line 72, and run `pytest test\games\test_log_file_export.py` it will update for you.

PR #8122, opened 2025-03-14
fix migration for banned starting pickups

SchwartzGandhi:


randovania/layout/preset_migration.py
```diff
@@ -1128,13 +1128,13 @@ def pop_from_base_config(field: str) -> None:
...
-    banned_pickups = ["Cannon Ball", "Unlimited Beam Ammo", "Unlimited Missiles", "Double Damage"]
-    pickup_config = preset["configuration"]["standard_pickup_configuration"]["pickups_state"]
     if game == RandovaniaGame.METROID_PRIME_ECHOES:
+        banned_pickups = ["Cannon Ball", "Unlimited Beam Ammo", "Unlimited Missiles", "Double Damage"]
+        pickup_config = preset["configuration"]["standard_pickup_configuration"]["pickups_state"]
         for pickup in banned_pickups:
             if pickup in pickup_config:
                 if "num_included_in_starting_pickups" in pickup_config[pickup]:
-                    pickup_config[pickup]["num_included_in_starting_pickups"] = 0
+                    pickup_config[pickup].pop("num_included_in_starting_pickups")
```
ThanatosGit, line 1137: I don't want to look it up: What changes with this change? **Not** talking about the obvious that it will remove the key instead of setting the amount to 0.
I'm asking because people could have saved the preset in the meantime with the first version.

ThanatosGit (changes requested):

randovania/layout/preset_migration.py
SchwartzGandhi: this was mainly to follow how the other pickups look. I reverted the change to this migration and made a new one instead.

randovania/layout/preset_migration.py
```diff
@@ -1128,15 +1128,25 @@ def pop_from_base_config(field: str) -> None:
...
     if game == RandovaniaGame.METROID_PRIME_ECHOES:
+        banned_pickups = ["Cannon Ball", "Unlimited Beam Ammo", "Unlimited Missiles", "Double Damage"]
+        pickup_config = preset["configuration"]["standard_pickup_configuration"]["pickups_state"]
         for pickup in banned_pickups:
             if pickup in pickup_config:
                 if "num_included_in_starting_pickups" in pickup_config[pickup]:
                     pickup_config[pickup]["num_included_in_starting_pickups"] = 0
 
 
+def _migrate_v102(preset: dict, game: RandovaniaGame) -> None:
```
henriquegemignani, line 1140: There's not really a reason for this migration. If something is wrong with the previous one, fix it since it hasn't hit stable yet.

henriquegemignani (changes requested):
Please add an rdvgame from current stable that has one of the starting pickups

randovania/layout/description_migration.py
```diff
@@ -615,6 +615,14 @@ def _migrate_v32(data: dict) -> None:
                     region_location[new_area_name] = region_location.pop(old_area_name)
 
 
+def _migrate_v33(data: dict) -> None:
+    banned_pickups = ["Cannon Ball", "Unlimited Beam Ammo", "Unlimited Missiles", "Double Damage"]
```
henriquegemignani, line 619: not sure any other game has these names, but best to put a game check

randovania/layout/preset_migration.py
ThanatosGit: I guess that was done because I was asking what was wrong with the one before (but I didn't say it has to be a new one atm).
There is a corner case for me where a new one is better even if didn't hit stable yet. If it was corrupted by the migration before and anyone saved it that way in the meantime, the corrupted one will also be used in stable.

test/test_files/log_files/prime2_starts_with_cannon_ball.rdvgame
```diff
@@ -0,0 +1,926 @@
+{
+    "schema_version": 30,
```
henriquegemignani, line 2: Create a `prime2` folder to add this inside

test/games/test_log_file_export.py
```diff
@@ -124,6 +124,8 @@ def pytest_generate_tests(metafunc: _pytest.python.Metafunc) -> None:
         "prime1_crazy_seed.rdvgame",  # chaos features
         "prime1_crazy_seed_one_way_door.rdvgame",  # same as above but 1-way doors
         "prime1_refills.rdvgame",  # Refill items + custom artifact count
+        # Prime 2
+        "prime2_starts_with_cannon_ball",
```
henriquegemignani, line 128: missing .rdvgame at the end

(and the `prime2/` prefix given the other comment)

henriquegemignani (changes requested):

randovania/layout/description_migration.py
```diff
@@ -615,6 +615,19 @@ def _migrate_v32(data: dict) -> None:
...
+    game_modifications = data["game_modifications"]
+
+    for game in game_modifications:
+        if game["game"] == "prime2":
+            banned_pickups = ["Cannon Ball", "Unlimited Beam Ammo", "Unlimited Missiles", "Double Damage"]
+            if "pickups" in game["starting_equipment"]:
+                starting_pickups = game["starting_equipment"]["pickups"]
+                for pickup in banned_pickups:
+                    if pickup in starting_pickups:
+                        starting_pickups.pop(pickup)
```
ThanatosGit, line 628: That's an arry not a dict.

randovania/layout/description_migration.py
```diff
@@ -615,6 +615,19 @@ def _migrate_v32(data: dict) -> None:
...
 
 
+def _migrate_v33(data: dict) -> None:
+    game_modifications = data["game_modifications"]
+
+    for game in game_modifications:
+        if game["game"] == "prime2":
+            banned_pickups = ["Cannon Ball", "Unlimited Beam Ammo", "Unlimited Missiles", "Double Damage"]
+            if "pickups" in game["starting_equipment"]:
+                starting_pickups = game["starting_equipment"]["pickups"]
```
ThanatosGit, line 625: @henriquegemignani Couldn't this also be in `game["starting_equipment"]["items"]` from a very old rdvgame due to _migrate_v14?

ThanatosGit (dismissed):

randovania/layout/description_migration.py
henriquegemignani: Yes, but then it's not pickups it's items directly and don't think it'll give an error with the new code

ThanatosGit: Don't know if understand correctly (maybe because I don't even know why it needs to be banned). Throwing no error is always a good thing but don't you need to remove it from the `items` in `starting_equiment` in that case?

henriquegemignani: It's perfectly fine to have it be a starting item, it does literally nothing.

ThanatosGit: Ok, then it's fine how it is I guess.

ThanatosGit:
@SchwartzGandhi Dunno how you created the patcher JSON file but there is a line https://github.com/randovania/randovania/blob/e5f4a095ef3da8dec2cb4a682c801a74e2d3a859/test/games/test_log_file_export.py#L72 you can simply activate, run the test locally, and then deactivate the line again. It looks like the patcher file is wrong because tests are still failing.

PR #8125, opened 2025-03-15
Fusion: Unlock Sector Hub

duncathan:
Adds a setting to unlock the doors in sector hub, as well as a new included preset with it enabled alongside my other proposed settings. Since this changes the fusion preset format, old presets will be broken (beta testers can deal with it)
image

This also slightly changes the way DLR is hidden. The tab is now always available, but the actual randomization settings are hidden unless running the client in preview mode (`tools\start_debug_client.bat`)

Normal client: 
image

Debug client:
image


randovania/games/fusion/presets/open_sector_hub.rdvpreset
```diff
@@ -0,0 +1,184 @@
+{
+    "schema_version": 102,
+    "base_preset_uuid": null,
+    "name": "Open Sector Hub",
+    "uuid": "680739a2-cbd8-4c3c-89e7-87ac171cbbb2",
+    "description": "A proposal for a new Starter Preset. Starts in the Sector Hub with 1 random starting item, and unlocks the doors in Sector Hub. This should significantly increase seed variety while remaining trickless.",
```
Miepee, line 6: I'm personally not interested in having a starter preset for fusion that starts at a different location.
Thus I would prefer it, if you'd omit the `A proposal for a new Starter Preset.` part. 

Having it as a second builtin preset is fine for me. Although I wonder if we should rephrase the last sentence, to either say "This significantly..." or just omitting it entirely.

duncathan: it's a proposal. you have yet to even explain why you're disinterested in the proposal

Miepee: For me, the starter preset should be made as intuitive for people coming from vanilla as much as possible. Changing the starting location for me goes against that idea.

Miepee: FWIW if @JeffGainsNGames and the other fusion devs are fine with that, I will concede on that point. 

It's just something I'm personally not a fan of.

duncathan: I can appreciate that preference, but I think it's better suited for discussions about the merits of this proposal once people have had the chance to try it than it is for vetoing the idea of even acknowledging this as a proposal

Miepee: My review comment here is only for the description of the preset.
I'm totally fine with this being another builtin preset that is shipped by default, just as long as it doesn't replace the starter preset. Also do realize that shipping such a preset is *really* helpful for letting people trying it out themselves.

duncathan: when the proposal gets approved or denied later, the description will be changed. I don't understand what the benefit is supposed to be of not acknowledging this proposal for what it is

JeffGainsNGames: > For me, the starter preset should be made as intuitive for people coming from vanilla as much as possible. Changing the starting location for me goes against that idea.

this is the same point I'm stuck on as well. starting in torvus for echoes could (theoretically for this argument) have more diverse seeds, but is not intuitive to a new player why they are starting there.

> I'm totally fine with this being another builtin preset that is shipped by default, just as long as it doesn't replace the starter preset (which this PR doesnt do). Also do realize that shipping such a preset is really helpful for letting people trying it out themselves.

I also agree that this is just a proposal for the settings and it will be way more helpful to test if we provide an actual preset vs having testers try to recreate it themselves.

> I don't understand what the benefit is supposed to be of not acknowledging this proposal for what it is

`A proposal for a new alternate preset`? might just be the implication that it *will* replace starter vs a proposed preset?

duncathan: > might just be the implication that it will replace starter vs a proposed preset?

thankfully, we have the perfect word to indicate that it's only a suggestion and hasn't actually been decided on yet - that word is "proposal" \:P

PR #8128, opened 2025-03-15
Some typing for randovania/gui/lib

henriquegemignani:


randovania/gui/lib/background_task_mixin.py
```diff
@@ -21,12 +22,12 @@ class BackgroundTaskMixin:
...
     _background_thread: threading.Thread | None = None
 
-    def _start_thread_for(self, target):
+    def _start_thread_for(self, target: typing.Callable[[], None]) -> None:
         randovania.games.prime2.patcher.csharp_subprocess.IO_LOOP = asyncio.get_event_loop()
         self._background_thread = threading.Thread(target=target, name=f"BackgroundThread for {self}")
         self._background_thread.start()
 
-    def run_in_background_thread(self, target, starting_message: str):
+    def run_in_background_thread(self, target, starting_message: str) -> None:
```
ThanatosGit, line 30: I'm just repeating mypy here but `target` is missing a type annotation.

ThanatosGit (changes requested):
Coudln't you also add the following to `[tool.mypy]` section?
- randovania/gui/lib/multiplayer_session_api.py
- randovania/gui/lib/data_editor_canvas.py
- randovania/gui/lib/connections_visualizer.py
- randovania/gui/lib/background_task_mixin.py (from a quick look that was not missing much)

Besides that only the obvious thing that tests are failing because of the change from the signal to the callable / emit.

henriquegemignani:
This PR started as a "let's add typing to the gui/lib folder" that I gave up because of stuff being too annoying.

> Coudln't you also add the following to `[tool.mypy]` section?
>     * randovania/gui/lib/multiplayer_session_api.py
>     * randovania/gui/lib/data_editor_canvas.py
>     * randovania/gui/lib/connections_visualizer.py
>     * randovania/gui/lib/background_task_mixin.py (from a quick look that was not missing much)

`randovania/gui/lib/multiplayer_session_api.py` I can't. There's some fun typing challenge for `handle_network_errors` that I'm leaving for later.
`randovania/gui/lib/background_task_mixin.py` is not finished. The `target`s needs some extra works to be correct.

randovania/gui/lib/game_exporter.py
```diff
@@ -54,12 +52,12 @@ def work(progress_update: ProgressUpdateCallable):
 
     except UnableToExportError as e:
         logging.warning(e.reason)
-        progress_update_signal.emit(f"Unable to export game: {e.reason}", 0.0)
+        background.progress_update_signal.emit(f"Unable to export game: {e.reason}", 0.0)
```
ThanatosGit, line 55: That error here is so confusing for me. This is because we cannot type a qt signal properly?

ThanatosGit (approved):

randovania/gui/lib/game_exporter.py
henriquegemignani: yeah, `QtCore.Signal` just doesn't seem to have the `emit` field or the correct signature or be valid typing if you do the `Signal(str, float)`. It's weird.

ThanatosGit: Alright. Then I press the merge button

ThanatosGit: If I understand correctly, I figured that it actually comes from something different.
The type stub for `Signal` is defined as 
```python
    @typing.overload
    def __get__(self, instance: PySide6.QtCore.QObject, owner: typing.Any | None, /) -> PySide6.QtCore.SignalInstance: ...
    @typing.overload
    def __get__(self, instance: None, owner: typing.Any | None, /) -> PySide6.QtCore.Signal: ...
```
`instance` in `BackgroundTaskMixin` is neither `None` nor a `PySide6.QtCore.QObject`. 
Could be "fixed" by inherting from `PySide6.QtCore.QObject` but that will introduce much more trouble because then you get `RuntimeError: You can't initialize an PySide6.QtWidgets.QMainWindow object in class MainWindow twice!`

duncathan: incidentally, this happens to be the only case in the codebase of a `Signal` being used on something that isn't a `QObject`

henriquegemignani: Might be the reason to implement our own `Signal`, I don't like the connectors being QObjects just for it.

PR #8131, opened 2025-03-16
Show which user is unauthorized when trying to login

Miepee:
Does server stuff need changelog entries? My assumption is no.


ThanatosGit:
The rendered template is shown to the user.
...but this is message will not be shown in stable because it's for dev builds. So, I'm fine with skipping changelog.

henriquegemignani (approved):

PR #8132, opened 2025-03-16
Fusion Reservoir West Room Fixes and Videos

M4RKERSM4SK:
Room: Reservoir West

Fixed the impossible connection of having only Varia and Jump Extends with the Water Up

Fixed the inconsistency of 2 of the Damage Runs Intermediate when all other connections (even the same ones back) was beginner. Beginner is for sure fitting for this room in terms of damage runs. You have to go out of your way to reach wires and getting out is obvious.

Fixed the wrong damage type for one of the tricks

Added in the missing trick videos for:

- Jump Extends Beginner
- Shinespark Tricks Beginner


randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1011,20 +1011,25 @@ Extra - room_id: [20]
           After Sector 4 Water Level Lowered
           All of the following:
               # Damage Run
-              Damage Runs (Intermediate)
+              Damage Runs (Beginner)
```
Miepee, line 1014: the reason this one should be intermediate is because the jump is surprisingly tight to pull off. Its not impossible but definitely a step above beginner, since its really easy to miss the ledge and jump into the electric outlet instead.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1011,20 +1011,25 @@ Extra - room_id: [20]
...
-          Any of the following:
-              Jump Extends (Beginner) or Have Any Jump Upgrade
-              Gravity Suit and Speed Booster and Shinespark Tricks (Beginner)
+          All of the following:
+              # Water Lowered allows the tricks without Gravity Suit
+              Any of the following:
+                  # Lower the Water
+                  After Sector 4 Water Level Lowered
+                  # Damage Run in the Water
+                  Gravity Suit and Damage Runs (Beginner) and Electrified Water Damage ≥ 30
```
Miepee, line 1027: there is an argument to be made here that if you combine this with jump extends/shinesparking that this one should be intermediate. at the very least shinesparking feels like it should be a step above.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
M4RKERSM4SK: Without Gravity I complety agree with you. I though I added a Gravity Suit and Damage Runs (beginner) here in addition to the intermediate cause with Gravity it was a peice of cake.

M4RKERSM4SK: I will get the additonal added here

randovania/games/fusion/logic_database/Sector 4 AQA.txt
Miepee: > Are you meaning the tricks are Intermediate or the damage run should be intermediate?

the damage run trick should be bumped to intermediate IMO.

doing both a jump extend or a shinespark have some decent punishment should you fail them. they definitely feel like a step up from most of our other beginner damager runs. It's not just "walking through", it's something that can make people fail the other trick more often because they are now under a small stress situation from the damage

randovania/games/fusion/logic_database/Sector 4 AQA.txt
M4RKERSM4SK: Ok added in it for more review, I beleive I have this calculated right

randovania/games/fusion/logic_database/Sector 4 AQA.txt
M4RKERSM4SK: I see what you mean now and with the punishment factor I agree. I bumped it up to intermediate.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1025,7 +1026,7 @@ Extra - room_id: [20]
                   # Lower the Water
                   After Sector 4 Water Level Lowered
                   # Damage Run in the Water
-                  Gravity Suit and Damage Runs (Beginner) and Electrified Water Damage ≥ 30
+                  Gravity Suit and Damage Runs (Intermediate) and Electrified Water Damage ≥ 30
```
Miepee, line 1029: have any jump upgrade should probably still be beginner tho.
dunno right now by heart how its with HJ, but grav+SJ should not give you any problems.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1012,10 +1012,11 @@ Extra - room_id: [20]
           After Sector 4 Water Level Lowered
           All of the following:
               # Damage Run
-              Damage Runs (Beginner)
+              Damage Runs (Intermediate)
               Any of the following:
                   Electrified Water Damage ≥ 31
-                  Electrified Water Damage ≥ 8 and Have Any Suit Upgrade
+                  Varia Suit and Electrified Water Damage ≥ 8
+          Gravity Suit and Damage Runs (Beginner) and Electrified Water Damage ≥ 8
```
Miepee, line 1019: For me this would be easier to read if it was:
```
ALL
- ANY
  - Damage Runs (Intermediate)
  - Damage Runs (Beginner) and Gravity Suit
- ANY
  - Electrified Water Damage >= 31
  - Electrified Water Damage and Have Any Suit Upgrade
```

M4RKERSM4SK: Yeah that is a lot nicer, changed to this

randovania/games/fusion/logic_database/Sector 4 AQA.txt
M4RKERSM4SK: Changed back to beginner, after more and more testing with it, I am firm now in the stance on this being beginner on the damage run

Miepee: hmm, fine. we technically shouldnt assume that a player has more health than whats given, but most of hte time when you fail the shinespark yuo do indeed have the ability to just get out of the water instead to farm for health.

PR #8144, opened 2025-03-17
Fix all mypy warnings for randovania/layout

henriquegemignani:
This PR depends on https://github.com/randovania/randovania/pull/8141.

### DIFF
--- a/randovania/layout/game_patches_serializer.py
+++ b/randovania/layout/game_patches_serializer.py
@@ -11,7 +11,7 @@ from randovania.game_description.db.area_identifier import AreaIdentifier
 from randovania.game_description.db.dock_node import DockNode
 from randovania.game_description.db.node_identifier import NodeIdentifier
 from randovania.game_description.db.pickup_node import PickupNode
-from randovania.game_description.game_patches import GamePatches
+from randovania.game_description.game_patches import GamePatches, StartingEquipment
 from randovania.game_description.hint import BaseHint
 from randovania.game_description.resources.pickup_index import PickupIndex
 from randovania.game_description.resources.resource_collection import ResourceCollection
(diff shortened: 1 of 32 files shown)


randovania/layout/versioned_preset.py
```diff
@@ -38,19 +39,19 @@ def __init__(self, original_exception: Exception):
...
         else:
-            msg = original_exception
+            msg = str(original_exception)
 
         super().__init__(msg)
         self.original_exception = original_exception
 
 
-class VersionedPreset:
+class VersionedPreset[Configuration: BaseConfiguration]:
```
duncathan, line 48: could you try using typing extensions to set a default for this?
```py
from typing_extensions import TypeVar

Configuration = TypeVar("Configuration", bound=BaseConfiguration, default=BaseConfiguration)

class VersionedPreset(Generic[Configuration]):
```

randovania/layout/game_patches_serializer.py
```diff
@@ -175,6 +224,7 @@ def decode_single(
     starting_location = NodeIdentifier.from_string(game_modifications["starting_location"])
 
     # Starting Equipment
+    starting_equipment: list | ResourceCollection
```
duncathan, line 227: shouldn't this be `game_patches.StartingEquipment`?

randovania/layout/versioned_preset.py
duncathan: possibly on `Preset` too

duncathan (changes requested):

randovania/layout/game_patches_serializer.py
henriquegemignani: I forgot that alias exist. Good catch

duncathan (approved):

PR #8148, opened 2025-03-18
Replace Qt `Signal` with own implementation

ThanatosGit:
Regarding: https://github.com/randovania/randovania/pull/8128#discussion_r1999699537

I have the feeling that I simplified it too much but would this be a solution?
(I skipped replacing all the signals for now and just did it for a few one)
At least it shows a type error for the `ExecutorToConnectorSignals` in AM2R 🙃

### DIFF
--- a/randovania/game_connection/connector/debug_remote_connector.py
+++ b/randovania/game_connection/connector/debug_remote_connector.py
@@ -2,12 +2,11 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from PySide6.QtCore import Signal
-
 from randovania.game_connection.connector.remote_connector import RemoteConnector
 from randovania.game_description import default_database
 from randovania.game_description.resources.inventory import Inventory
 from randovania.game_description.resources.resource_collection import ResourceCollection
+from randovania.gui.lib.signal import RdvSignal
 
 if TYPE_CHECKING:
     import uuid
@@ -22,15 +21,16 @@ class DebugRemoteConnector(RemoteConnector):
     _finished: bool = False
     _last_inventory_event: Inventory
     item_collection: ResourceCollection
-
-    RemotePickupsUpdated = Signal()
-    MessagesUpdated = Signal()
+    RemotePickupsUpdated: RdvSignal
+    MessagesUpdated: RdvSignal
 
     def __init__(self, game: RandovaniaGame, layout_uuid: uuid.UUID):
         super().__init__()
         self._game = game
         self._layout_uuid = layout_uuid
         self._last_inventory_event = Inventory({})
+        self.RemotePickupsUpdated = RdvSignal()
+        self.MessagesUpdated = RdvSignal()
 
         self.messages = []
         self.item_collection = ResourceCollection.with_database(default_database.resource_database_for(game))
--- a/randovania/gui/lib/signal.py
+++ b/randovania/gui/lib/signal.py
@@ -1,0 +1,17 @@
+import typing
+from collections.abc import Callable
+from typing import Generic, TypeVarTuple
+
+FuncArgs = TypeVarTuple("FuncArgs")
+
+
+class RdvSignal(Generic[*FuncArgs]):
+    def __init__(self) -> None:
+        self.subscribers: list = []
+
+    def emit(self, *func_args: *FuncArgs) -> None:
+        for x in self.subscribers:
+            x(*func_args)
+
+    def connect(self, func: Callable[[*FuncArgs], typing.Any]) -> None:
+        self.subscribers.append(func)
--- a/randovania/game_connection/connector/remote_connector.py
+++ b/randovania/game_connection/connector/remote_connector.py
@@ -3,10 +3,7 @@ from __future__ import annotations
 import enum
 import typing
 
-from PySide6 import QtCore
-
-from randovania.game_description.resources.inventory import Inventory
-from randovania.game_description.resources.pickup_index import PickupIndex
+from randovania.gui.lib.signal import RdvSignal
 from randovania.lib import enum_lib
 
 if typing.TYPE_CHECKING:
@@ -15,6 +12,8 @@ if typing.TYPE_CHECKING:
     from randovania.game.game_enum import RandovaniaGame
     from randovania.game_description.db.area import Area
     from randovania.game_description.db.region import Region
+    from randovania.game_description.resources.inventory import Inventory
+    from randovania.game_description.resources.pickup_index import PickupIndex
     from randovania.network_common.remote_pickup import RemotePickup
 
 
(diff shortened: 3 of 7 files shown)


randovania/gui/lib/signal.py
```diff
@@ -0,0 +1,17 @@
+import typing
+from collections.abc import Callable
+from typing import Generic, TypeVarTuple
+
+FuncArgs = TypeVarTuple("FuncArgs")
+
+
+class RdvSignal(Generic[*FuncArgs]):
+    def __init__(self) -> None:
+        self.subscribers: list = []
```
henriquegemignani, line 10: this should have typing for the list

randovania/gui/lib/signal.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def emit(self, *func_args: *FuncArgs) -> None:
+        for x in self.subscribers:
+            x(*func_args)
+
+    def connect(self, func: Callable[[*FuncArgs], typing.Any]) -> None:
```
henriquegemignani, line 16: return value should be None

Probably a good idea to make a type alias for the callable

randovania/game_connection/connector/debug_remote_connector.py
```diff
@@ -22,15 +21,16 @@ class DebugRemoteConnector(RemoteConnector):
     _finished: bool = False
     _last_inventory_event: Inventory
     item_collection: ResourceCollection
-
-    RemotePickupsUpdated = Signal()
-    MessagesUpdated = Signal()
+    RemotePickupsUpdated: RdvSignal
```
henriquegemignani, line 24: shouldn't this be `RdvSignal[]` or is that not a thing?

ThanatosGit: No, because then it expects `index or slice expression`

duncathan:
https://github.com/dgovil/PySignal
found this library, which is abandoned but has solid functionality. we could fork it and add proper typing?

randovania/game_connection/connector/debug_remote_connector.py
duncathan: strictly speaking it has to be `RdvSignal[()]`. if we upgrade to py3.13 we could use typevar defaults to make just `RdvSignal` work

ThanatosGit: Though after some troubleshooting, it probably should be `RdvSignal[()]` to pass mypy.

randovania/gui/lib/signal.py
ThanatosGit: Do you know how to do it?
If I just add something like `Function = Callable[[*FuncArgs], None]` next to the `FuncArgs` it doesn't really work for mypy. I guess because it doesn't know the params for the generic if you type alias outside of the class.
So, you get something like `Argument 1 to "connect" of "RdvSignal" has incompatible type "Callable[[str], None]"; expected "Callable[[VarArg(Any)], None]" `. 

Define it within the class doesn't work for me `Generic type alias within class cannot use bound type variables FuncArgs`

randovania/game_connection/connector/debug_remote_connector.py
henriquegemignani: Alternatively, the argument to RdvSignal could be arguments to Callable. So we'd do `RdvSignal[[int, str], None]`

Goes along with having return values of the listeners being the result of `emit`, but that's very not much not needed

randovania/gui/lib/signal.py
henriquegemignani: Hm.. yeah that can be a problem. It could potentially be a Protocol, but I don't think it helps much.
I imagine explicitly marking it as `TypeAlias` doesn't help?

duncathan:
https://gist.github.com/duncathan/d9a13f3f15914e1e08b323134380476d

thoughts?

henriquegemignani:
> https://gist.github.com/duncathan/d9a13f3f15914e1e08b323134380476d
> 
> thoughts?

I don't like everything around the `sender` functionality. From not thinking it's necessary, being very complex code, being wrong if it's ever emitted while emitting and likely breaks with threads.

I also don't know how useful this "signal is in the class" is. We've never really used that part of Qt's Signal.
Nor do I like the results being via an attribute in the signal, instead of just return value of emit.

duncathan:
> I also don't know how useful this "signal is in the class" is. We've never really used that part of Qt's Signal.

? it's how almost all of our signals are defined, wdym?

> I don't like everything around the sender functionality.

yeah I also didn't like that and would be happier to remove it. we do make use of qt's sender functionality in a couple places which is the only reason I hadn't removed it already

ThanatosGit:
The simple implementation here doesn't work as signal on a class because it will stay connected to already destroyed objects.
And yes, as dunc said, we nearly exclusively use it currently on a class, that can also be seen at the current state of the pr with me removing it from a class and shift it to the instance.

Not a deal breaker but the current state of the gist doesn't work with our tests because we explicitly set them in tests which will fail with `RuntimeError("Cannot reassign an RdvSignal")` and also some of the slots are just `MagicMock`.

duncathan:
> Not a deal breaker but the current state of the gist doesn't work with our tests because we explicitly set them in tests which will fail with RuntimeError("Cannot reassign an RdvSignal")

ah, that's a good reason to implement a real `__set__()` then lol

henriquegemignani:
> > I also don't know how useful this "signal is in the class" is. We've never really used that part of Qt's Signal.
> 
> ? it's how almost all of our signals are defined, wdym?

Because that's how Qt signals, but I don't think we connect to the class signal directly?
I'd rather create the signal per instance directly every time.

> 
> > I don't like everything around the sender functionality.
> 
> yeah I also didn't like that and would be happier to remove it. we do make use of qt's sender functionality in a couple places which is the only reason I hadn't removed it already

I'm fine using qt's signals for qt code directly.
But wow no wonder I didn't remember using `sender()` before. It's [7 years ago](https://github.com/randovania/randovania/commit/61d09a75f1bed581e238f3feb70ad35cab34c934#diff-305c99fa93f168e3f162903db145f06f59b04d31b3167ac95e393f3f03e160cfR75).

duncathan:
> I don't think we connect to the class signal directly?

that's because you can't. the descriptor setup is just some semantic sugar that makes adding a signal less verbose

henriquegemignani:
> The simple implementation here doesn't work as signal on a class because it will stay connected to already destroyed objects.
That's a case of using weakref I guess?
But could also be that we do need to deregister instead of trusting the GC to do it. Trusting the GC is prone to having weird errors.

> Not a deal breaker but the current state of the gist doesn't work with our tests because we explicitly set them in tests which will fail with `RuntimeError("Cannot reassign an RdvSignal")` and also some of the slots are just `MagicMock`.

Tests should be fine to no longer replace the signal with a mock. Should be pretty easy to connect to the signal with a mock and assert it was called, or assert your signal was connected to.

ThanatosGit:
> Tests should be fine to no longer replace the signal with a mock. Should be pretty easy to connect to the signal with a mock and assert it was called, or assert your signal was connected to.

Yeah, but the current gist cannot be connected with a MagicMock because of how `_get_slot` is implemented. Nothing that we couldn't change.

> > The simple implementation here doesn't work as signal on a class because it will stay connected to already destroyed objects.
> 
> That's a case of using weakref I guess? But could also be that we do need to deregister instead of trusting the GC to do it. Trusting the GC is prone to having weird errors.

I need to try it. Otherwise if we want to deregister then e.g. `DebugConnectorWindow` needs to remove what it does with `connector.InventoryUpdated.connect(self.update_inventory_table)`.

Or we just keep it on an per instance level or use the stuff from the gist. Maybe should decide what we want to do exactly.

henriquegemignani:
The only part I care about signals is being to add listeners to a thing, and when that thing fires my listeners are called. I like it enough to want to use it outside of GUI.

I don't care to replicate anything about Qt's Signal api and don't mind refactoring all the connectors to the new api.
I also don't mind to keep using Qt's Signals for Qt classes and our own outside of Qt.

I'll leave which signal implementation to use up to you.

duncathan:
I'll whip up a proper PR with my approach, for better comparison

PR #8154, opened 2025-03-19
Fusion Transmutation Trial Trick and Video Updates

M4RKERSM4SK:
Added in the missing beginner video for Transmutation Trial

Also added in a SWJ trick into the logic\db and video


randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -1694,7 +1694,10 @@ Extra - room_id: [51]
                   Can Freeze Enemies With Any Weapon
                   Any of the following:
                       Hi-Jump
+                      # Stand on Frozen Enemies: https://youtu.be/hkXa7OrjU7Q
                       Stand On Frozen Enemies (Beginner) and Wall Jump (Beginner)
+              # Single Wall Jump: https://youtu.be/DlAA9EITKuY
+              Wall Jump (Advanced) and Can Single Walljump
```
Miepee, line 1700: This needs toggling the SWJ off, so it needs the equipment manipulation misc resource at the very least.

But since SWJ will be an item later unrelated to screw, i'm not sure if we have the ability later to turn it off. So would be good to keep this in mind.

M4RKERSM4SK: Sounds good, will place that in. Whenever the direction is chosen on SWJs as well as equipment manip allowed in air, I can also correct it to whatever the direction is

Miepee (approved):

PR #8162, opened 2025-03-20
Add GameDatabaseView and use it in a few places

henriquegemignani:
This is my solution for the mess that is `filtered_database.py`, `patch_requirements` and all the stuff in [FactorioBootstrap](https://github.com/randovania/randovania/blob/main/randovania/games/factorio/generator/bootstrap.py#L37).

Everything that'd use a GameDescription/RegionList now instead uses a `GameDatabaseView`, which you can put a bunch of proxies in front, such as `LayerFilteredGameDatabaseView` or whatever else to handle the `patch_requirements`.

Of course, using these views during the generator/resolver can likely impact performance massively. This inpact is avoided via the [world-graph branch](https://github.com/randovania/randovania/tree/feature/world-graph), which actually depends on this PR in order to decently handle `Bootstrap.apply_game_specific`.

In this PR I've added the view and used it in a few places. Let me know what you think of the API. Still need to figure out certain things, like how `RegionList.node_name` and similar works.

### DIFF
--- a/randovania/game_description/filtered_game_database_view.py
+++ b/randovania/game_description/filtered_game_database_view.py
@@ -1,0 +1,38 @@
+from __future__ import annotations
+
+import typing
+
+from randovania.game_description.game_database_view import GameDatabaseView, GameDatabaseViewProxy
+
+if typing.TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from randovania.game_description.db.area import Area
+    from randovania.game_description.db.node import Node
+    from randovania.game_description.db.region import Region
+    from randovania.layout.base.base_configuration import BaseConfiguration
+
+
+class LayerFilteredGameDatabaseView(GameDatabaseViewProxy):
+    def __init__(self, original: GameDatabaseView, enabled_layers: set[str]):
+        super().__init__(original)
+        self.enabled_layers = enabled_layers
+
+    @typing.override
+    def node_iterator(self) -> Iterable[tuple[Region, Area, Node]]:
+        for entry in super().node_iterator():
+            if self.enabled_layers.intersection(entry[2].layers):
+                yield entry
+
+
+def filter_view_for_configuration(view: GameDatabaseView, configuration: BaseConfiguration) -> GameDatabaseView:
+    """
+    Creates a new GameDatabaseView that filters
+    :param view:
+    :param configuration:
+    :return:
+    """
+    return LayerFilteredGameDatabaseView(
+        view,
+        configuration.active_layers(),
+    )
--- a/randovania/game_description/game_database_view.py
+++ b/randovania/game_description/game_database_view.py
@@ -1,0 +1,304 @@
+from __future__ import annotations
+
+import abc
+from abc import ABC
+from typing import TYPE_CHECKING, override
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable, Iterator, Sequence
+
+    from randovania.game_description.db.area import Area
+    from randovania.game_description.db.dock import DockType, DockWeakness
+    from randovania.game_description.db.node import Node
+    from randovania.game_description.db.node_identifier import NodeIdentifier
+    from randovania.game_description.db.pickup_node import PickupNode
+    from randovania.game_description.db.region import Region
+    from randovania.game_description.hint_features import HintFeature
+    from randovania.game_description.pickup.pickup_database import PickupDatabase
+    from randovania.game_description.pickup.pickup_entry import PickupModel
+    from randovania.game_description.requirements.base import Requirement
+    from randovania.game_description.resources.item_resource_info import ItemResourceInfo
+    from randovania.game_description.resources.pickup_index import PickupIndex
+    from randovania.game_description.resources.resource_collection import ResourceCollection
+    from randovania.game_description.resources.resource_database import NamedRequirementTemplate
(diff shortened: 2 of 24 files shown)


randovania/game_description/game_database_view.py
```diff
@@ -0,0 +1,304 @@
...
+    @override
+    def node_from_pickup_index(self, index: PickupIndex) -> PickupNode:
+        return self._original.node_from_pickup_index(index)
+
+    @override
+    def area_from_node(self, node: Node) -> Area:
+        return self._original.area_from_node(node)
+
+
+def typed_node_by_identifier[NodeT: Node](game: GameDatabaseView, i: NodeIdentifier, t: type[NodeT]) -> NodeT:
```
duncathan, line 298: any reason this can't be a bound method on `GameDatabaseView`? possibly decorated as `@final`?

randovania/game_description/filtered_game_database_view.py
```diff
@@ -0,0 +1,39 @@
...
+        self.enabled_layers = enabled_layers
+
+    @typing.override
+    def node_iterator(self) -> Iterable[tuple[Region, Area, Node]]:
+        for entry in super().node_iterator():
+            if self.enabled_layers.intersection(entry[2].layers):
+                yield entry
+
+
+def filter_view_for_configuration(view: GameDatabaseView, configuration: BaseConfiguration) -> GameDatabaseView:
```
duncathan, line 28: couldn't this be a classmethod?

randovania/game_description/game_database_view.py
```diff
@@ -0,0 +1,304 @@
...
+
+class GameDatabaseView(ABC):
+    """
+    Provides access to the GameDescription and nested, with support for being filtered for Preset settings.
+
+    These APIs are all expected to be slow and shouldn't be used in any performance sensitive code.
+    """
+
+    @abc.abstractmethod
+    def node_iterator(self) -> Iterable[tuple[Region, Area, Node]]:
```
duncathan, line 145: this should probably be an `Iterator`. `RegionList.all_regions_areas_nodes` is an Iterator despite being typed as Iterable, lol

randovania/game_description/game_database_view.py
```diff
@@ -0,0 +1,304 @@
...
+    def __init__(self, original: GameDatabaseView):
+        self._original = original
+
+    @override
+    def node_iterator(self) -> Iterable[tuple[Region, Area, Node]]:
+        return self._original.node_iterator()
+
+    @override
+    def iterate_nodes_of_type[NodeT: Node](self, node_type: type[NodeT]) -> Iterator[tuple[Region, Area, NodeT]]:
+        return self._original.iterate_nodes_of_type(node_type)
```
duncathan, line 247: I also never know whether it's better to `return` or `yield from` in cases like this

randovania/game_description/game_database_view.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+class GameDatabaseView(ABC):
+    """
+    Provides access to the GameDescription and nested, with support for being filtered for Preset settings.
+
+    These APIs are all expected to be slow and shouldn't be used in any performance sensitive code.
+    """
+
+    @abc.abstractmethod
```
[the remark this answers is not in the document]
henriquegemignani: I started using it later, to make pycharm tell me if I forgot to implement a new method in the proxy

randovania/game_description/game_database_view.py
henriquegemignani: `return` is faster so I'll prefer that.

randovania/game_description/game_database_view.py
henriquegemignani: maybe the solution is just to have it all be `@final`. I was thinking we have the GameDatabaseView as the minimum API, and then something else to build higher level APIs

Right now I'm leaving `node_by_identifier` being re-implemented in GameDescription, but the goal is that'd be just iterate and search.

randovania/game_description/filtered_game_database_view.py
henriquegemignani: I was thinking we could have more filter layers, but that might be guessing too much into the future.

henriquegemignani:
> > Still need to figure out certain things, like how RegionList.node_name and similar works.
> 
> Ugh. `RegionList.node_name` would ideally just not exist tbh. If it weren't for dark world stuff that could just be a simple function/property on `NodeIdentifier`

Guess I have a new PR for Echoes is in the pipeline then.

randovania/game_description/game_database_view.py
```diff
@@ -160,6 +186,15 @@ def node_by_identifier(self, identifier: NodeIdentifier) -> Node:
         Raises KeyError if no node could be found.
         """
 
+    @final
+    def typed_node_by_identifier[NodeT: Node](self, i: NodeIdentifier, t: type[NodeT]) -> NodeT:
+        """
+        Wrapper for calling node_by_identifier, followed by an isinstance.
+        """
+        result = self.node_by_identifier(i)
+        assert isinstance(result, t)
```
duncathan, line 195: would be nice to raise an explicit TypeError rather than rely on the assert

duncathan (changes requested):
mypy

duncathan (approved):

PR #8166, opened 2025-03-22
Resolver: Fix unmet damage requirement chain disregarded as additional

Tyranisaur:
When a chain of requirements is unsatisfied, use all requirements between heal nodes to form additional requirements, in order to capture the accumulated damage requirements.

### DIFF
--- a/randovania/game_description/requirements/requirement_or.py
+++ b/randovania/game_description/requirements/requirement_or.py
@@ -88,7 +88,7 @@ class RequirementOr(RequirementArrayBase):
                 return Requirement.trivial()
             if potential_item != Requirement.impossible():
                 isolated_items.append(potential_item)
-        return isolated_items[0] if len(isolated_items) == 1 else RequirementAnd(isolated_items)
+        return isolated_items[0] if len(isolated_items) == 1 else RequirementOr(isolated_items)
 
     def as_set(self, context: NodeContext) -> RequirementSet:
         alternatives: set[RequirementList] = set()
--- a/randovania/game_description/requirements/requirement_or.py
+++ b/randovania/game_description/requirements/requirement_or.py
@@ -80,6 +80,16 @@ class RequirementOr(RequirementArrayBase):
 
         return RequirementOr(final_items, comment=self.comment)
 
+    def isolate_damage_requirements(self, context: NodeContext) -> Requirement:
+        isolated_items = []
+        for item in self.items:
+            potential_item = item.isolate_damage_requirements(context)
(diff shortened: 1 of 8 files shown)


randovania/resolver/resolver_reach.py
```diff
@@ -143,12 +144,24 @@ def calculate_reach(cls, logic: Logic, initial_state: State) -> ResolverReach:
...
                 elif target_node:
                     # If we can't go to this node, store the reason in order to build the satisfiable requirements.
                     # Note we ignore the 'additional requirements' here because it'll be added on the end.
                     if not requirement.satisfied(context, damage_health):
-                        requirements_excluding_leaving_by_node[target_node_index].append(requirement)
+                        full_requirement_for_target = (
+                            requirement
+                            if node.heal
+                            else RequirementAnd([requirement, satisfied_requirement_on_node[node.node_index]])
+                        )
```
duncathan, line 163: since you've repeated this twice, maybe break it out into a helper function

duncathan (changes requested):

randovania/resolver/resolver_reach.py
Tyranisaur: I moved it to a helper function, and then I later realized that duplicating this is an error. Keeping the helper function seems like a good idea though.

randovania/game_description/requirements/requirement_or.py
```diff
@@ -80,6 +80,16 @@ def simplify(self, keep_comments: bool = False) -> Requirement:
...
 
+    def isolate_damage_requirements(self, context: NodeContext) -> Requirement:
+        isolated_items = []
+        for item in self.items:
+            potential_item = item.isolate_damage_requirements(context)
+            if potential_item == Requirement.trivial():
+                return Requirement.trivial()
+            if potential_item != Requirement.impossible():
+                isolated_items.append(potential_item)
+        return isolated_items[0] if len(isolated_items) == 1 else RequirementAnd(isolated_items)
```
duncathan, line 91: shouldn't this be `RequirementOr`?

randovania/resolver/resolver_reach.py
```diff
@@ -46,6 +46,16 @@ def _is_requirement_viable_as_additional(requirement: Requirement) -> bool:
     )
 
 
+def _combine_damage_requirements(
```
duncathan, line 49: docstring pls

duncathan (changes requested):

duncathan (approved):

PR #8167, opened 2025-03-22
Add PickupNode.custom_index_group

henriquegemignani:
Tooling to split Echoes's Light/Dark world into regions, but without changing how generation works.

It's also a cool feature if any game wants to balance how the generator spreads pickups without making region changes.


randovania/game_description/game_migration.py
```diff
@@ -312,6 +312,18 @@ def _migrate_v28(data: dict, game: RandovaniaGame) -> None:
                     node["kind"] = hint_types[node["kind"]]
 
 
+def _migrate_v29(data: dict, game: RandovaniaGame) -> None:
+    for region in data["regions"]:
+        for area in region["areas"].values():
+            for node in area["nodes"].values():
+                if node["node_type"] == "pickup":
+                    custom_index_group = None
+                    if game.value == "prime2":
```
duncathan, line 321: would you mind adding a comment pointing out that cave story is specifically being excluded? (I would like it to be but it might seem like a mistake without acknowledging it)

duncathan (commented):
an integrity check forbidding redundant custom groups might be warranted? if you don't think it is then that's fine tho

duncathan:
> an integrity check forbidding redundant custom groups might be warranted? if you don't think it is then that's fine tho

moments after posting the review I've realized why this wouldn't make sense. carry on

henriquegemignani:
> > an integrity check forbidding redundant custom groups might be warranted? if you don't think it is then that's fine tho
> 
> moments after posting the review I've realized why this wouldn't make sense. carry on

I thought of one to prevent a group with empty name. We could have one to prevent a pickup from being a singleton to help prevent typos, but who knows if someone wants to make a location that the generator loves

randovania/game_description/game_migration.py
henriquegemignani: Is the in_dark_aether thing still actually being used in CS and not just a forgotten extra?

duncathan: only being used to create virtual regions. it's a relic of the time when connections between regions were much more limited. no reason not to convert them to real regions now

duncathan: but I would like it to be excluded from this specific migration since I would actually prefer if the new split regions are grouped separately by the generator

duncathan (approved):

PR #8173, opened 2025-03-24
Add more files to uv's cache-keys

henriquegemignani:


ThanatosGit:
I have read this in the documentation:
> For example, if a project specifies dependencies in pyproject.toml but uses setuptools-scm to manage its version, and should thus be rebuilt whenever the commit hash or dependencies change, you can add the following to the project's pyproject.toml:
```
[tool.uv]
cache-keys = [{ file = "pyproject.toml" }, { git = { commit = true } }]
```
But we have an edge case where the bulk generation job deletes the files but is on the same commit hash as it was from the job before?

ThanatosGit (approved):

henriquegemignani:
Yeah, the issue is uv not rebuilding if one of the generated files is deleted. For daily bulk generation, it's likely because we don't delete the workspace or cache, but a git clean or something runs. Self-hosted runners are different.

PR #8181, opened 2025-03-28
Fix dangerous settings being considered experimental setting

henriquegemignani:
Imagine there's a screenshot here


duncathan (approved):
could you imagine a screenshot for mii?

PR #8190, opened 2025-04-02
Fusion Traning Grounds Video and Clean Up Comments

M4RKERSM4SK:
Added

- The trick video for the beginner logic frozen enemies and wall jump

Changed

- Adjusted comments on other tricks to match and be consistent with what I was seeing in the rest of the db
- Adjusted the existing Frozen Enemies No Wall Jump video comment to the intermediate trick it was meant for (this should also fix the trick report)

I understand the tricks are close, but had a seed where only beginner logic was required and thought the intermedate was the beginner. Made a new video and adjustments in hopes this clears it all up

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 5 ARC.txt
+++ b/randovania/games/fusion/logic_database/Sector 5 ARC.txt
@@ -100,14 +100,14 @@ Extra - room_id: [3, 6]
   * L3 Hatch to Gerubus Gully/Door to Training Grounds
   * Extra - door_idx: (7,)
   > Door to Training Aerie
-      # https://youtu.be/Qe9eMPmoUcU
+      # Shinespark: https://youtu.be/Qe9eMPmoUcU
       Level 3 Keycard and Speed Booster and Shinespark Tricks (Intermediate) and Wall Jump (Advanced) and Disabled Door Lock Rando and Disabled Entrance Rando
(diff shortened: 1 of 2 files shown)


randovania/games/fusion/logic_database/Sector 5 ARC.txt
```diff
@@ -134,10 +134,11 @@ Extra - room_id: [3, 6]
       Any of the following:
           Space Jump
           All of the following:
-              # Freeze Chute Leeches: https://youtu.be/XeE_DkCnuGI
               Can Freeze Enemies With Any Weapon
               Any of the following:
+                  # Freeze Chute Leeches (No Wall Jump): https://youtu.be/XeE_DkCnuGI
```
Miepee, line 139: can you change the comment to `No Wall Jump, NHJ` or something? just so its clear that that ones specifically without HJ.

Miepee (commented):
disagree that they are close, doing the NHJ one is way more precise on where you need to hit the first chute leech.

randovania/games/fusion/logic_database/Sector 5 ARC.txt
M4RKERSM4SK: Yep, no problem, completed

Miepee (approved):

PR #8195, opened 2025-04-02
9.0 - Main to Stable

henriquegemignani:
Can someone review this small pr, kthxbye


Miepee:
1) after #8196

duncathan:
#8077 should really be attended to imo

Miepee:
that was my 2) if i hadn't fat-fingered my shift key lol

henriquegemignani:
Fixing the changelog can be done at any time since we only care about what's in `main` and what's attached to the GH release

duncathan:
that's fine, I just wanna make sure the changelog is cleaned up for the actual release

PR #8206, opened 2025-04-03
a try to fix issue #8071

Jason-Bee:
Issue https://github.com/randovania/randovania/issues/8071

wording might still need tweaking, but tried to describe it working for all games instead of just echoes examples, iam also unsure if this needs to be adressed in the changelog
picture


Miepee:
> iam also unsure if this needs to be adressed in the changelog

It does

Miepee (commented):
Few notes:
- we try our best to use the term Pickup over Item
- im not quite a fan of how Major is used here. My intuition is right now that they should be rephrased to either use a pronoun or the term Pickup
- gameworld feels better in lowercase

Cc @henriquegemignani cause he had a few notes from the last time this was attempted by Shale

Miepee:
I'm fine with a "disguised as generic" as a bridge for now. While the concepts itself may not apply to every game, I feel like making a fully generic text is not doable (would like to be proven wrong tho) and the trick tab is very likely to not be reread by people when trying different games so a unique-per-game text feels like itd hurt more.

Jason-Bee:
@Miepee something more like this then? 

image

randovania/gui/ui_files/preset_trick_level.ui
```diff
@@ -69,7 +69,7 @@
...
+           <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p align=&quot;justify&quot;&gt;Randovania has rules in place which guarantees that the game is completable regardless of the modifications made to the game. Here you can also configure which kind of game knowledge or skill it expects you to have, allowing for even more varied games.&lt;/p&gt;&lt;p align=&quot;justify&quot;&gt;No matter the level, it is always possible to softlock when you enter a room or area that you're unable to leave. For example, vanilla item rooms without the necessary pickup to escape, a pickup that enables Progression being past a room that requires it to traverse back.&lt;/p&gt;&lt;p align=&quot;justify&quot;&gt;However, it may be &lt;span style=&quot; font-style:italic;&quot;&gt;necessary&lt;/span&gt; to enter a later part of the gameworld to find a pickup that is also needed to traverse back.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
Miepee, line 72: > For example, vanilla item rooms without the necessary pickup to escape

I would personally just phrase it as "rooms without the necessary pickup to go back"

> a pickup that enables Progression being past a room that requires it to traverse back

This sounds really confusing right now.

> a later part of the gameworld

on second thought, it should probably be `game world`.

Jason-Bee: ill get to that later 👍

Jason-Bee:
image
so this is what i have right now, seems a lot better. wdy think Miepee?

randovania/gui/ui_files/preset_trick_level.ui
```diff
@@ -69,7 +69,7 @@
...
+           <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p align=&quot;justify&quot;&gt;Randovania has rules in place which guarantees that the game is completable regardless of the modifications made to the game. Here you can also configure which kind of game knowledge or skill it expects you to have, allowing for even more varied games.&lt;/p&gt;&lt;p align=&quot;justify&quot;&gt;No matter the level, it is always possible to softlock when you enter a room or area that you're unable to leave. For example, rooms without the necessary pickup to go back, Progression being past a point of no return.&lt;/p&gt;&lt;p align=&quot;justify&quot;&gt;However, it may be &lt;span style=&quot; font-style:italic;&quot;&gt;necessary&lt;/span&gt; to enter a later part of the game world to find a pickup that is also needed to traverse back.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
Miepee, line 72: "... **or** progression being past a point of no return."

Miepee, line 72: something about "that is also needed to traverse back" doesn't quite sit right with me, but can't point my finger at it.

Miepee:
Sorry Jay

PR #8209, opened 2025-04-04
Fix Speed Booster Upgrades being included as hintable items

ArcanoxDragon:
SBUs are now excluded from the credits (and thus the hinted item pool). The description was also changed to take randovania/open-dread-rando#398 into account, and to be less precise about the times since the minimum charge time is no longer exactly a multiple of 0.25 seconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Dread
 
+- Changed: The maximum number of Speed Booster Upgrade items that will have an effect is now 4 instead of 5 (any further upgrades have no effect). This is to prevent a quirk with inconsistent Speed Booster activation with very short charge times.
 - Fixed: The patching data being exported on race seeds.
+- Fixed: Speed Booster Upgrade items will no longer be hinted or included in the credits.
 
 #### Logic Database
 
--- a/randovania/games/dread/pickup_database/pickup-database.json
+++ b/randovania/games/dread/pickup_database/pickup-database.json
@@ -892,7 +892,7 @@
                 "am2r": "sItemSpeedBoosterUpgradeDread"
             },
             "preferred_location_category": "minor",
-            "description": "Each Speed Booster Upgrade reduces the amount of time Speed Booster takes to charge by one quarter of a second.\nDue to game limitations, the minimum allowed charge time is approximately half a second.",
+            "description": "Each Speed Booster Upgrade reduces the amount of time Speed Booster takes to charge by ~0.25 seconds.\nDue to game limitations, the minimum allowed charge time is approximately 0.5 seconds.",
             "progression": [
                 "SpeedBoostUpgrade"
             ],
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [9.1.x] - 2025-05-??
 
+### Metroid Dread
+
+- Changed: The maximum number of Speed Booster Upgrade items that will have an effect is now 4 instead of 5 (any further upgrades have no effect). This is to prevent a quirk with inconsistent Speed Booster activation with very short charge times.
+- Fixed: Speed Booster Upgrade items will no longer be hinted or included in the credits.
```
Miepee, line 13: They can still be hinted at. But its a *much* lower chance now.

randovania/games/dread/pickup_database/pickup-database.json
```diff
@@ -884,18 +884,19 @@
...
+                "expansion",
                 "misc"
             ],
             "model_name": "item_speedboostupgrade",
             "offworld_models": {
                 "am2r": "sItemSpeedBoosterUpgradeDread"
             },
             "preferred_location_category": "minor",
-            "description": "Speed Booster Upgrades reduce the amount of time it takes Speed Booster to activate by 0.25 seconds (for each upgrade collected).\nThe minimum charge time is 0.25s.",
+            "description": "Each Speed Booster Upgrade reduces the amount of time Speed Booster takes to charge by one quarter of a second.\nDue to game limitations, the minimum allowed charge time is approximately half a second.",
```
Miepee, line 895: While im here: i personally find reading 0.25 and 0.5 easier than writing those out into words. I would assume its also easier for other non-english natives.
But idk if its significant enough to actually get changed.

CHANGELOG.md
duncathan: The chance is effectively 0

randovania/games/dread/pickup_database/pickup-database.json
ArcanoxDragon: I didn't want to use exact values because I want to avoid the illusion of it being too precise. Given that the final upgrade reduces by _less than_ 0.25 seconds, and the minimum time is _slightly more than_ 0.5 seconds, I figured that approximated language would be better.

Miepee: 1) could just use "approximately 0.25" then
2) are these significant enough where someone would actually notice it?

CHANGELOG.md
Miepee: Was mostly thinking to prevent people reporting that as a bug when they read the changelog and try to craft a preset to specifically trigger them

But i guess the chance of that happening is also effectively 0 lol

Miepee:
You also do need to fix the tests.

randovania/games/dread/pickup_database/pickup-database.json
Tyranisaur: I'm not sure if has anything with not being a native english speaker, but I also think writing out the numbers with words is a bit jarring. Especially considering they were written with numbers before. Also, I don't think using words is a meaningful way to dampen the notion of the values being exact.

dyceron: Is it even necessary to mention how much it changes it by? Why not just say that the charge time is reduced?

PR #8213, opened 2025-04-05
Remove most functions for naming nodes from RegionList

henriquegemignani:
`RegionList.area_name` needs a bit more thinking, left for later

### DIFF
--- a/randovania/layout/game_patches_serializer.py
+++ b/randovania/layout/game_patches_serializer.py
@@ -51,7 +51,7 @@ def _pickup_assignment_to_item_locations(
         else:
             item_name = _ETM_NAME
 
-        items_locations[region.name][region_list.node_name(node)] = item_name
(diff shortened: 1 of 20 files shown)


randovania/layout/game_patches_serializer.py
```diff
@@ -51,7 +51,7 @@ def _pickup_assignment_to_item_locations(
         else:
             item_name = _ETM_NAME
 
-        items_locations[region.name][region_list.node_name(node)] = item_name
+        items_locations[region.name][f"{node.identifier.area}/{node.identifier.node}"] = item_name
```
duncathan, line 54: isn't this just `node.full_name()`?

henriquegemignani: Without region, yes. But I wanted to not use a thing mostly for visual feedback into the core serialization function.

duncathan: add a comment then pls

duncathan (approved):

PR #8215, opened 2025-04-06
add extra to pickupentry

duncathan:
this makes it a lot more convenient to access in the PDF and in connectors, where this is the most relevant. especially helpful for generated pickups, where the old pattern of getting a pickup definition by name wasn't possible


Miepee (commented):
I think the fusion pdf could be adjusted for this now too?

duncathan:
> I think the fusion pdf could be adjusted for this now too? cant remember the code too well there

fusion PDF was doing something a bit different where it was iterating through the ammo pickup configuration in the preset to do something - looked different enough that I decided to just leave it be

Miepee (approved):

PR #8216, opened 2025-04-06
Fusion Hide Required Main Fixes

M4RKERSM4SK:
Fixes https://github.com/randovania/randovania/issues/7959

Felt this implementation was the best middle ground. There is no nice way to add in a check for hiding the required mains in the item pool gui for ammo. In fact, there is a "fix me" for prime 1 that just hard checks if the game is prime 1 and then hides required mains.

Trying to add anything into the definitions/properties for ammo_pickups blew everything sky high. Changing ammo being not temporary (so item pool gui doesnt even do required mains) blew up logic and other stuff.

In the end, I decided to add in a value to extra for the ammo in the pickup database instead. Then added a check in the item pool gui for this value. This also safely didnt hurt other games.

I could probably fix the prime 1 hardcode as well by adding this extra value for the ammo in the pickup database for prime 1 in a seperate PR if this is a acceptable fix.

I also looked around and didnt see any mention of required mains in the difference tab or otherwise in the Fusion gui. Not sure if I missed something, but hopefully this should be it.

### DIFF
--- a/randovania/gui/preset_settings/item_pool_tab.py
+++ b/randovania/gui/preset_settings/item_pool_tab.py
@@ -388,10 +388,8 @@ class PresetItemPool(PresetTab, Ui_PresetItemPool):
                 require_main_item_check.stateChanged.connect(partial(self._on_update_ammo_require_main_item, ammo))
                 if self.game == RandovaniaGame.METROID_PRIME:
                     require_main_item_check.setVisible(False)
-                for extra, value in ammo.extra.items():
-                    if "hide_requires_main" in extra:
-                        if value:
-                            require_main_item_check.setVisible(False)
+                if ammo.hide_requires_main:
+                    require_main_item_check.setVisible(False)
                 layout.addWidget(require_main_item_check, current_row, 0, 1, -1)
                 current_row += 1
             else:
--- a/randovania/gui/preset_settings/item_pool_tab.py
+++ b/randovania/gui/preset_settings/item_pool_tab.py
@@ -388,6 +388,10 @@ class PresetItemPool(PresetTab, Ui_PresetItemPool):
                 require_main_item_check.stateChanged.connect(partial(self._on_update_ammo_require_main_item, ammo))
                 if self.game == RandovaniaGame.METROID_PRIME:
                     require_main_item_check.setVisible(False)
+                for extra, value in ammo.extra.items():
+                    if "hide_requires_main" in extra:
+                        if value:
+                            require_main_item_check.setVisible(False)
                 layout.addWidget(require_main_item_check, current_row, 0, 1, -1)
                 current_row += 1
             else:
--- a/randovania/gui/preset_settings/item_pool_tab.py
+++ b/randovania/gui/preset_settings/item_pool_tab.py
@@ -381,13 +381,10 @@ class PresetItemPool(PresetTab, Ui_PresetItemPool):
             add_column(pickup_spinbox)
             current_row += 1
 
-            # FIXME: hardcoded check to hide required mains for Prime 1
             if ammo.temporary:
                 require_main_item_check = QtWidgets.QCheckBox(pickup_box)
                 require_main_item_check.setText("Requires the main item to work?")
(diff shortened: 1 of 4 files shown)


randovania/gui/preset_settings/item_pool_tab.py
```diff
@@ -388,6 +388,10 @@ def add_column(widget: QtWidgets.QWidget) -> None:
                 require_main_item_check.stateChanged.connect(partial(self._on_update_ammo_require_main_item, ammo))
                 if self.game == RandovaniaGame.METROID_PRIME:
                     require_main_item_check.setVisible(False)
+                for extra, value in ammo.extra.items():
+                    if "hide_requires_main" in extra:
+                        if value:
```
ThanatosGit, line 393: Not that I had seriously thought about a better way to handle this stuff but this three lines are just `if ammo.extra.get("hide_requires_main", False)`, no?

M4RKERSM4SK: `.get()` will set the default value of the key to false if not found. Which for other games, I didn't research far enough to see how this would impact them. Thought it would be better to just check if its even there first then act.

Miepee: `.get()` doesnt set anything in the dictionary.
doing `my_dict.get("foo", False)` is sugar for
```py
if "foo" in my_dict:
  return_value = my_dict["foo"]
else:
  return_value = False
```

M4RKERSM4SK:
> So curious to know what "ammo_pickups blew everything sky high" exactly meant.

When I created a new property it just broke everywhere. Mainly code (maybe I am misinterpreting or too tired) that populated variables traversing properties of ammo_pickups. Almost like the new property just threw off what was expected? JSON errors on parsing it from what I could tell. I think this is maybe why the hard code for prime was even put in and a fix me for later?

Miepee:
Sounds like for me that you put the new field in a place where you would've needed to then adjust all the constructors as well.
Changing fields should definitely not be an issue since we did that quite often the last few months.

randovania/gui/preset_settings/item_pool_tab.py
M4RKERSM4SK: Is this not true?

```
.get(keyname, value)

Parameters

keyname	Required. The keyname of the item you want to return the value from

value Optional. A value to return if the specified key does not exist.
Default value None
```

Cause it was returning "false" when on other games that didn't have this in the ammo.extra dicts?

M4RKERSM4SK: I never set the optional to false :( it was returning none. Will give that a shot in the morning after sleep.

M4RKERSM4SK:
Sounds good, will see what I am doing wrong here and try again after some sleep and more specific errors to help guide this better

randovania/gui/preset_settings/item_pool_tab.py
Miepee: the docstring is true. as you can see there, it doesn't set anything in the dictionary. "A value to return" => function return value.
I think the way that the function is actually implemented underneath is something akin to
```py
def get(self, keyname, value = None):
  try:
    return self._dict_data[keyname]
  except KeyError
    return value
```

M4RKERSM4SK:
Ok found out my issue on the property change was not setting a defualt value causing a json decoder error across all games.

randovania/gui/preset_settings/item_pool_tab.py
M4RKERSM4SK: Very interesting that its part of catching the KeyError. Very good info and I for sure learned something today. Since this got changed with doin the property instead of extra, I think I am good to resolve this conversation.

Miepee (commented):
can you fix the prime 1 hardcode too while you'Re at it?

M4RKERSM4SK:
Sure, do you want the prime 1 fix in this PR or a seperate one for prime 1?

Miepee:
in this one

M4RKERSM4SK:
Added in prime 1 fix and remvoed hardcode

Miepee (approved):

PR #8224, opened 2025-04-08
CI: Support server_environment for production

henriquegemignani:
Turns our GH is nice and dynamically update the workflow dispatch parameters based on the selected branch.


Miepee (approved):
and once again!

PR #8226, opened 2025-04-08
Generator: reduce weight if an action only unlocks one other action.

Miepee:
Experimental. Opening now to get easier testing done on non-Prime1 games.
Afer I look at the results I will probably introduce a preset setting.

### DIFF
--- a/randovania/generator/filler/filler_library.py
+++ b/randovania/generator/filler/filler_library.py
@@ -110,14 +110,14 @@ class UncollectedState(NamedTuple):
 
         return unsafe_state - safe_state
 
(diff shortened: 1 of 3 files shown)


Tyranisaur:
Maybe instead of this, it would be an idea to contemplate how many pickup locations have been picked up already and haven't gotten an item assigned yet. Considering this, it's possible to make a judgement that if you have nowhere to place an item after the current action, then anything that only unlocks a hint is a bad idea. However, if you have some number of unused pickup locations, then doing something that only unlocks hints is more acceptable. Also, maybe it's worth to jack up the weight of events if you have a solid number of unused pickups.

PR #8240, opened 2025-04-10
Fix: Async room exporting spoilers for am2r+msr

Miepee:
Part 1 of 2


randovania/server/async_race/room_api.py
```diff
@@ -369,12 +370,15 @@ def join_and_export(sa: ServerApp, room_id: int, auth_token: str, cosmetic_json:
     preset = layout_description.get_preset(players_config.player_index)
     cosmetic_patches = preset.game.data.layout.cosmetic_patches.from_json(cosmetic_json)
 
+    # Overwrite spoiler property, as some games rely on this to provide another alternative to show spoilers.
+    layout_description = dataclasses.replace(
```
henriquegemignani, line 374: This changes the permalink, so this is wrong.

henriquegemignani (changes requested):

randovania/server/async_race/room_api.py
Miepee: oh. i'll come up with an alternative then.

henriquegemignani:
For that definition of `PatcherDataMeta`, it's invalid typing to not have all the keys.

henriquegemignani (approved):

henriquegemignani:
in case you wanted to, this is not the kind of change to cherry pick for a 9.0.2

PR #8241, opened 2025-04-10
Fix: patching data being exported on race seeds for some games

Miepee:
Part 2 of 2


henriquegemignani (changes requested):
This `_export_spoiler_json` field makes more sense being part of `_randovania_meta` created by the base PDF.

Miepee:
Hmm i thought that field got popped before it reaches the exporter? Need to look at it again.

Miepee:
Yeah, `_randovania_meta` gets popped off right before the patch data gets to the exporter: https://github.com/randovania/randovania/blob/bbdb25b6bf1bb5ee316be2618d241ce5991503d9/randovania/exporter/game_exporter.py#L80
The reason this was done IIRC was because certain games have a very strict schema, where if you provide any other value, the patcher would raise an exception.
So either we add the field in each games' PDF that wants to do something with it. Or we add a case in each games' exporter that doesnt want/cant have the `_randovania_meta` field.

henriquegemignani:
Just add an argument to `_do_export_game`?

Miepee:
Oh. yeah that'd work too.

henriquegemignani (approved):

PR #8243, opened 2025-04-11
Using GameDatabaseView even more

henriquegemignani:


randovania/games/samus_returns/generator/base_patches_factory.py
```diff
@@ -15,13 +15,15 @@
...
-    from randovania.game_description.game_description import GameDescription
+    from randovania.game_description.game_database_view import GameDatabaseView
 
 
 class MSRBasePatchesFactory(BasePatchesFactory[MSRConfiguration]):
     def dock_connections_assignment(
-        self, configuration: MSRConfiguration, game: GameDescription, rng: Random
+        self, configuration: MSRConfiguration, game: GameDatabaseView, rng: Random
     ) -> Iterable[tuple[DockNode, Node]]:
+        yield from super().dock_connections_assignment(configuration, game, rng)
```
ThanatosGit, line 25: What is the reason to call it here (and also in the other `base_patches_factory` classes)?
The parent class just do `yield from []`. 
Can't we just remove it and add `@override` annotation?

randovania/resolver/state.py
```diff
@@ -108,6 +114,25 @@ def game_state_debug_string(self) -> str:
...
         return self.damage_state.debug_string(self.resources)
 
+    def _advance_to(
+        self,
+        new_resources: ResourceCollection,
+        new_collected_resource_nodes: tuple[ResourceNode, ...],
+        damage_state: DamageState,
+        patches: GamePatches,
+    ) -> Self:
+        return State(
```
ThanatosGit, line 124: Annotate the class with `@final` such that mypy knows that `Self` is really always `State`.

ThanatosGit (changes requested):

randovania/games/samus_returns/generator/base_patches_factory.py
henriquegemignani: Whenever you re-implement a function from a parent class, you should have the `@override` annotation. It's purpose is to indicate you're doing so and error if such a method doesn't exist there, in case it's removed/renamed.

It's also good practice to always call super in such methods, to keep the original functionality working. It lets us change it in the future without having to go edit everything that is missing the super class.

ThanatosGit: The `@override` annotation is missing which was one of my points :p

> It's also good practice to always call super in such methods, to keep the original functionality working.

I wouldn't generalise it that way. An easy example is getter function were we have overrides and it is meaningless to call the parent method.
In other methods it is also a valid use case that you are overriding because you explicitly don't want to do what the parent class does.
Anyways, atm the super call doesn't really do anything and if the implementation in the parent class changes, I cannot tell you right now if all the sub classes which are overriding the method needs that behaviour additionally or not. I guess keeping it is fine but also removing it is fine. So, we can just keep it as it is.

henriquegemignani: An overridden method that doesn't call super should be the rare exception, not the norm.

ThanatosGit: `@override` has currently 200 hits in my local branch and most of them are not calling super because they are:
- just overwriting a getter method
- overwriting methods which are just implemented as a `raise NotImplementedError` stub
- don't care about the parent implementation e.g. things like `get_specific_pickup_precision_pairs` where the parent just returns an empty dict.

PR #8244, opened 2025-04-11
Generator: Avoid automatically collecting dangerous resources

Tyranisaur:


randovania/generator/reach_lib.py
```diff
@@ -124,7 +124,7 @@ def advance_reach_with_possible_unsafe_resources(previous_reach: GeneratorReach)
 
     previous_safe_nodes = set(previous_reach.safe_nodes)
 
-    for action in get_collectable_resource_nodes_of_reach(previous_reach):
+    for action in _get_safe_resources(previous_reach):
```
duncathan, line 127: uhhh are you sure this is correct? I'm pretty sure that the old behavior was doing what it intended to

Tyranisaur: If the old behavior intended to collect a dangerous resource for no gain here, then maybe that would be true, otherwise not.

duncathan: it's certainly intended to *check* dangerous resources. that's the entire point of the function

Tyranisaur: After taking the Araria EMMI first chase blob, it thinks it's completely fine to simply collect the arbitrary enky blob in this function, because doing that doesn't take away reachable nodes when it has flash shift. The problem is that this takes away the possibility of going into the invisible corpius room and placing anything that isn't morph ball or charge beam, otherwise it gets stuck.

duncathan: then the fix will have to involve the heuristic for whether a potentially dangerous action is actually safe. not sure what it would need to be to account for that situation

Tyranisaur: How about the version I just pushed, where the first recursive branch requires the action to not give dangerous resources, while the second recursion checks for going back to where you came from and taking all safe?

duncathan: oh that looks much more like a proper fix. checking for Dangerous resources is perfectly in line with what this function is supposed to do

duncathan (approved):

PR #8245, opened 2025-04-12
fix data editor links failing when special characters are present

duncathan:


test/gui/lib/test_data_editor_links.py
```diff
@@ -38,20 +42,24 @@ def test_click_on_link(skip_qtbot, trick_levels):
...
-def test_data_editor_href(echoes_game_description, text: str | None, area_id: AreaIdentifier):
+def test_data_editor_href(text: str | None, game: RandovaniaGame, area_id: AreaIdentifier):
     # Setup
-    region, area = echoes_game_description.region_list.region_and_area_by_area_identifier(area_id)
+    description = default_database.game_description_for(game)
+    region, area = description.region_list.region_and_area_by_area_identifier(area_id)
     if text is None:
         text = get_human_readable_region_and_area(region, area)
 
+    q = urllib.parse.quote
```
ThanatosGit, line 59: I'm not a fan of one letter long variable names besides the obvious loop variables but it's only used a few lines later once.

ThanatosGit (approved):

test/gui/lib/test_data_editor_links.py
Miepee: im surprised that ruff doesnt complain here tbh. we've had to do some renaming in mars-py cause it did there.

PR #8246, opened 2025-04-12
Generator: Fix not accounting for dangerous resource when evaluating

Tyranisaur:
When evaluating an action that gives a dangerous action, use reach.act_on, because this function accounts for the new dangerous resource in terms of removing nodes as safe in the reach.

### DIFF
--- a/randovania/generator/reach_lib.py
+++ b/randovania/generator/reach_lib.py
@@ -137,14 +137,12 @@ def advance_reach_with_possible_unsafe_resources(previous_reach: GeneratorReach)
             # print("Non-safe {} was good".format(logic.game.node_name(action)))
             return advance_reach_with_possible_unsafe_resources(next_reach)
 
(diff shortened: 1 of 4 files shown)


randovania/generator/reach_lib.py
```diff
@@ -152,15 +152,12 @@ def advance_reach_with_possible_unsafe_resources(previous_reach: GeneratorReach)
...
 
-def advance_to_with_reach_copy(base_reach: GeneratorReach, state: State) -> GeneratorReach:
+def advance_after_action(potential_reach: GeneratorReach) -> GeneratorReach:
     """
-    Copies the given Reach, advances to the given State and collect all possible resources.
-    :param base_reach:
-    :param state:
+    Collect all possible resources.
+    :param potential_reach:
     :return:
```
Miepee, line 159: if youre changing the docstring you might as well fill the param and return out :P

duncathan (approved):

PR #8251, opened 2025-04-15
adjust width of data visualizer and area view

JeffGainsNGames:
part of #8022 

some previews
image

image


Miepee:
would be nice to automatically save those, so that if you adjust them and restart RDV you dont have to adjust them again.

JeffGainsNGames:
Is that a request for that to be done here? Or just stating a desire?

Miepee (approved):
i guess the latter more than the former
Please write a changelog entry tho, as this *is* a gui change.

Miepee (approved):

PR #8253, opened 2025-04-18
Fix Echoes split migration with dock weaknesses

henriquegemignani:


Miepee:
Why were tests previously not failing? Do we not have a test case for this?

duncathan:
I'm not sure if we have an echoes DLR rdvgame we test, no

henriquegemignani:
It's actually from open save stations, which was also not tested.

henriquegemignani:
Added a test case

duncathan (approved):

duncathan (changes requested):
changelog

henriquegemignani:
It's a regression from a dev build.

duncathan (approved):
oh, I thought it was in stable

PR #8254, opened 2025-04-19
Fusion: fix guaranteed hints

### LINKED ISSUE #8120: Fusion: Guaranteed hints don't behave properly when starting with hinted item
only confirmed with charge hint so far, but I imagine if you have 0 shuffled metroids it probably has a similar issue
Image

JeffGainsNGames:
fixes #8120


ThanatosGit (commented):
I suggest to add or modify one of the existing tests to verify that it will be fixed and stay fixed.

PR #8259, opened 2025-04-21
OSRR Bump

ThanatosGit:
Just a draft to discuss some things


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [9.2.x] - 2025-06-??
+
+### Metroid: Samus Returns
+
+- Changed: Display the number of remaining DNA in the DNA message before entering the final boss.
+- Fixed: If Metroid Queen is the final boss and you don't have all DNA, the DNA message now appears.
+- Fixed: Minimap icon if something with hide model and changed to something with icons
```
ThanatosGit, line 14: No idea what is the desired behaviour, so I just wrote some non sense.

randovania/games/samus_returns/exporter/patch_data_factory.py
```diff
@@ -204,6 +223,33 @@ def game_enum(self) -> RandovaniaGame:
...
+            return "item_powerup"
+
+        def get_icon_from_cond_res(cond_res: ConditionalResources) -> str:
+            if not cond_res.resources:
+                return "item_nothing"
+            for resource in cond_res.resources:
+                return get_icon_from_item_res(resource[0])
+            return "item_powerup"
+
+        # TODO: Do something with self.configuration.pickup_model_style
```
ThanatosGit, line 242: What styles from all these different methods for hiding things, should show which icon?

randovania/games/samus_returns/exporter/patch_data_factory.py
```diff
@@ -204,6 +223,33 @@ def game_enum(self) -> RandovaniaGame:
...
+        # TODO: How do I know that it "is_remote" but actually coop ? Check the game enum in model???
```
ThanatosGit, line 243: For coop, `is_remote` is True, but it should not use `item_offworld` as a icon.
Is there a proper way to know if it is coop?
If I understand it correctly, we need to distinguish between `is_remote` for other worlds and use the `item_offworld` icon, but if it is `is_remote` but for the same world, we need to get the icon in the "normal" way?

ThanatosGit (commented):
@dyceron

dyceron (commented):
This seems fine to me. I guess this kinda handles the Metroids, but honestly, I don't know of a better solution unless you look at the caption.

dyceron (approved):

PR #8261, opened 2025-04-22
Hunters - PDF: Support starting items and nothing pickup

dyceron:


randovania/games/prime_hunters/exporter/patch_data_factory.py
```diff
@@ -52,16 +61,56 @@ def force_field_index_for_requirement(game: GameDescription, requirement: Layout
...
 
-    def _calculate_starting_inventory(self, resources: ResourceCollection) -> dict[str, str]:
-        result = {
-            # TODO: Take starting weapons into account
-            "weapons_string": "00000101",
-        }
+    def _calculate_starting_inventory(self, resources: ResourceCollection) -> dict[str, str | int]:
+        result: dict[str, str | int] = {}
+        starting_items = {}
+        resource_gain = [(resource, quantity) for resource, quantity in resources.as_resource_gain()]
```
duncathan, line 83: this appears to be redundant

randovania/games/prime_hunters/exporter/patch_data_factory.py
```diff
@@ -52,16 +61,56 @@ def force_field_index_for_requirement(game: GameDescription, requirement: Layout
...
+                starting_items.get(_ITEM_TO_ITEM_TYPE["Magmaul"], 0),
+                starting_items.get(_ITEM_TO_ITEM_TYPE["Judicator"], 0),
+                starting_items.get(_ITEM_TO_ITEM_TYPE["Imperialist"], 0),
+                starting_items.get(_ITEM_TO_ITEM_TYPE["Battlehammer"], 0),
+                1,  # Missiles
+                starting_items.get(_ITEM_TO_ITEM_TYPE["Volt Driver"], 0),
+                1,  # Power Beam
+            ]
+        )
+        starting_weapons = "".join(filter(str.isdigit, weapons))
```
duncathan, line 104: what's the point of the filter here?

randovania/games/prime_hunters/exporter/patch_data_factory.py
```diff
@@ -52,16 +61,56 @@ def force_field_index_for_requirement(game: GameDescription, requirement: Layout
...
+        result["weapons"] = starting_weapons
+
+        # The value of missiles ends up being starting + 1, so subtract 1
+        starting_missiles = starting_items.get(_ITEM_TO_ITEM_TYPE["Missile"], 1) - 1
+        starting_energy = 100 + (starting_items.get(_ITEM_TO_ITEM_TYPE["Energy Tank"], 0) * 100)
```
duncathan, line 109: are you sure this should be `100 + (etc)` instead of `99 + (etc)`?

duncathan (changes requested):

randovania/games/prime_hunters/exporter/patch_data_factory.py
dyceron: I need to pass a bitfield as a string to the patcher.

duncathan: I don't understand. Why does this require a filter?

dyceron:
Is there anything else you need me to fix?

randovania/games/prime_hunters/exporter/patch_data_factory.py
```diff
@@ -52,16 +61,53 @@ def force_field_index_for_requirement(game: GameDescription, requirement: Layout
...
+    def _calculate_starting_inventory(self, resources: ResourceCollection) -> dict[str, str | int]:
+        result: dict[str, str | int] = {}
+        starting_items = {}
+
+        for resource, quantity in resources.as_resource_gain():
+            try:
+                starting_items[item_type_for_item_resource(resource)] = quantity
+            except KeyError:
+                print(f"Skipping {resource} for starting inventory: no item id")
+                continue
```
duncathan, line 89: I'm not sure what the advantage of doing this is when you only ever index it by the name of the item. couldn't it just be like:
```py
starting_items = defaultdict(int)
starting_items.update({resource.long_name: quantity for resource, quantity in resources.as_resource_gain()})
```

randovania/games/prime_hunters/exporter/patch_data_factory.py
```diff
@@ -52,16 +61,53 @@ def force_field_index_for_requirement(game: GameDescription, requirement: Layout
...
+        weapons = [
+            starting_items.get(_ITEM_TO_ITEM_TYPE["Shock Coil"], 0),
+            starting_items.get(_ITEM_TO_ITEM_TYPE["Magmaul"], 0),
+            starting_items.get(_ITEM_TO_ITEM_TYPE["Judicator"], 0),
+            starting_items.get(_ITEM_TO_ITEM_TYPE["Imperialist"], 0),
+            starting_items.get(_ITEM_TO_ITEM_TYPE["Battlehammer"], 0),
+            1,  # Missiles
+            starting_items.get(_ITEM_TO_ITEM_TYPE["Volt Driver"], 0),
+            1,  # Power Beam
+        ]
```
duncathan, line 100: then if you do as I suggested above, this would instead be:
```py
weapons = [
    starting_items["Shock Coil"],
    starting_items["Magmaul"],
    starting_items["Judicator"],
    starting_items["Imperialist"],
    starting_items["Battlehammer"],
    1, # Missiles
    starting_items["Volt Driver"],
    1, # Power Beam
]
```
much cleaner

duncathan: although actually it would be safer to do:
```py
def starts_with_weapon(weapon_name: str) -> bool:
    return starting_items[weapon_name] > 0

weapons = [
    starts_with_weapon("Shock Coil"),
    starts_with_weapon("Magmaul"),
    starts_with_weapon("Judicator"),
    starts_with_weapon("Imperialist"),
    starts_with_weapon("Battlehammer"),
    True, # Missiles
    starts_with_weapon("Volt Driver"),
    True, # Power Beam
]
```
that way if you somehow have multiple copies of a weapon in the starting inventory, things won't break

randovania/games/prime_hunters/exporter/patch_data_factory.py
```diff
@@ -52,16 +61,53 @@ def force_field_index_for_requirement(game: GameDescription, requirement: Layout
...
+        starting_weapons = "".join(map(str, weapons))
+        result["weapons"] = starting_weapons
```
duncathan, line 102:
```py
fmt = "{:d}" * 8 # 8-bit bitfield
result["weapons"] = fmt.format(*weapons)
```

randovania/games/prime_hunters/exporter/patch_data_factory.py
```diff
@@ -52,16 +61,53 @@ def force_field_index_for_requirement(game: GameDescription, requirement: Layout
...
+
+        # The value of missiles ends up being starting + 1, so subtract 1
+        starting_missiles = starting_items.get(_ITEM_TO_ITEM_TYPE["Missile"], 1) - 1
```
duncathan, line 105: this is pretty icky. shouldn't this sorta case be handled in the patcher?

randovania/games/prime_hunters/exporter/patch_data_factory.py
duncathan: why? again, shouldn't the patcher handle things correctly instead of needing to hack around it in the pdf?

randovania/games/prime_hunters/logic_database/header.json
```diff
@@ -3,6 +3,13 @@
     "game": "prime_hunters",
     "resource_database": {
         "items": {
+            "Nothing": {
+                "long_name": "Nothing",
+                "max_capacity": 1,
+                "extra": {
+                    "weapon_id": 21
```
duncathan, line 10: shouldn't this be `item_type`, not `weapon_id`?

duncathan (changes requested):

randovania/games/prime_hunters/exporter/patch_data_factory.py
dyceron: I think i wrote it this way because when I was getting the value of missiles from the preset, it was always +1 the value that was set. I just want the amount of missiles given as configured by the launcher

randovania/games/prime_hunters/exporter/patch_data_factory.py
```diff
@@ -57,11 +61,34 @@ class HuntersPatchDataFactory(PatchDataFactory[HuntersConfiguration, HuntersCosm
...
+            starts_with_weapon("Volt Driver"),
+            True,  # Power Beam
+        ]
+        fmt = "{:d}" * 8  # 8-bit bitfield
+
+        result["weapons"] = fmt.format(*weapons)
+        result["missiles"] = starting_items["Missiles"]
+        result["ammo"] = 40
+
+        if starting_items["Energy Tank"] > 0:
```
duncathan, line 89: why the conditional?

randovania/games/prime_hunters/exporter/patch_data_factory.py
```diff
@@ -57,11 +61,34 @@ class HuntersPatchDataFactory(PatchDataFactory[HuntersConfiguration, HuntersCosm
     def game_enum(self) -> RandovaniaGame:
         return RandovaniaGame.METROID_PRIME_HUNTERS
 
-    def _calculate_starting_inventory(self, resources: ResourceCollection) -> dict[str, str]:
-        result = {
-            # TODO: Take starting weapons into account
-            "weapons_string": "00000101",
-        }
+    def _calculate_starting_inventory(self, resources: ResourceCollection) -> dict[str, str | int]:
+        result: dict[str, str | int] = {}
```
duncathan, line 65: when is a string value ever used?

randovania/games/prime_hunters/logic_database/header.json
```diff
@@ -25,9 +32,9 @@
                     "weapon_id": 1
                 }
             },
-            "Missile": {
-                "long_name": "Missile",
-                "max_capacity": 95,
+            "MissileLauncher": {
+                "long_name": "Missile Launcher",
+                "max_capacity": 99,
```
duncathan, line 37: this should be 1

duncathan (changes requested):
you should have a Nothing in one of your test files to ensure coverage

randovania/games/prime_hunters/exporter/patch_data_factory.py
dyceron: I figured if I didn't have to include it, it'd be fine but I'll revert it.

randovania/games/prime_hunters/exporter/patch_data_factory.py
dyceron: Wouldn't weapons qualify?

duncathan: ?

duncathan (approved):

PR #8263, opened 2025-04-22
Fusion: Fixes to Powamp Shaft Logic

M4RKERSM4SK:
The logic for this connection was impossible. It allowed to reach upper ladder gravless and no jump upgrade which ran by multiple people and the community, seems to be undoable by only Frozen Enemies advanced. I did enounter this in testing a seed and it was noncontinuable. 

- Fixed logic with what is possible
- Adjusted the difficulty level to reflect the tricks better
- Added in the videos

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 4 AQA.txt
+++ b/randovania/games/fusion/logic_database/Sector 4 AQA.txt
@@ -1708,19 +1708,18 @@ Extra - room_id: [35]
       Gravity Suit and Screw Attack
   > Upper Ladder
       Any of the following:
+          # No Gravity Suit: https://youtu.be/F6iQaCkrbV0
+          Stand On Frozen Enemies (Intermediate) and Can Freeze Enemies With Any Weapon and Have Any Jump Upgrade
           All of the following:
-              # Gravityless method
-              Stand On Frozen Enemies (Beginner) and Can Freeze Enemies With Any Weapon
-              Hi-Jump or Stand On Frozen Enemies (Advanced)
-          All of the following:
-              # Gravity method
               Gravity Suit
               Any of the following:
                   Space Jump
-                  # Shinespark
+                  # Shinespark: https://youtu.be/zDXCR7HINss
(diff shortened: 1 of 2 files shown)


randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1708,19 +1708,18 @@ Extra - room_id: [35]
...
               Any of the following:
                   Space Jump
-                  # Shinespark
+                  # Shinespark: https://youtu.be/zDXCR7HINss
                   Level 0 Keycard and Speed Booster and Shinespark Tricks (Beginner) and Disabled Door Lock Rando and Disabled Entrance Rando
-                  # Single walljump on the side and kill the powamps
+                  # Single Wall Jump: https://youtu.be/6iSEl5LF6aU
                   Screw Attack and Wall Jump (Intermediate) and Can Single Walljump
+                  # Stand On Frozen Enemies: https://youtu.be/eoniGf8_j70
+                  Stand On Frozen Enemies (Intermediate) and Can Freeze Enemies With Any Weapon
```
JeffGainsNGames, line 1722: it's weird to me that this is considered a frozen enemies trick since the room layout *I assume* was designed intentionally to use them as platforms
our base of `intended vanilla == trickless` is maintained by grav+SJ, so any less equipment *should* then be a trick so this PR is accurate to this rule and this comment is not blocking, just seems odd

JeffGainsNGames (approved):

randovania/games/fusion/logic_database/Sector 4 AQA.txt
M4RKERSM4SK: I thought about this too, and decided on Frozen Enemies trick. Reason being you have to freeze ones on the top in a certain order with Samus also a certain distance away from certain ones to make it work NHJ or NSJ otherwise you cant reach the top one. That is also why it is at intermediate, didnt fit the standard for beginner.

PR #8265, opened 2025-04-24
Fix CSHintDistributor

ThanatosGit:


duncathan (approved):
thanks Than

ThanatosGit:
I quickly added a super simple test to check if the nodes exist, so that future typos or forgotten changes are catched.
I'll leave the full-blown hint distributor tests to someone who knows CS :p

PR #8269, opened 2025-04-25
Hunters - Add Force Field Shuffle to UI

dyceron:
image
image
image

### DIFF
--- a/randovania/games/prime_hunters/layout/prime_hunters_configuration.py
+++ b/randovania/games/prime_hunters/layout/prime_hunters_configuration.py
@@ -19,7 +19,6 @@ class HuntersOctolithConfig(BitPackDataclass, JsonDataclass):
 class HuntersConfiguration(BaseConfiguration):
     octoliths: HuntersOctolithConfig
     force_field_configuration: ForceFieldConfiguration
-    shuffled_force_fields: bool
 
     @classmethod
     def game_enum(cls) -> RandovaniaGame:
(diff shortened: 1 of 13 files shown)


randovania/games/prime_hunters/layout/prime_hunters_configuration.py
```diff
@@ -19,6 +19,7 @@ class HuntersOctolithConfig(BitPackDataclass, JsonDataclass):
 class HuntersConfiguration(BaseConfiguration):
     octoliths: HuntersOctolithConfig
     force_field_configuration: ForceFieldConfiguration
+    shuffled_force_fields: bool
```
duncathan, line 22: how come you have both `shuffled_force_fields` and `force_field_configuration`?

dyceron: I couldn't figure out how to make `on_preset_changed` work, so this was what I thought of.

duncathan: what wasn't working?

dyceron: I think it was something to do with the preset not showing the changes when reopening? I can't remember. If there's another way to compare if its been changed vs how I'm doing it now, then it'd probably be cleaner.

duncathan (approved):

PR #8270, opened 2025-04-26
Randomprime v1.29.0

toasterparty:
Cherry-picks some changes from when we upgraded to the bugged v1.28.X version


CHANGELOG.md
```diff
@@ -72,6 +72,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### Metroid Prime
 
+- Added: Blast Shield Lock-On preset option
+- Added: Patch to fix Kerry Ann Odem's name in the credits on game version(s) where it is incorrectly spelled
+- Fixed: Main Plaza vault ledge door is now affected by door lock rando
+- Fixed: Elite Research Phazon Elite getting stuck on frame perfect cutscene skip
+- Fixed: Minor music transition fixes
+- Fixed: Chaos option export error when submerging rooms
+- Changed: [PAL] Remove extra door lock in Central Dynamo
+- Changed: [PAL] Load Hunter Metroid in Frost Cave
```
ThanatosGit, line 82: Now you merged main but as I expected it is under the headline of the already released version 9.1.0 :p

ThanatosGit (changes requested):

Miepee (changes requested):
i would like one another person approving on prime1 specifics actually.
because i'm not a fan ux wise of:
- the setting existing at all
- the checkbox being changed by a completely different setting

Miepee:
CC @JeffGainsNGames @Uncle-Reggie @duncathan

randovania/games/prime1/gui/ui_files/preset_prime_qol.ui
```diff
@@ -62,11 +62,8 @@
         </property>
         <item>
          <spacer name="top_spacer">
-          <property name="orientation">
-           <enum>Qt::Orientation::Vertical</enum>
-          </property>
           <property name="sizeType">
-           <enum>QSizePolicy::Policy::Fixed</enum>
+           <enum>QSizePolicy::Fixed</enum>
```
henriquegemignani, line 66: Seems like your Qt Creator is outdated.

Uncle-Reggie:
> i would like one another person approving on prime1 specifics actually. because i'm not a fan ux wise of:
> 
>     * the Blast Shield Lock-On preset option existing at all
> 
>     * the checkbox being changed by a completely different setting

I don't see the issue in the setting being force-enabled as long as it is mentioned that it does so, preferably in a tooltip. The settings being disabled by that option I do find to be somewhat unintuitive.

randovania/games/prime1/gui/preset_settings/prime_dock_rando_tab.py
```diff
@@ -0,0 +1,85 @@
...
+class PresetPrimeDockRando(PresetDockRando):
+    def __init__(self, editor: PresetEditor, game_description: GameDescription, window_manager: WindowManager):
+        super().__init__(editor, game_description, window_manager)
+
+        self.changes_box = QtWidgets.QGroupBox()
+        self.changes_box.setTitle("Door Changes")
+        self.changes_layout = QtWidgets.QVBoxLayout(self.changes_box)
+
+        extra_widgets: list[tuple[type[QtWidgets.QCheckBox | QtWidgets.QLabel], str, str]] = [
+            (QtWidgets.QCheckBox, "blue_save_doors_check", "Unlock Save Station Doors"),
```
henriquegemignani, line 31: Can you add a `,` after the last string in this line, so that all three are formatted the same?

henriquegemignani (approved):
Code-wide it looks fine. not 100% happy with the `getattr` but we already do that a bunch and I don't have a much better idea anyway

randovania/layout/preset_migration.py
```diff
@@ -1163,6 +1163,16 @@ def fix(identifier: dict) -> None:
...
 
 
+def _migrate_v103(preset: dict, game: RandovaniaGame) -> None:
+    if game != RandovaniaGame.METROID_PRIME:
+        return
+
+    dlr = preset["configuration"]["dock_rando"]["mode"] != "vanilla"
+    legacy = preset["configuration"]["legacy_mode"]
+    preset["configuration"]["blast_shield_lockon"] = dlr and not legacy
+    preset["configuration"]["blue_save_doors"] = dlr and not legacy
```
henriquegemignani, line 1173: With this, a preset that has DLR disabled but blue save doors on will be migrated to disabled. Is this intended?

randovania/games/prime1/gui/ui_files/preset_prime_qol.ui
toasterparty: 16.0.1 installed from GitHub. Not sure how to fix

duncathan:
> i would like one another person approving on prime1 specifics actually.
> because i'm not a fan ux wise of:
> - the Blast Shield Lock-On preset option existing at all
> - the checkbox being changed by a completely different setting

in my experience the latter practice leads to issues and user confusion. Cave Story does this in one or two spots and it's always been problematic. I strongly encourage finding another way to achieve whatever it is you're trying to do here

Uncle-Reggie:
> I don't see the issue in the setting being force-enabled as long as it is mentioned that it does so, preferably in a tooltip. The settings being disabled by that option I do find to be somewhat unintuitive.
> 
> I also think this is a divisive enough option to have a dedicated checkbox, or, force to always enabled behind the scenes and disable it in legacy mode.
Actually, having seen the windows being interacted with, I do not take issue with any of the decisions being made here and support them as they are. The setting needs to be an option, and the defaults associated with DLR are in good taste with how the two settings primarily go hand-in-hand.

randovania/games/prime1/gui/ui_files/preset_prime_qol.ui
henriquegemignani: if it's updated then leave it like that

PR #8273, opened 2025-04-27
Changed Options in Multi-World Client

DualVission:
Changed:
- Former "Everyone can claim worlds" to "Enable All Players to Claim Worlds"
- Former "Enable Co-Op" to "Enable Co-Op Worlds"
- File `multiplayer_session.ui`
- File `multiplater_session_window.py`

Added:
- Tool Tip for "Enable All Players to Claim Worlds"
- Tool Tip for "Enable Co-Op Worlds"

### DIFF
--- a/randovania/gui/multiplayer_session_window.py
+++ b/randovania/gui/multiplayer_session_window.py
@@ -370,10 +370,15 @@ class MultiplayerSessionWindow(QtWidgets.QMainWindow, Ui_MultiplayerSessionWindo
         self.update_multiworld_client_status()
         self.everyone_can_claim_check.setChecked(session.allow_everyone_claim_world)
         self.everyone_can_claim_check.setEnabled(self.users_widget.is_admin())
+        self.everyone_can_claim_check.setToolTip("By default, only admins can assign unclaimed worlds.")
         self.allow_coop_check.setChecked(session.allow_coop)
         self.allow_coop_check.setEnabled(self.users_widget.is_admin() and not_genned_yet)
         self.allow_coop_check.setText(
-            "Enable Co-Op" + ("" if not_genned_yet else " (can only be changed before generation)")
+            "Enable Co-Op Worlds" + ("" if not_genned_yet else " (can only be changed before generation)")
+        )
+        self.allow_coop_check.setToolTip(
+            ("" if not_genned_yet else "(Can only be changed before generation.)\n")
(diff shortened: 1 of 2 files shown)


DualVission:
I will add that I did not get the chance to test this. My system was refusing to let me run the server and client at the same time. Normally, I would avoid this, but since it is text changes, I felt more comfortable doing that this time.

duncathan:
I don't see what the name changes are accomplishing? The tooltips are probably fine but the new option names are just more verbose without actually providing more useful information

randovania/gui/ui_files/multiplayer_session.ui
```diff
@@ -146,14 +146,20 @@
            <item row="2" column="0">
             <widget class="QCheckBox" name="everyone_can_claim_check">
              <property name="text">
-              <string>Everyone can claim worlds</string>
+              <string>Everyone Can Claim Unclaimed Worlds</string>
```
ThanatosGit, line 149: Is there are reason why this is capitalized now? I seriously have no idea what exactly are our own rules for it. "World's" in this context is a randovania term and should be fine but dunno about the rest. (It is often all capitalized, no?)

randovania/gui/multiplayer_session_window.py
```diff
@@ -373,7 +373,11 @@ async def on_meta_update(self, session: MultiplayerSessionEntry):
         self.allow_coop_check.setChecked(session.allow_coop)
         self.allow_coop_check.setEnabled(self.users_widget.is_admin() and not_genned_yet)
         self.allow_coop_check.setText(
-            "Enable Co-Op" + ("" if not_genned_yet else " (can only be changed before generation)")
+            "Co-Op Worlds" + ("" if not_genned_yet else " (can only be changed before generation)")
```
ThanatosGit, line 376: You added the "can only be changed before generation" to the tooltip. I think that's very redundant now. As we now have a tooltip, I think it is sufficient to only have it there and it is much better than adding explanations dynamically to the label.

ThanatosGit (changes requested):
Please add a changelog entry mentioning that the wording and a tooltip was added clarifying what the co-op toggle does.

Miepee:
I'm agreeing with duncathan on this, that I don't see how these changes help communicating to the user on what the settings do.

PR #8285, opened 2025-05-04
Duplicate co-op setting

ThanatosGit:
Also add "Everyone can claim worlds"?


henriquegemignani (approved):
I'd add this for 9.1.1 since we're already making that release

ThanatosGit:
Alright, moved the changelog and figured that it really makes more sense to also duplicate the "Everyone can claim worlds".
If you duplicate a session, I expect that all settings are also duplicated.

PR #8293, opened 2025-05-09
Weighted Pre-Placement

JeffGainsNGames:
implementation of weighting up to individual games to decide

some sample data
image

### DIFF
--- a/randovania/resolver/bootstrap.py
+++ b/randovania/resolver/bootstrap.py
@@ -9,6 +9,7 @@ from randovania.game_description.db.resource_node import ResourceNode
 from randovania.game_description.resources.resource_collection import ResourceCollection
 from randovania.layout.base.trick_level import LayoutTrickLevel
 from randovania.layout.exceptions import InvalidConfiguration
+from randovania.lib import random_lib
 from randovania.resolver.state import State
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


randovania/resolver/bootstrap.py
```diff
@@ -250,29 +251,57 @@ def all_preplaced_pickup_locations(
...
     ) -> None:
         pre_placed_indices = list(pool_results.assignment.keys())
-        reduced_locations = [loc for loc in locations if loc.pickup_index not in pre_placed_indices]
+        reduced_locations = [loc for loc in locations.keys() if loc.pickup_index not in pre_placed_indices]
 
-        rng.shuffle(reduced_locations)
+        # working_locations is a copy of locations with only the keys in reduced locations
+        working_locations = {}
+        for loc in reduced_locations:
+            working_locations[loc] = locations[loc]
```
ThanatosGit, line 268: `reduced_locations` is a list with only the keys and `working_locations` contains the same keys but with the weighting again, then this block is just:
```python
reduced_locations = {loc: v for loc, v in locations.items() if loc.pickup_index not in pre_placed_indices}
```

randovania/resolver/bootstrap.py
```diff
@@ -250,29 +251,57 @@ def all_preplaced_pickup_locations(
...
+
+        # weighted locations is a list filled by selecting weighted elements from working locations
+        weighted_locations = []
+        for _ in reduced_locations:
+            loc = random_lib.select_element_with_weight_and_uniform_fallback(rng, working_locations)
+            weighted_locations.append(loc)
+            working_locations.pop(loc)
```
ThanatosGit, line 275: ...and this block would be 
```python
        while bool(reduced_locations):
            loc = random_lib.select_element_with_weight_and_uniform_fallback(rng, reduced_locations)
            weighted_locations.append(loc)
            reduced_locations.pop(loc)
```
because you iterate over it until it's empty

ThanatosGit (approved):
lgtm.

The fact that the tests now expect different `PickupIndex` for the configs, means it breaks the permalinks after merge, right? Because the same seed for the rng produces a different outcome, so the same permalink will produce a different result.
Iirc we break permalinks for new minor versions anyway, then it would only be a concern if this would be merged in a 9.1.2

JeffGainsNGames:
That would be my understanding. Any other reviewers necessary before I merge then?

randovania/resolver/bootstrap.py
```diff
@@ -250,29 +251,52 @@ def all_preplaced_pickup_locations(
...
+
+    def pre_place_pickups(
+        self,
+        rng: Random,
+        locations: list[PickupNode],
+        pool_results: PoolResults,
+        item_category: str,
+        game: RandovaniaGame,
+    ) -> None:
+        # calls pre_place_pickups_weighted with all weightings set to 1.0
```
henriquegemignani, line 295: Please make this a proper docstring

randovania/resolver/bootstrap.py
```diff
@@ -250,29 +251,52 @@ def all_preplaced_pickup_locations(
...
         game: RandovaniaGame,
     ) -> None:
         pre_placed_indices = list(pool_results.assignment.keys())
-        reduced_locations = [loc for loc in locations if loc.pickup_index not in pre_placed_indices]
+        reduced_locations = {loc: v for loc, v in locations.items() if loc.pickup_index not in pre_placed_indices}
 
-        rng.shuffle(reduced_locations)
+        # weighted_locations is a list filled by selecting weighted elements from reduced_locations
+        weighted_locations = []
+        while bool(reduced_locations):
```
henriquegemignani, line 267: the `bool` is not necessary.

henriquegemignani (changes requested):

henriquegemignani:
The permalink thing is perfectly fine. This is not the kind of change to backport for a hotfix.

henriquegemignani (approved):

PR #8295, opened 2025-05-11
Dread/logical energy parts

Tyranisaur:


randovania/games/dread/generator/bootstrap.py
```diff
@@ -40,8 +40,7 @@ def create_damage_state(self, game: GameDescription, configuration: DreadConfigu
         return EnergyTankDamageState(
             configuration.energy_per_tank - 1,
             configuration.energy_per_tank,
-            game.resource_database,
-            game.region_list,
+            game.resource_database.energy_tank,
         )
```
ThanatosGit, line 44: Doesn't this need to be `DreadDamageState` now? To me it looks like `DreadDamageState` is complety unused, so it is actually missing the key part of this PR?

Tyranisaur: Somehow forgot to add that back in after managing to commit after the precommit hook conflict fiasco 😅

duncathan:
I intend to take a look at this soonish

randovania/games/dread/layout/dread_damage_state.py
```diff
@@ -0,0 +1,112 @@
...
+    ) -> Self:
+        if self._energy_tank_difference(new_resources, old_resources) > 0:
+            return self._at_maximum_energy(new_resources)
+        elif (new_parts := self._energy_part_difference(new_resources, old_resources)) > 0:
+            if self._use_immediate_energy_parts:
+                result = self._duplicate()
+                result._energy = (
+                    self._energy
+                    + (self._energy_per_tank * new_parts // 4)
+                    + (self._energy_per_tank * (new_parts % 4)) % 4
```
ThanatosGit, line 78: Dunno what I'm looking at here xD, but if this is calculating a new energy value, this needs to be `+ (self._energy_per_tank * (new_parts % 4)) / 4` instead of the modulo operator at the end.
E.g. `new_parts = 2` => `(100 * (2 % 4)) % 4` => `(100 * 2) % 4` => `200 % 4` => `0` but it should be `200 / 4` => `50`

Tyranisaur: I suspected that there would be a more succinct way to write this, but I I ended up with this when I started to consider numbers like 4 and 5.

ThanatosGit: Just `self._energy_per_tank / 4 * new_parts` ? Would always be just `25 * new_parts`

Tyranisaur: The other thing I wanted to take into some kind of consideration was what happens when you don't have a multiple of 4 energy per tank. But I'm not sure what the game would do in those cases.

Tyranisaur: I ended up simplifying this by writing it in terms of `_maximum_energy`.

randovania/generator/filler/pickup_list.py
```diff
@@ -37,20 +37,43 @@ def interesting_resources_for_reach(reach: GeneratorReach) -> frozenset[Resource
         )
     )
     return game_description.calculate_interesting_resources(
-        satisfiable_requirements, reach.state.node_context(), reach.state.health_for_damage_requirements
+        satisfiable_requirements, reach.state.node_context(), reach.state.damage_state
     )
 
 
-def _unsatisfied_item_requirements_in_list(
+def _unsatisfied_item_requirements_for_danage(
```
duncathan, line 44: typo

randovania/resolver/damage_state.py
```diff
@@ -62,9 +60,9 @@ def apply_new_starting_resource_difference(
         """Creates a new state after new resources were added as if they were starting.
         Common difference: collecting a health upgrade fully heals you. But it won't do it here."""
 
-    def resource_requirements_for_satisfying_damage(self, damage: int) -> list[ResourceRequirement]:
+    def resource_requirements_for_satisfying_damage(self, damage: int) -> list[list[ResourceRequirement]]:
         """
         Determines what are the requirements for satisfying a damage requirement with the given value.
         :param damage:
-        :return: a list of resource requirements
+        :return: a list containing lists of resource requirements
```
duncathan, line 67: would be great to explain what these lists actually are

randovania/resolver/damage_state.py
```diff
@@ -62,9 +60,9 @@ def apply_new_starting_resource_difference(
         """Creates a new state after new resources were added as if they were starting.
         Common difference: collecting a health upgrade fully heals you. But it won't do it here."""
 
-    def resource_requirements_for_satisfying_damage(self, damage: int) -> list[ResourceRequirement]:
+    def resource_requirements_for_satisfying_damage(self, damage: int) -> list[list[ResourceRequirement]]:
```
duncathan, line 63: since you're in here, and mypy is complaining: please mark these functions with the `@abstractmethod` decorator from `abc`. if it still complains after doing that, add a `raise NotImplementedError` to each pls

randovania/games/dread/layout/dread_damage_state.py
```diff
@@ -0,0 +1,111 @@
...
+        num_tanks = resources[self._energy_tank]
+        num_parts = resources[self._energy_part_item]
+        num_whole_tanks = num_tanks + num_parts // 4
+        num_partial_parts = num_parts % 4
+        energy_from_partial = (self._energy_per_tank * num_partial_parts) // 4
+
+        return self._starting_energy + (self._energy_per_tank * num_whole_tanks) + energy_from_partial
+
+    def _duplicate(self) -> Self:
+        result = DreadDamageState(
```
duncathan, line 49: this should be `self.__class__(` rather than `DreadDamageState(` to achieve type safety (or you can use a type ignore directive)

randovania/games/dread/layout/dread_damage_state.py
```diff
@@ -0,0 +1,111 @@
...
+        new_resources: ResourceCollection,
+        old_resources: ResourceCollection,
+    ) -> int:
+        return new_resources[self._energy_part_item] - old_resources[self._energy_part_item]
+
+    @override
+    def apply_collected_resource_difference(
+        self, new_resources: ResourceCollection, old_resources: ResourceCollection
+    ) -> Self:
+        if self._energy_tank_difference(new_resources, old_resources) > 0:
```
duncathan, line 70: some comments briefly explaining each branch would be nice

randovania/games/dread/layout/dread_damage_state.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @override
+    def resource_requirements_for_satisfying_damage(self, damage: int) -> list[list[ResourceRequirement]]:
+        # A requirement for many "Energy Tanks" is added,
+        # which is then decreased by how many tanks is in the state by pickups_to_solve_list
+        # FIXME: get the required items for reductions (aka suits)
+        tank_count = 1 + (damage - self._starting_energy) // self._energy_per_tank
+        part_count = 1 + (damage - self._starting_energy) // (self._energy_per_tank // 4)
+        if not self._use_immediate_energy_parts:
+            part_count += (4 - (part_count % 4)) % 4
```
duncathan, line 94: the math in this function is kinda confusing in general, but this line in particular sticks out. is that final outer modulo even doing anything?

duncathan (changes requested):
so what options does the generator have to choose from now to satisfy energy requirements? could you post a generator debug log showcasing that?

Tyranisaur:
Hmm, so the redundant actions to place some combinations of energy tanks and energy parts are a result of the following.

The item requirements for damage are built up without looking at the current inventory or the total number of items in the pool, so there are branches for each plausible combination of tanks and parts. These are in the form of item requirements. Then this gets transformed into lists of pickups that solve the item requirements. The pickup lists do take into account your current inventory. However, the different combinations of energy tanks and parts still exists as separate branches. And when you are in the late game, you plausibly have a combination of tanks and parts. For example, you could be in a state where to satisfy the damage constraint with only energy tanks, you're missing two energy tanks. But to satisfy the damage with one less energy tank and the rest with energy parts, you could already have the energy parts, and only lack one energy tank. That is why you see both one energy tank and two energy tanks as options to achieve the same, even though the option that requires one energy tank should subsume the other.

The side effect of this, until resolved, seems to be that placing energy has higher weight than intended, because of the duplicated entries. I'm not sure how critical it is to fix that.

randovania/games/dread/layout/dread_damage_state.py
duncathan: > Would it be cleaner to remove it and add and part_count % 4 != 0 to the if?

I would prefer that

Tyranisaur:
> > The side effect of this, until resolved, seems to be that placing energy has higher weight than intended, because of the duplicated entries. I'm not sure how critical it is to fix that.
> 
> Based on this snippet, I would say pretty critical:
> 
> ```
> [P: Progressive Bomb] - 1.0
> [P: Energy Tank, P: Energy Tank] - 0.375
> [P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part] - 0.375
> [P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part, P: Energy Part] - 0.375
> [P: Energy Part, P: Energy Part, P: Energy Part] - 0.375
> [P: Energy Tank] - 0.375
> [E: region Artaria/area EMMI Zone Spinner/node Event - EMMI Spinner Rotated] - 0.375
> [E: region Artaria/area Chain Reaction Room/node Event - Chain Reaction Device] - 0.375
> [E: region Ghavoran/area Flipper Room/node Event - Super Missile Rotatable] - 0.375
> [P: region Ghavoran/area Cross Bomb Tutorial/node Pickup (Missile Tank)] - 0.75
> ```
> 
> that weight of 0.375 is the weight of a dangerous action that collects an event. this means that placing energy to do something dangerous is getting a ludicrously high effective weight of 1.875, dramatically higher than the only safe action here. I'm a bit worried about introducing behavior like this, and also realizing that a similar problem may exist in cave story atm

There's a couple of discussion points around this.

I would argue that most of the fights that get unlocked by energy in dread aren't actually dangerous. They just show up as dangerous for the weights for reasons discussed separately. I believe that even if the weight of placing energy is boosted for the wrong reasons, it's probably beneficial for seed variety to do so, as it will follow a progression in health requirements for the bosses.

I have some made changes that reduce the redundancies in requirements that arise from the combinations of ways to satisfy damage requirements. However, I believe we have unresolved design thinking about how to handle cases where some requirement combinations obsolete other combinations. Depending on which step in the machinery the redundancies are detected, we get different results already. Some ways to mitigate duplication here might also affect some other cases of requirements that subsume other requirements. Sometimes when one requirement is a superset of another requirement, then it also unlocks more things. And it's hard to make a judgement about whether or not the superset should be removed. If the weights of the actions end up being same the it probably would be reasonable, but it's not a given that we can determine that the weights are the same at the same step where we're currently considering which requirements to turn into actions.

It would probably make more sense to do a bigger refactor of how damage requirements are translated into energy tank requirements. The current logic spits out a total number of required energy tanks, not taking into account the current inventory or the number of tanks in the pool. Adding energy parts into this mix creates a combinatoric explosion of possibilities, as the damage can be spread across any combination of these two. The logic currently doesn't have the best tools to avoid deduplication when these combinations then later start to take the current inventory into account.

I thought about the idea of having it always place energy tanks first and then start placing energy parts once out of energy tanks to place. This would probably be a reasonable simplification to avoid some duplications, but hinges on the refactor mentioned above. It's also not a given that exhausting the tanks before placing parts will be good for seed quality.

In terms of adding support for considering damage reductions when placing items, it would also seem necessary to refactor the damage logic at least a little bit, but I think that's less of a concern than how damage reductions would also exhibit some degree of branching paths in requirements. 

I suspect that there's some level of awkwardness from how actions are considered and weighed on the premise that the action as a whole can consist of multiple pickups, but then always place one pickup per action anyway, and recalculate the actions.

There's a world where we take out the changes that make the generator place energy parts, but otherwise leave in the logical aspects of the energy parts for the benefits of the resolver. If we aren't in a position where we know how to weigh actions when there are multiple ways to satisfy damage requirements.

duncathan:
I intend to discuss the other parts of your post later, but for now: 
> There's a world where we take out the changes that make the generator place energy parts, but otherwise leave in the logical aspects of the energy parts for the benefits of the resolver. If we aren't in a position where we know how to weigh actions when there are multiple ways to satisfy damage requirements.

I'm not necessarily opposed to this, if need be. I do think it's a more global problem that this PR just happens to be highlighting rather than a specific problem with this PR

Tyranisaur:
I decided to solve the duplications differently. Instead of a general solution that removes pickup combinations that are proper supersets of other ones, I made the dread damage state not spit out redundant combinations of tanks and parts, based on current resources.

duncathan:
we'll merge in a few days, after the upcoming release

PR #8298, opened 2025-05-13
Generator: Increase debug level to print reach

Tyranisaur:
This piece produces a lot of output to the debug, and it seems less interesting than most of everything else at debug level 3.


Tyranisaur:
Should I make the cli help text for generator calls show that there's one more level with this?

henriquegemignani:
> Should I make the cli help text for generator calls show that there's one more level with this?

Yes. I thought the debug level also had an explicit limit 🤔

Miepee (approved):

PR #8303, opened 2025-05-14
AM2R: Add enemy extra data

Miepee:
type is dict[instance_id, enemy_name]


duncathan (approved):
I'm just gonna assume this is correct. I wish the keys could be real numbers but alas, json

PR #8306, opened 2025-05-16
Fusion: New locations

M4RKERSM4SK:
### Patcher Version Update

Updated Mars Patcher to 0.4.1

### DB

Changed the following event nodes to pickup nodes, the platforms can now grant pickups
- Save the Animals (Was a generic node for some reason, and changed to Major)
- Boiler Cooling Unit (Minor)
- Aux Power (Minro)

Added new pickup (minor) nodes, these locations now have a new tank pickup:
- Subzero Containment
- Quarantine Bay
- NE Stabilizer

Added/updated supporting logic for new locations (mainly just sub zero, l3 lock, making connections)
Updated corresponding map images (with the suggested image editor command from Miepee)
Added the extra data for the correct coords, sources to pass to the patcher
Removed the event resources since they are now pickup

### Default Preset

Changed OSH to default based on direction from Jeff
Updated the two presets (descriptions, starting E, etc.)
Updated the test describer to match the new default preset of OSH

### Testing
Generated and exported about 50 seeds on both presets
Played a seed to verify all the new pickups
Spot checked a few rdvgame/patcher jsons to make sure everything was passing correctly

Only thing I am unsure of is that in Central Reactor Core, there was trick dependent on "After Reactor Controls Activated". I went and tested the trick and it doesnt require the Aux Power platform to be down or activated, and I am pretty sure there was no layer changes after this activation anyways. Got the shinespark, so I dont think this was actually dependent on this event unless I am missing something big


randovania/games/fusion/logic_database/Main Deck.txt
```diff
@@ -1961,14 +1976,18 @@ Hint Features - 1-way Shutter
...
       Enabled Generator Workaround
-  > Save the Animals
+  > Pickup (Save the Animals)
       Trivial
 
-> Save the Animals; Heals? False
+> Pickup (Save the Animals); Heals? False
   * Layers: default
+  * Pickup 120; Category? Minor
   * FIXME: Causes issues with door lock rando.
```
Miepee, line 1985: you can remove the fixmes/todos that mention this in regards to being an event node

M4RKERSM4SK:
Events:

Animals

Aux

Boiler

M4RKERSM4SK:
Tanks:

honorad

Ridley

NEStabilizer

randovania/games/fusion/logic_database/Main Deck.txt
```diff
@@ -1449,9 +1461,9 @@ Extra - room_id: [49, 59]
                   Any of the following:
                       Wave Beam
                       # If you only have charge, you need to have access to the other side
-                      Charge Beam and After Reactor Controls Activated
+                      Charge Beam
               # Without Wave nor Charge, there's only one attempt and its hard.
-              After Reactor Controls Activated and Shinespark Tricks (Ludicrous)
```
Miepee, line 1454: the reason for relying on the event is that this is how we determine whether you have access to the other side.
if you only have charge, then in-game you cannot get to the other side, unless you explicitly activated reactor controls

M4RKERSM4SK: I am not following here, how does explicitly activating reactor controls mean you have access to the other side with only charge? 

Does activating the controls do something I am not aware? Otherwise wouldn't the event just have the same requirements as reaching it there from going the Yakuza route? not sure where charge fits in this

cluekitty: I believe it indicates that you entered the room from the auxiliary power controls side, or at least have the ability to be on the right side of the gate. When the gate is closed there is not enough room for a speed boost and you would need wave, or to enter from the right-side door, which only can be done after you enter through Yakuza's room if you don't have wave beam, and thus runs through the auxiliary power controls which is "trivial" from the room entrance, so RDV will assume you picked it up.

JeffGainsNGames: we could use a node requirement instead of an event iirc

M4RKERSM4SK: > thus runs through the auxiliary power controls which is "trivial" from the room entrance, so RDV will assume you picked it up

Yeah, so the event requirements would be what I am expecting here, but trying to understand this portion 

> if you only have charge, then in-game you cannot get to the other side

Isn't it reachable without charge too?

I don't mind putting back and event for this trick named like "Can Access Aux Controls"

JeffGainsNGames: charge is specifically to buffer the input to hit the gate weakpoint

and again just use a node requirement, it's close enough to an event requirement for this purpose (it'll be the pickupnode that's listed)
image

randovania/games/fusion/logic_database/Main Deck.txt
M4RKERSM4SK: Removed

randovania/games/fusion/logic_database/Main Deck.txt
M4RKERSM4SK: Added back in as pickup req as Jeff suggested

M4RKERSM4SK: Does putting in a pickup node requirement, mean there must be a required item for the trick?

```FAILED test/games/fusion/test_fusion.py::test_all_tricks_should_have_proper_requirements - AttributeError: 'NodeRequirement' object has no attribute 'items'```

Comment (JeffGainsNGames) [reply randovania/games/fusion/logic_database/Main Deck.txt]: I think this is just something wrong with the test itself

Comment (M4RKERSM4SK) [reply randovania/games/fusion/logic_database/Main Deck.txt]: ```requirement = Req region Main Deck/area Auxiliary Power Station/node Pickup (Reactor Control), resource = 'Jump Bombjump' ```
``` db = ResourceDatabase(game_enum=<RandovaniaGame.FUSION: 'fusion'>, item=[ItemResourceInfo(resource_index=0, long_name='Char...le Wall Jumping Physics', short_name='CanSWJ', resource_type=<ResourceType.MISC: 5>, extra=frozendict.frozendict({}))])```

Comment (duncathan) [reply randovania/games/fusion/logic_database/Main Deck.txt]: there should probably be a separate event rather than using a node requirement here. the connection that would use it doesn't require collecting the pickup at all

Comment (JeffGainsNGames) [reply randovania/games/fusion/logic_database/Main Deck.txt]: I'm lost at what the point of a NodeRequirement is then. Happy to go with event if this can be explained

quick search from discord was my reference for pointing m4rkers in that direction:
image

Comment (duncathan) [reply randovania/games/fusion/logic_database/Main Deck.txt]: node requirements are useful when you want a connection to explicitly require collecting a specific resource node. 

what you're doing here is a hacky shortcut for what you actually want. you want an event that is accessible from the right side of the gate, *within the same room as the gate*, that indicates you have access to that side of the gate. this has nothing whatsoever to do with the pickup in aux power

Comment (M4RKERSM4SK) [reply randovania/games/fusion/logic_database/Main Deck.txt]: Created an event on the right of the shutter in Silo checkpoint. Agreed that makes the most sense this way and it's a lot clearer in the trick now too that you have to have the ability to access the right of shutter.

**Comment by JeffGainsNGames** [comment]
feel free to pull this out of draft

**Comment by M4RKERSM4SK** [comment]
Pulled out of draft, thanks!

**Comment by M4RKERSM4SK** [comment]
Forgot to change the Animals to a Major location for RDV after discussion with Jeff. Its now corrected with last commit

**Comment by M4RKERSM4SK** [comment]
Updated to mars 0.4.1 for the minimap updates to match the new locations.

**Review by cluekitty — CHANGES_REQUESTED** [review]
There are changes in patcher 0.4.1 / asm 0.5.1 which make modifications to sector 6 Weapons Testing grounds. I don't see the logic updates for that in this PR. We've removed the Power Bomb blocks in this room entirely and thus traversing the room should now be trivial instead of "can requires pbombs". I have updated images for this room and for a room in sector 2 as well. The sector 2 room would not have logic changes as a result of the modifications.

File: randovania/games/fusion/gui/ui_files/games_tab_fusion_widget.ui [hunk 3]
```diff
@@ -147,6 +147,14 @@
 
 ### Gameplay
 
+- Added Pickups to the following Areas:
+    - Main Deck - Habitation Deck (Save the Animals)
+    - Main Deck - Auxiliary Power
+    - Main Deck - Sub-Zero Containment (near Ridley statue)
+    - Main Deck - Biohazard Containment
```
cluekitty, line 154: "Biohazard Containment" is not a room in the database. Perhaps this means Quarantine Bay?

Miepee (approved):

PR #8308, opened 2025-05-16
check for node requirements in fusion test

JeffGainsNGames:


test/games/fusion/test_fusion.py
```diff
@@ -4,6 +4,7 @@
```
Miepee, line 4: I'm fairly certain this whole test file should just be deleted
The integrity checks should have superseded this.

Miepee (changes requested):

test/games/fusion/test_fusion.py
JeffGainsNGames: close this PR and remove for AM2R as well? iirc it's had similar/same

Miepee: i never saw this reply: but yes

PR #8311, opened 2025-05-18
Prime: pre-placement of artifacts and phazon suit

JeffGainsNGames:
Allows pre-placement of Artifacts and Phazon Suit.

Data was gathered here: https://docs.google.com/spreadsheets/d/17mac3-c2I5fyqSZ1H7YKfTrnRSMGR6Zx8NAA4UhhhHc/edit?usp=sharing

image
(checking pre-place artifacts disables minimum actions spinbox)

Final values can be adjusted later, but in game testing will be important on the future of this setting so want to get it out.

### DIFF
--- a/randovania/games/prime1/logic_database/Chozo Ruins.json
+++ b/randovania/games/prime1/logic_database/Chozo Ruins.json
@@ -1089,7 +1089,8 @@
                         "default"
                     ],
                     "extra": {
-                        "position_required": false,
+                        "position_required": false, 
+                        "regional_weight": 0.60, 
                         "world_position": [
                             148.244965,
                             86.584686,
@@ -1124,6 +1125,7 @@
                         "default"
                     ],
                     "extra": {
+                        "regional_weight": 0.60, 
                         "world_position": [
                             125.546928,
                             25.281984,
@@ -1158,6 +1160,7 @@
                         "default"
                     ],
                     "extra": {
+                        "regional_weight": 0.60, 
                         "world_position": [
                             115.710464,
                             -13.073902,
@@ -1192,6 +1195,7 @@
                         "default"
                     ],
                     "extra": {
+                        "regional_weight": 0.60, 
                         "world_position": [
                             69.797874,
                             -35.624599,
@@ -2555,6 +2559,7 @@
                     ],
                     "extra": {
                         "position_required": false,
+                        "regional_weight": 0.60, 
                         "world_position": [
                             246.626495,
                             -47.550934,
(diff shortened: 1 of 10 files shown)


randovania/games/prime1/logic_database/Chozo Ruins.json
```diff
@@ -1089,7 +1089,8 @@
                         "default"
                     ],
                     "extra": {
-                        "position_required": false,
+                        "position_required": false, 
+                        "regional_weight": 0.60,
```
Tyranisaur, line 1093: Calling this `regional_weight` suggests that it should be defined per region instead of per pickup. I think it makes sense to give it a name that suggests the weighting isn't tied to regions. It probably makes sense to put `pre_placement` or similar in the name.

Also, there appears to be a bunch of trailing spaces.

Tyranisaur (commented):
Should steps be taken to avoid generation failures from "oops all accessible locations are artifacts" or this issue considered rare enough where it's astronomically unlikely to surface after automated retries?

JeffGainsNGames:
> Should steps be taken to avoid generation failures from "oops all accessible locations are artifacts" or this issue considered rare enough where it's astronomically unlikely to surface after automated retries?

From the bulk generations I did (with both settings enabled) failures on 1000 seeds was 15 generation failures, and 13 resolver failures. Totaled to (15+13)/1000 = 2.8% per individual generation, and ~3%/15 = 0.2% per automated retry.

Steps could be taken to reduce this for standard settings, like weighing hive totem/alcove at 0.0 to remove chances of placing over the early locations, or PMT at 0.0 for phazon suit.
But these options would detract significantly from settings like random start, random elevators, or higher trick settings.

randovania/games/prime1/logic_database/Chozo Ruins.json
JeffGainsNGames: yeah initial naming was due to each pickup node in each region having the same value, but given that might not always be the case depending on any future design, and no mention of what weight it's for (pre-placement) I agree could use better name.

Latter trailing spaces should be resolved when I fixed the database txt files.

henriquegemignani:
Hmmm I wonder if we can annotate which pre-assigned pickups are not mandatory assignments (compared to something being vanilla, or dna on bosses) then the generator, as a last resort instead of failing, would unassign one of these.
It could even be extended to unassigning unreachable pickups, to solve is one is set to inacessible locations (due to weird preset or random elevators)

JeffGainsNGames:
I'm assuming that's safely out of scope for this PR?

henriquegemignani:
For this pr, yeah. For this feature, maybe quite important.

JeffGainsNGames:
Realized the setting doesn't get reflected in the preset describer, assume that's a want?

JeffGainsNGames:
image

JeffGainsNGames:
oh yeah #8353

PR #8313, opened 2025-05-20
Add missing check for assigning specific location hints

clementga:
Currently, it seems disabling specific hint locations only just turns them all into jokes, but still assigns them making those locations impossible to hint through random hints.

Adding that conditions should hopefully clear that problem.


duncathan:
this isn't quite the right fix. it *should* be placing joke hints; but it should *not* be preventing those locations from being hinted if there's a joke hint

PR #8319, opened 2025-05-22
Add 4 lines tracker for dread

Naii-the-Baf:
Adds a tracker layout for dread with no progressive items that fits more nicely in the stream layout I usually use.
Mostly this is to have a tracker layout with split beams that isn't as wide as the default one, which fits neither in my screen nor in my stream layout.


JeffGainsNGames:
best to add a screenshot of the layout

Naii-the-Baf:
image

duncathan (approved):

PR #8325, opened 2025-05-27
Fusion: Fix Sub Zero Damage Run Logic

M4RKERSM4SK:
Hotdogged the "and" in the logic for the damage run on subzero containment.


Miepee (approved):
I wish we could catch these with tests :(

PR #8331, opened 2025-05-29
Generator: Fix not considering unlocked EventPickupNodes

Tyranisaur:


Tyranisaur:
> does this introduce a possibility of returning the same node multiple times? if it does, will that cause any problems anywhere?

I think the extent at which you end up double counting is that you have eventpickups count as both events and pickups, which strictly speaking is what they are. Other than that the debugger appears to make it look like the eventpickup node is part of the `base_state`, while neither of its `event_node` or `pickup_node` are.

randovania/generator/filler/filler_library.py
```diff
@@ -73,11 +74,14 @@ def from_reach_with_unsafe(cls, reach: GeneratorReach) -> Self:
...
-                node
-                for node in reach.iterate_nodes
-                if (node.node_index in base_state.nodes) and isinstance(node, res_type)
-            )
+            for node in reach.iterate_nodes:
+                if node.node_index in base_state.nodes:
+                    if isinstance(node, res_type):
+                        yield node
+                    elif (res_type is PickupNode) and isinstance(node, EventPickupNode):
+                        yield node.pickup_node
```
duncathan, line 82: you can just add a `# type: ignore[misc]` comment on each of these lines

duncathan (changes requested):

Miepee (commented):
I'm really unsure whether this PR is actually the proper fix. 
Is the eventpickups not being considered only an issue for dread? did this happen in all games?
I just find it pretty hard to believe that eventpickups were never (or were stopped being) considered for the generator, considering that some games are full of them.

(Also missing changelog entry for next release)

Tyranisaur:
> I'm really unsure whether this PR is actually the proper fix. Is the eventpickups not being considered only an issue for dread? did this happen in all games? I just find it pretty hard to believe that eventpickups were never (or were stopped being) considered for the generator, considering that some games are full of them.
> 
> (Also missing changelog entry for next release)

The bug that's being fixed is that when evaluating an action, if that action unlocks an eventpickup node, either safely or unsafely, then it is not considered for weights. Presumably the generator is already considering eventpickups as actions in their own right, and will collect eventpickups after having chosen an action that unlocks it safely.

I will add a changelog entry after the release, in order to skip merge conflicts in the changelog. But in the meanwhile, what level of detail is appropriate to list in the changelog?

Miepee:
> The bug that's being fixed is that when evaluating an action, if that action unlocks an eventpickup node, either safely or unsafely, then it is not considered for weights.

Hmmmm, I was sure that that worked for am2r tho. But maybe not and i never have looked at it precise enough.

>  But in the meanwhile, what level of detail is appropriate to list in the changelog?

they're fine being vague/non-detailed. Something like "fixed a bug where action weights were sometimes incorrectly calculated" is good enough. looking at existing entries it's varying a bit on how much detail we actually provide on generator changes.

duncathan (approved):

PR #8332, opened 2025-05-29
Hunter - Small DB look over

dyceron:
- Rename some nodes
- Adjust some connections
- Add new trick with video in High Ground
- Update trick documentation levels


dyceron:
Just need another core review

Miepee (approved):
needed to hunt for the pr again.
I have no clue who and when is gonna make the release, probably better to wait with merging until that is done

PR #8346, opened 2025-06-08
Dread: Connect back to release X event node after having the event

Tyranisaur:
When evaluating releasing the X as an action, breaking the blob would be considered unsafe because it involves leaving the event node, and there's no open path back to it.


XenoWars (commented):
This is a non-issue. The only way you can logically be at the Alcove node without having triggered the X event is if you are coming from reverse, which you are forced to break the blob first to even reach. Unnecessarily enabling the blob event to reach the X event from here bypasses the knowledge requirement that must be in place.

Tyranisaur:
> This is a non-issue. The only way you can logically be at the Alcove node without having triggered the X event is if you are coming from reverse, which you are forced to break the blob first to even reach. Unnecessarily enabling the blob event to reach the X event from here bypasses the knowledge requirement that must be in place.

There isn't a logic bug per se, but without this connection back, doing the bomb launcher before going into elun the other way leads to a funny situation where breaking the blob above from alcove being an unsafe action, because leaving the release X event node is already unsafe. The result is that the generator doesn't see that taking launcher before opening the plasma door will give you any pickups.

Admittedly, I only realized last night that the blob is still considered an unsafe action, because the connection out of the event node leads to the plasma door instead of the alcove node. Though, this part is alleviated by the "consider unsafe actions" logic. However, the missing connection back to the release X node is not alleviated by that.

XenoWars:
> There isn't a logic bug per se, but without this connection back, doing the bomb launcher before going into elun the other way leads to a funny situation where breaking the blob above from alcove being an unsafe action, because leaving the release X event node is already unsafe. The result is that the generator doesn't see that taking launcher before opening the plasma door will give you any pickups.

The connection back still needs to require beginner knowledge though which means in trickless seeds you will be forced to go through the launcher regardless and that connection wouldn't be doing anything.

> Admittedly, I only realized last night that the blob is still considered an unsafe action, because the connection out of the event node leads to the plasma door instead of the alcove node. Though, this part is alleviated by the "consider unsafe actions" logic. However, the missing connection back to the release X node is not alleviated by that.

How exactly does connecting to the plasma door make it dangerous? it's trivial to go between the door and alcove.

Tyranisaur:
> > There isn't a logic bug per se, but without this connection back, doing the bomb launcher before going into elun the other way leads to a funny situation where breaking the blob above from alcove being an unsafe action, because leaving the release X event node is already unsafe. The result is that the generator doesn't see that taking launcher before opening the plasma door will give you any pickups.
> 
> The connection back still needs to require beginner knowledge though which means in trickless seeds you will be forced to go through the launcher regardless and that connection wouldn't be doing anything.
> 

The connection back having an option where it's logical when you have the event collected means it's logical to go back to the event node when you already have the event collected. That means you either did the ledge warp, or have previously been at the event node. Either way, going back to the event node when you already have the event resource collected has no impact other than to make certain connections safer as seen by the generator and/or resolcer.

> > Admittedly, I only realized last night that the blob is still considered an unsafe action, because the connection out of the event node leads to the plasma door instead of the alcove node. Though, this part is alleviated by the "consider unsafe actions" logic. However, the missing connection back to the release X node is not alleviated by that.
> 
> How exactly does connecting to the plasma door make it dangerous? it's trivial to go between the door and alcove.

From the context of evaluating the action of taking the bomb launcher (collect the event), the generator acts on the event node, meaning it goes to the event node. From there it's considering what new resources are unlocked. New resources can be either safe or unsafe. The evaluation will automatically collect anything considered safe and expand the evaluation to consider everything unlocked by the safe actions. The problem begins from the fact that for the generator to consider anything as safe, it needs to have a connection both to the new resource and back again. Now in this context it is is standing on the "release X parasites" event node, having collected said event node. Going from the event node to the alcove node is already an unsafe "action" because the connection back to the event node requires beginner knowledge, so anything beyond this point is inherently unsafe unless there's already an open path all the way back around to the bomb launcher. This is what I'm fixing in this change. 

The next point is considering whether the event to break the blob is a safe action. The generator is kind of boneheaded on the default settings and isn't considering this from the context of having collected the event. So it will see that you satisfy the requirements to get to the event node, and it connects to the plasma door. However, the connection back to the alcove node requires having collected the blob event, which it hasn't at that point. Hence, breaking the blob from the alcove is considered unsafe in the context of evaluating the "release the x event". However, because the blob event node connects to the plasma door, that means that if you take the plasma door and either go through the chozo fight, or through the morph ball maze, then blowing up the blob with wave beam or similar is considered a safe action.

However, with the "consider unsafe actions" option enabled, then the generator will collect events when checking whether they're safe, so the latter problem goes away under that setting.

The effect of missing a connection back to the "release the x" node is that the weight of doing this action is less than it should be. It's weighted as an unsafe action that only unlocks something unsafe. Meanwhile, provided that you have enough items, the alternative action, which is to open the plasma lock, has twice as much weight, because that option is considered to unlock safe resources, and so the generator sees much more of what you unlock by taking that path. In effect, the result of this weighting would be contributing to vanilla plasma beam being more unlikely than it really should.

XenoWars:
> > The connection back still needs to require beginner knowledge though which means in trickless seeds you will be forced to go through the launcher regardless and that connection wouldn't be doing anything.
> 
> The connection back having an option where it's logical when you have the event collected means it's logical to go back to the event node when you already have the event collected. That means you either did the ledge warp, or have previously been at the event node. Either way, going back to the event node when you already have the event resource collected has no impact other than to make certain connections safer as seen by the generator and/or resolcer.

This one is on me. I thought it was requring the blob event.

> > How exactly does connecting to the plasma door make it dangerous? it's trivial to go between the door and alcove.
> 
> From the context of evaluating the action of taking the bomb launcher (collect the event), the generator acts on the event node, meaning it goes to the event node. From there it's considering what new resources are unlocked. New resources can be either safe or unsafe. The evaluation will automatically collect anything considered safe and expand the evaluation to consider everything unlocked by the safe actions. The problem begins from the fact that for the generator to consider anything as safe, it needs to have a connection both to the new resource and back again. Now in this context it is is standing on the "release X parasites" event node, having collected said event node. Going from the event node to the alcove node is already an unsafe "action" because the connection back to the event node requires beginner knowledge, so anything beyond this point is inherently unsafe unless there's already an open path all the way back around to the bomb launcher. This is what I'm fixing in this change.

But if this is the case, couldn't we just connect the event node back to the dock instead of it connecting to Alcove? And then not need the extra Alcove connection to the dock.

> The next point is considering whether the event to break the blob is a safe action. The generator is kind of boneheaded on the default settings and isn't considering this from the context of having collected the event. So it will see that you satisfy the requirements to get to the event node, and it connects to the plasma door. However, the connection back to the alcove node requires having collected the blob event, which it hasn't at that point. Hence, breaking the blob from the alcove is considered unsafe in the context of evaluating the "release the x event". However, because the blob event node connects to the plasma door, that means that if you take the plasma door and either go through the chozo fight, or through the morph ball maze, then blowing up the blob with wave beam or similar is considered a safe action.

So if we connect the blob event to Alcove instead of the door, will that mean Door to Blob to Alcove is a safe action or does it become unsafe? It should make the Alcove to Blob safe.

It feels like all we have to do is just change where the two events connect back to, that being the `Dock from Purple Drapes` and `Alcove` respectively.

Tyranisaur:
So there are a lot of places in the database that has this structure where the connection out of an event happens to go to a a node that causes the event to be weighted as an unsafe action when considering a resource that unlocks it. When I asked about it, dunc said hii thinks it's unreasonable for database maintainers to take this faulty event structure into account. The consider unsafe thing is a better solution for that. Though, like I allude
d to already, it won't help for the specific case I'm addressing here. 

But if we did want to make the simple solution in the database, what we would do is to have multiple trivial paths out of the event node, going to any relevant nodes that are otherwise connected by a requirement on having collected the event. That way, the event would be weighted as safe when doing it from any of the relevant directions.

Miepee:
I see both an approval and lots of database discussion - does that mean that the PR is not mergable then?

Tyranisaur:
Almost forgot that this wasn't in yet. I made it ready again.

henriquegemignani (approved):

PR #8358, opened 2025-06-17
Echoes: add instant morph to transport center

JeffGainsNGames:
I've seen people call this a BSJ and I really really don't like that term here


Miepee:
@JeffGainsNGames can you fix changelog?

PR #8360, opened 2025-06-17
Add check for major item count when major/minor split is enabled

Naii-the-Baf:


Naii-the-Baf:
> This line in the ui should also be changed to reflect this behaviour.

Done.

Miepee:
@Naii-the-Baf can you fix changelog?

Naii-the-Baf:
> @Naii-the-Baf can you fix changelog?

Done.

PR #8362, opened 2025-06-19
Fusion mini map coords

M4RKERSM4SK:
Added minimap coords to the Fusion Database

I have went through each room twice, but I still think I could miss some rooms, miss enter a number, etc. Would appreciate a double check. Easy way might be with the visualizer tool that Miepee was using to quickly be able to see any wrong spots.

Rooms did have overlap (according to mage room sizes and coords) on some rooms. For these rooms I took this approach

- If the overlap or extra space of the room went into a blank tile\space, I inlcuded it in the minimap coords
- If the overlap or extra space of the room went into a room that already had those coords, the coords were left out.

closes https://github.com/randovania/randovania/issues/8328


JeffGainsNGames:
image
this is the coordinate method, right?

Miepee:
~~No~~Yes. The coordinate method is: Up left is 0,0. Further to the right increases x, further down increases y.
grafik

EDIT: just realized that that's what you meant oops. Lack of origin point confused me.

randovania/games/fusion/logic_database/Sector 6 NOC.txt
```diff
@@ -1162,6 +1192,7 @@ Extra - unlocked_save_recharge_station: True
 Twin Caverns West
 Extra - map_name: Sector 630
 Extra - room_id: [30]
+Extra - minimap_coordinates: [{'x': 12, 'y': 8}, {'x': 12, 'y': 9}, {'x': 12, 'y': 10}, {'x': 12, 'y': 11}]
```
Miepee: cords are wrong

M4RKERSM4SK: Fixed, forgot the little room on the left, ty

PR #8363, opened 2025-06-20
Fusion: Launcher Adjustments

M4RKERSM4SK:
Issues reported:
- Launchers did not give ammo with multiple copies shuffled
- Starting with ammo tanks was not considered

With the new changes on the patcher/asm side for how launcher is handled, adjusted the RDV side to match and address the two issues above.

- Updated mars to patcher 0.5.0 in the reqs
- Initialized missiles and pbs to `0` since that is now possible/intended depending on settings
- Adjusted the cases to now consider launcher amount and ammos together as intented for starting
- Added MissileData and PowerBombData to match new patcher/asm changes
- Updated test files to reflect new chaanges

Tested (includes other stuff in the patcher/asm update)
- A bunch of starting ammo/launcher configs
- With starting ammo, checked it out in the status screen, worked good
- Increments of launchers
- New Color variations (Got a really cool emerald green SAX I haven't seen before)
- Color variations seem fixed for new locations tanks
- Credits updates  worked good
- Revealed Pillars (reported bug)
- Glass tube indicators

### DIFF
--- a/randovania/games/fusion/exporter/patch_data_factory.py
+++ b/randovania/games/fusion/exporter/patch_data_factory.py
@@ -156,19 +156,9 @@ class FusionPatchDataFactory(PatchDataFactory[FusionConfiguration, FusionCosmeti
             tank_dict[definition.extra["TankIncrementName"]] = state.ammo_count[0]
         tank_dict["EnergyTank"] = self.configuration.energy_per_tank
 
-        missile_launcher = next(
-            state
-            for defi, state in self.configuration.standard_pickup_configuration.pickups_state.items()
-            if defi.name == "Missile Launcher Data"
-        )
-        tank_dict["MissileData"] = missile_launcher.included_ammo[0]
-
-        pb_launcher = next(
-            state
-            for defi, state in self.configuration.standard_pickup_configuration.pickups_state.items()
-            if defi.name == "Power Bomb Data"
-        )
-        tank_dict["PowerBombData"] = pb_launcher.included_ammo[0]
+        for definition, state in self.configuration.standard_pickup_configuration.pickups_state.items():
+            if "LauncherIncrementName" in definition.extra:
+                tank_dict[definition.extra["LauncherIncrementName"]] = state.included_ammo[0]
 
         return tank_dict
 
--- a/randovania/games/fusion/exporter/patch_data_factory.py
+++ b/randovania/games/fusion/exporter/patch_data_factory.py
@@ -152,13 +152,13 @@ class FusionPatchDataFactory(PatchDataFactory[FusionConfiguration, FusionCosmeti
 
     def _create_tank_increments(self) -> dict:
         tank_dict = {}
-        for definition, state in self.patches.configuration.ammo_pickup_configuration.pickups_state.items():
-            tank_dict[definition.extra["TankIncrementName"]] = state.ammo_count[0]
+        for ammo_definition, ammo_state in self.patches.configuration.ammo_pickup_configuration.pickups_state.items():
+            tank_dict[ammo_definition.extra["TankIncrementName"]] = ammo_state.ammo_count[0]
         tank_dict["EnergyTank"] = self.configuration.energy_per_tank
 
-        for definition, state in self.configuration.standard_pickup_configuration.pickups_state.items():
-            if "LauncherIncrementName" in definition.extra:
-                tank_dict[definition.extra["LauncherIncrementName"]] = state.included_ammo[0]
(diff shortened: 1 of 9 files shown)


test/test_files/patcher_data/fusion/fusion/starting_items/world_1.json
```diff
@@ -2467,7 +2467,9 @@
     "TankIncrements": {
         "MissileTank": 5,
         "PowerBombTank": 2,
-        "EnergyTank": 100
```
Miepee, line 2470: can we modify this rdvgame to make you start with a missile expansion?

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -163,6 +155,21 @@ def _create_tank_increments(self) -> dict:
...
+            if defi.name == "Missile Launcher Data"
+        )
+        tank_dict["MissileData"] = missile_launcher.included_ammo[0]
+
+        pb_launcher = next(
+            state
+            for defi, state in self.configuration.standard_pickup_configuration.pickups_state.items()
+            if defi.name == "Power Bomb Data"
+        )
+        tank_dict["PowerBombData"] = pb_launcher.included_ammo[0]
```
Miepee: it would be cleaner to have something more like
```py
for definition, state in self.configuration.standard_pickup_configuration.pickups_state.items() if def.name in ["Power Bomb Data", "Missile Launcher Data"]:
  tank_dict[definition.extra["TankIncrementName"]] = state.included_ammo[0]
```

maybe even get rid of hte hardcoded data list in there, and just make it not do anything for pickups that don't have the extra field.

M4RKERSM4SK: The two launcher items do not have that extra data, which is why I assume this code was the way it was (I just snagged it from above where it was used for starting). 

I could look at adding a "LauncherIncrementName" to the launchers extra if this is still desired to reduce this down.  adding another thing to extra shouldnt impact anything else correct?

test/test_files/patcher_data/fusion/fusion/starting_items/world_1.json
M4RKERSM4SK: Tanks as starting can only be done as random starting item, trying to figure out how to do this without everything blowing up.

Even though I increased the random starting and added in a missile tank, it still is very angry at me somewhere

I am not sure if this is worth it for the test? Not sure how tests work if I were to generate a whole new RDV and roll until I get a missile tank randoming starting item

randovania/games/fusion/exporter/patch_data_factory.py
Miepee: correct.

test/test_files/patcher_data/fusion/fusion/starting_items/world_1.json
Miepee: having a different rdvgame replace the existing ones we *usually* want to avoid since it also helps test migrations
but since fusions in beta anyway, its whatever.

M4RKERSM4SK: Updated with a Missile tank and PB tank as starting to test both

randovania/games/fusion/exporter/patch_data_factory.py
M4RKERSM4SK: Ok, added in the extra data and did what you suggested on seeing if the extra data is there then add

M4RKERSM4SK:
Updated the test rdvgame and the cleaner way to do the launcher increments

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -163,6 +155,11 @@ def _create_tank_increments(self) -> dict:
         for definition, state in self.patches.configuration.ammo_pickup_configuration.pickups_state.items():
             tank_dict[definition.extra["TankIncrementName"]] = state.ammo_count[0]
         tank_dict["EnergyTank"] = self.configuration.energy_per_tank
+
+        for definition, state in self.configuration.standard_pickup_configuration.pickups_state.items():
```
Miepee: mypy doesnt like the reused variables here

M4RKERSM4SK: Adjusted the variables to reflect ammo vs standard pickup definitions

M4RKERSM4SK:
Ok cool, looked like mypy liked the rename

PR #8369, opened 2025-06-23
Fusion Arachnus Arena and Dark Stairwell Fix

M4RKERSM4SK:
Minimap Coord Adjustments:

- Adjusted arachnus arena to see how that looks with tools
- fixed a missed spot in dark stairwell


randovania/games/fusion/logic_database/Main Deck.json
```diff
@@ -7328,6 +7332,26 @@
                     38
                 ],
                 "minimap_coordinates": [
+                    {
```
Miepee: These added coordinates are wrong, what you had before here was correct.

M4RKERSM4SK: These are correct to mage, I plopped these in cause you had that gap in Justins software or whatever it was and I wanted to see if going true to mage is better, or making our own coords based on where it appears visually is the best course. This change was for you to experiement with how that looks in that software

Miepee: The software requires accurate coordinates from mage. I was under the impression that you had them wrong here, but I could also just have made a bug when writing the coordinate system

Miepee (approved):

randovania/games/fusion/logic_database/Main Deck.json
Miepee: I love confusing myself with map coordinates

M4RKERSM4SK: Ok, Concourse ventalation was wrong and it extended 1 too many, and was left out of aracnus arena. This should now be fixed with latest commit.

PR #8372, opened 2025-06-30
Hunters - Add configuration_id to PDF

dyceron:
Used by the patcher for RNG seeding

### DIFF
--- a/randovania/games/prime_hunters/exporter/patch_data_factory.py
+++ b/randovania/games/prime_hunters/exporter/patch_data_factory.py
@@ -168,7 +168,7 @@ class HuntersPatchDataFactory(PatchDataFactory[HuntersConfiguration, HuntersCosm
     def create_game_specific_data(self, randovania_meta: PatcherDataMeta) -> dict:
         starting_items = self._calculate_starting_inventory(self.patches.starting_resources())
         return {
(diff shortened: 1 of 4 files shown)


randovania/games/prime_hunters/exporter/patch_data_factory.py
```diff
@@ -168,6 +168,7 @@ def create_visual_nothing(self) -> PickupEntry:
     def create_game_specific_data(self, randovania_meta: PatcherDataMeta) -> dict:
         starting_items = self._calculate_starting_inventory(self.patches.starting_resources())
         return {
+            "configuration_id": self.description.shareable_hash,
```
Miepee: curious why you use this for the seed instead of `self.description.get_seed_for_world(self.players_config.player_index)`

dyceron: Would it provide a meaningfully different result?

Miepee: for MW games, using the seed hash makes it so that every hunters game has the same rng seed
using the world seed that rdv uses is more correct there.

I think the seed hash can also change during migrations, but not quite sure on that.

Miepee (approved):

PR #8374, opened 2025-07-01
Fusion: even more joke hints

Miepee:
More VIPs (Very Important Pull-requests)


randovania/games/fusion/exporter/joke_hints.py
```diff
@@ -38,10 +38,19 @@
     "Samus does not have an ADAM's apple.",
     "You must collect The Babys...",
     "You shouldn't MISSILE enemies when low on ammunition.",
-    "When seeing spiders, my Main Reaction is to flee.",
+    "When seeing spiders, my first Reaction is to flee.",
```
M4RKERSM4SK: Does removing Main take the punchline out of it being main reactor?

Also, instead of flee, would it be more tailored towards Yakuza if it was "jump"? Just a thought

Miepee: The room isnt called main reactor. its called reactor core.
i'm fine with using jump instead

randovania/games/fusion/exporter/joke_hints.py
```diff
@@ -21,12 +21,36 @@
...
     "Check this out! This Baby Metroid is so squishy!",
     "Have you seen Bob yet?",
+    "I forgot the next lines but they go:\nSecret tunnel~ Secret tunnel~ Through the Overgrown Spire~",
+    "Arachnus is, despite the name, not a member of the Arachnid family.",
+    "Samus, can you go to the fridge and get me some milk? I'm thirsty.",
+    "Yellow is greener than red but both are equally blue.",
+    "Keep a cool head while fighting the SA-X.",
+    "Quickly moving left and right is not a Fusion Dance.",
+    "Samus does not have an ADAM's apple.",
+    "You must collect The Babys...",
```
M4RKERSM4SK: Maybe the whole joke can just be "The Baby"

Babys I think is the possessive form, otherwise I don't know if babies would work instead to be more grammatically correct?

M4RKERSM4SK (approved):

randovania/games/fusion/exporter/joke_hints.py
```diff
@@ -21,12 +21,36 @@
...
     "Have you seen Bob yet?",
+    "I forgot the next lines but they go:\nSecret tunnel~ Secret tunnel~ Through the Overgrown Spire~",
+    "Arachnus is, despite the name, not a member of the Arachnid family.",
+    "Samus, can you go to the fridge and get me some milk? I'm thirsty.",
+    "Yellow is greener than red but both are equally blue.",
+    "Keep a cool head while fighting the SA-X.",
+    "Quickly moving left and right is not a Fusion Dance.",
+    "Samus does not have an ADAM's apple.",
+    "The Baby...",
+    "You shouldn't MISSILE enemies when low on ammunition.",
```
JeffGainsNGames: this doesn't seem like a joke hint and could interfere with gameplay

Miepee: Can you explain further? I'm not following

cluekitty: I could read `"You shouldn't MISSILE enemies when low on ammunition.",` as advice instead of a joke hint.

JeffGainsNGames: I read this as "you shouldnt use missiles on enemies when you are low on ammo", making me, a player, think "i shouldnt use my weapons to clear enemies for some strat or maybe some change I'm not aware of" 
rather than what I assume you are going for is "dont miss when low on ammo"

PR #8377, opened 2025-07-04
Refactor TeleporterConfiguration

dyceron:
Removes copy-pasting the same function for games that support shuffled teleporters


randovania/layout/lib/teleporters.py
```diff
@@ -108,6 +108,13 @@ def is_vanilla(self) -> bool:
     def has_shuffled_target(self) -> bool:
         return self.mode == TeleporterShuffleMode.ONE_WAY_ANYTHING
 
+    @property
+    def has_one_way_anywhere(self) -> bool:
+        if self.game in (RandovaniaGame.METROID_PRIME, RandovaniaGame.METROID_PRIME_ECHOES):
```
henriquegemignani: This should be implemented by having these games having their own class which overrides this method

randovania/layout/lib/teleporters.py
```diff
@@ -118,14 +125,14 @@ def editable_teleporters(self) -> list[NodeIdentifier]:
 
     @property
     def valid_targets(self) -> list[NodeIdentifier]:
-        if self.mode == TeleporterShuffleMode.ONE_WAY_ANYTHING:
+        if self.has_one_way_anywhere() and self.has_shuffled_target():
```
henriquegemignani: I don't get why the first condition needs to exist.

randovania/layout/lib/teleporters.py
```diff
@@ -137,16 +144,19 @@ def valid_targets(self) -> list[NodeIdentifier]:
             for identifier in self.editable_teleporters:
                 node = region_list.node_by_identifier(identifier)
                 if isinstance(node, DockNode) and node.dock_type in teleporter_dock_types:
-                    # Valid destinations must be valid starting areas
-                    area = region_list.nodes_to_area(node)
-                    if area.has_start_node():
+                    if self.has_one_way_anywhere():
```
henriquegemignani: The logic inside this if seems unrelated to the condition.

henriquegemignani (changes requested):

randovania/layout/lib/teleporters.py
```diff
@@ -118,14 +118,7 @@ def editable_teleporters(self) -> list[NodeIdentifier]:
 
     @property
     def valid_targets(self) -> list[NodeIdentifier]:
-        if self.mode == TeleporterShuffleMode.ONE_WAY_ANYTHING:
```
henriquegemignani: This can just be here, no need to remove.

henriquegemignani (changes requested):

randovania/games/common/prime_family/layout/lib/prime_trilogy_teleporters.py
```diff
@@ -30,3 +31,38 @@ def static_teleporters(self) -> dict[NodeIdentifier, NodeIdentifier]:
...
+            for identifier in original:
+                node = region_list.node_by_identifier(identifier)
+                area = region_list.nodes_to_area(node)
+                # Valid destinations must be valid starting areas
+                if area.has_start_node():
+                    result.append(identifier)
+                # Hack for Metroid Prime 1, where the scripting for Metroid Prime Lair is dependent
+                # on the previous room
+                elif area.name == "Metroid Prime Lair":
+                    original.append(
```
henriquegemignani: This should be `result`.

test/layout/test_teleporters.py
```diff
@@ -100,26 +157,44 @@ def test_data(request):
     )
```
henriquegemignani, line 167: Add a test function that calls `.valid_targets` to make sure that function is working properly

dyceron: I'll be completely honest. I have no clue how to actually do this. I'm setting something like `valid_targets = TeleporterConfiguration.valid_targets`, but then I don't know how to go from there. My assumption was I have to make up a target and somehow compare it to the valid targets based on the shuffle mode?

henriquegemignani: Something like this:
```python
config = PrimeTeleporterConfiguration(...)
targets = config.valid_targets
assert targets == [...]
```

Can likely use the existing instances of the teleporter config in the fixture already.

dyceron: Well I increased coverage, but not the way in the way I should have...

test/layout/test_teleporters.py
```diff
@@ -100,26 +157,57 @@ def test_data(request):
...
+    game = RandovaniaGame.METROID_PRIME_ECHOES
+    config = PrimeTrilogyTeleporterConfiguration(
+        mode=TeleporterShuffleMode.ONE_WAY_TELEPORTER,
+        skip_final_bosses=False,
+        allow_unvisited_room_names=False,
+        excluded_teleporters=TeleporterList((), game),
+        excluded_targets=TeleporterTargetList((), game),
+    )
+    targets = config.valid_targets
+    assert targets != [test_echoes_data]
```
henriquegemignani: This is wrong. This should be doing an `==` comparison, worst case maybe just a length one.

henriquegemignani (changes requested):

test/layout/test_teleporters.py
```diff
@@ -61,6 +65,66 @@ def _a(region, area, instance_id=None):
...
+        _g(b"\xc0", 6, "One-way, with cycles", 40, mode="one-way-teleporter"),
+        _g(
+            b"\xd8\x12",
+            16,
+            "One-way, with replacement; excluded 1 elevators",
+            40,
+            mode="one-way-teleporter-replacement",
+            excluded_teleporters=[
+                _a("Area 1", "Transport to Surface and Area 2", "Elevator to Surface East"),
+                _a("Area 4 Crystal Mines", "Transport to Area 4", "Elevator to Central Caves"),
```
Tyranisaur, line 106: Is this intentionally using the wrong area and node name resulting in one excluded teleporter, or is this test case supposed to exclude 2 teleporters?

dyceron: It was supposed to be two but the test kept throwing an error saying it was 1. Not sure if I just did something wrong there.

Tyranisaur: I think that makes sense with what I was expecting. I'm fixing it in #8426

PR #8387, opened 2025-07-11
fusion: add item jingles and infant metroid automessage

cluekitty:
Updates the mars-patcher package to add base accessibility patches, item jingles, and MessageIDs for infant metroids.

Message IDs are referenced as decimal in RDV, but Hex in asm. [Their values can be found here](https://github.com/MetroidAdvRandomizerSystem/mars-fusion-asm/blob/main/inc/enums.inc#L130)

~~Will depend on a patcher release which hasn't been published yet.~~

### DIFF
--- a/randovania/games/fusion/exporter/patch_data_factory.py
+++ b/randovania/games/fusion/exporter/patch_data_factory.py
@@ -68,8 +68,9 @@ class FusionPatchDataFactory(PatchDataFactory[FusionConfiguration, FusionCosmeti
             if text != self._placeholder_metroid_message:
                 item_message = {"Languages": dict.fromkeys(self._lang_list, text), "Kind": "CustomMessage"}
             else:
-                if message_id is not None:
-                    item_message = {"Kind": "MessageID", "MessageID": message_id}
+                if message_id is None:
+                    raise ValueError("Suppossed to use a message id but none given?")
+                item_message = {"Kind": "MessageID", "MessageID": message_id}
 
             # Shiny easter eggs
             if (
--- a/randovania/games/fusion/exporter/patch_data_factory.py
+++ b/randovania/games/fusion/exporter/patch_data_factory.py
@@ -64,13 +64,12 @@ class FusionPatchDataFactory(PatchDataFactory[FusionConfiguration, FusionCosmeti
             sprite = pickup.model.name
 
             item_message = {}
-            # Special case where we ignore metroid dna right now, because that needs more patcher work.
+            # Handles special case for infant metroids which use ASM to automatically determine message via a MessageID
             if text != self._placeholder_metroid_message:
                 item_message = {"Languages": dict.fromkeys(self._lang_list, text), "Kind": "CustomMessage"}
-
-            # Resources we want to specify a generic Message ID if we have not already provided custom messaging
-            if resource in ["InfantMetroid"] and not item_message and message_id is not None:
-                item_message = {"Kind": "MessageID", "MessageID": message_id}
+            else:
+                if message_id is not None:
+                    item_message = {"Kind": "MessageID", "MessageID": message_id}
 
             # Shiny easter eggs
             if (
--- a/randovania/games/fusion/exporter/patch_data_factory.py
+++ b/randovania/games/fusion/exporter/patch_data_factory.py
@@ -53,19 +53,24 @@ class FusionPatchDataFactory(PatchDataFactory[FusionConfiguration, FusionCosmeti
             if "source" in node.extra:
                 is_major = True
             if not pickup.is_for_remote_player and pickup.conditional_resources[0].resources:
-                resource = pickup.conditional_resources[0].resources[-1][0].extra["item"]
-                jingle = pickup.conditional_resources[0].resources[-1][0].extra.get("Jingle", "Minor")
-                message_id = pickup.conditional_resources[0].resources[-1][0].extra.get("MessageID", None)
+                conditional_extras = pickup.conditional_resources[0].resources[-1][0].extra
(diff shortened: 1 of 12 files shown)


randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -54,6 +54,8 @@ def _create_pickup_dict(self, pickup_list: list[ExportedPickupDetails]) -> dict:
                 is_major = True
             if not pickup.is_for_remote_player and pickup.conditional_resources[0].resources:
                 resource = pickup.conditional_resources[0].resources[-1][0].extra["item"]
+                jingle = pickup.conditional_resources[0].resources[-1][0].extra.get("Jingle", "Minor")
```
Miepee: Move pickup.cond.... Into a variable please

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -54,6 +54,8 @@ def _create_pickup_dict(self, pickup_list: list[ExportedPickupDetails]) -> dict:
...
+                message_id = pickup.conditional_resources[0].resources[-1][0].extra.get("MessageID", None)
             else:
                 resource = "None"
```
Miepee, line 62: I would redo the logic, so that the "custom_message" var below gets treated as the dict for ItemMessages. So that later down the line you can still say "major_entry["ItemMessages"] = custom_messages without having weird conditions or building a different dict there. Otherwise this reads really weird.
Very recommended to rename it for that

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -53,17 +53,24 @@ def _create_pickup_dict(self, pickup_list: list[ExportedPickupDetails]) -> dict:
...
-            custom_message = {}
+            item_message = {}
             # Special case where we ignore metroid dna right now, because that needs more patcher work.
             if text != self._placeholder_metroid_message:
-                custom_message = {"Languages": dict.fromkeys(self._lang_list, text), "Kind": "CustomMessage"}
+                item_message = {"Languages": dict.fromkeys(self._lang_list, text), "Kind": "CustomMessage"}
+
+            # Resources we want to specify a generic Message ID if we have not already provided custom messaging
+            if resource in ["InfantMetroid"] and not item_message and message_id is not None:
+                item_message = {"Kind": "MessageID", "MessageID": message_id}
```
Miepee: 1) you should update the commments
2) why is the second if logic so complicated? you can just make it an `else`

randovania/games/fusion/logic_database/header.json
```diff
@@ -488,7 +586,9 @@
             },
             "LowerWaterLevel": {
                 "long_name": "Sector 4 Water Level Lowered",
-                "extra": {}
+                "extra": {
+                    "MessageID": 38
```
Miepee: ?

test/test_files/patcher_data/fusion/fusion/all_hidden_with_nothing/world_1.json
```diff
@@ -29,6 +29,7 @@
             {
                 "Source": "MainDeckData",
```
Miepee, line 30: can you add another test case, that has these pickup style settings?
grafik

cluekitty: Are these test files typically generated or hand-crafted?

Miepee: hand crafted

Miepee: actually that depends what you mean by that
typically we just choose a preset with the settings we want, then take the first rdvgame that generates with that.

Sometimes in rare cases, we the manually modify the rdvgame afterwards if we need certain things from it that cant be done via the preset (e.g. a specific item at a specific location)

cluekitty: The question was mostly "Do I need to specify everything manually and type it out?" or "Do I need to have RDV Generate something and use that as the baseline (with possible edits)?" It sounds like the latter is the case here.

I do notice that the Seed hash for these are all fixed, would I need to specify a seed hash somehow during generation or would I just manually edit that after the fact?

Miepee: The seed hash (among a few other things) is mocked during the tests

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -53,17 +53,23 @@ def _create_pickup_dict(self, pickup_list: list[ExportedPickupDetails]) -> dict:
...
-            custom_message = {}
-            # Special case where we ignore metroid dna right now, because that needs more patcher work.
+            item_message = {}
+            # Handles special case for infant metroids which use ASM to automatically determine message via a MessageID
             if text != self._placeholder_metroid_message:
-                custom_message = {"Languages": dict.fromkeys(self._lang_list, text), "Kind": "CustomMessage"}
+                item_message = {"Languages": dict.fromkeys(self._lang_list, text), "Kind": "CustomMessage"}
+            else:
+                if message_id is not None:
+                    item_message = {"Kind": "MessageID", "MessageID": message_id}
```
Miepee:
```suggestion
            else:
                if message_id is None:
                    raise ValueError("Suppossed to use a message id but none given?")
                item_message = {"Kind": "MessageID", "MessageID": message_id}
```

PR #8388, opened 2025-07-11
Add 10 more joke hints

Miepee:


randovania/exporter/hints/joke_hints.py
```diff
@@ -16,4 +16,14 @@
     "For optimal play, ensure proper posture and a positive mental attitude.",
     "A useless hint can be found here.",
     "We're all very proud of you.",
+    "You've been playing the game for a while. Why not take a break?",
+    "Hints are never gonna give you up, never gonna let you down.",
+    "1, 2, 3, 4! Of randomization I want more!",
+    "If you have a randomized game that you like, consider sharing it!",
+    "Yahaha! You found this message!",
```
Tyranisaur: Is this too much of a botw/totk reference to fit as a generic hint?

MayberryZoom: If it's meant to be a Korok reference, why not just include the actual quote? "Yahaha! You found me!" Imo, this version feels clunky and it would be funnier/more recognizable that way.

Miepee: > Is this too much of a botw/totk reference to fit as a generic hint?

IMO, if we have fusion's secret message here, then this fits too. 

> why not just include the actual quote

My main concern was that "you found **me**" could be misinterpreted as finding something other than the joke message. But if you think that's fine, then I don't mind using it. Otherwise I could suggest "You found this joke hint" as a compromise.

Tyranisaur: I belive that the line from fusion is more neutral in a way where someone could use that phrase without the intention of referencing fusion. On the other side, I believe the "Yahaha" exclamation is a much more specific reference. 

The guidelines at the top of this list don't mention that hints should be generic and apply broadly, while hints pertaining to the individual games should be listed in the specific game. Moreover, I'm not sure how strongly the joke hints should be specific to the individual game (series). I have mixed feelings about the echoes hint that references raven beak.

Miepee: > I believe the "Yahaha" exclamation is a much more specific reference.

While that is true, I did specifically (try to) phrase this in a way, where the joke hint itself is understandable without needing to know that its referencing a specific thing.

> Moreover, I'm not sure how strongly the joke hints should be specific to the individual game (series). I have mixed feelings about the echoes hint that references raven beak.

This really feels like it should a case by case basis IMO. I'm not particularly fond of the RB hint in echoes either, but I also don't want to get rid of the other m references in am2r/fusion/msr either. But maybe the reason why I have less issues with those is that the reference there makes more sense? You do collect baby metroids in all 3 of them, but you don't interact with RB (or any chozo for that matter) in echoes.

MayberryZoom: > While that is true, I did specifically (try to) phrase this in a way, where the joke hint itself is understandable without needing to know that its referencing a specific thing.

For the record, I was totally unsure whether it was a reference or not until you confirmed it. I think the direct quote still works if you don't understand the reference, while also making it more clear where it comes from.

Tyranisaur: I don't feel strongly enough about it to block it either way, but I both don't see the problem with using the actual line from the game, and I would also prefer to remove the "Yahaha!" so that the reference won't be quite as strong.

Miepee: addressed

Tyranisaur (approved):

PR #8391, opened 2025-07-12
Attach rdvgame to PDF sentry errors

Miepee:
No clue on how to test locally that this actually works.
No changelog, since we don't add entries for telemetry stuff.

### DIFF
--- a/randovania/exporter/patch_data_factory.py
+++ b/randovania/exporter/patch_data_factory.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+import json
 import typing
 from random import Random
 from typing import TYPE_CHECKING
 
 import json_delta
+import sentry_sdk
 
 from randovania.exporter import pickup_exporter
 from randovania.exporter.hints.hint_exporter import HintExporter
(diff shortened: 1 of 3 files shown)


randovania/exporter/patch_data_factory.py
```diff
@@ -90,11 +92,19 @@ def create_data(self, custom_metadata: PatcherDataMeta | None = None) -> dict:
...
+                json.dumps(self.description.as_json(force_spoiler=True)).encode("utf-8"),
+                filename="rdvgame.json",
+                content_type="application/json",
+                add_to_transactions=True,
+            )
 
-        game_data = self.create_game_specific_data(randovania_meta)
-        json_delta.patch(game_data, self.patches.custom_patcher_data)
-        game_data["_randovania_meta"] = randovania_meta
+            randovania_meta = custom_metadata or self.create_default_patcher_data_meta()
```
MayberryZoom: Not sure if this is too much of a nitpick, but should this be indented into the `with` block? At the very least, not indenting it would make a cleaner commit on squash merge. To me, that would also communicate what the code does slightly more clearly.

MayberryZoom (changes requested):
Probably a good thing that the permalink is being generated during that test now, I don't think we want invalid presets in the test rdvgames.

Tests still failing (need to update patcher jsons), but other than that and my other comment, lgtm

randovania/exporter/patch_data_factory.py
Miepee: i suppose the indent might not actually be necessary here.
if you want to start a transaction for measuring purposes, then all the code has to be inside of the transaction scope.
But since we're only attaching a file to the current scope, we can probably get rid of it.

henriquegemignani (approved):

PR #8392, opened 2025-07-12
Ruff update

ThanatosGit:
Ignore PLC0415 [import-outside-top-level ](https://docs.astral.sh/ruff/rules/import-outside-top-level/#import-outside-top-level-plc0415) 
Ignore PLW1641  (should maybe fixed instead?) [eq-without-hash ](https://docs.astral.sh/ruff/rules/eq-without-hash/)
Fix PT028 [pytest-parameter-with-default-argument](https://docs.astral.sh/ruff/rules/pytest-parameter-with-default-argument/)  
Fix `SyntaxError: Cannot use type parameter lists on Python 3.11 (syntax was added in Python 3.12)` by targeting 3.12 with Ruff
Implement UP047  [non-pep695-generic-function](https://docs.astral.sh/ruff/rules/non-pep695-generic-function/) 
Fix UP040 [non-pep695-type-alias](https://docs.astral.sh/ruff/rules/non-pep695-type-alias/)
Fix UP046 [non-pep695-generic-class](https://docs.astral.sh/ruff/rules/non-pep695-generic-class/)
Fix UP035 (Import from `typing` instead: `override`) [deprecated-import](https://docs.astral.sh/ruff/rules/deprecated-import/)

Supersedes #8368

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -204,7 +204,6 @@ git_describe_command = [
 
 [tool.ruff]
 line-length = 120
-target-version = "py312"
 
 [tool.ruff.lint]
 select = [
@@ -231,7 +230,6 @@ extend-ignore = [
     "PLW0603",  # we use some globals here and there
     "PLW2901",  # TODO: should actually fix these
     "PLC0415",  # we explicitly use the import within a function to postpone the loading of a module
-    "PLW1641",  # allow implementation __eq__ without __hash__
     
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -204,7 +204,7 @@ git_describe_command = [
 
 [tool.ruff]
 line-length = 120
-target-version = "py311"
+target-version = "py312"
 
 [tool.ruff.lint]
 select = [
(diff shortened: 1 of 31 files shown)


pyproject.toml
```diff
@@ -204,7 +204,7 @@ git_describe_command = [
 
 [tool.ruff]
 line-length = 120
-target-version = "py311"
+target-version = "py312"
```
henriquegemignani: this can be removed, it discovers the target automatically from the required python version

pyproject.toml
```diff
@@ -230,6 +230,9 @@ extend-ignore = [
     "ISC001",  # may cause conflicts with formatter
     "PLW0603",  # we use some globals here and there
     "PLW2901",  # TODO: should actually fix these
+    "PLC0415",  # we explicitly use the import within a function to postpone the loading of a module
+    "PLW1641",  # allow implementation __eq__ without __hash__
```
henriquegemignani: was there too many cases? It can just be a `__hash__ = None`, but it's good to fix

henriquegemignani (changes requested):

pyproject.toml
ThanatosGit: I don't know how to do this correctly regarding mypy without slapping a `# type: ignore` everywhere. Otherwise it complains because of `__hash__ = None` is not compatible with`"Callable[[object], int]"` from `object`.
```
error: Incompatible types in assignment (expression has type "None", base class "object" defined the type as "Callable[[object], int]")  [assignment]
```

I've added a new commit that implements the `__hash__` for all classes. Basically, I just delegate it to the hash functions of the fields used for the `__eq__` comparison.

PR #8393, opened 2025-07-14
Fusion Fix Sanctum Missing Room Rame

M4RKERSM4SK:
From Bug Report
- Sanctum had a "Room name not provided"
- Was identified that Sanctum has 2 rooms and the transtion was occuring to the room id that was not getting passed
- Added in the other room ID to the DB

Tested
- Both Room IDs will now get passed to the patcher and both will display as Sanctum comming in from both entrances now

image

### DIFF
--- a/randovania/games/fusion/logic_database/Sector 2 TRO.json
+++ b/randovania/games/fusion/logic_database/Sector 2 TRO.json
@@ -9774,7 +9774,8 @@
             "extra": {
                 "map_name": "Sector 241",
                 "room_id": [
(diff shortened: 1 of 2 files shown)


cluekitty (changes requested):
The tests are failing, you may need to update the rdvgame files so that they have the room name.

M4RKERSM4SK:
Whoops, didnt send my test files commit (one day will get pycharm figured out)

Should be on the way shortly

PR #8394, opened 2025-07-15
Fusion Split Hints Operations having Bosses and Restricted having remaining

M4RKERSM4SK:
With discussion/feedback, it might be a better game experience/balance for the infant hints to be changed to the following:

- Infant Metroids hints that are on bosses are placed on operations deck terminal
- Inant Metroids hints that are not on bosses are placed in the restricted area terminal

Tested

- Infant hints are distrobuted to each terminal appropriately
- Tested if the hints held true in game by collecting them to confrim is was placed/hinted correctly
- Tested empty cases of each (either none on bosses or all on bosses)
- Other hints were still functional


M4RKERSM4SK:
My commit on the check on the extra data didnt go through, comming in a sec

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -242,23 +243,57 @@ def _create_nav_text(self) -> dict:
...
+                        fusion_boss_pickup_indexes.append(pickup_node.pickup_index)
+
+        artifacts_on_bosses = []
+        for boss_index in fusion_boss_pickup_indexes:
+            if boss_index in artifact_pickup_indexes:
+                artifacts_on_bosses.append(artifact_pickup_indexes[boss_index])
+
         if metroid_precision != SpecificPickupHintMode.DISABLED:
-            # loop through all metroids and place them alternating on Restricted then Operations Nav Rooms
+            # loop through all metroids and place ones on bosses on Operations Deck and the rest on Restricted Area
```
Miepee, line 270: Should we limit operations deck to max 2 hints still?
Considering bosses have higher chance than normal locations, feels like otherwise its more common to have like 4:1 (or higher) with operations deck vs labs, where its then prolly easy enough to juststumble over the last few.

M4RKERSM4SK: Given the data that Jeff had, the majority of seeds with 5 metorids on pre-placment will be with two bosses on average. When playing this felt good and also felt fine with even 1 or 3 on bosses. I havent rolled a bossless or 4-5 bosses yet. 

Even with 20 metroids, it was about 8 on bosses iirc

I honestly would like to see the feedback on how it feels in practice and on beta feedback before trying to limit the ops hint.

Here is Jeffs data
https://discord.com/channels/914291389293027329/1215675687936196608/1379513103406071970

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -242,23 +243,57 @@ def _create_nav_text(self) -> dict:
...
             for _, text in metroid_hint_mapping.items():
                 if "has no need to be located" in text:
                     continue
-                if hint_counter % 2 == 0:
-                    restricted_hint = restricted_hint + text + " "
-                else:
+                if _.long_name in artifacts_on_bosses:
```
Miepee: A variable called `_` implies "this variable is unused, but is declared for one reason or another".
If you actually use the variable like here, then you need to rename it.

(Altho for clarification this variable shouldnt have existed earlier and the code shouldve just done `for text in mapping.values()`)

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -242,23 +243,57 @@ def _create_nav_text(self) -> dict:
         hints = {}
         restricted_hint = ""
         operations_hint = ""
-        hint_counter = 0
+        artifact_locations = guaranteed_item_hint.find_locations_that_gives_items(
+            artifacts, self.description.all_patches, self.players_config.player_index
+        )
+        artifact_pickup_indexes = {}
```
Miepee: Your whole chain here is kinda weird
- first you make a (hinted index, resource long name) mapping
- then you go through *all* pickup locations in the name and make a mapping of which ones are on bosses
- then you go through all boss indices, and put the ones that are being hinted in another list
- and finally in the "split hints" section, you compare whether the resource long name is contained in the last list you did.

This seems way overcomplicated.
You can instead simplify the whole thing as:
- go through your hinted indices, and if its a boss index, put the resource on a list
- in the "split hints" section, see if your resource is in the list.

randovania/games/fusion/exporter/patch_data_factory.py
Miepee: Maybe im just unnecessarily worrying about a worst case then.

randovania/games/fusion/exporter/patch_data_factory.py
M4RKERSM4SK: fixed varaible name

randovania/games/fusion/exporter/patch_data_factory.py
M4RKERSM4SK: re-did it

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -242,23 +242,49 @@ def _create_nav_text(self) -> dict:
...
+        ]
+        artifacts_on_bosses = []
+
+        for resource, resource_locations in artifact_locations.items():
+            for _, resource_location in resource_locations:
+                if (
+                    self.game.region_list.node_from_pickup_index(resource_location.location).extra.get("source")
+                    in fusion_bosses
+                ):
+                    artifacts_on_bosses.append(resource.long_name)
```
Miepee: thoughts on 
```py
artifacts_on_bosses = [
  resource.long_name
  for resource, resource_locations in artifact_locations.items()
  for _, resource_location in resource_locations
  if self.game.region_list.node_from_pickup_index(resource_location.location).extra.get("source") in fusion_bosses
]
```
?
~~At the very least its better IMO to not have a seperate loop/nesting for unpacking the tuple~~ NVM mixed up the typing of the artifact_locations dict.

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -242,23 +242,49 @@ def _create_nav_text(self) -> dict:
...
-            # loop through all metroids and place them alternating on Restricted then Operations Nav Rooms
-            for _, text in metroid_hint_mapping.items():
+            # loop through all metroids and place ones on bosses on Operations Deck and the rest on Restricted Area
+            for metroid_resource, text in metroid_hint_mapping.items():
                 if "has no need to be located" in text:
                     continue
-                if hint_counter % 2 == 0:
-                    restricted_hint = restricted_hint + text + " "
-                else:
+                if metroid_resource.long_name in artifacts_on_bosses:
```
Miepee: dont use the long name - use the resource directly (which also means, save the resource directly into your list instead of the long name)

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -242,23 +242,49 @@ def _create_nav_text(self) -> dict:
...
+                else:
+                    restricted_hint = restricted_hint + text + " "
             restricted_hint = restricted_hint.rstrip()
             operations_hint = operations_hint.rstrip()
 
-            # special handling when there's no Metroids to hint on Operations Deck
-            if hint_counter == 1:
+            # special handling when there's no Metroids to hint on either terminal
+            if not operations_hint:
                 operations_hint = "This terminal was unable to scan for any [COLOR=3]Metroids[/COLOR]."
```
Miepee: make the string a variable and reuse it

M4RKERSM4SK:
Updated with list comprhension and the other suggestions, ready for re-review again

Miepee (approved):

PR #8395, opened 2025-07-15
Fusion update mars patcher

M4RKERSM4SK:
mars patcher to 0.6.1

Patcher side update included
- Audio fix on messages/jingles
- Block Graphics
- patcher map change for s5


Miepee (approved):
block graphics were already in, it includede a fix for zazabi speedway

PR #8403, opened 2025-07-18
Fusion Hue Shift Override Cosmetic Option

M4RKERSM4SK:
Added in GUI options to override the randomized hue shift defaults with a symetrical value of your choice

Reasoning
- Fashion Fusion, how else you going to match suit and beams combos

Min Max vs Single Value
- I found in testing that setting a min and max value to randomize between felt really boring and not really that great cosometic wise or really worth the effort. It was better to have full rando or a specific value set

Cosmetic Patches Dialog
- Made it to where "Override Hue Shift" and the Spinner options were disabled until "Randomize Hue Shift" is enabled. 
- I went with this functionality and override wording cause I wanted people to see that you can have full rando, but then if you want then get the option to "override" with your own spefic value. However if you want vanilla, just uncheck "Randomize Hue Shift" and both go back to greyed out
- Persisted the settings

Test files
- I updated the test files with what I could.
- If you add in the checkboxes that are disabled until a different checkbox is enabled, it fails cuase I dont think the tests were built for that. I didnt see any other options in the QT stuff to try and accomidate this specfic case in an automated fashion
- I didnt see any testing for other games that test spinners
- I didnt see any output json test that covers cosmetic settings

image

What is sent to patcher (example above with all use cases)

    "Palettes": {
        "Seed": 169563473,
        "Randomize": {
            "Beams": {
                "HueMin": 30,
                "HueMax": 30
            },
            "Tilesets": {
                "HueMin": 0,
                "HueMax": 360
            }
        },
        "ColorSpace": "Oklab"
    },


Miepee:
> setting a min and max value to randomize between felt really boring 

:(
I actually use min-max in am2r a lot in order to have a "full" random hue, but without having hues that are very close to vanilla's (e.g min 30 max 330), and would be a bit sad to loose that functionality.

Also, have you put any thought about the "symmetric" feature in the patcher? I cant fully remember how it work in detail but the schema description should explain it. I think biospark also has a more lengthy explanation somewhere in fusion-dev.

M4RKERSM4SK:
Got the min max values in and symmetric

image

M4RKERSM4SK:
Chanaged "Palette Shuflle" to "Color Rotation"

image

randovania/games/fusion/layout/fusion_cosmetic_patches.py
```diff
@@ -30,18 +30,31 @@ class ColorSpace(Enum):
...
     enemy_hue_min: int = 0
     enemy_hue_max: int = 360
+    enemy_hue_override_min: int = 0
+    enemy_hue_override_max: int = 0
     enable_tileset_palette: bool = False
+    enable_tileset_palette_override: bool = False
     tileset_hue_min: int = 0
     tileset_hue_max: int = 360
+    tileset_hue_override_min: int = 0
+    tileset_hue_override_max: int = 0
```
JeffGainsNGames, line 55: the original `suit_hue_min/max` (etc) were what I planned to use for this setting when I eventually got back to it
you can safely remove them in favour of your descriptive named variables.

JeffGainsNGames (changes requested):

randovania/games/fusion/layout/fusion_cosmetic_patches.py
M4RKERSM4SK: They are still used. When only `randomize hue shift` is selected, the values are sent to the patcher since they are the true min max. The override variables will retain the values set per spinner not only for the persistence but are sent only when the override checkbox is selected. I thought about hard coding the min max, but just incase if we needed to change I left those variables in.

JeffGainsNGames: I don't see any use of them besides this file
image

M4RKERSM4SK: for attr_name, palettes_key in palette_type_dict.items():
            if getattr(cosmetics, f"enable_{attr_name}_palette"):
                palettes[palettes_key] = {
                    "HueMin": getattr(cosmetics, f"{attr_name}_hue_min"),
                    "HueMax": getattr(cosmetics, f"{attr_name}_hue_max"),
                }
            if getattr(cosmetics, f"enable_{attr_name}_palette") and getattr(
                cosmetics, f"enable_{attr_name}_palette_override"
            ):
                palettes[palettes_key] = {
                    "HueMin": getattr(cosmetics, f"{attr_name}_hue_override_min"),
                    "HueMax": getattr(cosmetics, f"{attr_name}_hue_override_max"),
                }

JeffGainsNGames: yeah I'm a dummy I was catching on, ignore

JeffGainsNGames (approved):

PR #8405, opened 2025-07-18
Migrate uic call to own script called via pre-commit

henriquegemignani:


ThanatosGit (changes requested):
Looks fine so far to me but maybe add
```
ci:
    skip: [qtui]
```
because I don't think it makes sense to do it as part of the CI + it fails anyways because it doesn't have `uv` (see failed  check)

henriquegemignani:
> Looks fine so far to me but maybe add
> 
> ```
> ci:
>     skip: [qtui]
> ```
> 
> because I don't think it makes sense to do it as part of the CI + it fails anyways because it doesn't have `uv` (see failed check)

I didn't go back to this, but I'm thinking of just using the `python` language and have it install the package. Sidestep using uv, annoyingly.

Or do what you suggested, also works. Didn't think of that :p

ThanatosGit (approved):

PR #8406, opened 2025-07-19
MW Session: rename "generate without spoiler" to "for race"

### LINKED ISSUE #8401: MW: rename "gen without spoiler" to "gen for race"
solo: 
Image
multi:
Image

Miepee:
Fixes #8401


Tyranisaur (changes requested):
Probably good to reword the dialog message doen on line 991 to also use "generated for race" instead of "generated without spoiler".

Tyranisaur (approved):

PR #8409, opened 2025-07-22
Bump OSRR to v3.2.1

dyceron:
Bug fixes


CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Metroid: Samus Returns
 
 - Added: 10 more joke hints.
+- Fixed: Starting with extra Metroid DNA now counts toward the goal.
```
ThanatosGit, line 20: `toward` vs. `towards` is a british english vs. american english thing, right?

randovania/games/samus_returns/exporter/patch_data_factory.py
```diff
@@ -675,6 +675,7 @@ def _door_patches(self) -> list[dict[str, str]]:
 
     def _objective(self, config: MSRConfiguration) -> dict:
         return {
+            "placed_dna": config.artifacts.placed_artifacts,
```
ThanatosGit, line 678: Should we add the “toal_dna”? At the moment it's 39 always anyway, so Idc but wanted to mention it.

uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
```
ThanatosGit: Hmm the revision number goes down and it removed the "upload_time" everywhere.
I think you are using an outdated uv version.
That's the only thing which really must change before merging.

randovania/games/samus_returns/exporter/patch_data_factory.py
dyceron: I dont think it's necessary. We won't ever support more than 39 DNA here, at least currently.

CHANGELOG.md
dyceron: Yes it is. Toward is AE.

uv.lock
dyceron: Fixed

ThanatosGit (approved):

PR #8413, opened 2025-07-24
fusion: change hint settings

JeffGainsNGames:
as directed by Dunc here https://discordapp.com/channels/914291389293027329/1215675687936196608/1397279857456447538


M4RKERSM4SK (approved):
To be honest not sure what the `minimum_location_weight_for_hint_placement` does, but just in some spot checking for OSH, there looked to be better logical hints compared to what I was used to seeing on average. Vanilla seems still bad in some hints, but agreed with duncathan looked unrleated to the changes

dyceron (approved):

PR #8414, opened 2025-07-24
Echoes: Make Aerie safer

Tyranisaur:
* Add connection back to Dark Samus 2 event node from Bottom Floor.

### DIFF
--- a/randovania/games/prime2/logic_database/Sanctuary Fortress.txt
+++ b/randovania/games/prime2/logic_database/Sanctuary Fortress.txt
@@ -2151,6 +2151,9 @@ Hint Features - Abyss, Boss, Echo Visor Lock, Elevator, Portal
       Morph Ball Bomb and Bomb Space Jump (Expert) and Movement (Expert) and Use Screw Attack (Space Jump)
   > Elevator to Aerie Transport Station
       Echo Visor and After Aerie Pickup
+  > Event - Dark Samus 2
+      # Having a connection back the event node lets the genrator know when this sequence is safe.
+      After Dark Samus 2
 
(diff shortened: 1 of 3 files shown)


Tyranisaur:
Genned some seeds to look at the effect of this. https://docs.google.com/spreadsheets/d/1gKB6yVDSLPciEL69vMNx-oaxV8ynv7nHVzkHJclmc9A/edit?usp=sharing Not sure how much of this is noise and whether it would help to gen more than 1000 seeds. But it looks like the frequency of STK on Aerie is reduced from about 14% by about 5% Curiously, it's also showing a similar, but slightly smaller reduction of STK on 2Mos... Both of these locations show an increase in progression probability, more pronounced on aerie. Part of this is an increased likelihood of vanilla Echo, then a smaller increase of Cobalt, but otherwise mostly from temple keys and energy tanks.

randovania/games/prime2/logic_database/Sanctuary Fortress.txt
```diff
@@ -2151,6 +2151,8 @@ Hint Features - Abyss, Boss, Echo Visor Lock, Elevator, Portal
       Morph Ball Bomb and Bomb Space Jump (Expert) and Movement (Expert) and Use Screw Attack (Space Jump)
   > Elevator to Aerie Transport Station
       Echo Visor and After Aerie Pickup
+  > Event - Dark Samus 2
+      After Dark Samus 2
```
Miepee, line 2156: An explanation of this would also be appreciated

Miepee: To answer my own question:
image

PR #8422, opened 2025-08-01
Echoes: Make Amorbis safer

Tyranisaur:
Add connection back to Amorbis event node.

Similar to the other PR, for Aerie. Generation data added to https://docs.google.com/spreadsheets/d/1gKB6yVDSLPciEL69vMNx-oaxV8ynv7nHVzkHJclmc9A/edit?usp=sharing (same document).

### DIFF
--- a/randovania/games/prime2/logic_database/Dark Agon Wastes.txt
+++ b/randovania/games/prime2/logic_database/Dark Agon Wastes.txt
@@ -900,6 +900,8 @@ Hint Features - Boss, Temple
   > Event - Dark Agon Temple Key Gate
       # https://www.youtube.com/watch?v=yd91qBUxCeA
       Dark Agon Key 1 and Dark Agon Key 2 and Dark Agon Key 3 and Morph Ball Bomb and Morph Ball and Space Jump Boots and Bomb Space Jump (Expert) and Single Room Out of Bounds (Expert) and Slope Jump (Expert) and Standable Terrain (Expert) and Dark World Damage ≥ 175
+  > Event - Amorbis
+      After Amorbis
   > Amorbis Trigger
       All of the following:
           Knowledge (Intermediate)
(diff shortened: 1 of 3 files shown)


randovania/games/prime2/logic_database/Dark Agon Wastes.txt
```diff
@@ -900,6 +900,8 @@ Hint Features - Boss, Temple
   > Event - Dark Agon Temple Key Gate
       # https://www.youtube.com/watch?v=yd91qBUxCeA
       Dark Agon Key 1 and Dark Agon Key 2 and Dark Agon Key 3 and Morph Ball Bomb and Morph Ball and Space Jump Boots and Bomb Space Jump (Expert) and Single Room Out of Bounds (Expert) and Slope Jump (Expert) and Standable Terrain (Expert) and Dark World Damage ≥ 175
+  > Event - Amorbis
+      After Amorbis
```
Miepee, line 905: Can you elaborate on how this makes it safer? Isn't the generator already able to get to that node after amorbis is dead via safe zone -> amorbis trigger -> event amorbis?

Tyranisaur: safe zone -> amorbis trigger requires tricks.  

The culprit is this: for amorbis to be surface level safe, you need to be able to do the event and immediately leave. In its current form, I believe this is true with tricks and sufficient items, but otherwise you need the event from taking the dark agon energy. Now the problem is that inside the `advance_reach_with_possible_unsafe_resources` check that you would use to figure this out, you don't collect anything that isn't considered surface level safe. And here the problem is that you need to collect this event to be able to return to the amorbis event node. This PR changes this so that you always can always loop back to the amorbis event node, making the dark agon energy event safe, making it possible to get out of dark agon temple, which leads to the generator simply collecting all of these actions when all the requirements are satisfied.

PR #8423, opened 2025-08-01
Dread: Fix extra Grapple req to escape Grapple Beam Room

Tyranisaur:


Logarhythmic (approved):
Interesting how this one went unnoticed for so long.

PR #8425, opened 2025-08-01
Add_burenia_pseudo_video

Logarhythmic:
Added a video for a pseudo in burenia. Not sure if change is signifcant enough to warrant using the change log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed: It is now logical to get back to the lower door in Grapple Beam Room without Grapple Beam, making it logical to escape without Grapple Beam when doors are randomized.
 
+##### Burenia
+
+- Added: The pseudo to break the blob in the Hub to Darion now has a video.
+
 ## [10.0.0] - 2025-08-0?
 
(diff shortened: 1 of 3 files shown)


JeffGainsNGames:
> Not sure if change is significant enough to warrant using the change log

it might seem small, but any user facing change we do capture. some reference material to help:
https://github.com/randovania/randovania/blob/main/CONTRIBUTING.md#example-of-changes-that-should-have-changelog

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed: It is now logical to get back to the lower door in Grapple Beam Room without Grapple Beam, making it logical to escape without Grapple Beam when doors are randomized.
 
+##### Burenia
+
+- Added: The pseudo to break the blob in the Hub to Darion now has a video.
```
MayberryZoom: Imo we should be more specific here. I'm thinking something like:

> Added: A video for breaking the uppermost blob in Burenia Hub to Dairon with a pseudo-wave.

Also, I believe the changelog is conflicting so that should be fixed.

Logarhythmic: Changed.

MayberryZoom (approved):

JeffGainsNGames (approved):

PR #8426, opened 2025-08-02
Fix One-Way Anywhere failing to generate

Tyranisaur:


ThanatosGit:
If we were able to break it without a test failing, we maybe should a test to not have the same regression in the future again.

duncathan:
this change technically affects permalinks, though normal users currently have no way to generate a permalink that would be broken so it's probably fine to put this in a patch release?

Tyranisaur:
Looks like the tests were written kinda poorly, and didn't actually check the number of valid targets when running in one way anywhere or original connections. Also, I fixed the tests so the parameterized test aspect makes more sense, reducing the number of test executions from this file from 72 to 8.

PR #8430, opened 2025-08-04
Add hotkeys for Async Race Room

ThanatosGit:
Took me way too long because of the keyboard.Key vs. keyboard.KeyCode. Had an iteration where I saved the name and vk seperately and then later squashed everything together.
I'm very unsure if and how it works on other OS: https://pynput.readthedocs.io/en/latest/limitations.html#linux
but I decided to open it for discussion.

Since I'm lazy, I let Copilot generate the SPECIAL_KEY_DISPLAY_NAMES and reformulate my doc strings.

Screenshot 2025-08-03 221034
Screenshot 2025-08-03 220921


Miepee:
Since this doesnt work under linux and mac (needs root and/or doesnt work as global shortcuts) i'd definitely hide the button/functionality there.

But also, I was under the impression that instead of doing this, we wanted to hook it up to livesplit instead?

henriquegemignani:
> But also, I was under the impression that instead of doing this, we wanted to hook it up to livesplit instead?

Both this and that seems decent to have, potentially at the same time. Though that'd depend more on the users.

pyproject.toml
```diff
@@ -69,6 +69,9 @@ dependencies = [
 
     # For TypeVar default
     "typing-extensions>=4.4.0",
+
+    # For global hotkeys
+    "pynput",
```
henriquegemignani, line 74: This could be part of the `gui` group.

PR #8434, opened 2025-08-05
Migrate to Azahar

ThanatosGit:
Needs a new patcher version first to accept cci extensions

### DIFF
--- a/randovania/interface_common/persisted_options.py
+++ b/randovania/interface_common/persisted_options.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 from typing import TYPE_CHECKING
 
 from randovania.lib import json_lib, migration_lib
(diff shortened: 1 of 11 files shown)


dyceron:
Bumped the patcher version, so I think this PR is ready for review @Miepee

randovania/interface_common/persisted_options.py
```diff
@@ -154,6 +155,23 @@ def _msr_enable_remote_lua(options: dict) -> None:
...
+    if "output_preference" in options.get("game_samus_returns", {}):
+        as_dict: dict = json.loads(options["game_samus_returns"]["output_preference"])
+        if as_dict.get("selected_tab", "") == "citra":
+            as_dict["selected_tab"] = "azahar"
+        # dirty replacement: replaces the platform and the field name for the always empty dict tab options
+        tab_options = as_dict.get("tab_options", None)
+        if tab_options:
+            tab_options["azahar"] = {}
+            tab_options.pop("citra", None)
+        options["game_samus_returns"]["output_preference"] = json_lib.dumps_small(as_dict)
```
Miepee: we shouldn't rely on RDV functions here. probably better/safer to replace the json_lib call with 
`json.dumps(obj, separators=(",", ":"))`

dyceron: Can you explain the difference since I didn't write this?

Miepee: `json_lib` is randovania code and thus can change and break this migration. From our docs:
> ## Best practices for migrations
> - Avoid using any existing Randovania code, as these might change in the future
and break old migrations.

dyceron: Doesn't really answer my question but ok.

Miepee: I'm not sure how else to answer it?
We don't want to rely on any other code of randovania that isn't in this persisted_options file.

If you mean the functionality, they're the same. I just inlined `dumps_small`

dyceron: Changed it

randovania/interface_common/persisted_options.py
```diff
@@ -154,6 +155,23 @@ def _msr_enable_remote_lua(options: dict) -> None:
...
+    if "output_preference" in options.get("game_samus_returns", {}):
+        as_dict: dict = json.loads(options["game_samus_returns"]["output_preference"])
+        if as_dict.get("selected_tab", "") == "citra":
+            as_dict["selected_tab"] = "azahar"
+        # dirty replacement: replaces the platform and the field name for the always empty dict tab options
+        tab_options = as_dict.get("tab_options", None)
+        if tab_options:
+            tab_options["azahar"] = {}
+            tab_options.pop("citra", None)
+        options["game_samus_returns"]["output_preference"] = json.dumps(obj, separators=(",", ":"))
```
Miepee: You need to use `as_dict` here instead of `obj`

PR #8437, opened 2025-08-06
Add action to abandon a world

ThanatosGit:
Simple solution to abandon a world by collecting all locations.
Should be replaced in the future whenever we have a logical release of items.

Should there be some sort of state indicating that it was abandonend? That way it would already be in preperation for a logical release and we could mark it somehow in the ui as abandoned.

image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [10.1.0] - 2025-09-0?
 
+- Added: Action to abandon a world in a multiworld session.
+
 ### Metroid Dread
 
 #### Logic Database
--- a/randovania/gui/lib/multiplayer_session_api.py
+++ b/randovania/gui/lib/multiplayer_session_api.py
@@ -240,6 +240,11 @@ class MultiplayerSessionApi(QtCore.QObject):
         await self._session_admin_player(owner, admin_actions.SessionAdminUserAction.UNCLAIM, str(world_uid))
 
     @handle_network_errors
+    async def abandon_world(self, world_uid: uuid.UUID, owner: int) -> None:
+        self.logger.info("Abandon World %s from %d", world_uid, owner)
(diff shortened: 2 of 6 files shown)


ThanatosGit:
Looks like all the mypy erros on line `await self._session_api.abandon_world(world_uid, owner)` are caused by the `@handle_network_errors` annotation.

Should be generally fixed in a different PR.

MayberryZoom:
> Should there be some sort of state indicating that it was abandonend? That way it would already be in preperation for a logical release and we could mark it somehow in the ui as abandoned.

That seems useful to me, because I think it would be good to have abandoned worlds be claimable once this feature is fully fleshed out. If they're just waiting there periodically releasing items, I don't see any reason why they wouldn't be able to.

This could become a part of the "unclaimed" state, but there's probably a use case for having an unclaimed world not releasing items, so a new separate one makes more sense to me. And right now, there's no point in an abandoned world being claimable, so it's still useful for this intermediate implementation too.

henriquegemignani:
If you add this, then there's no point into having the fancier logical release. It's not how people end up using features.

The logical release isn't even hard to implement. The main blockers imo is that I wanted some fancier UI to indicate there's cpu intensive tasks being done by Randovania. Or some sort of task queue you can join to participate. Honestly I've been feature creeping it and procastinating instead of working on it.

ThanatosGit:
> If you add this, then there's no point into having the fancier logical release. It's not how people end up using features.

The fancier feature will nuke this implementation.
The users will have no choice.

I see this just as an intermediate solution and added it this way because it is easy to do and was asked about a handful of times.

henriquegemignani:
> The fancier feature will nuke this implementation. The users will have no choice.

I don't want to cause another outrage of users again when we do that. Sorry, rather not have it than remove it later.

JeffGainsNGames:
I feel like prior wording in this discussion strays from logical release being an *enhancement* to this PR, rather than removing a feature? If the outcome is the same (people can release games to not brick MW) I would hope people can not be frustrated with the method changing if the features goes from "release all now" to "release logically slowly", especially when most seem more concerned with the issue this aims to solve, rather than the method of solving it.

PR #8444, opened 2025-08-10
Refactor/change location field in layout description

Miepee:
Needs tests for hunters still fixed

### DIFF
--- a/randovania/layout/game_patches_serializer.py
+++ b/randovania/layout/game_patches_serializer.py
@@ -31,7 +31,8 @@ _NOTHING_PICKUP_NAME = "Nothing"
 def _pickup_assignment_to_pickup_locations(
     region_list: RegionList,
     pickup_assignment: PickupAssignment,
-    num_players: int,
+    current_world: int,
+    num_worlds: int,
 ) -> list[dict]:
     pickup_locations = []
 
@@ -42,13 +43,13 @@ def _pickup_assignment_to_pickup_locations(
         if node.pickup_index in pickup_assignment:
             target = pickup_assignment[node.pickup_index]
             item_name = target.pickup.name
-            if num_players > 1:
+            if num_worlds > 1:
                 target_world = target.player
             else:
                 target_world = 0
         else:
             item_name = _NOTHING_PICKUP_NAME
-            target_world = 0
+            target_world = current_world
 
         pickup_locations.append(
             {
@@ -72,11 +73,11 @@ def _find_area_with_teleporter(region_list: RegionList, teleporter: NodeIdentifi
     return region_list.area_by_area_location(teleporter.area_identifier)
 
 
-def serialize_single(player_index: int, num_players: int, patches: GamePatches) -> dict:
+def serialize_single(world_index: int, num_worlds: int, patches: GamePatches) -> dict:
     """
     Encodes a given GamePatches into a JSON-serializable dict.
-    :param player_index:
-    :param num_players:
+    :param world_index:
+    :param num_worlds:
     :param patches:
     :return:
     """
@@ -116,7 +117,9 @@ def serialize_single(player_index: int, num_players: int, patches: GamePatches) 
             }
             for dock, weakness in patches.all_dock_weaknesses()
         },
-        "locations": _pickup_assignment_to_pickup_locations(region_list, patches.pickup_assignment, num_players),
+        "locations": _pickup_assignment_to_pickup_locations(
+            region_list, patches.pickup_assignment, world_index, num_worlds
+        ),
         "hints": {identifier.as_string: hint.as_json for identifier, hint in patches.hints.items()},
         "game_specific": patches.game_specific,
     }
--- a/randovania/layout/description_migration.py
+++ b/randovania/layout/description_migration.py
@@ -727,7 +727,7 @@ def _migrate_v36(data: dict) -> None:
 def _migrate_v37(data: dict) -> None:
     assignment_re = re.compile(r"(.*) for Player (\d+)")
 
-    for game in data["game_modifications"]:
+    for world, game in enumerate(data["game_modifications"]):
         new_locations = []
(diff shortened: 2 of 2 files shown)


Miepee:
fixed hunters

henriquegemignani (changes requested):
No, I won't accept ints-as-strings for keys in the rdvgame file.

Miepee:
Right, forgot about that.
I initially had node_identifier.as_string as the key, but it felt really redundant since we also save .as_json as a field
I also tried having it as a list of dicts, but there were a few things where stuff then exploded because the order mattered and I didnt feel like trying to hack that out.

So I suppose as_string it is.

Miepee:
I didn't like as_string due to the duplication and tested out again how much actually breaks by having it as a list.
Turns out, we only had a few checks which checked for equality via old_rdvgame_dict == round_trip_rdvgame_dict
So I changed those tests to then exclude the list and compare equality on that one differently.
Feels cleaner, but lemme know what you think.

randovania/layout/game_patches_serializer.py
```diff
@@ -139,41 +149,33 @@ def _decode_pickup_assignment(
...
-            else:
-                pickup_name = target_name
-                target_player = 0
+        target_player = location["owner"]
 
-            node_identifier = NodeIdentifier.create(world_name, *area_node_name.split("/", 1))
-            node = region_list.typed_node_by_identifier(node_identifier, PickupNode)
-            pickup: PickupEntry | None
+        node_identifier = NodeIdentifier.from_json(location["node_identifier"])
+        node = region_list.typed_node_by_identifier(node_identifier, PickupNode)
```
Miepee: should I assert here that node == location["index"] ?

henriquegemignani: I'd actually just grab the node from the pickup index. Actually doesn't even need to care about grabbing the node.

Miepee: question then still remains tho, should we ensure that when loading the layout, that the specified index and identifier are the same? Or should stuff like
```json
{
  "node_identifier": {
    "region": "Great Temple",
    "area": "Electric",
    "node": "Boogaloo"
  },
  "index": 1
}
```
pass?

henriquegemignani: I would rather it just doesn't care, but maybe it's bad for people messing around? But I also would prefer no one to ever edit this file manually...

Miepee: yeah afaik its not toooo uncommon that people mess with it manually, and i'd rather have a sensible failsafe there then just in case someone does mess around with it.

randovania/layout/game_patches_serializer.py
```diff
@@ -34,26 +32,40 @@ def _pickup_assignment_to_item_locations(
...
+            owner = 0
 
-        # Not using `node.full_name` as this is not a user-facing string and needs to be consistent
-        items_locations[region.name][f"{node.identifier.area}/{node.identifier.node}"] = item_name
+        items_locations.append(
+            {
+                "node_identifier": node.identifier.as_json,
+                "index": node.pickup_index.index,
+                "pickup": item_name,
+                "owner": owner,
```
henriquegemignani: I'm torn between `owner` and `world`

randovania/layout/game_patches_serializer.py
```diff
@@ -34,26 +32,40 @@ def _pickup_assignment_to_item_locations(
     region_list: RegionList,
     pickup_assignment: PickupAssignment,
     num_players: int,
-) -> dict[str, dict[str, str]]:
-    items_locations: collections.defaultdict[str, dict[str, str]] = collections.defaultdict(dict)
+) -> list[dict]:
+    items_locations = []
```
henriquegemignani: If we're fixing things, this should be `pickup` instead of items. Same for the function name

randovania/layout/game_patches_serializer.py
```diff
@@ -139,41 +149,40 @@ def _decode_pickup_assignment(
...
+            continue
 
-            pickup_name_match = target_name_re.match(target_name)
-            if pickup_name_match is not None:
-                pickup_name = pickup_name_match.group(1)
-                target_player = int(pickup_name_match.group(2)) - 1
-            else:
-                pickup_name = target_name
-                target_player = 0
+        target_player = location["owner"]
```
henriquegemignani: `target_world`

randovania/layout/game_patches_serializer.py
```diff
@@ -139,41 +149,40 @@ def _decode_pickup_assignment(
...
 
-            node_identifier = NodeIdentifier.create(world_name, *area_node_name.split("/", 1))
-            node = region_list.typed_node_by_identifier(node_identifier, PickupNode)
-            pickup: PickupEntry | None
+        node_identifier = NodeIdentifier.from_json(location["node_identifier"])
+        node = region_list.typed_node_by_identifier(node_identifier, PickupNode)
```
henriquegemignani: Grab node from `pickup_index`, then check if the identifier of that node matches the one in the json

randovania/layout/game_patches_serializer.py
```diff
@@ -139,41 +149,40 @@ def _decode_pickup_assignment(
...
+
+        elif pickup_name != _NOTHING_PICKUP_NAME:
+            pickup = _get_pickup_from_pool(all_pools[target_player].to_place, pickup_name)
+        else:
+            pickup = None
 
-            if pickup is not None:
-                pickup_assignment[node.pickup_index] = PickupTarget(pickup, target_player)
+        if pickup is not None:
+            pickup_assignment[pickup_index] = PickupTarget(pickup, target_player)
```
henriquegemignani: If we're catching user errors, let's catch two entries for the same index.

randovania/layout/game_patches_serializer.py
Miepee: `world` sounds a bit meh for me. I would've preferred `target_world` over it, but that's longer than `owner`

PR #8446, opened 2025-08-10
Dread/update describer

Tyranisaur:
I thought it would be relevant to add test cases that would exercise the required mains descriptions, but it didn't look straight forward to change that part of the configuration from the existing test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added: The preset summary now lists when Flash Shift can be used with Flash Shift Upgrades can be used without the Main Flash Shift.
 - Changed: The preset summary now lists when Main Power Bomb is not required to use Power Bombs instead of the opposite.
+- Changed: The preset summary lists when Power Bomb limitations is disabled instead of when it's enabled.
 
 #### Logic Database
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Dread
 
+- Added: The preset summary now lists when Flash Shift can be used with Flash Shift Upgrades can be used without the Main Flash Shift.
+- Changed: The preset summary now lists when Main Power Bomb is not required to use Power Bombs instead of the opposite.
+
 #### Logic Database
 
 ##### Artaria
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Dread
 
+- Added: The preset summary now lists when Flash Shift can be used with Flash Shift Upgrades can be used without the Main Flash Shift.
```
Miepee: ?

randovania/layout/preset_describer.py
```diff
@@ -224,22 +224,27 @@ def progressive_items(self) -> ProgressiveItemTuples:
...
+            result[index] = ammo_item_default_main_required[index][1] != state.requires_main_item
 
     return result
 
 
-def message_for_required_mains(ammo_configuration: AmmoPickupConfiguration, message_to_item: dict[str, str]) -> dict:
-    item_names = list(message_to_item.values())
-    main_required = _require_majors_check(ammo_configuration, item_names)
-    return dict(zip(message_to_item.keys(), main_required))
+def message_for_required_mains(
```
henriquegemignani, line 241: I don't enjoy how the API ended up very awkward to use. Current use cases and hopefully future ones as well are all False/True for a given game, so we could move that flag to an additional argument (or different methods!)

henriquegemignani (changes requested):

henriquegemignani:
And could you add a better PR title? 😅

randovania/layout/preset_describer.py
Tyranisaur: I changed it to take the bool as a function level argument instead, and I took away the other parts of the refactor in a rebase.

randovania/games/cave_story/layout/preset_describer.py
```diff
@@ -33,7 +33,9 @@ def format_params(self, configuration: BaseConfiguration) -> dict[str, list[str]
             ],
             "Game Changes": [
                 message_for_required_mains(
-                    configuration.ammo_pickup_configuration, {"Missiles need main Launcher": "Missile Expansion"}
+                    configuration.ammo_pickup_configuration,
+                    {"Missiles need main Launcher": "Missile Expansion"},
+                    False,
```
henriquegemignani: Passing this bool via kwargs would make it more readable imo

Tyranisaur: Do you mean as a named function argument? Because I'm not familiar enough with kwargs to see how that machinery is applicable here.

henriquegemignani: Just passing the calls by naming the argument

Tyranisaur: Done. I fixuped this change since I also did the same to fix the way I forgot to change dread to use mains required by default, which was detected this time because I had fixed the Expansion vs Tank issue.

henriquegemignani (approved):

PR #8451, opened 2025-08-12
Prime1 Custom Items After Patcher Update

M4RKERSM4SK:
Resolved and merged conflicts and got this back up and running. Rebased again after toaster made a bunch of changes for the randomprime update awhile ago.

Everything works still in solo
- Required mains works
- Unlimited PBs and Missiles works

Multiworld still broken, I dont understand the prime1 connector really at all. Will need some real help on that part.


randovania/game_connection/connector/prime1_remote_connector.py
```diff
@@ -55,6 +59,31 @@ def _asset_id_format(self) -> str:
...
+    async def get_inventory(self) -> Inventory:
+        """Fetches the inventory represented by the given game memory."""
+
+        memory_ops = await self._memory_op_for_items(
+            [item for item in self.game.resource_database.item if item.extra["item_id"] < 1000]
+        )
+        ops_result = await self.executor.perform_memory_operations(memory_ops)
+
+        inventory = {}
+        for item, memory_op in zip(self.game.resource_database.item, memory_ops):
```
henriquegemignani, line 71: Please call `super().get_inventory()` instead of duplicating it's code here.

henriquegemignani (changes requested):

randovania/game_connection/connector/prime1_remote_connector.py
```diff
@@ -112,6 +141,16 @@ async def _patches_for_pickup(self, provider_name: str, pickup: PickupEntry, inv
             if delta == 0:
                 continue
 
+            if item.extra.get("unk2_bitmask_value"):
```
UltiNaruto, line 144: Use directly the item index instead of using Unknown Item 2

randovania/games/prime1/logic_database/header.json
```diff
@@ -290,25 +290,43 @@
...
                 }
             },
-            "MainPB": {
-                "long_name": "Main Power Bomb",
+            "UnlimitedMissiles": {
+                "long_name": "Unlimited Missiles",
                 "max_capacity": 1,
                 "extra": {
-                    "item_id": 1005
+                    "item_id": 1041,
```
UltiNaruto, line 297: Subtract 1000, you can check if `extras["unk2_bit_maskvalue"]` exists if you need a special handle case

randovania/games/prime1/logic_database/header.json
```diff
@@ -290,25 +290,43 @@
...
             },
-            "LockedPB": {
-                "long_name": "Locked Power Bomb",
-                "max_capacity": 99,
+            "UnlimitedPowerBombs": {
+                "long_name": "Unlimited Power Bombs",
+                "max_capacity": 1,
                 "extra": {
-                    "item_id": 1006
+                    "item_id": 1042,
```
UltiNaruto, line 305: Same as Unlimited Missiles

randovania/games/prime1/logic_database/header.json
```diff
@@ -290,25 +290,43 @@
...
+                    "unk2_bitmask_value": 2
                 }
             },
             "MissileLauncher": {
                 "long_name": "Missile Launcher",
                 "max_capacity": 1,
                 "extra": {
-                    "item_id": 1007
+                    "item_id": 1043,
```
UltiNaruto, line 313: Same as Unlimited Missiles

randovania/games/prime1/logic_database/header.json
```diff
@@ -290,25 +290,43 @@
...
+                    "unk2_bitmask_value": 4
+                }
+            },
+            "MainPB": {
+                "long_name": "Main Power Bomb",
+                "max_capacity": 1,
+                "extra": {
+                    "item_id": 1044,
```
UltiNaruto, line 321: Same as Unlimited Missiles

randovania/games/prime1/logic_database/header.json
henriquegemignani: Actually it's a good opportunity to change things so we don't have this magic rule that "item id above 1000 isn't actually an item". The correct thing is to just not have an item id in that case.

Some fixes will be needed in `PrimeRemoteConnector` but it's better than expanding the hack.

UltiNaruto: The item index is still needed since EPlayerItem has been extended with custom items, we're calling IncrPickup with the new item index and the game handles everything instead of doing it the hacky way by setting unknown item 2

M4RKERSM4SK:
After looking at these reviews and also shin's connector code:

The issue is that whatever is being passed from the MW connector to the prime world is wrong. It will play the hudmemo of unlimited power bombs revieved for example but nothing happens actually in game or its a missile instead of unlimited missles etc.

- What is unknown2 to the MW connector? Was unknown2 just an arbitraty spot to place a value to/from the patcher to do something based on that value in unknown2?

- Does unknown2 still even need to be used? I am looking at the patcher, and now there is actual item IDs to use for these special items. However in terms of `get_inventory` on line 77 it looks to me these new items are sent across this unkown2 item possibly? I am currious at all if we even need to have the get_inventory changes in here at all if unknow2 is not used as the bus anymore. 

- The 1000 subtraction takes place on line 150, is the connector or the header supposed to handle what the actual itemID value is here?

UltiNaruto:
- unknown2 holds the custom items you acquired, determined by bit flags (& 1 for missile launcher and so on)

- unknown2 is used only when reading since you can't call vanilla function and wait for the return value. btw on writing the custom item you use IncrPickup just like any normal pickup with the normal item ID

- just subtract any of item IDs in the item database that are made for custom items that have `unk2_bit_maskvalue`. it makes no sense to use >1000 IDs since we got `unk2_bit_maskvalue` for reading and we don't differentiate custom items for writing.

randovania/game_connection/connector/prime1_remote_connector.py
```diff
@@ -55,6 +59,31 @@ def _asset_id_format(self) -> str:
...
+            inv = InventoryItem(*struct.unpack(">II", ops_result[memory_op]))
+            if (inv.amount > inv.capacity or inv.capacity > item.max_capacity) and (item != self.multiworld_magic_item):
+                raise MemoryOperationException(f"Received {inv} for {item.long_name}, which is an invalid state.")
+            inventory[item] = inv
+
+        for item in self.game.resource_database.item:
+            if item.extra.get("unk2_bitmask_value"):
+                unknown2 = inventory.get(self.game.resource_database.get_item("Unknown2"))
+                if unknown2:
+                    item_value = (unknown2.capacity & item.extra["unk2_bitmask_value"]) >= 1
```
henriquegemignani, line 81: I personally don't like using a boolean somewhere that wants an int, even if booleans are ints...

JeffGainsNGames:
Link to thread for visibility https://discord.com/channels/914291389293027329/1297328657181835334

PR #8453, opened 2025-08-12
Space_Jump_Room_Logic_Update

Logarhythmic:
Updated logic for space jump room in the under water section. 

Added grapple movement to trick, magnet, flash, and slide jump as ways out. videos included.


CHANGELOG.md
```diff
@@ -31,6 +31,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Flipper Room: The Morph Ball Launcher now connects to the door to Navigation Station and is trivial, instead ot connecting to the door to Elun Transport Access and requiring that the Flipper has been rotated.
 - Changed: Flipper Room: Spinning the Flipper with the ledge warp from below now connects from the door to Elun Transport Access instead of from the Tunnel to Spider Magnet Elevator. It is now also logical to perform the ledge warp from below when Transport Randomizer is disabled.
 
+##### Ferenia
+
+- Fixed: Updated logic from Underwater Bottom in Space Jump Room to include more ways to reach Space Jump Room Access.
```
Miepee: If you just include more ways, thats an "added: "
"Fixed: " is if you change something existing that was broken/faulty

randovania/games/dread/logic_database/Ferenia.txt
```diff
@@ -800,10 +800,16 @@ Extra - asset_id: collision_camera_012
...
           Any of the following:
-              Grapple Beam
+              Spider Magnet
               All of the following:
                   Gravity Suit
-                  Space Jump or Simple IBJ
+                  Any of the following:
+                      Spin Boost or Simple IBJ
+                      # Slide Jump - https://youtu.be/3mfzhqqZBlc
+                      Single-wall Wall Jump (Beginner) and Slide Jump (Beginner)
```
Tyranisaur: So we deliberately increased all the single wall jumps because we don't consider this as a beginner trick. This should be intermediate.

It seems reasonable to use grapple without the grapple movement trick+ slide jump here.

randovania/games/dread/logic_database/Ferenia.txt
```diff
@@ -800,10 +800,16 @@ Extra - asset_id: collision_camera_012
...
+                      # Beginner Wall Jump - https://youtu.be/iHFyp9ON-pM
+                      Flash Shift and Wall Jump (Beginner)
```
Tyranisaur, line 813: I'm not necessarily opposed to having the wall jump on the flash shift here, but I feel like I should point out that it appears common practice to have a single wall jump combined with a flash shift as trickless in a bunch of places, including going to the left ledge in this room.

Tyranisaur (commented):
Something tells me it would be more reasonable to connect this via the right ledge instead of having it directly from the bottom.

randovania/games/dread/logic_database/Ferenia.txt
Logarhythmic: I added the grapple movement trick at the request of Xeno. I'm not sure what you mean with the slide jump.

Tyranisaur: Should be reasonable to use a trickless grapple to replace the single wall jump, still using slide jump to get from the first ledge to the top.

Logarhythmic: image

Tyranisaur (approved):

PR #8454, opened 2025-08-14
Add more joke hints

Miepee:


dyceron (commented):
I feel like these might be better as generic hints. They dont really fit the theme for Fusion.

randovania/games/fusion/exporter/joke_hints.py
```diff
@@ -53,4 +53,9 @@
     "Don't know what to do? Zeela vie.",
     "If something gets in your way, Choot it!",
     "Sometimes, it may not be possible to collect Evir-thing",
+    "You humans have some funny little brains. How do you get around in those things?",
+    "This hint isn't useful enough for you? Well, excuuuuuse me, princess!",
```
dyceron, line 57: Oops I didnt select them. These lines

Miepee: l 56 was chosen because adam is a computer (thus non human). Reading the hint in msr would feel a tad weird IMO and in CS definitely since quote is a robot.

l 57 was because samus does have the nickname of "princess" (especially in other m). it'd feel a bit weird IMO in CS where quote is "male", but I also don't know how common the nickname/teasing of calling someone entitled a princess is in english.

dyceron: Valid reasons. I'm good with this pr then

dyceron (approved):

PR #8456, opened 2025-08-14
IDE: always regenerate UI files when launching

### LINKED ISSUE #8450: Dev gui work no longer reflecting changes
Used to be able to edit the ui files and then just run\build with the changes relected during dev work

Can no longer see the changes reflected in the gui while doing dev work

I think this might have something to due with how uic is getting handled

To try and make sure it wasnt on my side I tried
- reinstalling QT creator
- whole brand new local repo
- new virtual env
- made sure nothing was open RDV wise before restart/building
- also ran uv run tools/uic_wrapper.py

Image

Miepee:
fixes #8450

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -9,6 +9,23 @@
             "command": "uv sync --all-extras --all-groups"
         },
         {
+            "label": "Regenerate UI",
+            "type": "shell",
+            "linux": {
+                "command": "${command:python.interpreterPath} '${workspaceFolder}/tools/uic_wrapper.py' '${workspaceFolder}' --all-ui-files"
+            },
+            "windows": {
+                "command": "& '${command:python.interpreterPath}' '${workspaceFolder}/tools/uic_wrapper.py' '${workspaceFolder}' --all-ui-files"
+            }
+        },
+        {
+            "label": "Launch Prerequisites",
+            "dependsOn": [
+                "Create venv with all exporters",
+                "Regenerate UI"
+            ]
+        },
+        {
             "label": "Run all tests on all cores",
             "type": "shell",
             "linux": {
--- a/.idea/runConfigurations/Regenerate_UI_files.xml
+++ b/.idea/runConfigurations/Regenerate_UI_files.xml
@@ -1,27 +1,24 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="GUI" type="PythonConfigurationType" factoryName="Python">
+  <configuration default="false" name="Regenerate UI files" type="PythonConfigurationType" factoryName="Python">
     <module name="randovania" />
     <option name="ENV_FILES" value="" />
     <option name="INTERPRETER_OPTIONS" value="" />
     <option name="PARENT_ENVS" value="true" />
     <envs>
       <env name="PYTHONUNBUFFERED" value="1" />
-      <env name="RUST_BACKTRACE" value="1" />
     </envs>
     <option name="SDK_HOME" value="" />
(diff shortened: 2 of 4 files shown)


.vscode/tasks.json
```diff
@@ -8,6 +8,23 @@
             "type": "shell",
             "command": "uv sync --all-extras --all-groups"
         },
+        {
+            "label": "Regenerate UI",
```
Miepee, line 12: This task doesnt quite work in vscode.
It does launch the script fine, but the call to pyside6-uic either fails or does nothing, i don't know which.

@ThanatosGit you wouldn't happen to have an idea as to why?

.vscode/tasks.json
```diff
@@ -8,6 +8,23 @@
...
+            "type": "shell",
+            "linux": {
+                "command": "${command:python.interpreterPath} '${workspaceFolder}/tools/uic_wrapper.py' '${workspaceFolder}' --all-ui-files"
```
henriquegemignani: This could just be `pre-commit run qtui -a`. Wouldn't even need to change the wrapper file too.

Miepee: that'd require pre-commit to be locally installed which might not always be the case right now i think?
but also do we want to always git add them when compiling? ive had some annoyances with that earlier

henriquegemignani: pre-commit should always be available, if not it's an uv call away

Miepee: Running pre-commit here is actually somewhat annoying due to 2 reasons:
- workflows where you don't (want to) stage your changes until you're ready to commit. This will cause no ui updates while you're trying to iterate on the ui.
- git clients which don't sync their staging area with the one from git

.vscode/tasks.json
ThanatosGit: The `uic_wrapper.py` should be fixed. The `try` `except` in the `uic_wrapper.py` is wrong. It uses a `ThreadPoolExecutor` which does not raise the exception to the outer thing. It would only do if you call `result()` on it (but IIRC that would wait for the `Future` to be resolved and make it sequential again).
Also if the file itself which you want to execute cannot be found it throws a `FileNotFoundError` and not a `subprocess.CalledProcessError`. The later will only be thrown if the found executable returns a non zero exit code.
I maybe should have read that file more carefully after the `Improve performance` commit. The burden of only read the differences after the initial review. 🙃

It probably fails because it cannot be found.
Dunno how Paths are set for all of these tools like pre-commit calling uv etc.
The easiest would be to just add it, I assume:
```
            "options": {
                "env": {
                "PATH": "${workspaceFolder}/.venv/Scripts;${env:PATH}"
                }
            },
```

Miepee: > It probably fails because it cannot be found.

Yup, that was it.
I've made it so it calls `result()` when the Future is done just to save anyone in the future more headaches.

henriquegemignani (approved):
There's gotta be a better way of handling generating python from the ui files that doesn't involve generating every file every time. Especially because this can cause annoyances with different Qt Creator versions as happened in #8493.

I guess we could maybe check modification date, if that's trustworthy. And maybe exclude the Qt Creator version from the generated file just in case too.

Another PR's job I guess.

Miepee:
Honestly i'm surprised that there is no way to pin the qt creator version that pyside6-uic uses. It'd still cause mass-changed files on a bump, but at least people couldn't do it accidentally.

I'll probably try in the future to just exclude the qt creator version from it.

henriquegemignani:
It'll use the one you have installed, I guess? Unless it's from the PySide version which then we could make sure to pin. I think I ignored it, which would be a problem.

PR #8461, opened 2025-08-15
Fusion Updates to Breeding Tank

M4RKERSM4SK:
Changes made
- Added in a trick with Space Jump and advanced movement in order to cross the tank without touching the water
- Add in the UWJ Videos that were missing
- Updated the Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ##### Sector 4 (AQA)
 
 - Fixed: Security Access - Fixed a trick video appearing incorrectly on the website.
+- Added: Breeding Tank - Added a movement advanced trick with Space Jump to cross the tank
 
 ##### Sector 6 (NOC)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ##### Sector 4 (AQA)
 
 - Fixed: Security Access - Fixed a trick video appearing incorrectly on the website.
-- Added: Breeding Tank - Added a movement advanced trick with Space Jump to cross the tank
+- Added: Breeding Tank - Added an advanced movement trick with Space Jump to cross the tank
 
 ##### Sector 6 (NOC)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed: Security Access - Fixed a trick video appearing incorrectly on the website.
 - Added: Breeding Tank - Added an advanced movement trick with Space Jump to cross the tank
+- Added: Breeding Tank - Added the Underwater Wall Jump trick videos
 
 ##### Sector 6 (NOC)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ##### Sector 4 (AQA)
 
 - Fixed: Security Access - Fixed a trick video appearing incorrectly on the website.
+- Added: Breeding Tank - Added an advanced movement trick with Space Jump to cross the tank
```
Miepee: I think this is now the 4th way i see of displaying a logic change in a specific room
Feels like we should at one point standardize that.

randovania/games/fusion/logic_database/Sector 4 AQA.txt
```diff
@@ -1552,14 +1552,22 @@ Extra - minimap_coordinates: [{'x': 7, 'y': 2}, {'x': 8, 'y': 2}]
   * L0 Hatch to Breeding Tank Access/Door to Breeding Tank
   * Extra - door_idx: (70,)
   > Door to Broken Bridge
-      Gravity Suit or Hi-Jump or Underwater Wall Jump (Intermediate)
+      Any of the following:
+          # Underwater Wall Jump: https://youtu.be/0xdEARW5qvU
+          Gravity Suit or Hi-Jump or Underwater Wall Jump (Intermediate)
+          # Space Jump Across: https://youtu.be/LFlbn5XPwCs
+          Space Jump and Movement (Advanced)
```
Miepee, line 1559: Thoughts on bumping the SJ crossing to expert due to the fact that if you fail it you (logically) have to reset and cant attempt again?
I haven't attempted the trick in-game to know how difficult its execution actually is.

zarakava: I've tried it a couple times and it is a bit tricky but not terrible. I don't know if HiJump affects it much. There is a save room immediately before as well. I do think Advanced is fine

CHANGELOG.md
M4RKERSM4SK: I was trying to just match what I saw. If there is a standard for changelog stuff that would be cool

randovania/games/fusion/logic_database/Sector 4 AQA.txt
M4RKERSM4SK: I also felt advanced was close to appropriate.
- Save station near by
- Can practice to get consistient
- Execution wise, not too tight on button presses. Didnt seem extra precise, but do need good timing

Understand the softlock risk, and I know this is apples to oranges but some of the advanced Shinesparks are also advanced based on execution and not on the reset risk. (aka Atmospheric stablizer NW)

If we need to consider reset risk into expert, I think it should be consistent across the board then

CHANGELOG.md
Miepee: Historically we've always had `- <type>: <room name>: <change>`
E.g. `- Fixed: Metroid Quarantine A: Getting the Missile Expansion via using the Spider Ball Track now requires Morph Ball.`

But probably fine to just leave it as is for now

Miepee (approved):

CHANGELOG.md
Miepee:
```suggestion
- Added: Breeding Tank - Added an Advanced Movement trick with Space Jump to cross the tank
```

PR #8464, opened 2025-08-16
Create WorldGraph

henriquegemignani:
In order to prevent this from dying, I'm proposing merging this to main as code that doesn't get used. Fun!

Actually it'll be used with a command line flag and just for the resolver, because that's all the code that got written. This and some tests failing is why this is draft. Still appreciate reviews.


randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+class NodeConnection(typing.NamedTuple):
+    target: WorldGraphNode
+    requirement: Requirement
+    requirement_without_leaving: Requirement
+
+
+@dataclasses.dataclass(slots=True)
+class WorldGraphNode:
```
Miepee, line 64: i think the docstrings are a bit messed up in this class? but also incomplete

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+
+    """"""
+    identifier: NodeIdentifier
+
+    """If passing by this node should fully heal."""
+    heal: bool
+
+    """
+    Which nodes can be reached from this one and the requirements for so.
+    Includes in-area connections, Dock connections, TeleporterNetwork connections and requirement_to_leave.
```
Miepee: so basically, includes every node, regardless of how its being accessed?

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+    """
+    resource_gain: list[ResourceQuantity]
+
+    """
+    A requirement that must be satisfied before being able to collect
+    """
+    requirement_to_collect: Requirement
+
+    """When set, leaving this node requires it to have been collected."""
+    require_collected_to_leave: bool
```
Miepee, line 95: I don't quite get what this is for / its usecase is?
Why do we need event/pickup nodes to be collected before leaving? Why do we not need that for teleporters/hints?

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+
+    """The pickup index associated with this node."""
+    pickup_index: PickupIndex | None
+
+    """"""
+    pickup_entry: PickupEntry | None
+
+    """"""
+    original_node: Node
```
Miepee: do we not have a class like a nodeidentifier that's storing Region/Area/Node instead of their string names?

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+    original_area: Area
+    original_region: Region
+
+    def is_resource_node(self) -> bool:
+        return len(self.resource_gain) > 0
+
+    def should_collect(self, context: NodeContext) -> bool:
```
Miepee, line 123: aren't should_collect and has_all_resources pretty much the same but negated?

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+    def has_all_resources(self, context: NodeContext) -> bool:
+        for resource, _ in self.resource_gain:
+            if not context.has_resource(resource):
+                return False
+        return True
+
+    def resource_gain_on_collect(self, context: NodeContext) -> ResourceGain:
+        yield from self.resource_gain
+        if self.pickup_entry is not None:
+            yield from self.pickup_entry.resource_gain(context.current_resources, force_lock=True)
```
Miepee, line 146: i thought `self.resource_gain` includes the pickup entries resources if it is one?

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+    nodes: list[WorldGraphNode]
+    node_provider: WorldGraphNodeProvider
+    node_by_pickup_index: dict[PickupIndex, WorldGraphNode]
+
+    def victory_condition_as_set(self, context: NodeContext) -> RequirementSet:
+        # TODO: calculate this just once
+        return self.victory_condition.as_set(context)
+
+
+def _get_dock_open_requirement(node: DockNode, weakness: DockWeakness) -> Requirement:
```
Miepee, line 205: note for me to see later why you can't just use the same functions in DockNode

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+        node.resource_gain.append((front_lock_resource, 1))
+        requirement_parts.append(ResourceRequirement.simple(front_lock_resource))
+        requirement_to_collect = _get_dock_lock_requirement(node.original_node, forward_weakness)
+
+    # Handle the different kinds of ways a dock lock can be opened from behind
+
+    if back_lock is not None:
+        back_lock_resource = NodeResourceInfo.from_node(target_node, context)
+
+        match back_lock.lock_type:
```
Miepee: not quite getting all of the cases here. E.g. FRONT_BLAST_BACK_IMPOSSIBLE adds both the back lock requirement and the forward lock requirement?

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+    elif isinstance(node.original_node, TeleporterNetworkNode):
+        raise NotImplementedError
+
+    for target_original_node, requirement in node.original_area.connections[node.original_node].items():
+        target_node = original_to_node.get(target_original_node.node_index)
+        if target_node is None:
+            continue
+
+        requirement_including_leaving = requirement
+        requirement = requirement.patch_requirements(damage_multiplier, context).simplify()
```
Miepee: wouldn't you want to switch these around? first simplify, then assign?

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+
+def create_node(
+    node_index: int,
+    patches: GamePatches,
+    original_node: Node,
+    area: Area,
+    region: Region,
+) -> WorldGraphNode:
+    """
+    Creates one WorldGraphNode based on one original node.
```
Miepee, line 362: i really don't like the term "original node" for a Node :(
makes it sound too confusing for me.

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+
+        if original_node.is_derived_node:
+            node_replacement[original_node] = None
+
+    for region, area, original_node in database_view.node_iterator():
+        original_node = node_replacement.get(original_node, original_node)
+        if original_node is None:
+            continue
+
+        nodes.append(create_node(len(nodes), patches, original_node, area, region))
```
Miepee: why do we call `len(nodes)` instead of `original_node.node_index`?

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+    if pickup_index is not None:
+        target = patches.pickup_assignment.get(pickup_index)
+        if target is not None and target.player == patches.player_index:
+            pickup_entry = target.pickup
+
+    return WorldGraphNode(
+        node_index=node_index,
+        identifier=original_node.identifier,
+        heal=original_node.heal,
+        connections=[],  # to be filled later
```
Miepee: i would maybe change "to be filled later" by "to be filled by create_graph" or something. and also explain why we cant/dont fill it in here.

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,393 @@
...
+    from randovania.game_description.resources.pickup_index import PickupIndex
+    from randovania.game_description.resources.resource_collection import ResourceCollection
+    from randovania.game_description.resources.resource_info import (
+        ResourceGain,
+        ResourceInfo,
+        ResourceQuantity,
+    )
+
+
+class NodeConnection(typing.NamedTuple):
```
Miepee: unsure if this should be renamed to `WorldGraphNodeConnection` to make it clearer that it belongs to here rather than a "generic" thing

randovania/graph/state.py
```diff
@@ -129,23 +139,26 @@ def _advance_to(
...
         """
-
-        if not node.should_collect(self.node_context()):
-            raise ValueError(f"Trying to collect an uncollectable node'{node}'")
+        context = self.node_context()
+        # if not (
+        #     node.should_collect(context)
+        #     and node.requirement_to_collect.satisfied(context, damage_state.health_for_damage_requirements())
+        # ):
+        #     raise ValueError(f"Trying to collect an uncollectable node'{node}'")
```
Miepee: why was this commented out?

randovania/graph/state.py
```diff
@@ -78,26 +80,34 @@ def copy(self) -> Self:
...
+                if isinstance(graph, WorldGraph):
+                    node_index = resource.resource_index - self.resource_database.first_unused_resource_index()
+                    node = graph.nodes[node_index]
+                    if node.pickup_index is not None:
+                        yield node.pickup_index
+                else:
+                    node = graph.node_by_identifier(resource.node_identifier)
+                    if isinstance(node, PickupNode) and node.pickup_index is not None:
+                        yield node.pickup_index
+
```
Miepee, line 98: I assume this uglyness is temporary?

randovania/resolver/logic.py
```diff
@@ -16,36 +16,47 @@
...
+    from randovania.game_description.resources.resource_info import ResourceInfo
+    from randovania.graph.state import State
     from randovania.layout.base.base_configuration import BaseConfiguration
     from randovania.resolver.damage_state import DamageState
-    from randovania.resolver.state import State
 
 
-def n(node: Node, with_region: bool = True) -> str:
-    return node.full_name(with_region=with_region) if node is not None else "None"
+def n(node: WorldGraphNode | Node) -> str:
```
Miepee: don't see the usage here: why was with_region removed?

randovania/graph/world_graph.py
henriquegemignani: If collecting has a requirement and collecting is required for leaving, then you can't cross this node without that requirement.
Collection being required helps reduce the graph: otherwise you need two branches, one for collecting and one for not. Mostly relevant if it's a dangerous resource but there's random cases in graph walking where it's convenient to be able to assume.

Also imo it's best for database design too. If you can skip the resource, then skip the node.

Miepee: > If collecting has a requirement and collecting is required for leaving

The requirement for events/pickups is trivial tho?

> Also imo it's best for database design too. If you can skip the resource, then skip the node.

I once heard the opposite: that in order to reduce more unnecessary edges (and potentially nodes), that even if a resource node is trivial to walk around, that its better to force it to go through it. E.g. Door1 <-> Pickup <-> Door2 instead of Door1 <-> Generic <-> Door2 and Generic <-> Pickup

henriquegemignani: > The requirement for events/pickups is trivial tho?

That's why it gets set to True there. Or at least should be :p

> 
> > Also imo it's best for database design too. If you can skip the resource, then skip the node.
> 
> I once heard the opposite: that in order to reduce more unnecessary edges (and potentially nodes), that even if a resource node is trivial to walk around, that its better to force it to go through it. E.g. Door1 <-> Pickup <-> Door2 instead of Door1 <-> Generic <-> Door2 and Generic <-> Pickup

I don't remember suggesting avoiding extra nodes. And the extra connections I remember talking about is avoiding redundant connections, that'd duplicate requirements.

randovania/graph/state.py
henriquegemignani: I think it's failing 😅 
Gonna re-check it.

randovania/graph/world_graph.py
henriquegemignani: The only ways to leave this node will be present in this list. No method calls or anything.
The requirements are also present here in their final form, no processing needed.

It's one of the important changes for performance, the other being just one class that is in a way [plain old data](https://en.wikipedia.org/wiki/Passive_data_structure).

randovania/graph/world_graph.py
henriquegemignani: Don't think so. And I think at this point and level I'd rather have it flat for performance. (Though nothing performance critical should be using these fields, so maybe it _should_ be in a pointer)

randovania/graph/world_graph.py
henriquegemignani: At the moment, the ones in DockNode has a lot of caching which I just don't want to touch.

Though the reason is I want to delete all that code from these classes, and this likely comes from a version of the branch where that's true.

henriquegemignani: Ah and it uses NodeContext which I don't have, just so it can fallback to GamePatches, which I don't want to here.

randovania/graph/world_graph.py
Miepee: Ok. The way the second sentence was phrased made me unsure whether there are some leave-methods that are *not* in here. Would be nice to replace that with a shorter sentence to make it clearer

randovania/graph/world_graph.py
Miepee: I was mostly just thinking that we *do* already have such a structure. But if we don't then let's leave it as-is

randovania/graph/world_graph.py
Miepee: Ohm i forgot about this note lol.
Thanks for explanation.

randovania/graph/world_graph.py
Miepee: i remembered that the WGNode.index isn't guaranteed to match the original_node.index
still, I would assume at least that accessing the index would be faster than getting the length

randovania/resolver/logic.py
henriquegemignani: I believe it's intentional for never printing nodes without the region in the logs.

Miepee: as long as nothing called it with_region=False then cool

randovania/graph/world_graph.py
henriquegemignani: Yeah and we have the same structure at the moment and I can't remember why.

randovania/graph/world_graph.py
henriquegemignani: No because the resources a pickup gives depends on the current resources, due to progressives.

randovania/graph/world_graph.py
henriquegemignani: The simplify is to improve on patch_requirements results since it replaces resource requirements with Impossible/Trivial.

randovania/graph/world_graph.py
henriquegemignani: The intent is to make the WGNode index be continuous and have a list of all WGNodes not have a single hole.
Plus I'd be surprised if length of a python list isn't O(1)...

Miepee: > The intent is to make the WGNode index be continuous and have a list of all WGNodes not have a single hole.

What does that matter for?

henriquegemignani: iterating over all nodes without worrying about None
indexing the list of all nodes without mypy complaining it can be None

randovania/graph/world_graph.py
Miepee: Would it right now do "more work" simplifying if the requirement_to_leave is not trivial?
Since it:
- would store the original requirement in req_including_leaving
- simplify the original requirement
- add the req_to_leave to req_including_leaving and simplify, which simplifies both the req_to_leave and a copy of the original requirement

if you'd have
```py
requirement = requirement.patch_requirements(damage_multiplier, context).simplify()
requirement_including_leaving = requirement
```
Then it wouldn't have to "re-simplify" the original requirement if req_to_leave is nontrivial, no?

henriquegemignani: Maybe? I haven't been concerned with optimising this step fully at the moment.

Further steps is trying to have WorldGraph only use RequirementSet or similar. Aka, have a requirement data structure 100% focused on performance. I might have to soonish to solve issues with these NodeRequirement all over the place that ends up being very ugly code.

randovania/graph/world_graph.py
henriquegemignani: For all the locks, if it has already been unlocked then you can cross. What differs is what situations allows you to unlock from the back.

randovania/graph/world_graph.py
henriquegemignani: What do you suggest?

randovania/graph/world_graph.py
Miepee: is there a point in the meantime to do `should_collect(): return not self.is_resource_node()`?

randovania/graph/world_graph.py
Miepee: Both "classic Node" and "non-wg-node" sound kinda bad. 
Maybe "DBNode, DBArea and DBRegion"?

randovania/graph/state.py
henriquegemignani: Reason: requirement_to_collect changed from a method to an attribute. Fixed by making it an attribute in the original class.

randovania/graph/world_graph.py
henriquegemignani: I don't like abbreviations like that on names

Miepee: For me it depends, but in this case it feels clearer IMO.
But maybe explaining that in the docstring for hte original_* is good enough:tm:

randovania/graph/world_graph.py
henriquegemignani: what do you mean messed up?

Miepee: AFAIK the docstring needs to be below a property.
But then you have:
```py
    """The index of this node in WorldGraph.nodes, for quick reference. Does not necessarily match Node.node_index"""
    node_index: int
    """"""
    identifier: NodeIdentifier
```
Which I'm pretty sure will give node_index an empty docstring, and whatever is above node_index the description of what was meant for node_index

randovania/graph/state.py
henriquegemignani: Hopefully.

randovania/graph/world_graph.py
henriquegemignani: Calling them `database_node` (realized I did that in State), `area` and `region`.

randovania/graph/world_graph.py
henriquegemignani: Gonna leave it for later, in case TeleporterNetworkNode needs it.

randovania/graph/world_graph.py
henriquegemignani: Changed it.

randovania/graph/world_graph.py
henriquegemignani: fixed

randovania/graph/world_graph.py
Miepee: does it warrant a comment to revisit or nah?

henriquegemignani: I'll leave one

randovania/graph/state.py
```diff
@@ -150,11 +150,11 @@ def collect_resource_node(self, node: ResourceNode | WorldGraphNode, damage_stat
         :return:
         """
         context = self.node_context()
-        # if not (
-        #     node.should_collect(context)
-        #     and node.requirement_to_collect.satisfied(context, damage_state.health_for_damage_requirements())
-        # ):
-        #     raise ValueError(f"Trying to collect an uncollectable node'{node}'")
+        if not (
```
Miepee: why commented out?

randovania/graph/world_graph.py
```diff
@@ -369,14 +373,16 @@ def create_graph(
...
+    original_to_node = {node.database_node.node_index: node for node in nodes}
     node_provider = WorldGraphNodeProvider(database_view, original_to_node)
 
     context = NodeContext(patches, resources, database_view.get_resource_database_view(), node_provider)
     for node in nodes:
-        if isinstance(node.original_node, HintNode | PickupNode | EventPickupNode):
+        if isinstance(node.database_node, HintNode | PickupNode | EventPickupNode):
             node.resource_gain.append((node_provider.get_node_resource_info_for(node.identifier, context), 1))
 
+        node.require_collected_to_leave
```
Miepee: ?

randovania/generator/dock_weakness_distributor.py
```diff
@@ -163,6 +163,7 @@ def __init__(self, game: GameDescription, configuration: BaseConfiguration, dock
 
     @classmethod
     def from_logic(cls, logic: Logic, dock: DockNode, target: DockNode) -> Self:
+        assert logic.game is not None
```
Miepee, line 166: if this were None, then wouldn't the call to `from_logic` fail since the type is specified as Logic?

randovania/graph/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+This module focuses on providing the tools needed for performant graph searches.
```
Miepee: is this supposed to get more extended?

randovania/resolver/hint_state.py
```diff
@@ -19,10 +20,12 @@ def __copy__(self) -> Self:
         other.pickup_available_indices_when_collected = dict(self.pickup_available_indices_when_collected)
         return other
 
-    def valid_available_locations_for_hint(self, state: State, game: GameDescription) -> list[PickupIndex]:
-        collected = list(state.collected_pickup_indices)
+    def valid_available_locations_for_hint(self, state: State, logic: Logic) -> list[PickupIndex]:
+        collected = list(state.collected_pickup_indices(logic.graph or logic.game))
```
Miepee: even more mypy

pyproject.toml
duncathan, line 1: we should definitely add `randovania/graph` to mypy

randovania/graph/state.py
```diff
@@ -78,26 +79,34 @@ def copy(self) -> Self:
...
             if count > 0 and isinstance(resource, NodeResourceInfo):
-                node = resource.to_node(context)
-                if isinstance(node, PickupNode):
-                    yield node.pickup_index
-
-    @property
-    def collected_hints(self) -> Iterator[NodeIdentifier]:
-        context = self.node_context()
+                if isinstance(graph, WorldGraph):
+                    node_index = resource.resource_index - self.resource_database.first_unused_resource_index()
```
duncathan: if this is the canonical way of getting the node index from a NodeResourceInfo, it should probably be a method on that class

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,451 @@
...
+        ResourceGain,
+        ResourceInfo,
+        ResourceQuantity,
+    )
+
+
+class WorldGraphNodeConnection(typing.NamedTuple):
+    target: WorldGraphNode
+    requirement: Requirement
+    requirement_without_leaving: Requirement
```
duncathan, line 56: some quick docstrings here would be nice. at a glance it's not obvious to mii what the difference between `requirement` and `requirement_without_leaving` is

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,451 @@
...
+
+@dataclasses.dataclass(slots=True)
+class WorldGraphNode:
+    """A node of a WorldGraph. Focused on being a very efficient data structures for the resolver and generator."""
+
+    node_index: int
+    """The index of this node in WorldGraph.nodes, for quick reference. Does not necessarily match Node.node_index"""
+
+    identifier: NodeIdentifier
+    """The NodeIdentifier of `database_node`."""
```
duncathan: is there a reason this exists instead of using `database_node.identifier`? if so, should mention that in the docstring

randovania/graph/world_graph.py
```diff
@@ -0,0 +1,451 @@
...
+        back_lock = back_weakness.lock
+
+    # Requirements needed to break the lock. Both locks if relevant
+    assert node.requirement_to_collect == Requirement.trivial()  # DockNodes shouldn't have this set
+    requirement_to_collect = Requirement.trivial()
+
+    # Requirements needed to open and cross the dock.
+    requirement_parts = [_get_dock_open_requirement(node.database_node, forward_weakness)]
+
+    #
```
duncathan: ?

henriquegemignani:
> do I understand correctly that eventually the plan will be to remove the various code branches in the generator/resolver that still work with database nodes instead of graph nodes?

Yeah, after this there's the generator, and then DLR, then making it all the default and then delete old code.

randovania/graph/state.py
henriquegemignani: At the exact moment I don't have a clue what will be the correct way because a decent way involves significantly bigger changes, but a method in WorldGraph is a good start.

duncathan: fine by mii

randovania/generator/dock_weakness_distributor.py
henriquegemignani: I'm asserting `logic.game`, not `logic`?
But this assert does mean I can't use WorldGraph here.

Miepee: ah, didn't pay enough attention there, whoops

randovania/graph/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""
+This module focuses on providing the tools needed for performant graph searches, while also providing
+a place to have post-processing of the logic database.
+
+The main classes are:
+- WorldGraph
+- WorldGraphNode
+- State
```
Miepee: is it common for these places to get a quick tldr/one-sentence summary?

PR #8465, opened 2025-08-16
Dread: Fix oversight with ledge warping out of diffusion beam room

Tyranisaur:
The logic is meant to break the missile door, then set up the ledge warp, then roll into the room by opening the door with a bomb. However, it's important that the bomb is not placed on the diffusion beam pickup.

### DIFF
--- a/randovania/games/dread/logic_database/Cataris.txt
+++ b/randovania/games/dread/logic_database/Cataris.txt
@@ -3198,7 +3198,7 @@ Extra - asset_id: collision_camera_044
           Heat/Cold Runs (Intermediate) and Heat Damage ≥ 50
   > Tunnel to Teleport to Dairon
       # https://youtu.be/Y3wgVEtidi0
-      Varia Suit and After Cataris - Ledge Warp Setup Outside Diffusion Room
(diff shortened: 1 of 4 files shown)


randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -3682,6 +3682,9 @@ Extra - asset_id: collision_camera_051
           Any of the following:
               Varia Suit
               Heat/Cold Runs (Intermediate) and Heat Damage ≥ 120
+  > Event - Ledge Warp Setup
```
Logarhythmic, line 3685: I know that the bomb is necessary to open the door once you've stored the warp, but what if the door was randomized to Grapple Doors for example? Should we not add `Not Door Rando` to this?

Tyranisaur: If the door is randomized then it's a two way door and you can go back out of the door in a way that makes is unproblematic.

Logarhythmic: good point

Logarhythmic (approved):
Looks a little weird at first with the requirements being split over two rooms, but i think its logically sound

Miepee (approved):

PR #8470, opened 2025-08-19
Fusion Change Softlock UI Text

M4RKERSM4SK:
This was brought up as an issue
- Anitsoftlock was misleading as it does not prevent all softlocks

Changed

- UI text from anti softlock to "Room Edits"
- Clarify what the option does
- List exactly what the room edits are

Wording change ideas from Reggie
List verified by Clue

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Metroid Fusion
 
 - Added: 5 joke hints.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Metroid Fusion
 
 - Added: 5 joke hints.
+- Changed: Anti-Softlock option under the Game Modifications -> Gameplay tab in the preset settings was changed to "Room Edits" and now includes an explanation and list of changes the option provides.
```
Miepee, line 49: The original intent of "anti softlocks" was to reduce non-obvious softlocks with no bombs.
Woild be nice if we can get that reflected in the name somehow. "Room edits" sounds a bit too generic there imo.

PR #8482, opened 2025-09-01
Fusion: Fix unable to export game in weird edge case

Miepee:
I have no clue how this can even happen considering we use is_file_validator right before. Couldn't find anything on the docs either, other than invalid path or no perms to open, neither of which I think are the case here
But apparently it does happen, so band-aid it is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added: 5 joke hints.
 - Added: Racetime integration with category https://racetime.gg/mfr-rdv
 - Fixed: Incorrect exporting behaviour, if the option to hide text of pickups was enabled.
+- Fixed: Unable to export a randomized game, in case the input file could not be read.
 - Removed: The "Anti-Softlock" option and instead made the following permanent changes:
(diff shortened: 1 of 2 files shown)


M4RKERSM4SK:
Were you able to try setting a file with bad permissions or something to actually hit the exception?

M4RKERSM4SK:
Only reason I ask, is I cant seem to get a senario where I can hit the execption, like maybe if I had this on a removable drive or something wonky happens there?

Miepee:
No, but we've had a bug report about it: https://discord.com/channels/914291389293027329/1410025448628289627

Apparently OneDrive can fuck with permissions. And it caused them to  not being able to open the export window at all, as it threw an exception before it fully initialized. Only way out was modifying the config file/messing with the input file.

M4RKERSM4SK:
Windows and onedrive for whatever reason can present the path name like C:\users\documents\file.ext similar to a sym link, but some one drive pathing is hidden underneath at least on windows 10. Especially if the file is "offloaded" from the local system due to not being access in awhile etc if the settings have "Free up space". Cool

Seems they did different folder stuff for windows 11.

I cant hit the exception on my windows 11 VM, but I guess good to have it here in case someone else hits it.

M4RKERSM4SK (approved):

JeffGainsNGames (approved):

PR #8483, opened 2025-09-01
Fusion: Show Seed Hash on title screen

Miepee:
grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Fusion
 
-- Added: The title screen will now display the Randovania and patcher version.
+- Added: The title screen will now display the Randovania and patcher version, as well as the Seed Hash.
 - Added: 5 joke hints.
 - Added: Racetime integration with category https://racetime.gg/mfr-rdv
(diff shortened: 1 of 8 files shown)


M4RKERSM4SK:
What would the centerting look like if the seed hash was all 1 line? Would it be too long? Center just seems off, but if there no better option guess thats it. 

Something like:

image

cluekitty:
> What would the centerting look like if the seed hash was all 1 line? Would it be too long? Center just seems off, but if there no better option guess thats it.
> 
> Something like:
> 
> image

I seem to recall some discord discussion about this that said we have a few words that would have the seed hash potentially be too long for one line, even with only two words. The max length of a line is 30 characters before it goes off-screen.

M4RKERSM4SK:
I forgot about max line lenth, thanks clue

Miepee:
The longest hash word from fusion is 14 characters long. If you take other games into account, then the longest hash word is 15 characters long.
Either way, not enough space to print all in one line. And arbitrarily cutting it off depending on the length seems eh.

M4RKERSM4SK:
Just adding biospark's comment so its known for furture case

Adding in language support on the title screen would cause this hash displayment to not fit.

Otherwise, I think this is about as good as center as it gets after looking at all the code etc. I am not sure if python center does anything better for it, but it looks like that discussion was already had.

M4RKERSM4SK (approved):

PR #8490, opened 2025-09-08
Drogyga HBJ

steven11sjf:
Ludicrous because it's a one-shot trick if you don't have spider magnet.


Logarhythmic:
Is is possible to WBJ and then SWJ to get back out of the water if you fail? I don't doubt it should be ludicrous, but I think it might be possible to escape the water if you fail. I'll check it when I get home.

image

Tyranisaur:
This is cool and all, but it's not a one-shot trick, because either you do this before fighting Drogyga when the water isn't there, meaning you can get back up to the ledge with Simple IBJ. Or you're doing this after fighting Drogyga, in which case you have Grapple Beam by construction.

I noticed that we're missing Single Wall Jumps to get out of the bottom here, both using Morph Ball and using Flash Shift.

MayberryZoom:
Can't you just bomb jump out of the water? I wouldn't call this ludicrous anyways, the bomb jump itself is expert at most imo. This is something that with practice becomes extremely consistent and someone could execute reliably (speaking from experience). We even have precedence for this, because an extremely similar one in the database (Ghavoran Left Entrance) is already rated at expert. If anything, that one is probably harder because you cover more distance (20 tiles instead of only 14) and need to IBJ higher to get over the pitfalls before going horizontal.

Also, this should really have a video.

steven11sjf:
Water bottom -> left of grapple blocks connection, since it's easier to follow in UI

stable:
image

this branch:
image

randovania/games/dread/logic_database/Burenia.txt
```diff
@@ -688,14 +690,12 @@ Extra - asset_id: collision_camera_007
...
               Any of the following:
+                  Lay Cross Bomb
                   All of the following:
-                      Water Bomb Jump (Beginner)
-                      Any of the following:
-                          Lay Cross Bomb
-                          Spider Magnet and Lay Normal Bomb
-                  Water Bomb Jump (Intermediate) and Lay Normal Bomb
+                      Lay Normal Bomb
+                      Spider Magnet or Single-wall Wall Jump (Beginner) or Water Bomb Jump (Intermediate)
```
Tyranisaur: First thing. This should be Intermediate Single Wall Jump. All instances of this trick were upped from beginner to intermediate  to underline that even the simplest applications of this takes a combination of button input that's more challenging to pull off than what we expect out of our beginner tricks. This application is further complicated by doing it directly out of a water bomb jump.

Second, and I think this was an underlying problem before: It's kind of a crime that the logic here is only viable after having defeated Drogyga. The logic here should allow getting this item before fighting Drog if you have enabled the corresponding tricks and have the sufficient items to where you could get out of the water if it was there. I think that means there should also be a branch here that involves IBJ.

The more I'm thinking about it, the less sure I am. It seems to be the case that in order to put the water here, you need to defeat Drogyga, and in order to defeat Drogyga, you need Grapple Beam due to the Grapple Block in Drogyga Access, but even in the face of Room Rando or literally starting in Drogyga Arena, you still need either Grapple Beam or Gravity Suit to do the fight. So maybe this connection could just be Grapple or Use Spin Boost or Simple IBJ or Speed Booster or Flash Shit and Wall Jump (Beginner).

steven11sjf: Isn't putting a start point in every room a fairly long-standing issue? I'm pretty sure it was added to the patcher/schema years ago but we just never got around to finishing the implementation.

Tyranisaur: Regardless, I brought that up to better underline my conclusion that it's safe to assume that either the water isn't there or you have either grapple or gravity.

Tyranisaur:
Now that we finally did the stable to main merge, can you resolve the merge conflict by moving the changelog entry to the upcoming version?

randovania/games/dread/logic_database/Burenia.txt
XenoWars: Regarding the second paragraph, Gravity Suit methods are already taken care of with or without Drogyga in the "And" group not shown in the images, though SWJ is not included and probably could be. I do agree there could be merit in removing the "After Drogyga" part with the Water Bomb Jump methods since Grapple, Gravity, and Before Drogyga are the only other possibilities and are already covered. Any method you can do while the water is there, you can also do with both Gravity and Before Drogyga methods.

Tyranisaur (approved):

PR #8497, opened 2025-09-15
Multiworld: Warn on LogicalPickupPlacement.All

Miepee:


randovania/layout/base/base_configuration.py
```diff
@@ -91,6 +91,9 @@ def dangerous_settings(self) -> list[str]:
         if self.consider_possible_unsafe_resources:
             result.append("Considering possible unsafe resources will increase generation time.")
 
+        if self.logical_pickup_placement == LogicalPickupPlacementConfiguration.ALL:
+            result.append("Requiring every pickup in the preset to be obtainable causes increased generation failures.")
```
Tyranisaur: The grammar of this sentence seems off to me. Can you make it ".. causes increased chance of generation failures."? Also, should we specify in some way that it's really only causing failures when paired with options that result in item pickups being inaccessible?

Miepee:
```suggestion
            result.append("Requiring every pickup in the preset to be obtainable causes increased chance of generation failures.")
```

Miepee: > Also, should we specify in some way that it's really only causing failures when paired with options that result in item pickups being inaccessible?

IMO, no. Checking that is most of the time very difficult for the one who generates or made the preset, as they essentially would have to be aware of all the games/their own games' DB's intricacies. 
E.g. you'd need to know this makes games impossible to gen on [Dread, am2r, CS] trickless, that prime trickless will be impossible if you have less than 15 missiles total, that echoes will be impossible if you have less than 50 ammo for both total etc.
And while it is relatively easy to check if a preset "works" by using it to gen a solo game, experiences have shown that basically no casual user does that.

PR #8502, opened 2025-09-18
Introduce "Beaten" game status

Miepee:
I'd like to make it a mandatory implementation. Is there anything that speaks against doing that?


henriquegemignani:
> I'd like to make it a mandatory implementation. Is there anything that speaks against doing that?

Yes. Go implement it for Echoes. :)

Miepee:
I'm not as familiar with Echoes' internals. Whats the difficult part there? I assume its the "read which room we're in" thing that you mentioned in the past?

Also didn't mean to make it mandatory *now*, but rather in the (near) future when our games have it implemented.

henriquegemignani:
> I'm not as familiar with Echoes' internals. Whats the difficult part there? I assume its the "read which room we're in" thing that you mentioned in the past?

We can only easily get the current world. Current area is at best an array index which is data we don't have, nor I'm 100% sure I trust. We'd certainly be limited to only adding new areas to the end or never messing with the list. (So no room rando where we move a room from Agon to Torvus)

> 
> Also didn't mean to make it mandatory _now_, but rather in the (near) future when our games have it implemented.

Unless you have someone who will go implement it for Echoes, that future will remain future.

Miepee:
Should be reviewable, although I need help with the migration.
It's throwing a `peewee.IntegrityError: FOREIGN KEY constraint failed` error and I don't know why or what I need to do to fix it. From what I can see, I have the same setup as last time we did a migration that added one column.

Miepee:
Managed to fix the migration by issuing sql commands instead of using the migrator's add column function

randovania/server/database.py
```diff
@@ -286,9 +286,7 @@ def create_session_entry(self) -> multiplayer_session.MultiplayerSessionEntry:
         # Get the worlds explicitly, as we return them and would also need for the user assocations
         worlds = {
             world.id: MultiplayerWorld(
-                id=world.uuid,
-                name=world.name,
-                preset_raw=world.preset,
+                id=world.uuid, name=world.name, preset_raw=world.preset, has_been_beaten=world.beaten
```
henriquegemignani: add a `,` at the end to keep the formatting

henriquegemignani (changes requested):

randovania/server/database.py
Miepee: done, thanks

randovania/server/multiplayer/world_api.py
```diff
@@ -273,6 +273,12 @@ def sync_one_world(
         association.last_activity = datetime.datetime.now(datetime.UTC)
         association.save()
 
+    if world_request.has_been_beaten and not world.beaten:
+        logger().info("Session %d, World %s has been beaten", world.session_id, world.name)
+        world.beaten = True
+        world.save()
```
Miepee, line 289: I'm not sure if saving here is the best thing to do tbh.
Honestly feel like i probably need another signal here to emit, like the inventory does above as we probably want to have those logged in the audit entry

randovania/gui/multiworld_client.py
```diff
@@ -118,6 +120,7 @@ def _create_new_sync_request(self) -> ServerSyncRequest:
                     collected_locations=self.database.get_locations_to_upload(uid),
                     inventory=None,
                     request_details=False,
+                    has_been_beaten=False,
```
Miepee: these two are somewhat of a placeholder. Ideally we would check for all games that were beaten at some point and upload that too, but I feel like adding two more entries to the WorldData feels wrong

randovania/server/multiplayer/world_api.py
```diff
@@ -273,6 +273,12 @@ def sync_one_world(
         association.last_activity = datetime.datetime.now(datetime.UTC)
         association.save()
 
+    if world_request.has_been_beaten and not world.beaten:
```
henriquegemignani, line 286: This case should update the last activity.

Session history should somehow show when a world was beaten, but that one is purely a list of actions. Audit log could do it, but that's a weird place. Guess audit will have to do? 🤔

henriquegemignani: Also a test case for this.

randovania/network_common/world_sync.py
```diff
@@ -14,6 +14,7 @@ class ServerWorldSync:
     collected_locations: tuple[int, ...]
     inventory: bytes | None  # frozendict[str, InventoryItem], but pre-encoded
     request_details: bool
+    has_been_beaten: bool
```
henriquegemignani, line 17: Server API version needs to be raised

tools/old_data.db
henriquegemignani, line 1: hey

henriquegemignani (changes requested):
Needs some tests overall for the flag.

randovania/server/multiplayer/world_api.py
henriquegemignani: The save can be skipped as you're adding to `worlds_to_update`

randovania/gui/multiworld_client.py
```diff
@@ -99,26 +99,26 @@ def _create_new_sync_request(self) -> ServerSyncRequest:
...
-                    collected_locations=locations,
-                    inventory=None,
-                    request_details=False,
-                )
+            if uid not in sync_requests:
+                upload_data = self.database.get_data_to_upload(uid, provide_even_on_no_change=False)
+                if not upload_data:
+                    continue
+
+                sync_requests[uid] = upload_data
```
henriquegemignani: probably would just keep this inside the `if`, and invest the condition.

randovania/gui/multiworld_client.py
```diff
@@ -99,26 +99,26 @@ def _create_new_sync_request(self) -> ServerSyncRequest:
...
-                sync_requests[uid] = ServerWorldSync(
-                    status=GameConnectionStatus.Disconnected,
-                    collected_locations=self.database.get_locations_to_upload(uid),
-                    inventory=None,
-                    request_details=False,
-                )
+                upload_data = self.database.get_data_to_upload(uid, provide_even_on_no_change=True)
+                if not upload_data:
+                    continue
+                sync_requests[uid] = upload_data
```
henriquegemignani: same thing

henriquegemignani (changes requested):

henriquegemignani (changes requested):
In `test_world_api.py`, assert that the World is correct updated (or not updated) when the beaten flag is sent.

And that it can only be beaten once.

Miepee:
Done

PR #8504, opened 2025-09-19
Enable sentry-sdk logging support

henriquegemignani:


randovania/monitoring/__init__.py
```diff
@@ -138,6 +140,8 @@ def traces_sampler(sampling_context):
         event_scrubber=HomeEventScrubber(),
         before_breadcrumb=before_breadcrumb,
         before_send=before_send,
+        enable_logs=True,
```
Miepee, line 143: this will enable logs for the client too, yes?

henriquegemignani: Yes. I kept thinking and thinking if that's a good idea or not. In the end, as long as the user data filtering works it's probably fine

PR #8505, opened 2025-09-19
Lock RDS to below 0.29.0, OPR below 0.16, make empty parametrize fatal.

henriquegemignani:
Newer versions are not safe ~~for work~~.


pyproject.toml
```diff
@@ -98,8 +98,8 @@ exporters = [
 
     # Metroid Prime 1/2
     "ppc-asm>=1.2.1",
-    "open-prime-rando>=0.13.0",
-    "retro-data-structures>=0.24.0",
+    "open-prime-rando>=0.13.0,<0.16",
```
Miepee: a comment as to why they have to be limited would be cool

PR #8512, opened 2025-09-20
Fix a lot of mypy errors in multiplayer_session_window

Miepee:
The rest ~15 that are left are more annoying to fix as they need a few logic changes


henriquegemignani (commented):
Can you either fix or ignore the remaining mypy issues, so the file is added to the required files for mypy?

randovania/gui/lib/multiplayer_session_api.py
```diff
@@ -26,11 +28,11 @@
 OriginalFunc = typing.Callable[Param, RetType]
 
 
-def handle_network_errors[**Param, RetType](
-    fn: typing.Callable[typing.Concatenate[MultiplayerSessionApi, Param], RetType],
-) -> typing.Callable[Param, RetType]:
```
duncathan, line 31: hmm. seeing this concatenate has mii questioning if my signature was correct. maybe it should be:
```py
def handle_network_errors[T, **P](
    fn: Callable[Concatenate[MultiplayerSessionApi, P], Coroutine[Any, Any, T]],
) -> Callable[Concatenate[MultiplayerSessionApi, P], Coroutine[Any, Any, T | None]]:
    @functools.wraps(fn)
    async def wrapper(self: MultiplayerSessionApi, *args: P.args, **kwargs: P.kwargs) -> T | None:
```
I'm not sure if the change is necessary, tho

Miepee:
I originally thought about fixing a lot of them in this PR but I wanted to keep this PR smaller to not mix "a lot of really trivial changes" with "decent amount of logic changes" during review. But if you'd rather have them fixed here, then sure.

henriquegemignani:
I'd rather not have logic changes just for typing >.>

Miepee:
We do need logic changes tho, as we currently just straight up ignore typing in this file :D 
Lots of nullables that we pretend are never null

henriquegemignani:
A lot of them technically just need an assert, because they depend on that case.

Miepee:
Assert wouldn't be quite enough. We have seen things slip through every now and then with things that "shouldn't" be None suddenly being None in real world usage.

henriquegemignani:
> Assert wouldn't be quite enough. We have seen things slip through every now and then with things that "shouldn't" be None suddenly being None in real world usage.

Personally that's a different task to adding typing to a file.

Miepee:
Which sums up the reason why I wanted to do that in a different PR.

henriquegemignani:
Yeah but you can still just ignore specific errors and have the file be checked.

Miepee:
Done

randovania/gui/lib/multiplayer_session_api.py
Miepee: I don't really see a difference with or without your suggestion.
In both cases they behave the same, in regards to mypy complaining, on the current code and if i plop the decorator on a function that doesnt have `self: MultiplayerSessionApi` 

The behaviour being: no complains

randovania/gui/multiplayer_session_window.py
```diff
@@ -1,11 +1,14 @@
+# mypy: disable-error-code="union-attr"
```
duncathan, line 1: I don't love doing this if it's something actually within our power to work around

randovania/gui/lib/multiplayer_session_api.py
```diff
@@ -26,11 +28,11 @@
...
+def handle_network_errors[T, **P](
+    fn: Callable[P, Coroutine[Any, Any, T]],
+) -> Callable[P, Coroutine[Any, Any, T | None]]:
```
duncathan, line 33: https://github.com/randovania/randovania/blob/76f51f214ed5dd9f34131f18bab878b48005dcb6/randovania/server/server_app.py#L51

maybe this should be pulled out into something in `lib` so that it can be reused wherever needed? I find it a lot easier to read

randovania/gui/multiplayer_session_window.py
```diff
@@ -477,7 +484,7 @@ def update_game_tab(self):
             # FIXME: this triggers on every meta update as opposed to just when the generation status gets updated
             # common_qt_lib.alert_user_on_generation(self, self._options)
 
-    def _describe_action(self, action: MultiplayerSessionAction):
+    def _describe_action(self, action: MultiplayerSessionAction) -> tuple[str, str, str]:
```
duncathan, line 487: something more semantic like a namedtuple would be nice here. at a glance i don't have a clue what these elements are

randovania/gui/multiplayer_session_window.py
```diff
@@ -762,7 +770,8 @@ def generate_layout(progress_update: ProgressUpdateCallable):
                     layout.save_to_file(last_multiplayer)
 
                 self.update_progress("Finished generating, uploading...", 100)
-                await uploader(layout)
+                # FIXME: Fix decorator typing
```
duncathan, line 773: why not just fix this instead of ignoring?

duncathan (changes requested):

PR #8508, opened 2025-09-19
Update mars patcher to 0.8.0

Miepee:
Implementation of insta morph to come in a future pr

PR #8515, opened 2025-09-21
Fusion: Add instant morph option

Miepee:
Blocked on #8508


Miepee:
> Add functionality to differences tab

I thought the differences tab was for changes that are always present? And not changes that the user can toggle

Miepee:
I can't see for example the open sector hub mentioned there either.

M4RKERSM4SK:
Warp to start audio issues are still present.

1. If you warp to start after warping to start, it kills all back ground music

2. If you warp to start, the save jingle no longer plays and you can move instantly

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Fusion
 
+- Added: Option that allows quick-morphing via the SELECT button.
```
M4RKERSM4SK, line 34: You were worried about this option being burried or not noticed. Do you want the readme to combat that by saying something like the "the option can be found in..."

M4RKERSM4SK: Sorry meant changelog

Miepee: well the option is on by default, so i think anyone who is technical enough to read the changelog, knows how to find it in the preset option

M4RKERSM4SK: Sounds good

PR #8519, opened 2025-09-21
Change QObject to RDVSignal in non-gui modules

Miepee:


randovania/game_connection/builder/connector_builder.py
```diff
@@ -2,15 +2,15 @@
...
 
 if TYPE_CHECKING:
     from randovania.game_connection.connector.remote_connector import RemoteConnector
     from randovania.game_connection.connector_builder_choice import ConnectorBuilderChoice
 
 
-class ConnectorBuilder(QtCore.QObject):
-    StatusUpdate = QtCore.Signal(str)
+class ConnectorBuilder:
+    StatusUpdate = RdvSignal[str]()
```
henriquegemignani: I think it's `RdvSignal[[str]]` no?

Miepee: NOt sure, but probably based on the other ones that are int he connector

henriquegemignani (approved):

henriquegemignani:
These can likely be moved to a separate "module" as `interface_common` kinda requires gui anyway.

Miepee:
`interface_common` afaik no, but it looked like there were a few inside `gui` that dont rely on a gui and could be generalized out.

PR #8521, opened 2025-09-22
Prime 1: Improve Elevator Rando containing credits as a target

Miepee:
The TLDR; is: let's try to ensure that we always have the possibility of having the credits as a viable target.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Adjusted fit of main menu text to accommodate larger seed hashes
 - Changed: Adjusted fit of DLR Blast Shields over doors
 - Changed: Research Core cutscene camera angle will now be consistent regardless of the height of the placed pickup
+- Changed: Improved chance of successfully generating a game when shuffling everything on One-Way Anywhere elevator randomization.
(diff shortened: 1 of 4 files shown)


Miepee:
Did some stats with this on the Alternative Prime Preset (it has random start, and 1-3 starting items).

Before:
Invalid starting state means, that the "verify if game is beatable" caught it before it ever went into generation.

| Total | Successful | Invalid Starting State | Failed Generation | 
|--------|--------|--------|-----|
| 7000 | 1848 | ~3630 | ~1530  |
| 100% | 26,4% | 51,8% | 21,8% |

Out of the 1848 seeds that successfully genned, i draw a graph of how common a specific gen-order length is:
grafik


After:
I'm not certain whether "invalid starting state" still happened. My terminal history wasn't big enough to figure that out. So if they did, they're bundled with failed generation here.
| Total | Successful | Failed Generation | Impossible for Solver |
|--------|--------|-----|-----|
| 7000 | 3595 | ~3156  | 249 |
| 100% | 51,3% | 45,0% | 3,5%  |

Genorder length graph, keep in mind that these are 3695 seeds here:
grafik

TLDR;
- Doubled success rate
- About equal results in terms of seed quality (or maybe slightly better? Hard to tell, could be too little data)
- Generation time went from ~8h to ~1.5h, so around a fifth of the original generation time

PR #8531, opened 2025-09-25
Improve some configs that are deprecated

henriquegemignani:


Miepee (approved):
Change codspeed's Walltime config too please either in this PR or another.

Miepee:
Hmmm, looking at the last codspeed PR I can't see it complaining about what I had in memory. So maybe I'm thinking of a different thing, or it was fixed.

PR #8534, opened 2025-09-26
migrate server to FastAPI

### LINKED ISSUE #3754: Server: Replace Flask with an async server
This allows us to also use `socketio.AsyncServer` instead.
Main advantage is to have bigger similarity between client and server code. Currently server uses sync, but client uses async.
Better integration between the discord bot with the server is awkward with how different the two are.

Main idea of replacement framework: [Sanic](https://sanic.readthedocs.io/en/stable/)

duncathan:
Fixes #3754 

TODO:

- [x] Migrate functionality
- [x] Fix mypy errors in `randovania/server`
- [x] Fix tests
- [x] Documentation
- [x] Review to make sure there haven't been regressions
- [x] Reimplement prometheus cuz I forgot about it until writing this checklist
- [x] Probably rename the `flask` command 
- [x] Any leftover TODO comments


henriquegemignani:
I'm fine with leaving prometheus and renaming the flask command to a follow up PR.

randovania/cli/server.py
```diff
@@ -15,14 +15,51 @@ def flask_command_logic(args: Namespace) -> None:
...
 
-    from randovania.server import app
+    from randovania.server.server_app import ServerApp
 
-    server_app = app.create_app()
-    server_app.sa.sio.run(server_app, host="0.0.0.0")
+    server_app = ServerApp(randovania.get_configuration())
+
+    if args.mode == "dev":
+        host = "127.0.0.1"
```
henriquegemignani: I run my dev server with `0.0.0.0` >:(

Is there no other things to set for a dev server mode? Like auto-reload, verbose stack traces and so on? Flask has an entire thing where the HTML you get on an exception literally contains a python console.

randovania/server/multiplayer/session_admin.py
```diff
@@ -63,37 +63,37 @@ def verify_has_admin(
...
-def _verify_not_in_generation(session: MultiplayerSession):
+def _verify_not_in_generation(session: MultiplayerSession) -> None:
     if session.generation_in_progress is not None:
         raise error.InvalidActionError("Session game is being generated")
 
 
 def _get_preset(preset_bytes: bytes) -> VersionedPreset:
     try:
-        preset = VersionedPreset.from_bytes(preset_bytes)
+        preset: VersionedPreset = VersionedPreset.from_bytes(preset_bytes)
```
henriquegemignani, line 96: Why is this type declaration needed?

randovania/server/database.py
```diff
@@ -213,7 +219,7 @@ def get_layout_description_as_json(self) -> dict | None:
         return None
 
     def get_layout_description_as_binary(self) -> bytes | None:
-        if self.layout_description_json is not None:
+        if self.layout_description is not None:
```
henriquegemignani: Assign to a variable to avoid creating it twice.

randovania/server/discord_auth.py
```diff
@@ -0,0 +1,86 @@
...
+        self.role_id = str(config["role_id"])
+        async with aiohttp.ClientSession() as self.session:
+            self.session.headers["Authorization"] = "Bot {}".format(config["token"])
+            yield self
+
+    async def verify_user(self, user_id: int) -> bool:
+        url = f"https://discordapp.com/api/guilds/{self.guild_id}/members/{user_id}"
+        async with self.session.get(url) as r:
+            try:
+                result = await r.json()
```
henriquegemignani: Only call `.json()` if the response is OK, otherwise we're not sure it's actually a json. (a random 502 in the way can likely be a pure HTML for example)

randovania/server/discord_auth.py
```diff
@@ -0,0 +1,86 @@
...
+                return True
+
+
+@asynccontextmanager
+async def discord_oauth_lifespan(_app: RdvFastAPI) -> Lifespan[CustomDiscordOAuthClient]:
+    config = randovania.get_configuration()
+
+    client_id = config["discord_client_id"]
+    client_secret = config["server_config"]["discord_client_secret"]
+    redirect_uri = f"{config['server_address']}/login_callback"
```
henriquegemignani, line 76: Adding a TODO is perfect for now, but this could be asking fastapi to get the URL for the route that handles this.

randovania/server/discord_auth.py
```diff
@@ -0,0 +1,86 @@
...
+        if config is None:
+            yield None
+            return
+
+        self = cls()
+        self.sa = _app.sa
+        self.guild_id = config["guild_id"]
+        self.role_id = str(config["role_id"])
+        async with aiohttp.ClientSession() as self.session:
+            self.session.headers["Authorization"] = "Bot {}".format(config["token"])
```
henriquegemignani, line 51: Ewww, I really abused the bot's token in the server side to implement this? Guess it does let us skip extra permissions in OAuth.

pyproject.toml
```diff
@@ -126,10 +114,14 @@ gui = [
...
+    "fastapi-discord>=0.2.7",
+    "fastapi-socketio-handler>=0.4.0",
+    "prometheus-fastapi-instrumentator",
+    "fastapi",
+    "uvicorn",
+    "jinja2",
+    "itsdangerous",
     "eventlet",
     "peewee",
     "requests-oauthlib",
```
henriquegemignani, line 131: Still used?

pyproject.toml
```diff
@@ -126,10 +114,14 @@ gui = [
...
-    "prometheus-flask-exporter",
+    "sentry-sdk[fastapi]>=2.35.0",
+    "fastapi-discord>=0.2.7",
+    "fastapi-socketio-handler>=0.4.0",
+    "prometheus-fastapi-instrumentator",
+    "fastapi",
+    "uvicorn",
+    "jinja2",
+    "itsdangerous",
     "eventlet",
```
henriquegemignani: Still used?

randovania/server/multiplayer/session_admin.py
duncathan: I think it's actually a mypy bug. I have no idea but without it I get a `var-annotated` error

randovania/cli/server.py
duncathan: there was an autoreload thing but it was weird and I forgot to check back on it. the host was just the recommended settings, I can adjust if you'd prefer

BashPrime:
From what I've looked at things look kosher? Saw some hanging todos that I'll bring attention to.

pyproject.toml
duncathan: yes

randovania/server/discord_auth.py
duncathan: this is tricky to do, because that route doesn't exist yet at the time this code is run

duncathan:
Ready for review and merge

PR #8535, opened 2025-09-26
Remove benchmark flag from test_reach_size_from_start_echoes

henriquegemignani:


henriquegemignani:
It's also so fast there's no reason to exclude from the rest of the suite.

PR #8536, opened 2025-09-26
Fix opening tar in bulk generation

henriquegemignani:


henriquegemignani:
I checked the documentation, but turns out it was talking about `tarfile.open` and not the constructor!

PR #8537, opened 2025-09-28
Gracefully support disallowing the bot from sending messages.

henriquegemignani:


randovania/discord_bot/preset_lookup.py
```diff
@@ -370,6 +370,12 @@ async def on_message(self, message: discord.Message):
         if message.author == self.bot.user:
             return
 
+        # Support for disabling the bot from replying by just disallowing it from sending messages.
+        permissions = message.channel.permissions_for(message.guild.me)
+        if not permissions.send_messages:
```
Miepee, line 375: Should we try to inform the admin/server owner about it? Since it can be both accidental misconfiguration and intentional

henriquegemignani: How would you be able to inform the owner, and how would you do it in a way that doesn't happen every time a message is sent to these channels. No persistence on bot's side allowed.

Miepee: Idk, check for magic word in channel topic and if not present, dm them? 😅

But since we do have access to the logs, i suppose we might as well look at them manually every now and then and inform the server mods if something weirds going on

henriquegemignani: It's not like this change would make a difference. This sort of misconfiguration before would cause lots of exceptions in the log, now it's just some messages.

randovania/discord_bot/preset_lookup.py
```diff
@@ -370,6 +370,12 @@ async def on_message(self, message: discord.Message):
...
+            logging.info("Bot not allowed to send messages at #%s", message.channel.name)
```
Miepee: Cab you include the guild name in thelog too?

Miepee (changes requested):

Miepee (approved):

PR #8539, opened 2025-09-28
Fix typing error in multiplayer_session_window

Miepee:
Fixes https://randovania.sentry.io/issues/6908998327 and related


henriquegemignani (changes requested):
Can you add a test case that would catch this?

Supposedly unrelated, but your fix is actually wrong :P

PR #8541, opened 2025-09-28
Add extra to logging calls in preset_lookup.py

henriquegemignani:


Miepee (approved):
Mypy complains a bit here, but thats all "what if the bot was DMed" which right now i dont think is supported anyway

henriquegemignani:
> Mypy complains a bit here, but thats all "what if the bot was DMed" which right now i dont think is supported anyway

It is actually! I doubt anyone uses, but everything you'd use the sandbox channel you can actually do in a DM with the bot.

randovania/discord_bot/preset_lookup.py
```diff
@@ -248,7 +243,15 @@ async def _try_get[T](
         return decoder(json.loads(data.decode("utf-8")))
     except Exception as e:
         if not isinstance(e, UnsupportedVersion):
-            logging.exception(f"Unable to process {attachment.filename} from {message}")
+            logging.exception(
+                "Unable to process %s from %s",
```
Miepee, line 247: somewhat of a nit, but wouldnt this log be somewhat unreadable if someone sends a message with only the attachment (meaning that `content` would be blank/empty)?

randovania/discord_bot/preset_lookup.py
```diff
@@ -260,16 +263,16 @@ async def _try_get[T](
         return None
 
 
-async def _get_presets_from_message(message: discord.Message):
+async def _get_presets_from_message(message: discord.Message) -> AsyncGenerator[Preset[BaseConfiguration]]:
     for attachment in message.attachments:
         filename: str = attachment.filename
         if filename.endswith(VersionedPreset.file_extension()):
-            result = await _try_get(message, attachment, VersionedPreset)
+            result = await _try_get(message, attachment, lambda d: VersionedPreset(d).get_preset())
```
Miepee: mypy wants type annotation here. Maybe similar bug to what was mentioned in #8534 ?

henriquegemignani: In this case it's tricky because we don't have `Preset` defaulting to `Preset[BaseConfiguration]` (I think?) so it really needs a hint.

PR #8549, opened 2025-10-02
Fusion: optimize blank line in credits

Miepee:
The fusion credits wants to write one blank line. Previously this was done by writing an empty string as a new line.
However it is much more memory efficient for the game, to just tell the previous element to include one more newlines.
This doesn't *fully* make the "Credits exceeded length" dissapear, but it postpones it to where 99% of people won't encounter it.
Before this change, I was able to trigger it by only including a few duplicates, after this change in order to trigger it, I have to shuffle ~95 majors (out of 127 locations). 

Fixes https://randovania.sentry.io/issues/6916113458/


henriquegemignani (changes requested):
If this PR doesn't need a test file update, you need an extra test file.

Miepee:
It does need test file changes, i only realized after sending the PR off <_<
Latest commit fixes it

henriquegemignani (approved):

PR #8555, opened 2025-10-03
Use a separate socketio_path for the server

henriquegemignani:


randovania/server/socketio.py
```diff
@@ -21,7 +21,7 @@
 @asynccontextmanager
 async def fastapi_socketio_lifespan(_app: RdvFastAPI) -> Lifespan[SocketManager]:
     async with SocketManager(socketio_path=None) as manager:
-        manager.mount_to_app(_app, _app.sa.configuration["socketio_path"])
+        manager.mount_to_app(_app, _app.sa.configuration["server_config"].get("socketio_path", "/socket.io/"))
```
duncathan, line 24: maybe default to the client path, rather than a hard-coded default?

henriquegemignani: Ehh it'd break the existing config which is partly what I don't want to touch. Honestly we correct thing is no longer have this variable at all and rely of only knowing the prefix to use.

duncathan (approved):

PR #8558, opened 2025-10-03
Prime: remove some dangerous requirements

JeffGainsNGames:
chozo ice temple, trick can be done before and after
control tower, before to the event it leads to, and we can pretty safely ignore research core
dynamo access, make the fight an event on the side that can be bypassed without plasma or accessed with plasma, which preserves the function of the room in the database
image
image


randovania/games/prime1/logic_database/Phazon Mines.txt
```diff
@@ -1348,9 +1348,7 @@ Extra - aabb: [57.918835, -49.629333, -88.16487, 129.05498, -6.0125637, -65.8200
   * Extra - world_position: (129.04474, -14.07471, -83.17842)
   * Extra - world_rotation: (0.0, -0.0, 0.0)
   > Door to Omega Research
-      No Plasma Beam or After Elite Pirate Fight (Dynamo Access)
-  > Event - Elite Pirate Fight
-      Plasma Beam
+      Trivial
```
Miepee: This room gets now changed (drastically? cant tell) and would probably be good to mention it in the changelog.
You essentially changed the requirements from
```
ANY:
- No Plasma Beam
# Kill Pirate
- ALL:
  - Plasma Beam
  - ANY:
    - Charge Beam
    - Combat (Beginner)
    - 200 HP
```
To `Trivial` for going to the door, and for killing the pirate you removed the combat stuff.
Also, Is the Elite Pirate Fight event even used for anything?

JeffGainsNGames: probably could improve the changelog, will wait for any other discussions
I guess I could add the combat requirements next to plasma, but the whole event as you mention next is not used for anything other than unlocking the doors. which we can't use because it introduces a `before`
the only reason I left it in is because it preserves the "behaviour" of the room a bit better

duncathan: I would be surprised if this bit of logic has ever been relevant. The odds of being down here with Plasma, no charge, combat disabled, and only 0 or 1 etanks are pretty miniscule

Miepee: IIRC it can become relevant if you have some nasty mines start locations. Either way, should probably get a little more descriptive in the changelog.

Samuel6710: I believe this logic can be relevant in alternate race presets so I don't see the low odds of it occurring being a valid reason for not considering implementing that logic.

JeffGainsNGames: removed from PR, ongoing discussion in https://discordapp.com/channels/914291389293027329/1423725452538679366

CHANGELOG.md
```diff
@@ -7,7 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## [10.3.0] - 2025-11-??
 
-- To be filled.
+### Metroid Prime
+
+#### Logic Database
+
+##### Phendrana Drifts
+
+- Fixed: Chozo Ice Temple: Removed unnecessary dangerous requirements.
```
Samuel6710, line 21: Should these be in the changelog?

Miepee: We've done both in the past, so no real guideline.

JeffGainsNGames:
additionally removed dangerous beetles
added some documentation to chozo ice temple since I was already touching it

PR #8559, opened 2025-10-03
Dread: Add missing rotatable requirement on Screw Attack Room Spark

Tyranisaur:


Logarhythmic (approved):
The change log took me a few reads to understand what was going on. I might have said -

"Using a Shinespark to reach Start Point 2 from the door to Transport to Burenia was missing a requirement on having activated the Rotatable."

PR #8561, opened 2025-10-04
AM2R: rename tester node

Miepee:


Miepee:
Arena is probably better yeah.

PR #8569, opened 2025-10-05
Add explicit categories for Trivial/Impossible in the ConnectionsEditor

henriquegemignani:
Super special cases for these two. Hopefully nothing blows up...

image

### DIFF
--- a/randovania/gui/dialog/connections_editor.py
+++ b/randovania/gui/dialog/connections_editor.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import typing
+from typing import override
 
 from PySide6 import QtGui, QtWidgets
 from PySide6.QtCore import Qt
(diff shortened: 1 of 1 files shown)


randovania/gui/dialog/connections_editor.py
```diff
@@ -426,22 +461,43 @@ def __init__(
...
         else:
             requirement_type = type(requirement)
         signal_handling.set_combo_with_value(self.requirement_type_combo, requirement_type)
+        self._create_editor(requirement)
 
-        if isinstance(requirement, ResourceRequirement):
+    def _create_editor(self, requirement: Requirement) -> None:
+        current_data = self.requirement_type_combo.currentData()
+
+        if current_data == CUSTOM_MAPPING[Requirement.trivial()]:
```
Miepee, line 487: i'm not quite sure i follow this. Isn't `CUSTOM_MAPPING[Requirement.trivial()]` == `CUSTOM_MAPPING[Requirement.impossible()]`?

henriquegemignani: No. Each instance of `object()` is an unique object, with it's own hash and equality is if it's the same instance.

Miepee: ah

Miepee (approved):

PR #8571, opened 2025-10-05
Move Custom UI Widgets into own module

Miepee:
Better name suggestions welcome.


henriquegemignani:
What's the difference between this and `randovania.gui.widgets`?

Miepee:
Oh didnt even think about that. Good point, the ones in this MR should all be moved there.

randovania/gui/lib/connections_visualizer.py
```diff
@@ -66,6 +66,10 @@ def create_tree_items_for_requirement(
 
 
 class ConnectionsVisualizer:
+    """
+    A widget that's meant for showing a human-readable form of a Connection/Requirement.
```
henriquegemignani, line 70: This is not a widget :)

randovania/gui/widgets/flow_layout.py
```diff
@@ -48,6 +48,10 @@
 
 
 class FlowLayout(QLayout):
+    """
+    A custom layout that adjusts the position of widgets within depending on the width of the application window.
```
henriquegemignani, line 52: Flows are not Widgets 🤔

Miepee: They're not, but they fit here based on feel? IDK what the proper naming for "thing you would use in qt designer or when manually creating a ui" other than "widget". From quickly looking at qt documentation i couldnt find anything more helpful either there. Layouts and widgets are classified as QtObjects there and thats not quite how i'd like to call this.
Maybe "qt" stuff or something?

randovania/gui/lib/connections_visualizer.py
Miepee: moved back

henriquegemignani (approved):

PR #8573, opened 2025-10-06
Supply obfuscator secret via env var for the discord bot

henriquegemignani:
Will this actually work is a mystery we find out after merging.

### DIFF
--- a/randovania/lib/obfuscator.py
+++ b/randovania/lib/obfuscator.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import os
 
 import cryptography
 from cryptography.fernet import Fernet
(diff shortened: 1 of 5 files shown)


randovania/lib/obfuscator.py
```diff
@@ -22,12 +23,19 @@ class InvalidSecret(Exception):
 
 
 def _get_fernet() -> Fernet:
-    if _secret is None:
+    secret = _secret
+
+    if secret is None:
+        env_var = os.environ.get("OBFUSCATOR_SECRET", "")
```
Miepee, line 29: theoretically allows end user clients to use their own secrets but like... lol.

Miepee (approved):

randovania/lib/obfuscator.py
henriquegemignani: I thought about it, but it's only when using a build that doesn't have one. Aka, they built from source. Which they could've just built with one instead.

PR #8574, opened 2025-10-06
Add manually triggered workflow for testing mostly ARM images

henriquegemignani:


.github/workflows/other-image-test.yml
```diff
@@ -0,0 +1,121 @@
+name: Other Image Test
```
Miepee: Since there's a lot of copy-paste in here, I suppose there is no nice/easy way of combining it with our main python workflow?

.github/workflows/arm-python.yml
Miepee, line 1: rename to `arm-python.yml` please if this stays as a seperate workflow

.github/workflows/other-image-test.yml
henriquegemignani: Not that I'm aware......

.github/workflows/arm-python.yml
henriquegemignani: I was thinking to keep this for testing building for weird platforms (like Switch or Android APK) but I guess it'll always boil down to arm lmao

And these would be significantly different builds I guess

PR #8589, opened 2025-10-09
Document Prime Connector more

Miepee:


randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -135,16 +143,30 @@ def _current_status_world(self, world_asset_id: bytes | None, vtable_bytes: byte
...
     @property
     def multiworld_magic_item(self) -> ItemResourceInfo:
+        """
+        The item that is used for tracking which locations have been collected and how many sent pickups have been
+        received.
+
+        Every item in the Prime games has both an `amount` and a `capacity`. The former stores how much you have of that
+        item, while the latter stores how much in total you can carry. For example, 5 / 10 Missiles.
+        Randovania will store which location was last collected in the magic item's amount, and how many pickups have
+        been received in the item's capacity.
```
toasterparty:
```suggestion
        The game will increment magic item's `amount` by `pickup_index+1` for each collected location containing an off-world item.
        Randovania will store the number of remote items sent to the game in magic item's `capacity`
```

Miepee: Hmm, what about:
```md
The game will increment the magic item's `amount` and `capacity` by `pickup_index+1` for each 
collected location containing an off-world item. If the connector detects that `amount` is bigger
than zero, it will decrement both by `pickup_index+1` to detect the collected location.
Additionally, the connector will also store the number of remote items sent to the game in the 
magic item's `capacity`. It will however only be read if `amount` is zero.
```

I'm not sure if that is too much implementation detail for "here's what the magic item does", but we don't have any other easily accessible place on where to describe it.

henriquegemignani: Who reads this file is people looking for implementation details.

> `collected location containing an off-world item`

Not exactly. Echoes does that for every location. It's prime 1 that doesn't do it.

toasterparty: My main concern was this bit is misinformation "Randovania will store which location was last collected in the magic item's amount". Randovania doesn't, the game does.

As long as that's clear I'm happy

Miepee: Adjusted

randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -177,6 +207,10 @@ async def check_for_collected_location(self) -> bool:
 
         magic_inv = InventoryItem(*struct.unpack(">II", op_result))
         if magic_inv.amount > 0:
+            # Amount is bigger than 0, so we have collected an item.
+            # The PDF added pickup_index+1 to both the amount and capacity, so subtract 1 back from the amount
+            # to get the pickup index and emit it.
+            # Afterward, subtract the amount from both the magic item's amount and capacity.
```
henriquegemignani: This `subtract` here is confusing. We're not doing on our side, we're sending a command to the game to decrement these values directly in the player's inventory.

randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -200,14 +234,21 @@ async def receive_remote_pickups(
...
+        """
+        Checks via the multiworld magic item whether a new pickup from the server needs to be sent to the game.
+        :return: Returns True if a new pickup to the game was sent, False otherwise.
+        """
 
         in_cooldown = self.message_cooldown > 0.0
         multiworld_magic_item = self.multiworld_magic_item
         magic_inv = inventory.get(multiworld_magic_item)
-        if magic_inv is None or magic_inv.amount > 0 or magic_inv.capacity >= len(remote_pickups) or in_cooldown:
+        # Skip sending if game has collected a location that we didn't check yet,
```
henriquegemignani: I don't like this kind of comment. It's only telling me what the code is doing and nowhere it touches the why.

There's 3 different conditions and it explain why these conditions are important.

randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -200,14 +234,21 @@ async def receive_remote_pickups(
         inventory: Inventory,
         remote_pickups: tuple[RemotePickup, ...],
     ) -> bool:
-        """Returns true if an operation was sent."""
+        """
+        Checks via the multiworld magic item whether a new pickup from the server needs to be sent to the game.
+        :return: Returns True if a new pickup to the game was sent, False otherwise.
```
henriquegemignani, line 240: I'd like to be precise. Pickups are never sent to the game, only code to be executed.

Miepee: tried rephrasing, not sure if it sounds too awkward now.

PR #8593, opened 2025-10-13
Add `benchmark print` command, optimize other commands

henriquegemignani:


randovania/cli/commands/benchmark.py
```diff
@@ -143,13 +143,45 @@ def compare_reports(
...
+    return [param.get_preset(0).game for param in parameters]
+
+
+def _extract_generator_params_bytes(link: str) -> bytes:
+    encoded_param = link.encode("utf-8")
+    encoded_param += b"=" * ((4 - len(encoded_param)) % 4)
+
+    b = base64.b64decode(encoded_param, altchars=b"-_", validate=True)
+    data = PermalinkBinary.parse(b).value
+    return data.generator_params
```
Miepee, line 175: don't we have a function for this somewhere in Permalink? Permalink.from_str(link).parameters or something like that?

randovania/cli/commands/benchmark.py
```diff
@@ -274,6 +302,10 @@ def add_commands(sub_parsers: typing.Any) -> None:
     repeat_parser.add_argument("input_file", type=Path, help="The session to repeat.")
     repeat_parser.set_defaults(func=repeat_logic)
 
+    print_parse: ArgumentParser = sub_parsers.add_parser("print", help="Print a report files")
```
Miepee: "a report files"?

randovania/cli/commands/benchmark.py
henriquegemignani: Yes but it's slow

Miepee: worth a comment then?

test/cli/commands/test_benchmark.py
```diff
@@ -77,7 +74,7 @@ def test_compare_reports() -> None:
         None,
     ]
 
-    benchmark.compare_reports(parameters, reference, results)
+    benchmark.compare_reports(games, reference, results)
```
Miepee, line 77: Funny type error here. Why python doesnt implicitly convert ints to floats is beyond me.
But maybe it'd be better to change the types from `list[float | None]` to `list[Number | None]`?

henriquegemignani: Ah it's in the tests. I remember running mypy and was wondering why I missed something.

Ints aren't floats, I'm glad python typing complains about it.

PR #8594, opened 2025-10-13
Change distances_to_node to use a GameDatabaseView

henriquegemignani:
And a few other things by consequence.

### DIFF
--- a/randovania/game_description/game_database_view.py
+++ b/randovania/game_description/game_database_view.py
@@ -212,7 +212,8 @@ class GameDatabaseView(ABC):
     @final
     def find_area_by_identifier(self, identifier: AreaIdentifier) -> tuple[Region, Area]:
         """
-        Searches an Area by the given identifier, using `node_iterator`.
+        Searches for an Area with the given identifier, using `node_iterator`. Slow operation.
+        Raises KeyError if no area could be found.
         """
         identifier_tuple = identifier.as_tuple
         for region, area, _ in self.node_iterator():


randovania/exporter/hints/pickup_hint.py
```diff
@@ -42,12 +42,11 @@ class PickupHint:
     pickup_name: str
 
 
-def _calculate_determiner(pickup_assignment: PickupAssignment, pickup: PickupEntry, region_list: RegionList) -> str:
+def _calculate_determiner(pickup_assignment: PickupAssignment, pickup: PickupEntry, game_view: GameDatabaseView) -> str:
     name_count: dict[str, int] = collections.defaultdict(int)
-    for i in range(region_list.num_pickup_nodes):
-        index = PickupIndex(i)
-        if index in pickup_assignment:
-            pickup_name = pickup_assignment[index].pickup.name
```
Miepee, line 50: I'm looking at this, doesn't this technically fix a potential bug too?
On main, we used to just to iterate from 0 until num_pickup_nodes, but wouldn't that mean that it could skip pickupnodes if there are empty spaces?
e.g. 2 pickup nodes, index 0 and index 120 would've only checked index 0 and index 1 here
i have no clue whether that's something that can happen in practice in our games, but probably?

randovania/game_description/game_database_view.py
```diff
@@ -208,6 +209,17 @@ def typed_node_by_identifier[NodeT: Node](self, i: NodeIdentifier, t: type[NodeT
             return result
         raise KeyError(f"Node at {i} is an {type(result)}, not {t}")
 
+    @final
+    def find_area_by_identifier(self, identifier: AreaIdentifier) -> tuple[Region, Area]:
```
Miepee, line 213: does this being slower than the region_list's one due to lack of cache matter?

randovania/exporter/hints/pickup_hint.py
henriquegemignani: Yes, it's a possible bug indeed. With the drastic result of your pickup being hinted with `A` instead of `The` or vice versa, certainly a blocker.

I remember Super Metroid being a game that had non-continuous pickup indices (because it caused bugs when added years ago 😄 ), but not sure about any other game.

randovania/game_description/game_database_view.py
henriquegemignani: My goal is that nothing that cares about performance should use GameDatabaseView.

But I like to document things that search in a slow way as things that are slow, so users know to not call multiple times.

PR #8601, opened 2025-10-14
Prime: Use Beam templates over beam resources

Miepee:


CHANGELOG.md
```diff
@@ -60,6 +61,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added: Chozo Ice Temple: Documentation for reaching the Chozo Statue NSJ.
 - Fixed: Chozo Ice Temple: Removed unnecessary dangerous requirements.
 - Fixed: Control Tower: Removed unnecessary dangerous requirements.
+- Fixed: Research Entrance: ???
```
Miepee: Image
These requirements are for there and back. But with the way this is laid out, I neither know what this is trying to model, now if I'm actually fixing a logic bug here or not.

Samuel6710: This is modeling the pirate fight that happens on your first pass through. I'm surprised that there isn't an event for that, unless it was removed?

Miepee: I at least didnt remove it. And i dont think the game ever had an event removed either.

Miepee: ok, turns out this was not fixing a logic bug. so no need to add to changelog.

PR #8603, opened 2025-10-14
Create centralized storage of Presets for server

henriquegemignani:
This table stores all presets identified by their sha256 hash, ensuring that duplicated presets do not use more storage than needed. We also store presets compressed.

There's some basic tooling for the server storing the preset received as is, without updating to latest version. One step further to supporting old clients in latest server.

A follow-up from this is an API for getting a Preset from a given hash, and then adding that hash to the `ServerWorldResponse` for `ServerWorldSync`, combined with the client storing a database of hash -> preset locally too. Working towards changing the received pickups list being a list of pickup names. Which not only removes the garbage code of encoding PickupEntry, but also is another necessary step for supporting old clients in latest server.


henriquegemignani:
I'm not sure what to think of the database migration. It'll take forever to run in prod, a bit scared. Also would like write it in a way without `World.preset` existing, because I also want to drop that column.

PR #8604, opened 2025-10-15
Update dependencies

henriquegemignani:
Goal: update everything that doesn't need to change a single line of code.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ dependencies = [
     # Versioning presets
     "dulwich>=0.20,<0.23.0",   # Causes a crash with pyinstaller on windows when using ProcessPoolExecutor
     # error reports
-    "sentry-sdk<2.41.0",
+    "sentry-sdk<2.41.0",  # Needs code fixes to update. Some import got renamed?
     # Patch GamesPatches JSON from rdvgame
     "json-delta",
     # For the Version class


henriquegemignani:
Hmm this changes open-samus-returns-rando

pyproject.toml
```diff
@@ -48,9 +48,9 @@ dependencies = [
     # Immutable dicts, used in many places
     "frozendict",
     # Versioning presets
-    "dulwich>=0.20",
+    "dulwich>=0.20,<0.23.0",   # Causes a crash with pyinstaller on windows when using ProcessPoolExecutor
     # error reports
-    "sentry-sdk",
+    "sentry-sdk<2.41.0",
```
Miepee: why do we need to restrict the version?

pyproject.toml
```diff
@@ -97,7 +97,7 @@ exporters = [
 ]
 
 gui = [
-    "PySide6-Essentials>=6.5",
+    "PySide6-Essentials==6.9.1",
```
Miepee, line 100: why do we need to lock on the specific one?

pyproject.toml
```diff
@@ -166,7 +166,7 @@ installer = [
 ]
 
 typing = [
-    "mypy",
+    "mypy<1.16.0",  # anything newer needs code changes
```
Miepee, line 169: what kind of code changes? wouldnt it be better to make those first and then update mypy?

pyproject.toml
henriquegemignani: broke stuff. didn't look into it

pyproject.toml
henriquegemignani: gotta at least to re-generate the ui files. That wasn't in the scope of my pr

pyproject.toml
henriquegemignani: "Goal: update everything that doesn't need to change a single line of code."

pyproject.toml
Miepee: can this get a comment?

PR #8611, opened 2025-10-16
Use zstd for compressing presets and rdvgames

henriquegemignani:


Miepee (commented):
looks fine minus mypy and codspeed failing

Miepee (approved):

PR #8616, opened 2025-10-17
Restore Session: Catch the correct exception for discord

henriquegemignani:


randovania/server/user_session.py
```diff
@@ -174,8 +174,8 @@ async def restore_user_session(sa: ServerApp, sid: str, encrypted_session: bytes
         await sa.sio.save_session(sid, {})
         raise
 
-    except (KeyError, peewee.DoesNotExist, json.JSONDecodeError, InvalidTokenError) as e:
-        # InvalidTokenError: discord token expired and couldn't renew
+    except (KeyError, peewee.DoesNotExist, json.JSONDecodeError, fastapi_discord.exceptions.Unauthorized) as e:
+        # Unauthorized: bad discord access token. Expired?
```
Miepee, line 178: do we not have a test for this? come this didn't scream before?

henriquegemignani: Sure, but all we could do is raise some random exception in our test and check that we caught that one. No way to really test if the third party code is throwing the exception we want.

Miepee: oh. i thought we were throwing the exception somewhere. bleh.

Miepee (approved):

PR #8618, opened 2025-10-17
Fusion: Add more location features

Miepee:
grafik
grafik

### DIFF
--- a/randovania/games/fusion/logic_database/header.json
+++ b/randovania/games/fusion/logic_database/header.json
@@ -2937,7 +2937,7 @@
             "long_name": "Mimic",
             "hint_details": [
                 "",
-                "guarded by an Item-Mimic"
+                "guarded by a Tank-Mimic"
             ]
         },
         "geron": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Fusion
 
-- Added: 3 new hint location features: guarded by an Item-Mimic, near a Geron and in a room with extreme temperatures.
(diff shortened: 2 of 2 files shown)


randovania/games/fusion/logic_database/header.json
```diff
@@ -2930,8 +2930,31 @@
...
                 "",
-                "in a room with more than one pickup."
+                "in a room with more than one pickup"
             ]
+        },
+        "mimic": {
+            "long_name": "Mimic",
+            "hint_details": [
+                "",
+                "guarded by an Item-Mimic"
```
JeffGainsNGames: thoughts on `near` vs `guarded`?

Miepee: i feel like guarded sounds nicer? all of the mimics are always right in front of the item

JeffGainsNGames: sure sounds nicer, was just thinking `near` is same format as many other hints so could be easier to parse? 
really minor so not blocking just thought I'd raise it in case anyone had strong feelings, in the end players will likely just be catching `Item-Mimic`

Miepee: I think that `guard` is easier cause then you see that it *is* something else then all the other hints.
`Item-Mimic` was a term that I was actually unsure on. I mostly just didn't want to use `Mimic` directly to avoid the confusion of "all X are mimicking enemies!"

JeffGainsNGames: Pickup-mimic
Expansion-mimic
Tank-mimic
Imposter 👀
...
Item-mimic is probably the most immediately recognized

Miepee: tank mimic sounds slightly nicer?
also gets rid of "item" which we don't really want to use

JeffGainsNGames: I just don't know if users will connect missile/pb-tank over army tank first
After reading it once sure

PR #8619, opened 2025-10-17
Prime: Some debug code to help figuring out why magic item increases by 2

Miepee:
The magic item should only ever increase by 1. 
However sometimes it increases by 2 and we don't know why. Let's add some logging to figure out if 1) this is the place where it comes from 2) what happens within execute_remote_patches that causes it to happen

### DIFF
--- a/test/game_connection/connector/test_prime1_remote_connector.py
+++ b/test/game_connection/connector/test_prime1_remote_connector.py
@@ -160,7 +160,9 @@ async def test_interact_with_game(connector: Prime1RemoteConnector, depth: int, 
     connector.executor.is_connected.return_value = True
     connector.executor.disconnect = MagicMock()
 
-    connector.get_inventory = AsyncMock()
+    inventory_mock = MagicMock(spec=Inventory)
+    inventory_mock.get = MagicMock(return_value=MagicMock(spec=InventoryItem))
+    connector.get_inventory = AsyncMock(return_value=inventory_mock)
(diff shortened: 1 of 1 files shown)


randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -270,6 +270,16 @@ async def receive_remote_pickups(
...
         await self.execute_remote_patches(patches)
+
+        # Safety/debugging check to see if this is the spot where the magic item randomly increases by 2 instead of 1.
+        expected_capacity = magic_inv.capacity + 1
+        memory_ops = await self._memory_op_for_items([multiworld_magic_item])
+        op_result = await self.executor.perform_single_memory_operation(*memory_ops)
+        assert op_result is not None
+        magic_inv = InventoryItem(*struct.unpack(">II", op_result))
+        if magic_inv.capacity != expected_capacity:
+            self.logger.warning(f"Magic capacity should be '{expected_capacity}' but it was '{magic_inv.capacity}'")
```
henriquegemignani: Good chance this will always trigger on Dolphin. Probably not on Nintendont because network slow enough for a new frame to happen.

When `execute_remote_patches` ends, all it means we've finished writing to game memory the code to run. At the earliest it'll run next frame for the game.

Miepee: ah. meh. reworked to check after the game has ran the code

PR #8622, opened 2025-10-18
 Mercury Steam: support game beaten

Miepee:
Draft.
Needs an MSR PR merged, a dread PR opened+merged and a dread implementation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,7 @@ exporters = [
     "factorio-randovania-mod>=0.8.0",
 
     # Metroid Dread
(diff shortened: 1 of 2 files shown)


Miepee:
Still being blocked by https://github.com/randovania/open-dread-rando/pull/408 + but otherwise reviewable

henriquegemignani (approved):
I think there's room for this to be ran with old games so we we receive `nil` as the beaten state, but it should be fine. Unless that variable already existed before.

henriquegemignani:
Oh, changelog?

PR #8625, opened 2025-10-19
Bot: Report that an rdvgame is for racing

henriquegemignani:


test/discord_bot/test_preset_lookup.py
```diff
@@ -238,3 +240,46 @@ async def test_reply_for_preset(mocker):
...
+    layout = dataclasses.replace(
+        layout,
+        generator_parameters=dataclasses.replace(
+            layout.generator_parameters,
+            spoiler=has_spoiler,
+        ),
+    )
+
+    # Run
+    from randovania.discord_bot import preset_lookup
```
Miepee, line 260: whats the reasoning for importing in the test, rather than outside of it?

Miepee (approved):

test/discord_bot/test_preset_lookup.py
henriquegemignani: It can help with mock patcher, but I just followed what the other tests did in this file. Don't remember if there anything matters in this case.

Miepee: sounds familiar haha

PR #8626, opened 2025-10-19
Build universal2 executables for macOS

henriquegemignani:
Test build from CI

### DIFF
--- a/randovania.spec
+++ b/randovania.spec
@@ -75,6 +75,7 @@ exe = EXE(
     upx=False,
     icon=icon_path,
     console=True,
+    target_arch="universal2",
 )
 coll = COLLECT(exe, a.binaries, a.zipfiles, a.datas, strip=False, upx=False, name="randovania")
 app = BUNDLE(


henriquegemignani:
> PyInstaller.utils.osx.IncompatibleBinaryArchError: /Users/runner/work/randovania/randovania/.venv/lib/python3.12/site-packages/_cffi_backend.cpython-312-darwin.so is not a fat binary! For details about this error message, see: https://pyinstaller.org/en/stable/feature-notes.html#macos-multi-arch-support

From that doc:

> The error message will typically be either "{name} does not contain slice for {target_arch}!" (when trying to build a universal2 program and the collected binary is a thin single-arch file) or "{name} is incompatible with target arch {target_arch} (has arch: ...)!" (when trying to build program for foreign architecture in a partial universal2 environment and the collected binary is a thin single-arch file for the non-target architecture).

> In such cases, creating frozen application for the selected target architecture will not be possible unless the problem of missing arch slices is manually addressed; for example, by downloading the wheel corresponding to the missing architecture, and stiching the offending binary files together using the lipo utility. You can also use 3rd party utilities, such as delocate-merge from the [delocate](https://pypi.org/project/delocate) project, to merge two single-arch wheels for a package into a universal2 wheel, and then install the merged wheel into your build environment.

Aka this is going to be really, really annoying to do.

PR #8633, opened 2025-10-19
Echoes: Support doing new patcher only for Echoes

henriquegemignani:
~~The title is a lie, because when it gets to actually exporting we do nothing.~~ it actually exports. Not that you get anything usable but that's on OPR

### DIFF
--- a/randovania/games/prime2/exporter/game_exporter.py
+++ b/randovania/games/prime2/exporter/game_exporter.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import copy
 import shutil
+import tempfile
+from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


henriquegemignani:
Alright let's get this in :) @duncathan

randovania/games/prime2/exporter/patch_data_factory.py
JeffGainsNGames, line 711: can you explain why elevators are checking for new patcher but not things like translator gates?

duncathan: this is necessary to work around a bug that comes up when using both patchers

henriquegemignani: I would say it's half that, half part of the old plan of slowly migrating features to be handled by the new patcher.

PR #8634, opened 2025-10-19
Change Fusion to use the same generic backends as other games

Miepee:


randovania/exporter/hints/credits_spoiler.py
```diff
@@ -67,40 +67,69 @@ def generic_credits(
...
+        all_patches, players_config
+    )
 
     starting_pickups = starting_pickups_with_count(all_patches[players_config.player_index])
     for pickup, count in starting_pickups.items():
         if count < 1:
             continue
 
-        if pickup not in major_pickups_spoiler:
-            major_pickups_spoiler[pickup] = []
```
Miepee, line 94: @henriquegemignani These lines cause it to add minor pickups to the spoiler. Was that intended? I currently left the functionality but got surprised seeing missile and pb tanks showing up in the credits

henriquegemignani: These lines in particular? So the missiles don't exist in `major_pickups_spoiler`, but exists in `starting_pickups`?

Miepee: yeah

henriquegemignani: With the conditions you listed, the only way for that to happen is if these minors are random starting pickups.
If that's not the case, then I don't see how it could be this line here.

Miepee: That is the case yes. I'm asking if  it was intentional whether minor pickups should be listed in the credits if they're random starting items. It feels weird since the credits usually only ever has majors.

henriquegemignani: I haven't ever thought about that case. Probably, could change `starting_pickups_with_count` to filter for only those that should show on credits.

Miepee: Cool, then i'll make a seperate PR for that.

randovania/exporter/hints/credits_spoiler.py
```diff
@@ -67,40 +67,67 @@ def generic_credits(
...
         if count > 1:
             msg += f" ({count} copies)"
-        major_pickups_spoiler[pickup].append(msg)
+        details[pickup].append(msg)
+
+    for pickup, entries in details.items():
+        if not entries:
+            entries.append("Nowhere")
+
+    return dict(sorted(details.items(), key=sort_pickup))
```
henriquegemignani: This could just be a `list[tuple[....]]` instead of a dict.

randovania/exporter/hints/credits_spoiler.py
```diff
@@ -67,40 +67,67 @@ def generic_credits(
...
-    def sort_pickup(p: PickupEntry) -> tuple[float, str]:
+    def sort_pickup(tup: tuple[PickupEntry, list[OwnedPickupLocation | str]]) -> tuple[float, str]:
+        p = tup[0]
         return major_pickup_name_order.get(p.name, math.inf), p.name
 
-    details = get_locations_for_major_pickups_and_keys(all_patches, players_config)
-    major_pickups_spoiler = {
-        pickup: [entry.export(namer, use_player_color) for entry in entries] for pickup, entries in details.items()
-    }
+    details: dict[PickupEntry, list] = get_locations_for_major_pickups_and_keys(all_patches, players_config)
```
henriquegemignani: Maybe just change the return value of `get_locations_for_major_pickups_and_keys` to be `list[OwnedPickupLocation | str]`?

test/test_files/patcher_data/fusion/fusion/starting_items/world_1.json
```diff
@@ -3639,6 +3639,26 @@
         {
             "LineType": "White1",
             "Text": "Aquarium Kago Storage",
+            "BlankLines": 1
+        },
+        {
+            "LineType": "Red",
+            "Text": "Missile Tank",
```
henriquegemignani: Didn't you want to change so minors don't show up?

Miepee: I did in another pr. I just have to rebase this one.

randovania/exporter/hints/credits_spoiler.py
Miepee: I thought about it, but it feels weird having a purposefully wrong signature only for it to make it easier in the single place where we use the function

randovania/exporter/hints/credits_spoiler.py
Miepee: Wouldn't you then loose out on the pickup entry tho?
Or do you mean a `tuple[PickupEntry, list[...]]`?

henriquegemignani: The later. Can just drop the `dict` call and all the `.items()` calls in the callers.

randovania/exporter/hints/credits_spoiler.py
henriquegemignani: The single place which is also the single use :)

randovania/exporter/hints/credits_spoiler.py
Miepee: Done

randovania/exporter/hints/credits_spoiler.py
```diff
@@ -69,40 +69,68 @@ def generic_credits(
...
```
henriquegemignani: the explicit hint here shouldn't be needed anymore

randovania/games/fusion/exporter/patch_data_factory.py
```diff
@@ -371,12 +371,17 @@ def _create_nav_text(self) -> dict:
...
+        majors = credits_spoiler.generic_credits(
+            self.configuration.standard_pickup_configuration, self.description.all_patches, self.players_config
         )
 
-        for pickup, locations in majors.items():
+        for pickup, locations in majors:
             for location in locations:
+                # Special case for special messages
+                if isinstance(location, str):
+                    elements[pickup.name].append({"World": None, "Region": "", "Area": location})
```
henriquegemignani, line 382: update some test case to hit this

randovania/exporter/hints/credits_spoiler.py
```diff
@@ -69,40 +69,68 @@ def generic_credits(
...
 
         msg = "As a random starting item"
         if count > 1:
             msg += f" ({count} copies)"
-        major_pickups_spoiler[pickup].append(msg)
+        details[pickup].append(msg)
+
+    for pickup, entries in details.items():
+        if not entries:
+            entries.append("Nowhere")
```
henriquegemignani, line 105: can you update some test case to hit this?

henriquegemignani (changes requested):

randovania/exporter/hints/credits_spoiler.py
Miepee: I mostly added it cause the old code had the same thing.
But honestly, I don't even know *how* it can be hit. 
The `details` dict provided by `get_locations_for_major_pickups_and_keys` only has a pickup key, if it also has an entries value. 
The starting pickups loop above only ever adds a new pickup to the dictionary with one message

So from what I can see, this was some weird failsafe that maybe got triggered in the past that nowadays can never get hit?

henriquegemignani: ... it's a failsafe supposed to catch a case which I guess the case never handled properly. Cool.

Either an intentionally excluded pickup would show up as this, but plando where a pickup is in the pool but never placed.
I do have ideas of features that could make this relevant, so it's kinda neat to handle it. But guess it can be issue for the future time.

PR #8642, opened 2025-10-19
Optimize World Creation on layout import

Miepee:
Need to fix tests still.
I mostly opened the PR to get feedback on whether this approach is even okay to begin with.
The thought behind it was for clients to not update themselves every time when we're mass sending actions, because one session update is costly client side.
I initially thought about having some kind of "do-not-update-session" event, that we send first to the clients. But that seemed annoying and kinda risky? E.g. what happens if you just somehow miss the event where you were told to start listening to events again.

So instead I just opted for an option during the server admin_session function to determine whether we emit or not. Making that part of `*args` doesn't really work, and afaik i can't provide it as a named argument easily

### DIFF
--- a/randovania/gui/lib/multiplayer_session_api.py
+++ b/randovania/gui/lib/multiplayer_session_api.py
@@ -128,14 +128,14 @@ class MultiplayerSessionApi(QtCore.QObject):
         self.logger = SessionIdLoggerAdapter(_base_logger, self)
 
     async def _session_admin_global(
-        self, action: admin_actions.SessionAdminGlobalAction, *args: JsonType | bytes
+        self, action: admin_actions.SessionAdminGlobalAction, *args: JsonType | bytes, emit_event: bool = True
     ) -> JsonType | None:
         assert self.widget_root is not None
         try:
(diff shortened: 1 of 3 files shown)


henriquegemignani:
What about an endpoint directly about uploading a list of worlds to create at once, so we don't need that dumb for loop in the client? :)

Miepee:
Was thinking about that. But I assumed that that wouldn't work since I thought the add_audit_entry was the function causing the meta event, which I only found out later and forgot to revisit oops.

But also I think we need the client loop anyway? If only to create the world names. Unless you meant to move the loop to the server, and have it figure out on what names are valid to create with instead.

Miepee:
The one case about having a specific non-emitting parameter, is that we could potentially easily use it for other places too. E.g. right below said client loop there is another loop to update the presets which could benefit from this

henriquegemignani:
If the issue is the client keeps updating the UI, then add some sort of cooldown?
After an ui update happens, it could start a 5s timer where further updates are not applied, just saved.

Miepee:
Ah yeah that should work

PR #8654, opened 2025-10-21
Use certifi as HTTPS certificate source everywhere

henriquegemignani:


randovania/lib/http_lib.py
```diff
@@ -0,0 +1,13 @@
+import ssl
+
+import aiohttp
+import certifi
+
+
+def http_session(headers=None) -> aiohttp.ClientSession:
```
Miepee: mypy?

henriquegemignani: fixed

Miepee (approved):

PR #8660, opened 2025-10-22
Create universal2 Pyinstaller executable for macOS

henriquegemignani:


tools/bundle_macos_universal_wheels.py
```diff
@@ -0,0 +1,362 @@
+import asyncio
```
Miepee, line 19: can this file get a comment somewhere at the top of the file indicating what its for?

tools/bundle_macos_universal_wheels.py
```diff
@@ -0,0 +1,362 @@
...
+    # Find the oldest macOS version among the wheels for the universal2 wheel
+    min_version = None
+    for wheel_data in matching_wheels:
+        wheel_info = wheel_data["wheel_info"]
+        arch, version = parse_macos_platform(wheel_info["platform_tag"])
+        if min_version is None or version < min_version:
+            min_version = version
+
+    # Replace the platform tag with universal2 using the oldest version
+    major, minor = min_version
```
Miepee, line 202: mypy error :)

tools/bundle_macos_universal_wheels.py
```diff
@@ -0,0 +1,362 @@
...
+            if e.stderr:
+                print(f"    stderr: {e.stderr}")
+            return False
+
+
+def is_pure_python_wheel(wheel_info: dict) -> bool:
+    return wheel_info["abi_tag"] == "none" and wheel_info["platform_tag"] == "any" and wheel_info["python_tag"] == "py3"
+
+
+async def main() -> int:
```
Miepee, line 270: is the tldr of this file
- go through all packages
- search on pypi whether universal wheels exist
  - if yes install those
- if no, try if we can build a universal wheel / fetch intel and arm wheels
- put all the new wheels into specific folder

?

henriquegemignani: yeah. 

For building an universal executable we need:
- An universal python executable, we have that
- Every installed binary wheel needs to be universal2

tools/bundle_macos_universal_wheels.py
henriquegemignani: I hate these mypy errors that are just an extra assert.

tools/bundle_macos_universal_wheels.py
henriquegemignani: Have a comment

PR #8661, opened 2025-10-22
Generator: Get rid of duplicate actions

Miepee:
Did benchmarks for only 2 games cause i was lazy:
```
python -m randovania development benchmark compare before after
                    Difference | Fixes | Failures |     Mean |    Stdev |   Median
                 Metroid Dread |     0 |        0 |   -2.628 |    3.367 |   -2.506
                 
                 
python -m randovania development benchmark compare b a         
                    Difference | Fixes | Failures |     Mean |    Stdev |   Median
                Metroid Fusion |     1 |        3 |    0.171 |    1.667 |    0.195
```

Reason for fusion being different is because it hit a case of having duplicate actions. And since those got removed probably increased the chances of placing only keycards and locking itself out.

### DIFF
--- a/randovania/generator/filler/pickup_list.py
+++ b/randovania/generator/filler/pickup_list.py
@@ -174,16 +174,16 @@ def get_pickups_that_solves_unreachable(
 
     all_lists = _requirement_lists_without_satisfied_resources(state, possible_sets, uncollected_resources)
 
-    result = []
+    # Hacky way to get rid of duplicates. Can't use a set as a set's order is not deterministic.
+    result = {}
     for requirement_list in sorted(all_lists, key=lambda it: it.as_stable_sort_tuple):
         pickups = pickups_to_solve_list(pickups_left, requirement_list, state)
         if pickups is not None and pickups:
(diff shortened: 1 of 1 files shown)


Tyranisaur:
It's worth pointing out that the `single_set` config is basically always true, because it was migrated to true, and we don't expose it in the gui as far as I can tell. So there shouldn't be any duplicate requirements before starting to find pickups that solve them. So exactly how you can end up with true duplicates in these actions is a bit of a mystery to me. Otherwise, what I know we have are actions that use sets of items that are true subsets of other actions' items. And if you wanted to get rid of that kind of duplication, you would need to prune out any combination of items that is a true superset of another set of items.

Miepee:
> So there shouldn't be any duplicate requirements before starting to find pickups that solve them.

Henrique suggested it might be because of slightly different ammo requirements (ammo >=1 vs ammo >=2).
I haven't confirmed it, but it does sound plausible as to why Fusion has the Missile Launcher/Data showing up by itself twice (e.g. without it combined with something else).

henriquegemignani (approved):

PR #8662, opened 2025-10-23
keybearer hints choose a random pickup feature

duncathan:
quick and dirty. if this feels better to people we can begin the tuning phase. gaussian params can change as always, and I would like to explore bringing back "missile-related" and the like for these if there's a way we can keep it out of the pool for translator hints

### DIFF
--- a/test/games/prime2/generator/test_echoes_hint_distributor.py
+++ b/test/games/prime2/generator/test_echoes_hint_distributor.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import copy
 import dataclasses
 from random import Random
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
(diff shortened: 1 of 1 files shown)


JeffGainsNGames (approved):

randovania/games/prime2/pickup_database/pickup-database.json
```diff
@@ -186,21 +184,24 @@
             "hint_details": [
                 "a ",
                 "targeting system"
-            ]
+            ],
+            "hidden": true
```
Miepee, line 188: i'm curious: do we need to carry these around with us?
or could we say that migrate these away and just say you're EoL on "old rdvgames" if you generated them during this dev-testing period?

duncathan: we can remove them at any time if so desired

PR #8663, opened 2025-10-23
Don't import caver on startup

henriquegemignani:
This also reduces errors when running the GUI with just the `gui` extra.


henriquegemignani:
> Whats the reasoning for keeping the patcher references in gui/dialog/game_export_dialog.py? Feel like if we're already making our own enum, we might as well use it everywhere

I'm not sure I follow, since that file now no longer imports `caver` at all?

Miepee:
Ignore me, somehow missed the file during review

PR #8670, opened 2025-10-24
Hints: avoid picking same feature if precision is same

JeffGainsNGames:
image

feature chooser would always return the first item in the list of `possible_features` if multiple features had the same precision

put the shuffle as late as possible to preserve the debug prints sorting


Miepee:
changelog entry?

PR #8678, opened 2025-10-25
Admin Panel: Provide toggling admin status of a Member

Miepee:
grafik
grafik

Credits to the people who keep joining a 5+ world session and then leave not even midway through without giving anyone else the ability to claim worlds

### DIFF
--- a/randovania/server/multiplayer/web_api.py
+++ b/randovania/server/multiplayer/web_api.py
@@ -28,6 +28,11 @@ def RdvFileResponse(data: str, filename: str) -> Response:
     return Response(data, headers={"Content-Disposition": f"attachment; filename={filename}"})
 
 
+# For endpoints, there are 2 ways to  gate them behind being accessible only by an RDV-Admin:
+# - pass in an AdminDep as an argument
+# - pass in RequireAdminUser as a dependency in the decorator.
+
+
 @router.get("/sessions", response_class=HTMLResponse)
 def admin_sessions(
     sa: ServerAppDep, user: AdminDep, request: Request, page: Annotated[int, Query(ge=1)] = 1


randovania/server/multiplayer/web_api.py
```diff
@@ -139,6 +139,34 @@ def download_world_preset(user: AdminDep, world_id: int) -> Response:
...
 
 
+@router.post("/session/{session_id}/user/{user_id}/toggle_admin", response_class=HTMLResponse)
+def toggle_admin_status(sa: ServerAppDep, request: Request, session_id: int, user_id: int) -> HTMLResponse:
+    try:
+        membership = MultiplayerMembership.get_by_ids(user_id, session_id)
+        user = User.get_by_id(user_id)
+    except MultiplayerMembership.DoesNotExist:
+        raise HTTPException(status_code=404, detail="Membership not found")
+    except User.DoesNotExist:
```
henriquegemignani, line 165: Kinda redundant as MultiplayerMembership requires the user id to exist.

randovania/server/multiplayer/web_api.py
```diff
@@ -139,6 +139,34 @@ def download_world_preset(user: AdminDep, world_id: int) -> Response:
...
+        raise HTTPException(status_code=404, detail="User not found")
+
+    membership.admin = not membership.admin
```
henriquegemignani, line 170: This needs to add an audit log entry to the session.

randovania/server/multiplayer/web_api.py
```diff
@@ -139,6 +139,34 @@ def download_world_preset(user: AdminDep, world_id: int) -> Response:
     return RdvFileResponse(world.preset, f"{session.name} - {world.name}.rdvpreset")
 
 
+@router.post("/session/{session_id}/user/{user_id}/toggle_admin", response_class=HTMLResponse)
```
henriquegemignani: Missing a dep to `RequireAdminUser`

randovania/server/multiplayer/web_api.py
Miepee: Yeah. Only really added it as a failsafe, so that if this for whatever reason ever decides to throw, that the function still returns cleanly

randovania/server/multiplayer/web_api.py
Miepee: Done

PR #8684, opened 2025-10-27
Web api for viewing async rooms

Miepee:
grafik
grafik

Still need to fix tests, but wanted to open the pr already beforehand

Nice usecase about it is that it helps us admins get the spoiler log in case we need to investigate something while the race is still going

### DIFF
--- a/randovania/server/async_race/web_api.py
+++ b/randovania/server/async_race/web_api.py
@@ -1,0 +1,89 @@
+from math import ceil
+from typing import Annotated
+
+from fastapi import APIRouter, HTTPException, Query, Request, Response
+from fastapi.responses import HTMLResponse
+
+from randovania.lib import json_lib
+from randovania.server.database import (
+    AsyncRaceEntry,
+    AsyncRaceRoom,
+)
+from randovania.server.server_app import AdminDep, ServerApp, ServerAppDep
+
+router = APIRouter()
+
+
+def RdvFileResponse(data: str, filename: str) -> Response:
+    return Response(data, headers={"Content-Disposition": f"attachment; filename={filename}"})
+
+
+@router.get("/async-race-rooms", response_class=HTMLResponse)
+def admin_async_rooms(
+    sa: ServerAppDep, user: AdminDep, request: Request, page: Annotated[int, Query(ge=1)] = 1
+) -> HTMLResponse:
+    page_count = ceil(AsyncRaceRoom.select().count() / 20)
+
+    if not page_count:
+        return sa.templates.TemplateResponse(
+            request,
+            "basic_page.html.jinja",
+            context={
+                "title": "Async Race Rooms",
+                "header": "Async Race Rooms",
+                "content": "<p>No Async Race Rooms.</p>",
+            },
+        )
+
+    if page > page_count:
+        raise HTTPException(status_code=404, detail="Async Race Room page not found")
+
+    order = AsyncRaceRoom.creation_date.desc()  # type: ignore[attr-defined]
+    paginated_query = AsyncRaceRoom.select().order_by(order).paginate(page, 20)
+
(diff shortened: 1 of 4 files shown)


randovania/server/async_race/web_api.py
```diff
@@ -0,0 +1,89 @@
...
+from randovania.server.database import (
+    AsyncRaceEntry,
+    AsyncRaceRoom,
+)
+from randovania.server.server_app import AdminDep, ServerApp, ServerAppDep
+
+router = APIRouter()
+
+
+def RdvFileResponse(data: str, filename: str) -> Response:
```
duncathan, line 17: this shouldn't be redefined

randovania/server/async_race/web_api.py
```diff
@@ -0,0 +1,89 @@
...
+    return Response(data, headers={"Content-Disposition": f"attachment; filename={filename}"})
+
+
+@router.get("/async-race-rooms", response_class=HTMLResponse)
+def admin_async_rooms(
```
duncathan, line 22: enough of this code comes from the session page that it should probably be split out into a dedicated pagination function

randovania/server/templates/web_api/async_room.html.jinja
```diff
@@ -0,0 +1,94 @@
...
+                        <td>
+                            {% if entry.finish_datetime != None %}
+                                {{ entry.finish_datetime.strftime("%H:%M:%S.%f %Z, %d %B %Y") }}
+                            {% else %}
+                                Has not finished yet
+                            {% endif %}
+                        </td>
+                        <td>
+                            {% if entry.start_datetime != None and entry.finish_datetime != None %}
+                                {{ (entry.finish_datetime - entry.start_datetime)}}
```
duncathan, line 77: is this the correct value? shouldn't it be accounting for pauses?

randovania/server/templates/web_api/asyncs_list.html.jinja
```diff
@@ -0,0 +1,54 @@
...
+                <td>Creator</td>
+                <td>Creation Date</td>
+            </thead>
+            <tbody>
+                {% for room in rooms %}
+                    <tr>
+                        <td><a href="{{ url_for('admin_async_room', async_room_id=room.id) }}">
+                            {{ room.name }}
+                        </a></td>
+                        <td>{{ room.layout_description.all_patches[0].game.game.long_name }}</td>
```
duncathan, line 29: shouldn't this account for multiworld asyncs?

duncathan (changes requested):

randovania/server/templates/web_api/async_room.html.jinja
Miepee: oh hm. didn't think about that. probably should.

randovania/server/templates/web_api/asyncs_list.html.jinja
Miepee: probably should.

randovania/server/async_race/web_api.py
```diff
@@ -0,0 +1,89 @@
...
+@router.get("/async-race-room/{async_room_id}", response_class=HTMLResponse)
+def admin_async_room(sa: ServerAppDep, user: AdminDep, request: Request, async_room_id: int) -> HTMLResponse:
+    try:
+        async_room: AsyncRaceRoom = AsyncRaceRoom.get_by_id(async_room_id)
+    except AsyncRaceRoom.DoesNotExist:
+        raise HTTPException(status_code=404, detail="Async Room not found")
+
+    user_entries = list(AsyncRaceEntry.select().where(AsyncRaceEntry.room == async_room))
+
+    print(user_entries)
```
henriquegemignani, line 64: print

PR #8690, opened 2025-10-28
REST API access from clients

henriquegemignani:
Plus improve db access in FastAPI tests.

### DIFF
--- a/randovania/network_client/network_client.py
+++ b/randovania/network_client/network_client.py
@@ -134,6 +134,18 @@ emyPxgcYxn/eR44/KJ4EBs+lVDR3veyJm+kXQ99b21/+jh5Xos1AnX5iItreGCc=
 -----END CERTIFICATE-----"""
 
 
+def _apply_default_rest_headers(kwargs: _RequestOptions) -> None:
+    """Adds the default headers we should send to every REST request."""
+    headers = kwargs.get("headers")
+    if headers is None:
+        kwargs["headers"] = headers = {}
+    else:
+        assert isinstance(headers, dict)
+
+    if "Accept" not in headers:
+        headers["Accept"] = "application/json"
+
+
 class NetworkClient:
     sio: socketio.AsyncClient
     _current_user: CurrentUser | None = None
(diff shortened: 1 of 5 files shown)


randovania/network_client/network_client.py
```diff
@@ -713,3 +726,30 @@ def allow_reporting_username(self) -> bool:
...
         self._allow_reporting_username = value
         self._update_reported_username()
+
+    def server_get(
+        self,
+        url: str,
+        headers: http_lib.LooseHeaders | None = None,
+        **kwargs: Unpack[_RequestOptions],
+    ) -> _RequestContextManager:
+        """Perform HTTP GET request to Randovania's REST Server."""
```
duncathan: it's not super obvious at a glance that these are returning async context managers - maybe include an example snippet in each docstring showing the `async with` usage

randovania/server/server_app.py
```diff
@@ -406,38 +427,66 @@ async def server_app(request: fastapi.Request) -> ServerApp:
...
-                if user is None or (needs_admin and not user.admin):
-                    raise fastapi.HTTPException(status_code=403, detail="User not authorized")
-            else:
-                user = User.get()
+    user_id: int | None = request.session.get("user_id")
+    discord_token = request.session.get("discord_oauth_token")
 
-            return user
+    # TODO: check for user tokens
+    # maybe make sure tokens have a prefix so we know it's token vs encrypted session?
```
duncathan: could you clarify this TODO? is this related to a change in behavior, or did the old function have this deficiency as well?

randovania/server/user_session.py
```diff
@@ -329,18 +338,53 @@ async def browser_discord_login_callback(
...
             "user": user,
-            "is_admin": user.admin or sa.app.debug,
+            "is_admin": user.admin,
         },
     )
 
 
+@router.get("/guest_login")
+async def guest_login(sa: ServerAppDep, request: Request) -> Response:
+    if not sa.app.debug:
```
duncathan, line 358: this is inconsistent with the check performed by the existing login method: https://github.com/randovania/randovania/blob/76f51f214ed5dd9f34131f18bab878b48005dcb6/randovania/server/user_session.py#L119-L121

randovania/server/user_session.py
```diff
@@ -329,18 +338,53 @@ async def browser_discord_login_callback(
...
+
+    return sa.templates.TemplateResponse(
+        request,
+        "user_session/guest_login.html.jinja",
+    )
+
+
+@router.post("/guest_login")
+async def guest_login_post(sa: ServerAppDep, request: Request, name: typing.Annotated[str, Form()]) -> Response:
+    if not sa.app.debug:
```
duncathan, line 369: likewise

randovania/network_client/network_client.py
henriquegemignani: I copy pasted these from aiohttp itself :p
But I also find it confusing there too

randovania/server/server_app.py
henriquegemignani: Feature change. Before you only have an encrypted session you got from the login method, but for the access token I'm skipping the login and allowing you to just send the token in the header directly.

randovania/server/user_session.py
henriquegemignani: Intentional. I don't believe it's possible to do via a web form, or at the very least loses all meaning. Can work via purely API, but this was more for pure-browser in Web api.

I also can't remember why I went with that elaborate method. I follow what it's doing, it's safe against capturing a request and replaying it later, plus needs a shared secret between host and client. Only reason I can think of is making it somewhat safe to use in production, which I'm fine removing.

Oh and I do intend to remove the socketio flow, just in follow up PRs.

duncathan: if the follow up pr will also be removing the guest secret stuff, I suppose that's fine then

randovania/network_client/network_client.py
duncathan: I think a quick snippet would still be helpful. just like
```py
async with self.server_get("/guest_login") as response: 
    status = result.status
    body = await response.json()
```

henriquegemignani:
I broke tests by increasing coverage in the last commit :p

duncathan:
will merge after release

PR #8695, opened 2025-10-29
LiveSplit One integration

henriquegemignani:
It's not fully wip but I still need some tests. And maybe make the integration a little bit more robust.

### DIFF
--- a/test/server/async_race/test_room_api.py
+++ b/test/server/async_race/test_room_api.py
@@ -792,7 +792,6 @@ async def test_get_livesplit_url(test_client, simple_room):
     user = User.get_by_id(1235)
 
     test_client.sa.get_current_user = AsyncMock(return_value=user)
-    test_client.sa.sio.enter_room = AsyncMock()
 
     url = await room_api.get_livesplit_url(
         test_client.sa,
@@ -803,7 +802,6 @@ async def test_get_livesplit_url(test_client, simple_room):
     token = url.split("/")[-1]
     decoded = test_client.sa.decrypt_str(base64.urlsafe_b64decode(token))
     assert decoded == "1235/1"
-    test_client.sa.sio.enter_room.assert_awaited_once_with(sid, "async-race-1-1235", namespace="/")
 
 
 async def test_get_livesplit_url_not_member(test_client, simple_room):


randovania/server/async_race/room_api.py
```diff
@@ -502,7 +520,110 @@ async def submit_proof(sa: ServerApp, sid: str, room_id: int, submission_notes:
...
+    await sa.sio.emit(
+        signals.ASYNC_RACE_ROOM_UPDATE,
+        (await room.create_session_entry(sa, user)).as_json,
+        namespace="/",
+        to=_get_async_race_socketio_room(room, user),
+    )
+
+
+async def emit_async_room_message(sa: ServerApp, room: AsyncRaceRoom, sid: str, message: str) -> None:
+    pass
```
toasterparty: Dead code?

randovania/server/async_race/room_api.py
```diff
@@ -502,7 +520,110 @@ async def submit_proof(sa: ServerApp, sid: str, room_id: int, submission_notes:
...
+async def get_livesplit_url(sa: ServerApp, sid: str, room_id: int) -> str:
+    room = AsyncRaceRoom.get_by_id(room_id)
+    user = await sa.get_current_user(sid)
+    entry = database.AsyncRaceEntry.entry_for(room, user)
+    if entry is None:
+        raise error.NotAuthorizedForActionError
+
+    await sa.ensure_in_room(sid, _get_async_race_socketio_room(room, user))
+
+    token = base64.urlsafe_b64encode(sa.encrypt_str(f"{user.id}/{sid}")).decode("ascii")
```
toasterparty: Is it fine for these URLs to live forever? Is there a world where we want to have a TTL on these?

randovania/server/async_race/room_api.py
```diff
@@ -502,7 +520,110 @@ async def submit_proof(sa: ServerApp, sid: str, room_id: int, submission_notes:
...
+    websocket: WebSocket,
+    room_id: int,
+    token: str,
+) -> None:
+    app: RdvFastAPI = websocket.app
+
+    # Extract the user id and sid from an encrypted value.
+    # The only way to get a valid token is to call `get_livesplit_url`, which checks if the user is an
+    # active member of the race, solving the lack of authentication needed in this endpoint.
+    user_id_str, sid = app.sa.decrypt_str(base64.urlsafe_b64decode(token)).split("/")
```
toasterparty: Is it intentional that you don't check the sid here?

randovania/server/async_race/room_api.py
henriquegemignani: Oh, yeah. I gave up on this feature. Was thinking of having the server tell the client of some errors, like "bad data from livesplit received" but didn't figure out the client side of how to display

randovania/server/async_race/room_api.py
henriquegemignani: Huh, yeah it ended up unused.
I could have the `emit_async_room_update` be emitted directly to the `sid`, but I don't see why not send to the room.
Could actually change the sid with the `room_id` to confirm the token is for this room in particular.

Also made me realise that if you re-open the client, the sid will change and you won't be in the room, so you stop receiving the updates caused by livesplit, which would _continue_ to work.

randovania/server/async_race/room_api.py
henriquegemignani: The user never deletes the URL from their live split, then goes do something else after and presses "start-over" which will undo their finished status from the async?

Though I can't think of how a TTL would fix this case.

It's not a security critical URL, but helps mitigate against someone playing with the URL and breaking things.

randovania/server/async_race/room_api.py
henriquegemignani: Changed to room id instead of sid, solving actually using the token for a different room.

PR #8701, opened 2025-10-31
Use REST api for Login

henriquegemignani:


randovania/server/server_app.py
```diff
@@ -84,7 +85,6 @@ class ServerApp:
     db: peewee.SqliteDatabase
     metrics: Instrumentator
     fernet_encrypt: Fernet
-    guest_encrypt: Fernet | None = None
```
duncathan, line 87: I assume that the security implications here are "guest login is officially considered non-secure"?

henriquegemignani: When making this PR I failed to remember my thought process on having the workflow for guest auth, since HTTPS solves all the issues.

But right now I think I can think of one reason: restricting who can use guest login to these that know the secret the server is using for it. But that could just as easily be handled by a password?

But anyway, guest login has always been non-secure as you could keep creating new accounts.

PR #8706, opened 2025-11-01
Store WorldData files in a per-version folder

henriquegemignani:


randovania/interface_common/world_database.py
```diff
@@ -85,36 +87,42 @@ def __init__(self, persist_path: Path):
...
         )
 
-    async def load_existing_data(self):
-        for f in self._persist_path.glob("*.json"):
+    async def load_existing_data(self) -> None:
+        paths = [self._persist_path]
+        for version in range(2, CURRENT_VERSION + 1):  # versioned folder started at 2
+            paths.append(self._persist_path.joinpath(str(version)))
+
+        for f in itertools.chain.from_iterable([path.glob("*.json") for path in reversed(paths)]):
```
Miepee, line 117: since its iterating through all data in reverse, does this not cause a scenario where e.g. different versions have the same world uuids, causing rdv to only read the earlier one?

henriquegemignani: The last element of `paths` is the current version, so it'll read current version first. And it doesn't re-read the same UUID. I also added a test precisely for that case.

Miepee: ah i'm blind, whoops

Miepee (approved):

PR #8710, opened 2025-11-02
Add new rule that DockNode can only connect to a single area node

henriquegemignani:
We've had issues here and there regarding using DockNodes as path inside of an Area. 
Prime/Echoes also has weird issue with one-way anywhere elevators + door lock rando where taking the elevator can be logically a way of breaking a door lock.
Some weirdness in new graph is caused by routing through docks with locks you can't currently break.

This database migration creates a node called `Front of <>` for every Dock Node that has more that one area connection or is spawn point.

Is this a good idea? I'll fix the tests further if we think this is good.


henriquegemignani:
Updated to limit the migration and rule to:
- Dock Nodes with a the default weakness that has a lock
- Dock Nodes of a type that door lock rando allows to be shuffled into a weakness with a lock

Also changed the `valid_starting_location` migration to be only for Prime 1 and 2, as that's mostly a concern in these games.

henriquegemignani:
I'll try this differently.

PR #8712, opened 2025-11-02
Optimize Generator: Use Requirement instead of RequirementSet

henriquegemignani:
Small test sample size for starter preset:
Dread: 11.3s -> 9.3s
Echoes: 5.5s -> 4.5s
Prime 1: 3.1s -> 2.6s

### DIFF
--- a/randovania/generator/old_generator_reach.py
+++ b/randovania/generator/old_generator_reach.py
@@ -8,7 +8,6 @@ from typing import TYPE_CHECKING, NamedTuple, Self, override
 from randovania.game_description.db.resource_node import ResourceNode
 from randovania.game_description.requirements.base import Requirement
 from randovania.game_description.requirements.requirement_and import RequirementAnd
-from randovania.game_description.requirements.requirement_set import RequirementSet
 from randovania.game_description.requirements.resource_requirement import ResourceRequirement
 from randovania.generator import graph as graph_module
 from randovania.generator.generator_reach import GeneratorReach
@@ -18,6 +17,8 @@ if TYPE_CHECKING:
 
     from randovania.game_description.db.node import Node, NodeContext, NodeIndex
     from randovania.game_description.game_description import GameDescription
+    from randovania.game_description.requirements.requirement_set import RequirementSet
+    from randovania.game_description.resources.resource_info import ResourceInfo
     from randovania.generator.filler.filler_configuration import FillerConfiguration
     from randovania.resolver.state import State
 
@@ -41,7 +42,7 @@ def _extra_requirement_for_node(game: GameDescription, context: NodeContext, nod
 class GraphPath(NamedTuple):
     previous_node: Node | None
     node: Node
-    requirement: RequirementSet
+    requirement: Requirement
 
     def is_in_graph(self, digraph: graph_module.BaseGraph) -> bool:
         if self.previous_node is None:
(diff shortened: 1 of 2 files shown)


randovania/generator/old_generator_reach.py
```diff
@@ -128,7 +128,7 @@ def _potential_nodes_from(self, node: Node) -> Iterator[tuple[Node, RequirementS
 
             yield (
                 target_node,
-                requirement.patch_requirements(1.0, context).as_set(context),
+                requirement,
```
Miepee, line 131: I thought that a RequirementSet was specifically optimized for speed?

randovania/generator/old_generator_reach.py
```diff
@@ -371,11 +376,11 @@ def unreachable_nodes_with_requirements(self) -> dict[Node, RequirementSet]:
                 to_check.append((target_node, requirement))
 
         for node, requirement in to_check:
-            requirements = requirement.patch_requirements(context)
-            if node in results:
-                results[node] = results[node].expand_alternatives(requirements)
+            requirements = requirement.patch_requirements(1.0, context).as_set(context)
```
Miepee, line 379: in addition to the above comment, if RequirementSet is slow then why wasn't this changed to return a normal Requirement?

randovania/generator/old_generator_reach.py
```diff
@@ -343,20 +343,25 @@ def act_on(self, node: ResourceNode) -> None:
...
                         edges_to_remove.append((source, target))
 
             for edge in edges_to_remove:
                 self._digraph.remove_edge(*edge)
 
         self.advance_to(new_state)
 
-    def unreachable_nodes_with_requirements(self) -> dict[Node, RequirementSet]:
-        results: dict[Node, RequirementSet] = {}
+    def unreachable_nodes_with_requirements(self) -> dict[NodeIndex, RequirementSet]:
```
Miepee, line 363: I don't get the reasoning why we want to return NodeIndex as the key instead of Node
Isn't that more annoying for the caller?

randovania/generator/old_generator_reach.py
henriquegemignani: Not exactly. I think the most important part it's an `or` of `ands`. And it ensures there's nothing like `(A and B) or (A and B and C)`. So depending on how much redundancy your requirements have, it can be a lot faster! Factorio in particular likes it.

But precisely because of that process, it's really slow to create a RequirementSet, so we don't want to do it during generator at all if possible.

randovania/generator/old_generator_reach.py
henriquegemignani: It forces Nodes to be hashable, which they aren't in WorldGraph :p

randovania/generator/old_generator_reach.py
Miepee: So when is a good place to create them then?

randovania/generator/old_generator_reach.py
henriquegemignani: I haven't investigated this time around but I've messed with this in the past to know it's kinda important.
Since `RequirementSet` removes redundant alternatives, the conversion in this function produces different results. Also however is calling this (which ends up being the `get_pickups_that_solves...`) actually wants just a list of alternatives, which actually means RequirementSet.

Basically we actually want to have these be RequirementSet, but for the current structure it's faster to convert as late as possible. Ideally it's be fetched directly from the node, but we'd need to stop doing these random `RequirementAnd` for that.

randovania/generator/old_generator_reach.py
henriquegemignani: Must: you care about having the `flat list of alternatives`.
Bad: When you want to combine requirements.
Never: when all you'll do is call `satisfied` once and discard. This PR is about this actually! By delaying to `unreachable_nodes_with_requirements`, a lot of times we just don't do the `as_set` operation anymore.

Miepee (approved):

PR #8713, opened 2025-11-02
Add graph implementation using rustworkx

henriquegemignani:
Improves performance further in some cases.
Prime 1: 2.6s -> 1.9s
Dread: unchanged


henriquegemignani:
Guess we need some benchmarks to get some report for codspeed.

PR #8716, opened 2025-11-03
Change all exporters to use fixed version dependency.

henriquegemignani:
Upon further reflection, making sure the exporters are only updated explicitly is how we actually want so let's keep to that rule.


pyproject.toml
```diff
@@ -66,38 +66,38 @@ dependencies = [
...
 
     # Metroid Planets
-    "planets-yapr>=1.0.0",
+    "planets-yapr==1.0.0",
 
     # Metroid Prime 1
-    "py_randomprime>=1.30.2",
-    "random-enemy-attributes>=1.0.3",
+    "py_randomprime==1.30.2",
+    "random-enemy-attributes==1.0.4.3",
```
Miepee, line 89: find it funny how this ended up getting an extra digit

PR #8723, opened 2025-11-03
Remove Resolver job, prepare Benchmark job to be required

henriquegemignani:
They're running the same tests and with recent changes the Resolver job is quite long!


henriquegemignani:
Actually this is going to break the merge queue. Ugh.

PR #8729, opened 2025-11-04
Keep original json for LayoutDescription for attaching to Sentry

henriquegemignani:
For server, as serializing the layout (and particularly the permalink string) can be expensive.

### DIFF
--- a/randovania/exporter/patch_data_factory.py
+++ b/randovania/exporter/patch_data_factory.py
@@ -88,8 +88,12 @@ class PatchDataFactory[Configuration: BaseConfiguration, CosmeticPatches: BaseCo
 
     def _attach_to_sentry(self) -> None:
         with sentry_sdk.isolation_scope() as scope:
+            description_json = self.description.original_dict
+            if description_json is None:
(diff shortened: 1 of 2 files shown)


randovania/exporter/patch_data_factory.py
```diff
@@ -88,8 +88,12 @@ def create_default_patcher_data_meta(self) -> PatcherDataMeta:
 
     def _attach_to_sentry(self) -> None:
         with sentry_sdk.isolation_scope() as scope:
+            description_json = self.description.original_dict
+            if description_json is None:
```
Miepee, line 92: Does the layout get saved by the server? At least I'm assuming something like that, as otherwise this would be always None I think.
Also does the layout for multis get created based off of a json dict?

henriquegemignani: Yes and yes. It comes from this function https://github.com/randovania/randovania/blob/a4e6054e1fb60c40d341d27cc0a4ec0abc489248/randovania/server/database.py#L158

And `from_bytes` calls `from_json_dict`.

PR #8732, opened 2025-11-04
Remove Pillow's AVIF module from releases, saving 7+ MB.

henriquegemignani:


tools/create_release.py
```diff
@@ -104,6 +104,25 @@ async def download_dotnet() -> None:
     script_path.unlink()
 
 
+def remove_unnecessary_modules() -> None:
+    """Removes python modules that should not be included in the final build."""
+
+    # Remove Pillow's support for AVIF images. We currently do not use this format, and it's 7+ mb.
+    try:
```
Miepee, line 111: why cant we just install an avif-less pillow package?

henriquegemignani: https://github.com/python-pillow/Pillow/issues/9101

henriquegemignani: And just deleting the file is kinda recommend. And the code does handle it properly too.
https://github.com/python-pillow/Pillow/pull/8869#issuecomment-2791718428

We could also build the wheels ourselves, but ew :P

Miepee (approved):

PR #8734, opened 2025-11-04
ResolverReach: don't call _combine_damage_requirements when damage is 0

henriquegemignani:
For a sample Prime 1 seed.
| | Classic| World Graph |
|--------|--------|--------|
| Before | 1.392s| 1.022s|
| After | 0.968s | 0.634s|


henriquegemignani:
Not exactly correct.

PR #8738, opened 2025-11-05
Try to ignore macOS/Linux on merge queue, try #2

henriquegemignani:
(commit is intentionally wrong to try something in CI)


henriquegemignani:
this might be it

Miepee (approved):

henriquegemignani:
why does it work when I excluded `pull_request` but it doesn't when I exclude `merge_queue`???
I guess it's time to spin up a test repository

PR #8739, opened 2025-11-05
Shutdown NetworkClient on Qt shutdown

henriquegemignani:


test/network_client/test_network_client.py
```diff
@@ -39,6 +39,19 @@ def client(tmp_path, mocker: pytest_mock.MockerFixture):
     return client
 
 
+async def test_shutdown(tmp_path):
+    # Explicitly not mocking http_session
+    # The assert is not leaking a resource.
```
Miepee, line 44: will this test fail if hte resource is leaked?

henriquegemignani: Yes, it triggers a warning and our tests runs with warnings to errors.
It'll likely not be in this exact test though...

duncathan:
does this make #8722 redundant?

henriquegemignani:
You can see how it already had the `await client.http.close()` but it wasn't enough. I don't know if `disconnect_from_server` makes a difference.
I'm also not even 100% sure that this new `test_shutdown` won't leak...

PR #8740, opened 2025-11-05
Keep track of what requirements have isolated damage already

henriquegemignani:


henriquegemignani:
I can make `dread/starter_preset.rdvgame` not slower, but the speedup for other cases not as big. The difference is to still calculate `isolated_satisfied` even when `damage == 0`. I imagine it's helpful to do so when you keep backtracking all the time, which you do in the case of an impossible game like `dread/starter_preset.rdvgame`.

I think it's better to be faster in the possible case, but not completely sure.

PR #8744, opened 2025-11-07
Feature/create preset button

### LINKED ISSUE #6183: Make the preset menu more beginner friendly
Currently, the preset selection menu is very unintuitive to a new user. I've had many cases of people completely new to Randovania struggling to understand how to work it (including myself in the past). Something should be done to try remedy this.

The main issue seems to be confusion over how to make a new preset, so I think that's where the focus it should be. My immediate suggestion is to add a "New Preset" button somewhere, but maybe there's something else that can be done.

Logarhythmic:
- partially fixes #6183

image

### DIFF
--- a/randovania/gui/widgets/select_preset_widget.py
+++ b/randovania/gui/widgets/select_preset_widget.py
@@ -201,11 +201,12 @@ class SelectPresetWidget(QtWidgets.QWidget, Ui_SelectPresetWidget):
         monitoring.metrics.incr("gui_preset_customize_clicked", tags={"game": self._game.value})
 
         old_preset = self._current_preset_data.get_preset()
+        parent_uuid: uuid.UUID | None
         if self._current_preset_data.is_included_preset:
             parent_uuid = old_preset.uuid
             old_preset = old_preset.fork()
         else:
-            parent_uuid: uuid.UUID | None = self._options.get_parent_for_preset(old_preset.uuid)
+            parent_uuid = self._options.get_parent_for_preset(old_preset.uuid)
 
         await self._customize_preset_with_dialog(old_preset, parent_uuid)
 


randovania/gui/ui_files/select_preset_widget.ui
```diff
@@ -23,6 +23,13 @@
    <property name="bottomMargin">
     <number>0</number>
    </property>
+   <item row="0" column="0" colspan="2">
+    <widget class="QPushButton" name="create_new_preset_button">
+     <property name="text">
+      <string>Create New Preset</string>
```
Miepee:
```suggestion
      <string>＋ Create New Preset</string>
```

randovania/gui/widgets/select_preset_widget.py
```diff
@@ -254,6 +255,12 @@ def _on_duplicate_preset(self):
         new_preset = VersionedPreset.with_preset(old_preset.get_preset().fork())
         self._add_new_preset(new_preset, parent=old_preset.uuid)
 
+    def _on_create_new_preset(self):
```
Miepee: typing

randovania/gui/widgets/select_preset_widget.py
```diff
@@ -254,6 +255,12 @@ def _on_duplicate_preset(self):
...
+        # Create a new preset based on the starter preset for the current game
+        starter_preset = self._window_manager.preset_manager.default_preset_for_game(self._game)
```
Miepee, line 270: how does this function with games that have multiple builtin presets, such as fusion/cs?

randovania/gui/widgets/select_preset_widget.py
```diff
@@ -254,6 +255,12 @@ def _on_duplicate_preset(self):
...
+        new_preset = VersionedPreset.with_preset(starter_preset.get_preset().fork())
+        self._add_new_preset(new_preset, parent=None)
```
Miepee: What does this do from a UI pov? If I skimmed over it correctly it just adds a new preset to the list?
I'd rather personally have it so that it brings up the customize-preset-dialog, where if declined does nothing and if accepted adds it (as a non-parent maybe too?)

Logarhythmic: changed it to open the customize dialogue and doesn't actually make the preset until you click save preset

randovania/gui/widgets/select_preset_widget.py
Logarhythmic: https://github.com/user-attachments/assets/180754f1-f32e-4cac-9fbd-491969aec4ec

randovania/gui/ui_files/select_preset_widget.ui
```diff
@@ -26,7 +26,7 @@
    <item row="0" column="0" colspan="2">
     <widget class="QPushButton" name="create_new_preset_button">
      <property name="text">
-      <string>Create New Preset</string>
+      <string>+ Create New Preset</string>
```
Miepee: please use the ＋ symbol that I posted, not a normal +.

randovania/gui/widgets/select_preset_widget.py
```diff
@@ -254,6 +255,33 @@ def _on_duplicate_preset(self):
...
+            self._logic_settings_window.raise_()
+            return
+
+        monitoring.metrics.incr("gui_preset_create_new_clicked", tags={"game": self._game.value})
+
+        # Create a new preset based on the starter preset for the current game
+        starter_preset = self._window_manager.preset_manager.default_preset_for_game(self._game)
+        new_preset = starter_preset.get_preset().fork()
+
+        editor = PresetEditor(new_preset, self._options)
```
henriquegemignani: Instead of duplicating the entirety of `_on_customize_preset`, you'll want to extract the common logic into a single function that takes a `VersionedPreset` and then call that function from both this and `_on_customize_preset`.

randovania/gui/widgets/select_preset_widget.py
```diff
@@ -201,6 +193,22 @@ async def _on_customize_preset(self):
...
+            return
+
+        monitoring.metrics.incr("gui_preset_customize_clicked", tags={"game": self._game.value})
+
+        old_preset = self._current_preset_data.get_preset()
+        if self._current_preset_data.is_included_preset:
+            parent_uuid = old_preset.uuid
+            old_preset = old_preset.fork()
+        else:
+            parent_uuid = self._options.get_parent_for_preset(old_preset.uuid)
```
Miepee: you have a mypy error here
get_parent_for_preset can return None if the preset doesn't have a parent

randovania/gui/widgets/select_preset_widget.py
```diff
@@ -254,6 +263,15 @@ def _on_duplicate_preset(self):
         new_preset = VersionedPreset.with_preset(old_preset.get_preset().fork())
         self._add_new_preset(new_preset, parent=old_preset.uuid)
 
+    @asyncSlot()
+    async def _on_create_new_preset(self) -> None:
+        monitoring.metrics.incr("gui_preset_create_new_clicked", tags={"game": self._game.value})
```
Miepee, line 268: fwiw, this doesnt actually do anything right now. Maybe hopefully in the near future tho

Miepee (approved):

PR #8748, opened 2025-11-08
LayoutDescription: Fill serialized patches cache on creation

henriquegemignani:


randovania/layout/layout_description.py
```diff
@@ -176,6 +176,10 @@ def get_preset(i: int, p: dict) -> Preset:
             original_dict=original_dict,
         )
 
+        # Fill the cache so calling shareable_word_hash is not slow by needing to re-serialize
+        object.__setattr__(result, "__cached_serialized_patches", json_dict["game_modifications"])
```
Miepee, line 180: why do we need to do this, instead of `result.__cached_serialized_patches = json_dict["game_modifications"]`?

henriquegemignani: `LayoutDescription` is a frozen dataclass, so we're not allowed to change attributes.
Except of course we can hack around that, which is done this way.

Miepee (approved):

PR #8754, opened 2025-11-09
Dread/boss energy adjustments

Tyranisaur:


CHANGELOG.md
```diff
@@ -17,8 +17,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed: Prime 1's Refill Pickups will now use more appropriate models.
 
 ### Metroid Dread
+
 - Changed: Energy Parts are now logical.
 
+#### Logic Database
+
+- Changed: The energy requirements for the Corpius, Kraid, Silver Chozo Robot, Gold Chozo Robot and Gold Chozo Robot have been reduced, so that when using medium damage strictness there available energy is enough to satisfy these fights.
```
Miepee: "There available energy"?

randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -2056,7 +2056,7 @@ Extra - asset_id: collision_camera_020
                       Missiles ≥ 24
                       Missiles ≥ 15 and Combat (Intermediate)
           # Energy Requirements
-          Combat (Beginner) or Normal Damage ≥ 198
+          Combat (Beginner) or Normal Damage ≥ 130
```
Logarhythmic, line 2059: Whats the goal here? The old test was using 1 for the damage_strickness so we're reducing it to match at a 1.5 level which is the default?

I'm assuming that's the case. 130 means 130*1.5 = 195 vs the 198 that was the original value.

Miepee: > Whats the goal here?

Make the vanilla game to resolve as possible, using default settings

And our default settings use a 1.5x damage multiplier

Tyranisaur: Yes, what I did for these was to start at the apporiate 98 + 100 for each energy tank, then account for the 1.5 multiplier. Then for later bosses I also accounted for the 25% damage reduction from varia suit. I didn't adjust the values for late game bosses where you would have gravity suit, but if I did I would have accounted for  a then 50% damage reduction. If I would have adjusted the later bosses, then I would likely increase their required energy, which I'm not sure is necessary.

PR #8757, opened 2025-11-11
Various optimizations for generator

henriquegemignani:
- Using rustworkx for the graph algorithms, as not-python is faster than python
   - We lose some time with calculating the path cost by doing it in python. Good spot for native module!
- Calculate less stuff when creating a UncollectedState
- Add `GeneratorReach.set_of_reachable_node_indices` as it has a way of doing that faster than calling is_reachable for each node

### DIFF
--- a/randovania/generator/old_generator_reach.py
+++ b/randovania/generator/old_generator_reach.py
@@ -258,8 +258,6 @@ class OldGeneratorReach(GeneratorReach):
         def weight(source: int, target: int, attributes: graph_module.GraphData) -> int:
             return _is_collected(target)
 
-        # Having to calculate the costs in python is slower,
-        # but calculating the paths is so much faster it's worth it.
         self._reachable_costs = self._digraph.shortest_paths_dijkstra(
             self._state.node.node_index,
             weight=weight,


randovania/generator/graph.py
```diff
@@ -84,11 +104,13 @@ def edges_data(self) -> Iterator[tuple[int, int, GraphData]]:
             for target, requirement in data.items():
                 yield source, target, requirement
 
+    has_multi_source_dijkstra = True
```
Miepee: what does this do?

randovania/generator/graph.py
```diff
@@ -165,3 +187,73 @@ def strongly_connected_components(self) -> Iterator[set[int]]:
...
+    def remove_edge(self, previous_node: int, next_node: int) -> None:
+        self._graph.remove_edge(previous_node, next_node)
+
+    def has_edge(self, previous_node: int, next_node: int) -> bool:
+        return self._graph.has_edge(previous_node, next_node)
+
+    def __contains__(self, item: int) -> bool:
+        return item in self._added_nodes
+
+    def edges_data(self) -> Iterator[tuple[int, int, GraphData]]:
```
Miepee: whats this tuple int int?

randovania/generator/old_generator_reach.py
```diff
@@ -222,52 +222,94 @@ def _calculate_safe_nodes(self) -> None:
...
-            weight=weight,
-        )
+        if self._digraph.has_multi_source_dijkstra:
+            costs, self._reachable_paths = self._digraph.multi_source_dijkstra(
+                {self._state.node.node_index},
+                weight=weight,
+            )
+            self._reachable_costs = typing.cast("dict[int, int | None]", costs)
+        else:
+            # Having to calculate the costs in python is slower,
```
Miepee: does rustworkx not support calculating the costs?

randovania/graph/world_graph.py
```diff
@@ -87,6 +87,13 @@ class WorldGraphNode:
     - EventNode: the event
     - HintNode/PickupNode: the node resource
     - DockLockNode: the dock node resources
+
+    These resources must all provide exactly 1 quantity each.
+    """
+
+    resource_gain_bitmask: int = dataclasses.field(init=False, default=0)
+    """
+    Bitmask of all ResourceInfo indices granted by this node for fast checking.
```
Miepee, line 96: might be helpful to add how to check the bitmask to see if it contains a certain resrourceinfo?

randovania/graph/world_graph.py
```diff
@@ -115,14 +122,26 @@ class WorldGraphNode:
...
-        return len(self.resource_gain) > 0
+        return bool(self.resource_gain)
 
     def full_name(self, with_region: bool = True, separator: str = "/") -> str:
         """The name of this node, including the area and optionally region."""
         return self.identifier.display_name(with_region, separator)
 
     def should_collect(self, context: NodeContext) -> bool:
+        return not self.has_all_resources(context)
         result = False
```
Miepee: unreachable code?

randovania/graph/world_graph.py
```diff
@@ -137,6 +156,7 @@ def should_collect(self, context: NodeContext) -> bool:
         return result
 
     def has_all_resources(self, context: NodeContext) -> bool:
+        return self.resource_gain_bitmask & context.current_resources.resource_bitmask == self.resource_gain_bitmask
         for resource, _ in self.resource_gain:
```
Miepee: unreachable code?

randovania/graph/world_graph.py
```diff
@@ -87,6 +87,13 @@ class WorldGraphNode:
     - EventNode: the event
     - HintNode/PickupNode: the node resource
     - DockLockNode: the dock node resources
+
+    These resources must all provide exactly 1 quantity each.
```
Miepee, line 91: does this not cause issues with e.g. pickup nodes that grant you multiple things?

randovania/generator/old_generator_reach.py
henriquegemignani: As far as I could tell, the shortest paths function support using weights, but it returns the paths but not the cost of these paths.

Miepee: > A dictionary of the shortest paths from the provided node where the key is the node index of the end of the path and the **value is the cost/sum of the weights of path**

?

randovania/graph/world_graph.py
henriquegemignani: The resources granted by a pickup in a pickup node até not added to the node itself.

randovania/generator/old_generator_reach.py
henriquegemignani: We're calling the one they returns a dict of index to list of indices, indicating the path. I know we use both info, but can't remember on top of my head if the path is essential

randovania/graph/world_graph.py
Miepee: Ah, ok.

randovania/generator/graph.py
henriquegemignani: two node indices, representing an edge.

duncathan: a type alias for node indices is probably a good idea tbh

randovania/generator/graph.py
```diff
@@ -31,39 +32,42 @@ def copy(self) -> typing.Self:
...
-        sources: set[int],
-        weight: Callable[[int, int, GraphData], int],
-    ) -> tuple[dict[int, int], Mapping[int, Sequence[int]]]:
+    def edges_data(self) -> Iterator[tuple[NodeIndex, NodeIndex, GraphData]]:
+        """Iterates over all edges that were added via `add_edge`."""
         raise NotImplementedError
 
     def shortest_paths_dijkstra(
         self,
         source: int,
```
Miepee: source should also be nodeindex, no?

randovania/generator/old_generator_reach.py
```diff
@@ -260,25 +258,17 @@ def _is_collected(target: int) -> int:
...
-            self._reachable_costs = typing.cast("dict[int, int | None]", costs)
-        else:
-            # Having to calculate the costs in python is slower,
-            # but calculating the paths is so much faster it's worth it.
-            self._reachable_paths = self._digraph.shortest_paths_dijkstra(
-                self._state.node.node_index,
-                weight=weight,
-            )
-            self._reachable_costs = {}
+        # Having to calculate the costs in python is slower,
```
Miepee: isnt this comment outdated now?

PR #8759, opened 2025-11-11
Change ResourceInfo into a protocol

henriquegemignani:


randovania/gui/docks/resource_database_editor.py
```diff
@@ -23,8 +22,13 @@
 from randovania.lib import frozen_lib
 
 if typing.TYPE_CHECKING:
+    from _typeshed import DataclassInstance
+
     from randovania.game_description.db.region_list import RegionList
     from randovania.game_description.resources.resource_database import ResourceDatabase
+    from randovania.game_description.resources.resource_info import ResourceInfo
+
+    class DataclassResourceInfo(ResourceInfo, DataclassInstance, typing.Protocol): ...
```
duncathan, line 31: why do this separately?

henriquegemignani: The only thing that cares about the resource info being a dataclass is these `EditableTableModel`. And I don't want to add this rule to the `ResourceInfo` protocol itself.

I was hoping to have custom resource info classes for world graph, though I'm not so sure if it's relevant anymore. But I had made this branch already so it's here!

duncathan: this feels like a bit of a hack but i guess it's fine

duncathan (approved):

randovania/gui/docks/resource_database_editor.py
henriquegemignani: Doesn't feel like it. The UI editor as we have can only change types that are both dataclasses and ResourceInfo. Which kinda implies anything stored by ResourceDatabase needs to be a dataclass, if we want an UI editor.

But still gives room for Requirements/ResourceCollection and others to use ResourceInfo that aren't dataclasses.

PR #8765, opened 2025-11-13
Limit how many actions are sent for session history

henriquegemignani:


Miepee (approved):

Miepee:
Technicalyl should get a changelog entry, as it stops new clients from looking at a full session history

henriquegemignani:
I don't plan making a release.

PR #8771, opened 2025-11-14
Async races: don't error in transitions that do nothing

henriquegemignani:


henriquegemignani:
It's some weird condition that all it does is show an error to the user, with no negative consequences.

PR #8773, opened 2025-11-15
Improve performance of expand_graph

henriquegemignani:
Tiny difference, but it's a bit faster. Matters more with consider unsafe.

### DIFF
--- a/randovania/generator/old_generator_reach.py
+++ b/randovania/generator/old_generator_reach.py
@@ -67,8 +67,8 @@ class GraphPath:
 
 
 class _SafeNodes:
-    as_list: list[int]
-    as_set: set[int]
+    as_list: list[NodeIndex]
+    as_set: set[NodeIndex]
 
     def __init__(self, component: Collection[NodeIndex]):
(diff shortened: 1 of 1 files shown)


randovania/generator/old_generator_reach.py
```diff
@@ -43,27 +43,37 @@ def _extra_requirement_for_node(
...
-            digraph.add_edge(self.previous_node.node_index, self.node.node_index, data=self.requirement)
+            digraph.add_edge(self.previous_node, self.node, data=self.requirement)
 
 
-class _SafeNodes(typing.NamedTuple):
+class _SafeNodes:
     as_list: list[int]
     as_set: set[int]
 
+    def __init__(self, component: Collection[NodeIndex]):
```
duncathan, line 73: might as well update the types for `as_list` and `as_set` to use node index too. also would it be worth giving this class slots?

henriquegemignani: This one is barely accessed so slots wouldn't matter here.

PR #8782, opened 2025-11-16
advance_after_action: don't attempt to collect safe resources in the new reach

henriquegemignani:


randovania/generator/reach_lib.py
```diff
@@ -19,6 +19,12 @@
     ResourceNodeT = typing.TypeVar("ResourceNodeT", bound=ResourceNode)
 
 
+def _get_reach_class() -> type[GeneratorReach]:
+    from randovania.generator.old_generator_reach import OldGeneratorReach
+
+    return OldGeneratorReach
```
Miepee, line 25: are you planning to at one point use a different reach for this?

henriquegemignani: There's some years old plans, similar to the world graph thing. I'm not refactoring it out myself, but if someone else wants feel free :p

randovania/generator/reach_lib.py
```diff
@@ -131,9 +129,10 @@ def advance_after_action(previous_reach: GeneratorReach) -> GeneratorReach:
             # print("Non-safe {} was good".format(action.full_name()))
             return advance_after_action(next_reach)
 
-        next_next_state = next_reach.state.copy()
+        # No need to call collect_all_safe_resources_in_reach on this new reach, as we've already collected everything
+        # at the start of this loop
+        next_reach = _get_reach_class().reach_from_state(game, next_reach.state.copy(), previous_reach.filler_config)
```
Tyranisaur, line 134: What makes this reach different from the one that was in this variable before reassigning it here?

henriquegemignani: It's a reach starting from the action being checked, instead of being a reach from initial node that moves to action.

Now why creating a new reach here ends up different from the old one? Not entirely sure. There's a lot of avoiding changing things/cache invalidation going around the `act_on_node` that maybe it's some bug there?

PR #8792, opened 2025-11-18
Prime: Adjust fighting elite pirates

Miepee:


randovania/games/prime1/logic_database/header.txt
```diff
@@ -72,12 +73,12 @@ Templates
...
                       Any of the following:
                           Missile ≥ 10 and Disabled Hard Mode
-                          Power Beam ≥ 20 and Enabled Hard Mode
+                          Missile ≥ 20 and Enabled Hard Mode
               All of the following:
                   # Normal Fight
                   Normal Damage ≥ 200
                   # Weaponry
-                  Charge Beam or Combat (Beginner) or Shoot Plasma Beam
+                  Charge Beam or Shoot Plasma Beam
```
toasterparty, line 84: There's a bit of a middle ground missing.

Something for beginner combat that doesn't have enough missiles to do the missile spam strat.

Like consider this player:
- Power beam only
- Doesn't know how to do the AI manipulation
- Does know the game well enough to jump over the shockwave attacks and keep backpedaling/dashing to avoid melee damage

They can probably do the fight while taking less than 200 damage. How much damage is that?

Miepee: > Power beam only

I explicitly wanted to mark power beam only fights as intermediate. They're already intermediate if you know the cheese strat just because of how long the fight is. Having the same fight but longer because of no manip doesn't seem like it should be beginner to me.

Miepee: I could however see a better middle ground for a player who knows how to dodge the pirates' attacks to have a fewer health requirement

toasterparty: up to you if you want to work out that middle ground, otherwise mergeable

Miepee: i'll go work something out for it.

Miepee: Added a beginner combat alternative with less health

PR #8794, opened 2025-11-18
Prime: Revise Chozo Ghosts requirements

Miepee:


randovania/games/prime1/logic_database/header.txt
```diff
@@ -81,6 +81,13 @@ Templates
                   # Dodging
                   Space Jump Boots or Combat (Beginner)
 
+* Kill Chozo Ghosts:
+      All of the following:
+          Shoot Power Beam
+          Any of the following:
+              Combat (Intermediate)
+              Charge Beam and Normal Damage ≥ 98
```
Samuel6710: Shouldn't this be `or`?

randovania/games/prime1/logic_database/header.txt
```diff
@@ -81,6 +81,13 @@ Templates
                   # Dodging
                   Space Jump Boots or Combat (Beginner)
 
+* Kill Chozo Ghosts:
+      All of the following:
+          Shoot Power Beam
+          Any of the following:
+              Combat (Intermediate)
```
Samuel6710, line 88: I'm torn on whether this should be beginner or intermediate.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1883,8 +1883,9 @@ Extra - aabb: [225.01727, 76.59607, 54.75615, 343.10974, 209.55296, 138.8626]
                   Missile ≥ 230 and Disabled Hard Mode
   > Event - Chozo Ghosts
       All of the following:
-          After Sunchamber Chozo Ghosts Layer Activated and Shoot Power Beam
-          Charge Beam or Combat (Intermediate)
+          # Kill Chozo Ghosts
+          After Sunchamber Chozo Ghosts Layer Activated and Kill Chozo Ghosts
```
Samuel6710: This fight involves 3 ghosts whereas HOTE only fights 1, but the logic is the same. I think it would be worth to let Sunchamber ghosts have its own logic.

randovania/games/prime1/logic_database/header.txt
Miepee: No, its supposed to be an AND.
I do not want people to fight ghosts hitless on trickless. They should at least have ~100HP.

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: Making sunchamber its own logic feels weird. Aside from first pass hote, all fights are either 2 or 3 instances
Hote first pass is 1. 
Hote second(third?) Pass is 3
Sunchamber is 3
Training chamber is 2
Life grove is 2 or 3, cant remember
Furnace, ruined shrine, arboreterum all also have 2 or 3

If anything i could understand making first pass hote its own block. But it being a worse case means that you'll automatically have all the requirements to fight the non-required ghosts

randovania/games/prime1/logic_database/header.txt
Samuel6710: I feel like there should be a middle ground, like 
`Charge Beam and Combat (Beginner)`
Having hitless ghosts be intermediate when they only have 3 unique attacks and one of them has a very distinct and lengthy charge window that can easily be interrupted with charge beam is funky to me.

randovania/games/prime1/logic_database/header.txt
Miepee: Me too, but decided to go for the worse case here. 
For reference one ghost takes 50 shots, so a total fight takes between 50 and 150 shots
For reference, an elite pirate takes ~300 shots and is also at combat intermediate. But an elite pirate is also constantly damageable while ghosts go in and out of existance which introduces waiting time and makes the fight annoyingly longer.
Could maybe make it beginner with x-ray? Idk.
It's also harder to do them damageless if you're fighting multiple ghosts at once, instead of just one.

randovania/games/prime1/logic_database/header.txt
Miepee: Fair, but also how many people do them hitless in their vanilla playthrough? I could maybe be convinced to lower the damage requirement so that you dont need any Etanks, or make it so that its easier to dodge with xray (aka when you can actually see all the ghosts instead of being surprise-attacked by some random dude behind you)

Miepee: Decided to do a few casual fights against sunchamber ghosts. Took ~120 damage with charge beam and ~190 without.
Definitely think the ~100 (~150 with multiplier) is fine. It's actually somewhat difficult to do multiple ghosts hitless, especially without X-Ray as it's easy to loose sight of where they rush off too and then get surprise attacked (or bump into them)
I also tried doing it with X-Ray, but it didn't really make dodging that much easier IMO, since you don't see more than a few meters in front of you. The fight did go by a lot quicker though since you can now always damage ghosts.

Samuel6710:
I think that’s fair then (reply to https://github.com/randovania/randovania/pull/8794#discussion_r2542587848)

randovania/games/prime1/logic_database/header.txt
henriquegemignani: Would an X-Ray alternative to health be interesting? It does have the consequence of making X-Ray be placed for more reasons.

Miepee: I tried to answer that in my comment:
> I also tried doing it with X-Ray, but it didn't really make dodging that much easier IMO

I could see *lowering* the health requirements with x-ray, but not fully replacing it as an alternative

Miepee: Tried doing a few X-Ray+Charge Beam fights. Took around ~20 Damage, so I can provide that as an alternative

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: Did a few more different ghost fights, and yeah, i feel like HotE definitely should be its own fight. Its difficulty is just so much easier

Miepee:
Adjusted logic based on feedback. Gonna fix the tests once I get an approval

toasterparty (approved):

PR #8796, opened 2025-11-18
Resolver: Don't keep track of node path

henriquegemignani:
Before:
Took 126.125 seconds. Hash: GUZLKAZ6

After:
Took 108.654 seconds. Hash: GUZLKAZ6

### DIFF
--- a/randovania/resolver/energy_tank_damage_state.py
+++ b/randovania/resolver/energy_tank_damage_state.py
@@ -51,19 +51,24 @@ class EnergyTankDamageState(DamageState):
         return result
 
     def _at_maximum_energy(self, resources: ResourceCollection) -> Self:
-        result = self._duplicate()
-        result._energy = result._maximum_energy(resources)
-        return result
+        maximum_energy = self._maximum_energy(resources)
+        if maximum_energy == self._energy:
+            return self
+        else:
+            result = self._duplicate()
+            result._energy = maximum_energy
+            return result
 
     @override
     def is_better_than(self, other: DamageState | None) -> bool:
         if other is None:
             return True
-        assert isinstance(other, EnergyTankDamageState)
-        return self._energy > other._energy
+        return self._energy > other._energy  # type: ignore[attr-defined]
 
     @override
     def apply_damage(self, damage: int) -> Self:
+        if damage <= 0:
+            return self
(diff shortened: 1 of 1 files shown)


henriquegemignani:
One more dumb commit:
`Took 88.923 seconds. Hash: GUZLKAZ6`

randovania/resolver/energy_tank_damage_state.py
```diff
@@ -51,19 +51,24 @@ def _duplicate(self) -> Self:
...
     def is_better_than(self, other: DamageState | None) -> bool:
         if other is None:
             return True
-        assert isinstance(other, EnergyTankDamageState)
-        return self._energy > other._energy
+        return self._energy > other._energy  # type: ignore[attr-defined]
 
     @override
     def apply_damage(self, damage: int) -> Self:
+        if damage <= 0:
```
Miepee, line 70: is there ever a usecase where someone wants to apply negative damage?

henriquegemignani: Breaking things.

duncathan:
is the node path used for anything other than logging?

henriquegemignani:
tracker

randovania/resolver/energy_tank_damage_state.py
```diff
@@ -51,19 +51,24 @@ def _duplicate(self) -> Self:
         return result
 
     def _at_maximum_energy(self, resources: ResourceCollection) -> Self:
-        result = self._duplicate()
-        result._energy = result._maximum_energy(resources)
-        return result
+        maximum_energy = self._maximum_energy(resources)
```
Miepee, line 54: dumb idea, but i suppose theres no easy way to only call maximum_energy, if the resources that are used in said function have changed in the resourcecollection? Aka, avoid calling _maximum_energy, unless we have to.

henriquegemignani: I'd put a really high chance of being slower

Miepee:
Will approve once mypy is addressed

henriquegemignani:
> Will approve once mypy is addressed

It's the mypy-modified and it doesn't tell us where it happened. I don't have a clue what to fix.

PR #8799, opened 2025-11-19
World Graph: cache patched requirements

henriquegemignani:
Before: Took 85.774 seconds. Hash: GUZLKAZ6
After: Took 72.951 seconds. Hash: GUZLKAZ6

But this feels a bit like a hack. Total memory usage on CLI was under 120mb so quite negligible.

### DIFF
--- a/randovania/graph/world_graph.py
+++ b/randovania/graph/world_graph.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import collections
 import copy
 import dataclasses
+import functools
 import typing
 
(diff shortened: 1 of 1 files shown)


Miepee:
What makes this feel like a hack?

henriquegemignani:
Partly because I guess the hash was wrong before and it caused issues. Partly because unbounded cache (but it was better than not unbounded, I tested).
But I guess it's precisely for this kind of thing that I made the `Requirement` class immutable.

PR #8802, opened 2025-11-20
Prime: Add wallboost escape to alcove.

Miepee:


Miepee:
It breaks a few things, mostly whenever there's some non-door connection to them. 
One Way Anywhere is the easiest example of it. Any elevator that leads onto a dock causes the generator to think that it can destroy the lock.

PR #8810, opened 2025-11-21
Planets (Zebeth): QoL patches support

UltiNaruto:
Supports `Open Missile Door with 1 missile` in logic.
`Allow downward shots` still need logic to be done.

QoL patches tab :

image

### DIFF
--- a/randovania/games/planets_zebeth/gui/ui_files/preset_planets_zebeth_patches.ui
+++ b/randovania/games/planets_zebeth/gui/ui_files/preset_planets_zebeth_patches.ui
@@ -106,7 +106,7 @@
     <item>
      <widget class="QLabel" name="allow_downward_shots_label">
       <property name="text">
-       <string>Allows the player to shoot below.&lt;br&gt;&lt;span style=&quot; font-weight:700;&quot;&gt;Warning :&lt;/span&gt; This isn't in logic yet.</string>
+       <string>Allows the player to shoot below.</string>
       </property>
      </widget>
     </item>
--- a/randovania/games/planets_zebeth/gui/generated/preset_planets_zebeth_patches_ui.py
+++ b/randovania/games/planets_zebeth/gui/generated/preset_planets_zebeth_patches_ui.py
@@ -103,6 +103,6 @@ class Ui_PresetPlanetsZebethPatches(object):
         self.open_missile_doors_with_one_missile_check.setText(QCoreApplication.translate("PresetPlanetsZebethPatches", u"Open Missile doors with one Missile", None))
         self.open_missile_doors_with_one_missile_label.setText(QCoreApplication.translate("PresetPlanetsZebethPatches", u"With this option enabled, only one missile is required to open red doors instead of 5.", None))
         self.allow_downward_shots_check.setText(QCoreApplication.translate("PresetPlanetsZebethPatches", u"Allow downward shots", None))
-        self.allow_downward_shots_label.setText(QCoreApplication.translate("PresetPlanetsZebethPatches", u"Allows the player to shoot below.<br><span style=\" font-weight:700;\">Warning :</span> This isn't in logic yet.", None))
+        self.allow_downward_shots_label.setText(QCoreApplication.translate("PresetPlanetsZebethPatches", u"Allows the player to shoot below.", None))
     # retranslateUi
 


randovania/games/planets_zebeth/exporter/patch_data_factory.py
```diff
@@ -101,14 +101,16 @@ def _create_hash_dict(self) -> dict:
         }
 
     def _create_game_config_dict(self):
+        c = self.patches.configuration
+        assert isinstance(c, PlanetsZebethConfiguration)
```
Miepee: I'm fairly certain you don't need to assert this type as it's already ensured. At least if you use `self.configuration`.

randovania/games/planets_zebeth/gui/preset_settings/planets_zebeth_patches_tab.py
```diff
@@ -15,27 +12,21 @@
...
-        self.include_extra_pickups_check.setText("Include Extra Pickups")
-        self.root_layout.addWidget(self.include_extra_pickups_check)
-
-        self.setCentralWidget(self.root_widget)
+        self.setupUi(self)
 
         # Signals
-        for f in _FIELDS:
-            self._add_persist_option(getattr(self, f"{f}_check"), f)
+        self.warp_to_start_check.stateChanged.connect(self._persist_option_then_notify("warp_to_start"))
```
Miepee: might be more readable to do what am2r in its gameplay_tab does.

randovania/games/planets_zebeth/gui/preset_settings/planets_zebeth_patches_tab.py
```diff
@@ -45,15 +36,9 @@ def tab_title(cls) -> str:
...
-                editor.set_configuration_field(attribute_name, value)
-
-        signal_handling.on_checked(check, persist)
-
     def on_preset_changed(self, preset: Preset) -> None:
         config = preset.configuration
         assert isinstance(config, PlanetsZebethConfiguration)
-        for f in _FIELDS:
-            typing.cast("QtWidgets.QCheckBox", getattr(self, f"{f}_check")).setChecked(getattr(config, f))
+        self.warp_to_start_check.setChecked(config.warp_to_start)
```
Miepee: see above

randovania/games/planets_zebeth/gui/ui_files/preset_planets_zebeth_patches.ui
```diff
@@ -0,0 +1,132 @@
...
+     <widget class="QCheckBox" name="allow_downward_shots_check">
+      <property name="text">
+       <string>Allow downward shots</string>
+      </property>
+     </widget>
+    </item>
+    <item>
+     <widget class="QLabel" name="allow_downward_shots_label">
+      <property name="text">
+       <string>Allows the player to shoot below.&lt;br&gt;&lt;span style=&quot; font-weight:700;&quot;&gt;Warning :&lt;/span&gt; This isn't in logic yet.</string>
```
Miepee: fwiw since you're in dev there's technically no need to point that out.

randovania/games/planets_zebeth/layout/planets_zebeth_configuration.py
```diff
@@ -20,7 +20,9 @@ class PlanetsZebethArtifactConfig(BitPackDataclass, JsonDataclass):
 class PlanetsZebethConfiguration(BaseConfiguration):
     artifacts: PlanetsZebethArtifactConfig
     energy_per_tank: int = dataclasses.field(metadata={"min": 1, "max": 1000, "precision": 1})
-    include_extra_pickups: bool
+    warp_to_start: bool
```
Miepee, line 23: What is the usecase of warp to start in planets?
In all other games, the reason why warp to start exists is because you can permanently brick your save file if you go to the wrong place and save. Is that possible to do in Planets as well?

I don't remember planets quite exactly but AFAIK it:
- saves the game every time you collect any pickup (which means ponrs dont exist)
- when you die, it puts you at the elevator of the area you're in

Which from what i know should make it impossible to brick your save file somewhere. I don't remember any situation where you can get *into* an area but not *out*.

UltiNaruto: Warp to start is purely QoL on Planets. 
Collecting pickups indeed saves the game, but I changed how dying works, you respawn at the last entrance instead of the last top elevator of the current region you are in. (ex: you took the elevator from Ridley to Norfair, you won't respawn at the elevator that goes from Norfair to Brinstar if you die)

Miepee: Hmm. might be worth revisiting but probably more in testing than in this PR.

randovania/games/planets_zebeth/gui/preset_settings/planets_zebeth_patches_tab.py
```diff
@@ -39,6 +46,12 @@ def header_name(cls) -> str | None:
     def on_preset_changed(self, preset: Preset) -> None:
         config = preset.configuration
         assert isinstance(config, PlanetsZebethConfiguration)
-        self.warp_to_start_check.setChecked(config.warp_to_start)
-        self.open_missile_doors_with_one_missile_check.setChecked(config.open_missile_doors_with_one_missile)
-        self.allow_downward_shots_check.setChecked(config.allow_downward_shots)
+        for f in self._CHECKBOX_FIELDS:
+            typing.cast("QtWidgets.QCheckBox", getattr(self, f"{f}_check")).setChecked(getattr(config, f))
+
+    def _add_checkbox_persist_option(self, check: QtWidgets.QCheckBox, attribute_name: str) -> None:
```
Miepee, line 52: This function now gets redefined for the 6th time.
I'm letting it pass here, but usually this should be refactored out to somewhere so that all games can use it

Miepee (approved):

PR #8818, opened 2025-11-22
Prime 1: Clean TAN to Vault Access

Miepee:


toasterparty (changes requested):
I know you're just adding them inertly for now, but I would prefer the spawn points were added programmatically with a script as I would rather review that script then 200 new nodes one at a time.

Specifically, a script to idempotently modify the database json to:
- Add a node named "Spawn Point" with coordinates provided by a helper dict
- Change the default node from the current to the one named "Spawn Point"
- Add a trivial connection from spawn point to the old default node

Miepee:
Good to know. Thought doing that step by step might be nicer for review but i don't care either way.

Miepee:
They do, lots of things in fusion/am2r/msr(/dread?) Uses those

PR #8825, opened 2025-11-24
Fix State creation to not be weird

henriquegemignani:


henriquegemignani:
All methods of `Node` classes are gone!

randovania/resolver/bootstrap.py
```diff
@@ -140,49 +135,44 @@ def trick_resources_for_configuration(
...
         self,
-        resources: ResourceCollection,
-        game: GameDescription,
+        minimal_logic: MinimalLogicData,
+        resource_database: ResourceDatabaseView,
         standard_pickups: StandardPickupConfiguration,
-    ) -> None:
-        resource_database = game.resource_database
+    ) -> ResourceGain:
+        resource_gain = []
```
Miepee: mypy \o/

randovania/resolver/bootstrap.py
```diff
@@ -140,49 +135,44 @@ def trick_resources_for_configuration(
 
             yield trick, level.as_number
 
-        return static_resources
```
Miepee, line 143: yield + return in the same function confuses me lol
what did the return of an empty dict do before?

henriquegemignani: I don't have a clue

Miepee: great xP

Miepee (approved):

PR #8830, opened 2025-11-25
Add Bitmask

henriquegemignani:
Unused, as using this in ResourceCollection reduces performance without the followup GraphRequirement. At least without compilation, which I'm not enabling with this PR. I'll leave the shenanigans for that for another one.


henriquegemignani:
Merging everything at once with #8847 will be better.

PR #8832, opened 2025-11-25
Fusion: Remove beginner movement from easy S6 climbs

Miepee:
Reasoning:
- If you're ever required to go down (e.g. did *not* accidentally fall down), then it's always possible to reload the room should you have accidentally shot the wrong block
- Reloading the room if you "messed up" is something we already expect with freezing logic. E.g if you accidentally kill an enemy you need to climb a room.
- We already assume that the player knows in which blocks blue-X are and thus which blocks to not shoot.
- Helps the generator


JeffGainsNGames (approved):

JeffGainsNGames:
changelog conflict

PR #8837, opened 2025-11-26
Prime 1: Sunchamber and co rework

Miepee:
Will fix tests once the core logic has approval

### DIFF
--- a/randovania/games/prime1/logic_database/Chozo Ruins.txt
+++ b/randovania/games/prime1/logic_database/Chozo Ruins.txt
@@ -1943,7 +1943,7 @@ Extra - aabb: [225.01727, 76.59607, 54.75615, 343.10974, 209.55296, 138.8626]
                           # NSJ
                           Combat/Scan Dash (Advanced)
                       # Health Requirements
-                      Combat (Ludicrous) or Normal Damage ≥ 50
+                      Normal Damage ≥ 50
               Any of the following:
                   # Requirements for destroying the Bomb Slots
                   Can Use Bombs
@@ -1969,8 +1969,8 @@ Extra - aabb: [225.01727, 76.59607, 54.75615, 343.10974, 209.55296, 138.8626]
       Trivial
   > Event - Flaahgra
       All of the following:
-          # Flahgra can be defeated without shooting the first dish if coming from Sun Tower Access and before starting the fight
-          Morph Ball and Knowledge (Beginner)
+          # Flahgra can be defeated by hitting the bomb slots from the outside, which are active before starting the fight
+          Morph Ball and Knowledge (Intermediate)
           Any of the following:
               # Requirements for destroying the bomb slots
               All of the following:
--- a/randovania/games/prime1/logic_database/Chozo Ruins.json
+++ b/randovania/games/prime1/logic_database/Chozo Ruins.json
@@ -11708,15 +11708,6 @@
                                                                                             "amount": 50,
                                                                                             "negate": false
                                                                                         }
-                                                                                    },
-                                                                                    {
-                                                                                        "type": "resource",
-                                                                                        "data": {
-                                                                                            "type": "tricks",
-                                                                                            "name": "Combat",
-                                                                                            "amount": 5,
-                                                                                            "negate": false
-                                                                                        }
                                                                                     }
                                                                                 ]
                                                                             }
@@ -12002,14 +11993,14 @@
                         "Event - Flaahgra": {
                             "type": "and",
                             "data": {
-                                "comment": "Flahgra can be defeated without shooting the first dish if coming from Sun Tower Access and before starting the fight",
+                                "comment": "Flahgra can be defeated by hitting the bomb slots from the outside, which are active before starting the fight",
                                 "items": [
                                     {
                                         "type": "resource",
                                         "data": {
                                             "type": "tricks",
                                             "name": "Knowledge",
-                                            "amount": 1,
+                                            "amount": 2,
                                             "negate": false
                                         }
                                     },


randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1304,7 +1308,11 @@ Extra - aabb: [279.84372, -45.27577, 55.823124, 300.77795, 16.10456, 70.44933]
   * Extra - world_position: (284.20767, 16.104553, 60.87069)
   * Extra - world_rotation: (0.0, 0.0, 90.12323)
   > Door to Arboretum
-      Trivial
+      Any of the following:
+          # Deal with Weed and Vines
+          Shoot Any Beam
+          Combat (Beginner) and Normal Damage ≥ 26
+          Combat (Advanced) and Normal Damage ≥ 6
```
Samuel6710: Was this tested? I'm not sure how you'd make it past the vines and only take 6 damage, much less so with Space Jump.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1886,55 +1903,97 @@ Extra - aabb: [225.01727, 76.59607, 54.75615, 343.10974, 209.55296, 138.8626]
...
+                  All of the following:
+                      # Shoot down the dishes to make the tentacles retract
+                      Any of the following:
+                          Shoot Power Beam
+                          All of the following:
+                              # Shoot dishes with only Missiles
+                              Combat (Beginner) and Shoot Any Beam
+                              Any of the following:
+                                  Missile ≥ 10
+                                  Missile ≥ 4 and Combat (Advanced)
```
Samuel6710, line 1930: I feel like `Combat (Advanced)` might be a little high here? The logic does expect you to never use missiles to stun Flaahgra but I think the health requirements cover this well. I'd suggest lowering it to intermediate.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1886,55 +1903,97 @@ Extra - aabb: [225.01727, 76.59607, 54.75615, 343.10974, 209.55296, 138.8626]
...
+                      Any of the following:
+                          # Health requirements
+                          Combat (Intermediate) or Normal Damage ≥ 80
+                          Combat (Beginner) and Normal Damage ≥ 25
+                  All of the following:
+                      # Jump directly to the Bomb Slots. Dashing is done from dishes/tentacles
+                      Knowledge (Beginner)
```
Samuel6710: After some testing and thinking more critically about the discussion between Reggie and I, I determined that to bomb the slot, it nessecitates entering morph ball mid air onto a standable to bomb the slots or landing on a nearby vine and then morphing and bombing (does not work with every slot). I would make this `Standable Terrain (Intermediate)` because of this requirement but also because of the risk of knockback from the vines and damage from poison water plus follow ups from Flaahgra which can easily stack up. Advanced is also arguable too.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1886,55 +1903,97 @@ Extra - aabb: [225.01727, 76.59607, 54.75615, 343.10974, 209.55296, 138.8626]
...
+                      Any of the following:
+                          All of the following:
+                              # With Space Jump
+                              Space Jump Boots
+                              Combat/Scan Dash (Intermediate) or L-Jump (Beginner)
+                          # NSJ
+                          Combat (Intermediate) and Combat/Scan Dash (Advanced)
+                      # Health Requirements
+                      Combat (Advanced) or Normal Damage ≥ 35
```
Samuel6710: Damageless is theoretically possible, but not probable. You're at risk of so many damage opportunities that damageless is likely `Ludicrous`. I'd say also bump the `Normal Damage` requirement to `50` since trickless combat and 70 health with the typical damage leniency feels a bit strict, even with this advanced trick, since avoiding subsequent damage from Flaahgra's attacks is more difficult than if you were bombing the slots after lowering the solar panels.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1886,55 +1903,97 @@ Extra - aabb: [225.01727, 76.59607, 54.75615, 343.10974, 209.55296, 138.8626]
...
+                  All of the following:
+                      # Jump directly to the Bomb Slots. Dashing is done from dishes/tentacles
+                      Knowledge (Beginner)
+                      Any of the following:
+                          All of the following:
+                              # With Space Jump
+                              Space Jump Boots
+                              Combat/Scan Dash (Intermediate) or L-Jump (Beginner)
+                          # NSJ
+                          Combat (Intermediate) and Combat/Scan Dash (Advanced)
```
Samuel6710: What's the idea behind `Combat (Intermediate)` here?

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1886,55 +1903,97 @@ Extra - aabb: [225.01727, 76.59607, 54.75615, 343.10974, 209.55296, 138.8626]
...
+                          Combat (Intermediate) or Normal Damage ≥ 80
+                          Combat (Beginner) and Normal Damage ≥ 25
+                  All of the following:
+                      # Jump directly to the Bomb Slots. Dashing is done from dishes/tentacles
+                      Knowledge (Beginner)
+                      Any of the following:
+                          All of the following:
+                              # With Space Jump
+                              Space Jump Boots
+                              Combat/Scan Dash (Intermediate) or L-Jump (Beginner)
```
Samuel6710, line 1942: It is possible to use Flaahgra itself as a dash point to a bomb slot, but it's extremely difficult. If this is considering using the Scan Visor, then you'll want to include it too as an item requirement for dashing. This goes for the NSJ dash as well.

Samuel6710:
also update changelog

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: I did test all of the damage requirements, yes.
Wish i had recorded them though cause i dont remember if i did them with or without SJ
Its definitely tricky. Wouldnt mind bumping it to expert if you think its appropriate

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: The reasoj its advanced is:
1. For the last round, youre at the mercy of rng. Because sometimes (very rarely) flaaghra can just decide to not drop any missiles at all
2. If it were intermediate, then that'd mean that doing the fight with only 4 missiles damageless would be intermediate, which doesnt feel appropriate

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: I'll let you db maintainers figure this one out. 
Do note however that intermediate+ will move it out of tournament/weekly logic

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: The dashes consider every visor here, yes.
Is dashing from flaaghra easier than from tentacles? Because if no, see the db comment:
> dashing is done from dishes/tentacles

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: Remnant that i forgot to remove it seems. Originally the SJ way was beginner combat too

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: When i tried it, i got on average ~15 damage per attempt. Sometimes a little bit more, sometimes a little bit less
So i dont mind bumping the number up here.

randovania/games/prime1/logic_database/Chozo Ruins.txt
Samuel6710: I think Expert would be appropriate if 6 damage involves avoiding the vine damage.

randovania/games/prime1/logic_database/Chozo Ruins.txt
Samuel6710: That's fair, I think this is fine then

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: Do you want me to change all vine-avoidances to expert or only this one?

Samuel6710: The ones in Sunchamber Lobby are extremely difficult to navigate over, whereas Sunchamber Access has verticality that can be taken advantage of, so maybe that one is lower with SJ?

randovania/games/prime1/logic_database/Chozo Ruins.txt
Samuel6710: The tentacles can only be locked onto with Scan Visor, and dashing off the panels is trickier than dashing off the tentacles since you can see where you're going with the tentacles and don't have to bend to reach the slots, but you do have to do that with the panels.

Miepee: I dont get your issue here. If dashing off of flaaghra is more difficult than tentacles/dishes, there is no need to include it.

And every visor *is* already in the requirements. See the very top with "Morph and Has Any Visor"

randovania/games/prime1/logic_database/Chozo Ruins.txt
Uncle-Reggie: Moving a discussion I was very clear about to GH to get new light on it is not good practice. The Flaahgra quick fight does not require standable terrain. It is not going to be included.

randovania/games/prime1/logic_database/Chozo Ruins.txt
Uncle-Reggie: Miepee is correct. If a trick is made obsolete by an easier version of the trick or with less inventory reqs, we don't model it in the DB. The "rare Reggie strat" for Echoes is one of those instances. I obsoleted a BSJ and removed a bomb requirement from Torvus Grove with that trick. If it's easier to dash from the dishes or tentacles, there's no need to model the harder dash off Flaahgra.

Edit: If there is a missing Scan req for specifically the tentacle dash, Sam is right, and I just missed that.

Samuel6710: I'm saying that dashing off the panels is just as difficult as dashing off Flaahgra and that dashing off the tentacles is easier, so there should be a scan visor requirement.

randovania/games/prime1/logic_database/Chozo Ruins.txt
Samuel6710: It became obvious that the discussion had moved on from being good faith. I not only presented video evidence to back up my statements, but did further testing specifically for this review. When I asked evidence be presented to the contrary, it was denied, and the counterargument was a vague explanation provided with "I know more than you, accept what I'm saying." My testing, explanations, and video clips all make it clear that standable terrain is required. Please recognize that this logic is not changed at your discretion, it's a team effort amongst all of us.

Samuel6710: > Do note however that intermediate+ will move it out of tournament/weekly logic

This is true, but with the vanilla fight requirements, I don't think it's much of a problem. Beginner doesn't feel quite right when many of the standables are seams and both Flaahgra and the tentacles make it trickier.

Uncle-Reggie: Luckily, a PWE render was provided and I was able to piggyback off of that to provide my reasoning. My expertise in game knowledge is what qualifies me for this role. I have been playing for longer and I do, in fact, have more knowledge and experience than you do. Continuing to say there is no evidence supporting is disingenuous considering there is now proof to back up my statement.

You insisting this to be true does not make it true. I do have final say before toaster on all matters Prime. His decisions override mine.

I fundamentally disagree with your statement that standable terrain may be used because it absolutely isn't mandatory at all. In all your testing, you could have also attempted to not use it and rather easily succeeded. But since you're being obstinate, fine. Later this weekend I'll record something after I go through all the trouble to set up my recoding stuff just to prove you wrong, because that's an effective use of my time and energy.

Samuel6710: My testing did include trying to avoid using the standables. I spent a solid 10-15 ish minutes just throwing myself at the bomb slot and moving around and morphing at earlier and later times.

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: How much easier difficulty wise should it be tho?
Right now with SJ a dash is always intermediate and NSJ advanced
Are you saying it should be this?
Sj + scan = intermediate
Sj + any other visor = advanced
Nsj+ scan = advanced
Nsj + scan = expert


Cause even tho dashing from the tentacle may be *slightly* easier than the dishes, i dont think its "we differentiate both by an entire difficulty" easier.
And IMO dashing off of flaaghra is way harder than the dishes. That plant can move like crazy and is in a really awkward position. Whereas with the dish, you lock onto it, move into the correct position, dash, and maybe bend very slightly.

Samuel6710: It's important to consider that the logic expects you to do the fight without scan visor using one dish as your dash point for all the bomb slots. There are multiple spots where you will lose the lock on and not be able to see where you are going, plus having to avoid flaahgra while getting to your dash spot. The tentacles are significantly more flexible than using the panel. I think the scan dash logic should be as follows:

SJ + Scan = Beginner Dash
SJ = Intermediate Dash
NSJ + Scan = Advanced Dash
NSJ = Expert

Miepee: > without scan visor using one dish as your dash point for all the bomb slots.

This isn't true! It expects you to have *two* dishes! Because the first one is required to always be shot down.
And the two dishes are always in a mirroring pattern, e.g. something like this:
grafik

Samuel6710: That's a good point. I overlooked that. I still think it is more difficult than the tentacle dashes, but based on your diagram I can see why you'd put them at the same trick difficulty.

Miepee: yeah, I won't deny that the tentacles are easier, but the difference isnt stark enough to put them into seperate categories.
Which brings me to this:
> SJ [+ Scan] = Beginner Dash

I don't consider the dash beginner, just due to the fact that it's a significant step up from other beginner dashes. The area you have to land on is somewhat small, messing it up can mean lots of damage from acid and/or pant and/or tentacle and you also have to do the dash in a boss fight while potentially being attacked

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1886,55 +1903,97 @@ Extra - aabb: [225.01727, 76.59607, 54.75615, 343.10974, 209.55296, 138.8626]
...
+                          All of the following:
+                              # Shoot dishes with only Missiles
+                              Combat (Beginner) and Shoot Any Beam
+                              Any of the following:
+                                  Missile ≥ 10
+                                  Missile ≥ 4 and Combat (Advanced)
+                      Any of the following:
+                          # Health requirements
+                          Combat (Intermediate) or Normal Damage ≥ 80
+                          Combat (Beginner) and Normal Damage ≥ 25
```
toasterparty, line 1934: or Combat Advanced damageless

randovania/games/prime1/logic_database/Chozo Ruins.txt
toasterparty: I argue that knowing that these 4 bomb slots are unlike every single other in the game in that they an be activated without actually slotting the ball inside them warrants knowledge intermediate. It's more knowledge than just knowing PBs work in addition to bombs, but it's less niche than knowing about wavebuster strategies.

toasterparty: This trick does not involve standable terrain.

The Standable Terrain trick is for unmorphed movement. The trick requires wedging the morph ball into the armpit of each bomb slot. Unlike the standing player, morph ball is affected by friction and so there is more to the player's vertical position calculation than just "if feet on stadable, no go down". To check a hunch, I built an iso without the force triggers on each of the vines. It turns out the vine at 9 o'clock is actually impossible to bomb from the outside NSJ without the help of the force trigger to keep the ball in place, despite the fact that the player can do "standable terrain" next to the slot.

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: Just in case, does that mean "change the damagless from combat intermediate to combat advanced" or am i to interpret that as "toaster didn't see the damageless combat intermediate"?

toasterparty: "toaster didn't see the damageless combat intermediate"

fuck that `or` thing in an `Any` block gets me every time

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: ok i redid the vine dodging.
Sunchamber Access I've left as is, but added a beginner dodge with SJ, as it turns out to be really easy.

Sunchamber lobby i've bumped to expert. SJ does not make it easier because of the ceiling.

You *can* get consistent at it but the vines can very easily just corner you and hit you unexpectedly.

Miepee:
> also update changelog

Donezo

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1886,55 +1905,97 @@ Extra - aabb: [225.01727, 76.59607, 54.75615, 343.10974, 209.55296, 138.8626]
...
 
 > In Arena (Back); Heals? False
   * Layers: default
   * Extra - world_position: (275.376892, 193.421051, 80.388443)
+  > Door to Sun Tower Access
+      Trivial
   > Event - Flaahgra
       All of the following:
           # Flahgra can be defeated without shooting the first dish if coming from Sun Tower Access and before starting the fight
           Morph Ball and Knowledge (Beginner)
```
toasterparty: This still requires bombing from outside the slot, so by the same logic we applied earlier, knowledge intermediate.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1886,55 +1905,97 @@ Extra - aabb: [225.01727, 76.59607, 54.75615, 343.10974, 209.55296, 138.8626]
...
+                  Can Use Bombs
+                  Power Bomb ≥ 4 and Knowledge (Beginner) and Can Use Power Bombs
           All of the following:
+              # Wavebuster fight. It deals direct damage to Flaaghra
               Charge Beam and Wavebuster and Combat (Intermediate) and Knowledge (Advanced) and Shoot Wave Beam
               Any of the following:
                   Missile ≥ 460
                   Missile ≥ 230 and Disabled Hard Mode
+                  # Flaaghra gives refills after getting up due to a dish getting knocked down
+                  Missile ≥ 20 and Combat (Ludicrous)
```
toasterparty, line 1958: Did you or someone else check damageless here? I was under the impression you had two options: Drop into the acid for 1-2 damage per slot or hold in the armpit for one application of contact damage

Miepee: It may depend on the bomb slot?
I was sometimes able to do a bomb slot damageless. But I didn't test (nor do I know anyone else who did) doing *every* bomb slot damage less. I think sam said its theoretically doable to do every one damageless, and I just took that at face value

Miepee: Ok after bashing my head against this, i can not do that with the final 2 bomb slots.
Want me to remove the ludicrous for now until we get an actual proof of this?

Miepee: Looking further into it, it's also very likely to be comlpetely impossible with PBs since they take a tad longer to hit the bomb slot.  Unless i'm just missing something (which wouldn't be too unlikely)

toasterparty: Remove for now to be safe, then add in if someone makes a skilled effort to solve for minimum damage

PR #8836, opened 2025-11-26
Add Caltrops as damage for Blank game

henriquegemignani:

PR #8838, opened 2025-11-26
Add GraphRequirement

henriquegemignani:
Needs #8830 and #8836. Will rebase after merging both.


henriquegemignani:
Merging everything at once with #8847 will be better.

PR #8841, opened 2025-11-26
Generator: Don't walk into an immediate softlock

### LINKED ISSUE #477: Generator: do not consider actions that lead to immediate softlock
Case A: Placing Violet Translator in GFMC for default seeds.
Case B: Advancing to Ing Cache 1 when Light Beam does not existing in the seed

Potential solution:

The generator calculates the weight of all options by creating a reach assuming you did that action
We could run the _calculate_progression_pickups on that reach and abort actions that have 0 progression pickups
well, 0 progression pickups and 0 uncollected_resource_nodes/usable_progression_pickups

Miepee:
Fixes #477 

Tried to test this with echoes a bit and didn't see anything immediately wrong.
Not sure by how much this increases the time again, but I guess we'll see

### DIFF
--- a/randovania/generator/filler/retcon.py
+++ b/randovania/generator/filler/retcon.py
@@ -115,7 +115,7 @@ class EvaluatedAction(typing.NamedTuple):
         return EvaluatedAction(self.action, new_reach, self.multiplier, self.offset)
 
 
-def _evaluate_action(base_reach: GeneratorReach, action_weights: ActionWeights, action: Action) -> EvaluatedAction:
+def _evaluate_action(player_state: PlayerState, action_weights: ActionWeights, action: Action) -> EvaluatedAction:
     """
     Calculates the weight offsets and multipliers for the given action, as well as the reach
     you'd get by collecting all resources and pickups of the given action.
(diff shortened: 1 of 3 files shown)


randovania/generator/filler/retcon.py
```diff
@@ -134,6 +136,14 @@ def _evaluate_action(base_reach: GeneratorReach, action_weights: ActionWeights,
...
             potential_reach.act_on(resource)
         multiplier *= action_weights.DANGEROUS_ACTION_MULTIPLIER
+        pickups_that_solve, uncollected_resource_nodes, _ = player_state.calculate_potential_actions(potential_reach)
+        # If our potential reach does not allow us to progress in any way, don't try to consider it.
+        if (
+            not pickups_that_solve
+            and not uncollected_resource_nodes
+            and not potential_reach.victory_condition_satisfied()
+        ):
+            multiplier = 0.0
```
henriquegemignani, line 146: Can you do some basic testing to see how often this case shows up? I think you can just do a `print("THE CASE")`, run with no debugging and just do a `| grep -c CASE` or something.

If it's often enough, would be worth trying to skip evaluating this action unless we end in the backup weights case.

Miepee: I tried doing a benchmark for all games.
It only seemed to happen for fusion and out of the 100 gens appeared 126 times. (And even that high amount i think is due to some bug?)
Generally it seems to only appear if you heavily mess with your custom presets

PR #8842, opened 2025-11-26
Disable the GC during generation, for speedups

henriquegemignani:


randovania/generator/filler/runner.py
```diff
@@ -23,6 +24,8 @@ async def run_filler(
     player_pools: Sequence[PlayerPool],
     world_names: list[str],
     status_update: Callable[[str], None],
+    *,
+    disable_gc_during: bool = True,
```
Miepee, line 28: any specific reason why its an argument instead just always assuming this is true?

henriquegemignani: If we need to disable it because this sure feels unsafe.

PR #8843, opened 2025-11-27
Prime: Sun Tower revisit

Miepee:


randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1464,18 +1467,16 @@ Extra - aabb: [203.54953, 323.97684, 5.845402, 266.79828, 366.83566, 99.515015]
...
                       Combat (Ludicrous)
-                      All of the following:
-                          Shoot Any Beam
-                          Any of the following:
-                              Missile ≥ 8
-                              Missile ≥ 4 and Disabled Hard Mode
+                      Missile ≥ 4 and Shoot Any Beam
   > Event - Unlock spider track
       Scan Visor and Shoot Super Missile
   > Event - Activate Sunchamber Ghosts From Bottom
```
Miepee: This node cane be removed now, as it doesn't actually achieve anything logically.
But I'm not sure if by how much it hinders readability?

Miepee:
Thanks for the links
Based off of the image present in the second discord link, i decided to actually remove the "trigger ghosts from bottom" node as it was just way too confusing with all the weird cases.

I also looked into the generation logs and looks like that hasn't changed signifcantly. Both were done on trickles.
Before:
- Collects plant
- Is forced to collect "trigger ghosts from top" as thats the only action and it cannot go down otherwise
- Before going down, will try to either place more items or collect more pickups. All of which are dangerous

After:
- Collects plant
- Has the option on either placing more items, collect more pickups, or trigger "ghosts from top". All of which are dangerous. Ghosts being an event, has the lowest weight at 0.375 (0.5 * 0.75). However, if it picks the ghosts event, then nothing majorly actually happens.

So both cases are basically the same: It either places items to unlock more locations, or it goes to a new location.
Also checked just in case and both are also still doing a backwards safe plant (aka with bombs+spider+charge+supers, collecting plant, the ghost trigger events, ghosts itself, wild artifact all at once)

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1496,11 +1497,11 @@ Extra - aabb: [203.54953, 323.97684, 5.845402, 266.79828, 366.83566, 99.515015]
       All of the following:
           Morph Ball and Knowledge (Intermediate)
           Any of the following:
-              # https://youtu.be/spZZY6E95ZU
-              Complex Bomb Jump (Intermediate) and Normal Damage ≥ 20 and Can Use Bombs
-              # Roll down to the gap in the track and back up to trigger ghosts, skipping super missiles. This connection makes fighting Flaahgra without Spider Ball a dangerous action.
+              # Early Wild: https://youtu.be/spZZY6E95ZU
+              Complex Bomb Jump (Advanced) and Normal Damage ≥ 20 and Can Use Bombs
```
Samuel6710: I'd personally leave this in intermediate despite the punishment for failing the trick because there are many more palyers who are comfortable with early wild but not UBJs/HBJs compared to players who know both. The punishment is bad, but not enough to lug it in with the most brutal timing based tools the game, in my opinion.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1496,11 +1497,11 @@ Extra - aabb: [203.54953, 323.97684, 5.845402, 266.79828, 366.83566, 99.515015]
...
+              # Roll down to the gap in the track and back up to trigger ghosts, skipping super missiles.
               Can Use Spider Ball
-              # https://youtu.be/pEruOGbEteI
+              # Buffer an Instant Unmorph Jump: https://youtu.be/pEruOGbEteI
```
Samuel6710: "Buffer" implies holding the button until it registers, I'd say `Roll into the corner with vines and fall and Instant Unmorph Jump` or something similar.

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: What are other advanced complex bomb jumps?
Failing this trick may not be brutal in terms of execution, but it is brutal in terms of  punishment. Not only does failing it mean that you have to reset and backtrack ~15+min (depending on your trick, dlr and other settings this number can very much go higher) just to have another attempt at it, but also saving after failing has the chance to permanently brick your save file. 

I know that @jeffgainsngames also wanted to have it as Advanced. Would be curious to know other maintainers thoughts.

Uncle-Reggie: This cleanup is removing more and more gateways tricks in between logic stages. Early Wild is an intermediate bomb jump. It is the easiest complex bomb jump in the game. Yes it sucks to fail, but this is a gateway trick into the rest of the category.

Gateway tricks are important. Sometimes we have a trick that could be either but it rated slightly lower because it helps push up into the next category.

Miepee: > Sometimes we have a trick that could be either but it['s] rated slightly lower 

This is something I very much disagree with
1) Having a trick "that could be either" generally means that the trick difficulties are defined too loosely. Part of Jeff's (and mines) push, was to actually formally define them as strict as we can, so that we get rid of "could be either" wherever possible.
2) In the past where we had cases where "it could be either", the precedent was basically almost always to push them onto the *higher* difficulty, not lower one. I do think that "gateway tricks" are helpful to some extent, but sometimes I feel like in the Prime games that is pushed a little bit too far. I've heard now from i-dunno-how-many-people-anymore that they often times do not enable a trick at beginner/intermediate precisely because there are a few weird outliers that they either cannot or have immense difficulties performing. Which is precisely why we have the precedent of rating them higher, so that these things don't happen.


---

Ultimately, if the majority of Database Maintainers want to keep this at intermediate, then sure, I'll leave it there. But for me the fact that 1) a fail either means a game reset with at least re-fighting one boss or a bricked save file and 2) you are basically forced to do that trick on every run where you have it enabled (unless you have spider(+supers) to do the easier strat which is **way** rarer to happen on higher difficulties) just makes me want to push it the higher difficulty for safety and for consistency with other games

JeffGainsNGames: good to confirm as well, we do have a beginner CBJ based on HotE (which also requires expert Bomb Jump?), not sure which it is but I believe it's this method <https://youtu.be/EuJuuxaXnvk?si=hvsR6AeW4a6xVeWR&t=30> or the 2x DBJ thingy (also shown in that link)
image
pretty sure I remember something about this classification, but not enough to be confident

Samuel6710: That beginner complex bomb jump is the triple bomb jump. It originally didn't have that complex bomb jump trick requirement, but was added in because of the bomb refill timer abuse that's required

Uncle-Reggie: The removal of gateway tricks is antithetical to the original design of the DB for Prime when it was made. There were many tricks we purposefully placed in the positions they were in to be gateways. If we want to just clean cut everything, then fine. It seems like the idea of having those tricks is causing more problems than the good we intended, and I don't feel like doing this back and forth every time we chop another one down, so I'm gonna tune out for a while.

Uncle-Reggie: What doesn't make sense about this reclassification is that it's the only Intermediate Complex Bomb Jump. EW is your chance to learn how to abuse the bomb timer before having to do harder things like HBJs and UBJs. You're skipping the learning curve and lumping it with all the other advanced tricks, which it is significantly easier to do than.

JeffGainsNGames: I would not argue the difficulty of Early Wild vs other complex bomb jumps (HBJs, UBJs) from a purely mechanical point of view, but the fact that EW can only be attempted once (assuming no lucky recovery) and has a much greater time to retry was my inclination to place it at a higher difficulty based on our documentation:
image
I will also note we have the tools available for practicing EW, which at least makes learning the trick friendlier than without (not all players can access savestates). And while practice world does offer HBJ/UBJ practice, in-game examples might be slightly different timing (at least for HBJs and collision)
My initial thought was to make the Triple Bomb Jump (above) the initial trick in CBJ at Intermediate (not realizing it *was* already in the database as Beginner (though also requires Expert Bomb Jumps)) since it's inputs are relatively less complex and time to retry is minimal. So it's current existence requires me to re-assess my own thinking.

Miepee: Something that we do quite commonly (altho i'm not sure if we have it mentioned somewhere in the docs) is that we skip out a Beginner level for a trick alltogether as an indicator of "this trick is really for skilled players only, and newbies should not enable this". Could potentially do the same with Complex Bomb Jumps too, similar how we do it already with single room OoB or IS.

Does make me trend towards leaving Early Wild as intermediate for now until we get to reworking complex bomb jumps. I do remember there being a few discussions every now and then on the current categorizations, and there are few enough of them (~30 or so) where they can all be reassessed in one swoop

toasterparty: When I play, I sometimes set Complex Bomb Jump to Intermediate. If EW was moved up to Advanced, I would not feel pressure to learn more randomizer bomb jumps and instead would just set CBJ to Disabled, ruining EW for me.

It's been mentioned a few times that EW is a one-shot trick, but that's not entirely true compared to something like Magma Pool dash. If you're anywhere in the ballpark of the correct analog stick angle, you'll fall back down to the Eyon and get another attempt. IIRC you get an average of about 5 attempts per etank.

If you want to disincentivize the generator/player from exposing/being exposed to dangerous logic such as going down Sun Tower without the means to get back up, then you should do that by expanding the "Dangerous Gravity Suit Logic" option to mean all logic and normalize that checkbox across all games (as as opposed to burying it under higher trick difficulties)

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1472,22 +1475,18 @@ Extra - aabb: [203.54953, 323.97684, 5.845402, 266.79828, 366.83566, 99.515015]
...
-                      All of the following:
-                          Shoot Any Beam
-                          Any of the following:
-                              Missile ≥ 8
-                              Missile ≥ 4 and Disabled Hard Mode
+                      Missile ≥ 4 and Shoot Any Beam
   > Event - Unlock spider track
       Scan Visor and Shoot Super Missile
-  > Event - Activate Sunchamber Ghosts From Bottom
-      After Sun Tower Spider Track Unlocked and After Flaahgra and Can Use Bombs and Can Use Spider Ball
```
toasterparty, line 1490: Please compare the progression placement probability for item locations before/after this change to validate that we haven't introduced a big placement bias here.

randovania/games/prime1/logic_database/Chozo Ruins.txt
duncathan: > If you want to disincentivize the generator/player from exposing/being exposed to dangerous logic such as going down Sun Tower without the means to get back up, then you should do that by expanding the "Dangerous Gravity Suit Logic" option to mean all logic and normalize that checkbox across all games (as as opposed to burying it under higher trick difficulties)

Dread has already done this and it worked well over there

Miepee: Out of curiosity, is there anything else in prime that you would place under a hypothetical checkbox like this? I can't really think of any other cases that exist of the top of my head (other than gravity stuff) where if you chose to not do a trick, that you may be permanently locked out of doing it.

toasterparty: Almost everything that's a `before` event fits this logic category. For example, you can't collect the Research Core item too early if logic needs you to climb IRW without SJ.

I also think the generator can do optimized hell runs using etanks as health refills, something you can ruin a save with if you collect them and save at the wrong station.

There's also a few ubiquitous world PoNR thresholds to consider like Vent Shaft before Boost, Omega Research before Spider, Far Phen before Space Jump, etc. but that's only dangerous if you have warp-to-start disabled which is not recommended.

Miepee: Ah, so we have quite a few cases of those. Cool.

randovania/games/prime1/logic_database/Chozo Ruins.txt
Miepee: I compared 1k starter preset on ce9b5c6ae2ee4f472cc7ca43290aee2e0c6d48f0 vs 6fca333 (main vs last commit of this pr as of time of writing)
There's a few things that change, but nothing that seems like the cause of this PR. The quick rundown:
- Missile expansions are ±4% on a lot of places. Etanks on a lot less places
- Vanilla bombs is -3% on this PR (went from 9.5% to 6.5%)
- Shore Tunnel and Fungal Hall Access Artifacts are +3%, Research Core is -3%
- "Progression Probabilty" jumps around in ways I don't quite fully understand. Phen Shorelines Ice Wall and 
PMT are +3%, Shore Tunnel, PCA and Storage Depot A are -3%.
- All changes are within [-4.5%; 4.5%]

Would intuitively say that it's just margin of error.

toasterparty: > Would intuitively say that it's just margin of error.

I'm inclined to agree.

If vanilla bombs actually went down because of this, that'd make this net positive

PR #8844, opened 2025-11-27
Disable GC during the resolver

henriquegemignani:


randovania/resolver/logic.py
```diff
@@ -70,11 +73,19 @@ def get_attempts(self) -> int:
...
+    def resolver_quit(self) -> None:
+        if self.disable_gc:
+            gc.enable()
+
     def start_new_attempt(self, state: State, max_attempts: int | None) -> None:
         if max_attempts is not None and self._attempts >= max_attempts:
             raise ResolverTimeoutError(f"Timed out after {max_attempts} attempts")
 
         self._attempts += 1
+        if self.disable_gc and self._attempts % 50 == 0:
```
Miepee, line 89: Any kinds of measurements done for this, or just "felt right"?

henriquegemignani: full gc was stupidly slow even if only after the re-enable. gc(0), gc(1) and random attempt values didn't have a significant cpu time. Didn't bother looking at memory profiler.

PR #8847, opened 2025-11-28
Move core loop of Resolver and Generator to native plus GraphRequirement

henriquegemignani:
Depends on #8838.

### DIFF
--- a/randovania/lib/cython_helper.pxd
+++ b/randovania/lib/cython_helper.pxd
@@ -24,7 +24,8 @@ cdef extern from *:
             obj = set_obj;}
         
         ~PyRef() {
-            Py_XDECREF(obj);obj = NULL;
+            Py_XDECREF(obj);
+            obj = NULL;
         }
 
         PyRef(const PyRef& other)  {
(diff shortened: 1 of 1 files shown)


randovania/lib/cython_helper.pxd
```diff
@@ -0,0 +1,73 @@
...
+
+        PyRef(const PyRef& other)  {
+            Py_XINCREF(other.obj); 
+            obj = other.obj;
+        }
+        PyRef(PyRef&& other) {obj = other.obj; other.obj = NULL;}
+
+        PyRef& operator=(const PyRef& other) {
+            Py_XDECREF(obj); 
+            Py_XINCREF(other.obj);
```
Tyranisaur: I think it would be safer to swap the order of these, so you do the increment before the decrement, for better safety in the face of self assignment.

Tyranisaur (approved):

randovania/lib/cython_helper.pxd
henriquegemignani: Can actually check if it's the same obj and skip the reference counting!

PR #8848, opened 2025-11-28
Prime: Ruined shrine revision

Miepee:


randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -538,12 +548,12 @@ Extra - aabb: [-95.53676, -38.609608, -1.0693345, -22.250599, 43.303684, 30.2103
   > Door to Ruined Shrine Access
       Any of the following:
           Space Jump Boots
-          # https://www.youtube.com/watch?v=nJVx6BVzXNc
-          Combat/Scan Dash (Intermediate) and Standable Terrain (Intermediate)
-          L-Jump (Intermediate) and Standable Terrain (Intermediate)
+          # NSJ escape: https://www.youtube.com/watch?v=nJVx6BVzXNc
+          Combat/Scan Dash (Intermediate) and L-Jump (Intermediate) and Standable Terrain (Intermediate)
```
Miepee: After doing this trick for a while, i wonder whether either the l-jump or the standable should be advanced.
Because the jump to the last standable doesnt feel intuitive enough for me to be intermediate. You need to be far enough to the left, so that when you jump you dont bump on the sloped wall which decreases your height

Uncle-Reggie: There is borderline no cost of failure. It's a bit tricky, but you've got all the time in the world to pull it off and if you fall you can just do it again.

Miepee: My reasoning for bumping it is *because* it is tricky. It just doesn't feel at a similar level compared to other intermediate tricks.
If we use the very broad guidelines that jeff brought up:
> Intermediate tricks require a good foundation of a tricks concepts to execute but are still easy to pick up and repeat. Seeing a video is helpful but precise inputs likely aren't required.
> Advanced tricks require practice but are able to be performed consistently by skilled players. A video with inputs is likely required to understand the key aspects of the trick.

The trick feels precise enough where intermediate just feels too low. For the longest time I was *very* inconsistent at getting it (with like a 90% failure rate or something), and I only now, after practicing the trick a bunch and looking at the collision in PWE, have a better understanding of what I actually need to do the trick successfully.

Uncle-Reggie: Each category has a couple gateway tricks to the next level if the cost of failure is super low like this one is. If you think it's that hard, then bump it, but the cost of failure is why it's rated where it is.  You've got infinite retries with no cost of failure.

Samuel6710: I personally still struggle a bit when it comes to that jump, the direction your momentum is being pushed so that you don't hug the slope and lose it makes it tricky enough to be advanced I'd say.

Miepee: > the cost of failure is why it's rated where it is. You've got infinite retries with no cost of failure.

Good to know the background for it, thanks. 
I'll bump it up then, cause it can be super frustrating to keep failing a trick with no clear indication of what one is doing wrong, even if the trick is easily repeatable. The Observatory dash comes to mind for me as a comparison which is also easily repeatable without any failure cost, but still being rated advanced.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -553,22 +563,27 @@ Extra - aabb: [-95.53676, -38.609608, -1.0693345, -22.250599, 43.303684, 30.2103
...
           Bomb Jump (Beginner) and Can Use Bombs
           All of the following:
-              Standable Terrain (Intermediate)
+              # NSJ escape the pit, then dash over the risen wall. Dash either off of Beetles or something scannable.
+              L-Jump (Intermediate) and Standable Terrain (Intermediate)
               Any of the following:
-                  Combat/Scan Dash (Advanced)
-                  Scan Visor and Combat/Scan Dash (Intermediate)
+                  Combat/Scan Dash (Expert)
+                  Scan Visor and Combat/Scan Dash (Advanced)
```
Samuel6710, line 586: Advanced seems a little high here? What's the idea behind bumping it up?

Miepee: My reasoning for bumping it, is that dashing over the wall is significantly harder than dashing over the beetle-fight-trigger. Not only do you have to do a better dash, but if you fail you also have to repeat the NSJ escape

randovania/games/prime1/logic_database/Chozo Ruins.txt
toasterparty: It's an intermediate L-Jump and an intermediate dash without a doubt.

I could see a case for advanced standable terrain since it's a cut above "just stand on this one tiny thing and you're good"

Miepee: > It's an intermediate L-Jump and an intermediate dash without a doubt.

I wouldn't exactly call it "without a doubt" if there are quite a few racers (according to sam and jeff) that still struggle with that
EDIT: nevermind, misread

Miepee: Bumped it to advanced standable

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -483,29 +499,23 @@ Extra - aabb: [-95.53676, -38.609608, -1.0693345, -22.250599, 43.303684, 30.2103
...
+  > Top of Half Pipe near pickup
+      Any of the following:
+          Can Use Boost Ball
+          # Use petruding root: https://youtu.be/93RLiYdob9Q
+          Space Jump Boots and Standable Terrain (Beginner)
+          # NTSC-only NSJ: https://youtu.be/w0TIiZwv3Gk
+          L-Jump (Intermediate) and Standable Terrain (Advanced)
+          Complex Bomb Jump (Ludicrous) and Can Use Bombs
+          # Slope Jump up
+          Space Jump Boots and Slope Jump (Advanced)
```
toasterparty, line 518: TIL

PR #8870, opened 2025-12-13
Better use of Cython for Generator/Resolver

henriquegemignani:
- Improve ResourceCollection.duplicate
- Generator: don't use rustworkx, use our own Cython code adopting networkx code. This lots of python overhead.
- Less use of unordered_map, it's slow
- GraphRequirement: special case checking the trivial set by pointer, as it's used WAY too often
- Improve graph duplication a lot for DLR.

### DIFF
--- a/randovania/resolver/resolver.py
+++ b/randovania/resolver/resolver.py
@@ -69,7 +69,7 @@ def _is_dangerous_event(state: State, action: ResolverAction, dangerous_resource
         return False
 
     resource_indices = action.resource_gain_bitmask.get_set_bits()
-    if resource_indices.size() <= 1:
+    if len(resource_indices) <= 1:
         # Only one resource and one is guaranteed to be an event.
         return True
 
--- a/randovania/lib/bitmask.py
+++ b/randovania/lib/bitmask.py
@@ -143,6 +143,7 @@ if cython.compiled:
         def get_set_bits(self) -> vector[cython.size_t]:
             """Gets a list of all set bit indices."""
             result: vector[cython.size_t] = vector[cython.size_t]()
+            # For non-Cython callers, this results in a regular list
 
             idx: cython.size_t
             for idx in range(self._masks.size()):
@@ -228,9 +229,9 @@ else:
         def is_subset_of(self, other: BitmaskInt) -> cython.bint:
             return self._mask & other._mask == self._mask
 
-        def get_set_bits(self) -> vector[int]:
+        def get_set_bits(self) -> list[int]:
             """Gets a list of all set bit indices."""
-            result: vector[int] = []
+            result: list[int] = []
 
             mask_str = bin(self._mask)[2:]
             idx = len(mask_str)


randovania/game_description/resources/resource_collection.py
```diff
@@ -46,13 +50,37 @@ def __init__(self, resource_database: ResourceDatabaseView, resource_array: vect
...
+        result: ResourceCollection = ResourceCollection.__new__(ResourceCollection)
+        result.resource_bitmask = Bitmask.create_native()
+        result._existing_resources = {}
+        result._resource_database = resource_database
+
         if cython.compiled:
             result._resize_array_to_fit(count)
         else:
             result._resource_array = vector([0]) * count
+            result._damage_reduction_cache = unordered_map[cython.size_t, cython.float]()
```
Tyranisaur, line 62: For what it's worth, `std::unordered_map` has a `reserve` function` that seems applicable here, but I don't think it's very important.

randovania/graph/graph_requirement.py
```diff
@@ -243,11 +253,37 @@ def __str__(self) -> str:
...
+    @cython.exceptval(check=False)  # type: ignore[call-arg]
+    def _find_other_idx(self, resource_index: cython.int) -> cython.int:
+        """
+        Searches self._other_resources for an entry with the given resource_index.
+        Returns the index of the found entry, or -1 if not.
+        """
+        i: cython.int
+        for i in range(self._other_resources.size()):
+            if self._other_resources[i].first == resource_index:
+                return i
```
Tyranisaur, line 266: Is it viable to use `enumerate` to get a bit neater syntax for these loops?

randovania/resolver/resolver.py
```diff
@@ -60,15 +60,22 @@ def _simplify_additional_requirement_set(
...
+        resource_indices = action.resource_gain_bitmask.get_set_bits()
+        if len(resource_indices) > 1:
+            mapping = state.resource_database.get_resource_mapping()
+            for resource_index in resource_indices:
+                resource = mapping[resource_index]
+                if resource.resource_type == ResourceType.EVENT and resource in dangerous_resources:
+                    return True
+        else:
+            return True
+    return False
```
Tyranisaur: Could flip the first two if's to use early returns and avoid the nesting

Tyranisaur (approved):

randovania/graph/graph_requirement.py
henriquegemignani: Don't think so, the intent is to generate a proper c++ loop here.

Can test and check it later

henriquegemignani: image

Very bad.

randovania/game_description/resources/resource_collection.py
henriquegemignani: The damage reduction cache is only has a few keys at maximum (one for every damage resource). I'd say the ResourceCollection objects are created a lot more often than growing this cache will matter, especially because we end clearing it every time we add a new resource (for now...)

PR #8871, opened 2025-12-14
CI: Use file-based compilation instead of env var

henriquegemignani:


Miepee (approved):
nvm read the rest on discord

PR #8874, opened 2025-12-15
Dread: Remove Diffusion Abuse trick to push Wide Block Purple EMMI Arena

Tyranisaur:
This trick cannot be executed here when defeating the EMMI because you go into the Omega Cannon mode when you hold the aim button.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ##### Ferenia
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -49,6 +49,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed: Minor adjustment related to Kraid to increase seed variety.
 
+##### Ferenia
+
+- Fixed: Removed Diffusion Abuse trick to push Wide Beam Block in Purple EMMI Arena.
```
Miepee: adding reasoning might be good here? just so that its clear why thats a fix

PR #8891, opened 2025-12-30
Add remote activation nodes

### LINKED ISSUE #3690: Add RemoteActivationNode
This node lets you collect a different resource node without having to directly visit it. The remote node does not need to be at the same Area or even World. After remote collection, the state will still be at the remote activation node.

Should support at least:
- EventNodes
- LogbookNode
- DoorLockNode

Miepee:
fixes #3690
Added a test room into BDG which does seem to generate and resolve successfully.

Known issues:
- renaming a node in the data visualizer that a RAN references causes the reference to not fully update
- playthrough tab lists both the original node and the reference for pickups
- some very ugly hack in regards to WG's node_by_pickup_index where i just wanted some quick fix to get something to work (which i'm not sure is actually correct and will probably conflict with #8870?)

I have not tested remote activating anything but pickups/events (so hints/docklocks/teleporternetwork. couldn't think of anything nice in the crane scenario)


henriquegemignani:
> renaming a node in the data visualizer that a RAN references causes the reference to not fully update

https://github.com/randovania/randovania/blob/main/randovania/game_description/editor.py#L154-L210

Edit: should've read the PR first :P
but it's that function indeed that fixes all references to a node after it gets renamed

randovania/generator/filler/player_state.py
```diff
@@ -335,6 +336,10 @@ def build_available_indices(
     def get_pickup_node(node: Node | None) -> PickupNode:
         if isinstance(node, EventPickupNode):
             return node.pickup_node
+        if isinstance(node, RemoteActivationNode):
```
henriquegemignani: I'd say we just skip the remote nodes in this case. All we care is all pickup indices, so just finding all PickupNodes is enough,

randovania/graph/world_graph.py
```diff
@@ -739,7 +752,11 @@ def _adjust_graph_for_patches(
 
     # Add dock connections
     for node in graph.nodes:
-        if isinstance(node.database_node, DockNode):
+        db_node = node.database_node
+        if isinstance(node.database_node, RemoteActivationNode):
+            db_node = graph.node_identifier_to_node[node.database_node.remote_node]
```
henriquegemignani: Why?

henriquegemignani (changes requested):

Miepee:
> but it's that function indeed that fixes all references to a node after it gets renamed

yeah, i probably just missed something there, that i haven't investigated too deeply yet

randovania/graph/world_graph.py
Miepee: My thought process is that if a RAN references a docknode, that it should get the resourceGains that stem from the lock. 
But looking at the _create_dock_connection in a bit more detail again, the resulting connection from it would just be totally wrong I think.

Would Dread (/MSR?) have any use for that right now out of curiosity? I feel like it's a huge pitfall, especially with DLR

Miepee:
Also i'm really suspicous of the fact that *every* game's hash is affected. Need to look into what caused that too.

randovania/graph/world_graph.py
henriquegemignani: All you'd need to do is grant all resources the referenced node grants.

henriquegemignani:
> Also i'm really suspicous of the fact that _every_ game's hash is affected. Need to look into what caused that too.

Same. Really weird. _maybe_ migration number, bu development permalinks shouldn't care about that.

randovania/graph/world_graph.py
Tyranisaur: Dread has a couple of instances where remote activating door locks would be a cleaner than the hack we do today where we have some connections to represent unlocking doors through ceilings/floors. It would need to be kept outside of door lock randomization, because it's going to be hard to express the ability to open doors for that purpose.

To the degree that I know SR, I can think of some instances where there's similar constructs there.

duncathan: Dread would love to have the ability to model unlocking *specific lock types* remotely. Only power, charge, wide, plasma, and wave doors can be unlocked through walls. Without the ability to specify that, then as Ty said it would just be a bit of a cleanup for our existing no-DLR connections

duncathan:
Can't do a proper review yet but one immediate thing: I don't actually like the name, despite it being our working name for it forever lol. "Activation" is a weird word for this. I'd rather something like RemoteNode, RemoteResourceNode, RemoteCollectionNode

henriquegemignani:
> Can't do a proper review yet but one immediate thing: I don't actually like the name, despite it being our working name for it forever lol. "Activation" is a weird word for this. I'd rather something like RemoteNode, RemoteResourceNode, RemoteCollectionNode

Out of these names, I prefer `RemoteCollectionNode`.

PR #8897, opened 2026-01-04
Fusion: Fix geron weakness logic+migration

Miepee:
Looking at this more, I really do want to keep the shinespark/speedbooster vulnerability for gerons. Lots of cool skill expression that's just gone, but afaik not easily fixable as its intertwined with screw :(

I didn't touch the drain pipe walk, because not sure yet if we want to fix that or not.

### DIFF
--- a/test/test_files/patcher_data/fusion/fusion/starting_items/world_1.json
+++ b/test/test_files/patcher_data/fusion/fusion/starting_items/world_1.json
@@ -786,19 +786,19 @@
                 "Room": 39,
                 "BlockX": 4,
                 "BlockY": 6,
-                "Item": "MissileTank",
-                "ItemSprite": "MissileTank",
-                "Jingle": "Minor",
+                "Item": "SuperMissiles",
+                "ItemSprite": "SuperMissiles",
+                "Jingle": "Major",
                 "ItemMessages": {
                     "Kind": "CustomMessage",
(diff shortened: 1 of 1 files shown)


randovania/layout/preset_migration.py
```diff
@@ -1213,7 +1213,9 @@ def _migrate_v109(preset: dict, game: RandovaniaGame) -> None:
 
 def _migrate_v110(preset: dict, game: RandovaniaGame) -> None:
     if game == RandovaniaGame.FUSION:
-        preset["configuration"]["adjusted_geron_weaknesses"] = True
+        knowledge_enabled = preset["configuration"]["trick_level"]["specific_levels"].get("Knowledge")
+        val = False if knowledge_enabled else True
+        preset["configuration"]["adjusted_geron_weaknesses"] = val
```
JeffGainsNGames, line 1218: just to confirm, the idea here is only migrate the setting to True when knowledge is disabled? Makes sense since this is a setting targeting trickless

Miepee: yes.
could also redo this once we have a parameter to differentiate between "is this preset contained in rdvgame" vs "is this a normal preset", because ideally we'd like this to target everyone and not just trickless.

JeffGainsNGames (approved):

PR #8909, opened 2026-01-15
Optimize WeightedLocations.can_fit

henriquegemignani:
With a 40-world echoes multi:
```
Before: Took 133.969 seconds. Hash: ACEXZGLS
 After: Took  80.975 seconds. Hash: ACEXZGLS
```

### DIFF
--- a/randovania/generator/filler/weighted_locations.py
+++ b/randovania/generator/filler/weighted_locations.py
@@ -31,7 +31,7 @@ class WeightedLocations:
     def __str__(self) -> str:
         return str(self._items)
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"WeightedLocations({self._items!r})"
 
     def is_empty(self) -> bool:


randovania/generator/filler/weighted_locations.py
```diff
@@ -17,9 +17,9 @@
 
 
 class WeightedLocations:
-    _items: dict[tuple[PlayerState, PickupIndex], float]
+    _items: list[tuple[PlayerState, dict[PickupIndex, float]]]
```
Miepee, line 20: how often is count_for_player (or something similar where you need to get stuff from a specific player) invoked? does it make sense to have this as a `dict[PlayerState, dict[PickupIndex, float]]`?

randovania/generator/filler/weighted_locations.py
```diff
@@ -31,55 +31,71 @@ def __eq__(self, other: object) -> bool:
     def __str__(self) -> str:
         return str(self._items)
 
+    def __repr__(self):
```
Miepee: missing return annotation in case you didnt see

randovania/generator/filler/weighted_locations.py
henriquegemignani: `PlayerState` isn't hashable

Miepee: ah. that'd do it.

randovania/generator/filler/weighted_locations.py
henriquegemignani: ahh I hate annotations on these methods, they need to always return str >.>

PR #8911, opened 2026-01-15
Improve UncollectedState creation

henriquegemignani:
same 40 multi of the other pr: before 80.652, after 66.027

### DIFF
--- a/randovania/graph/state.py
+++ b/randovania/graph/state.py
@@ -84,9 +84,6 @@ class State:
             copy.copy(self.hint_state),
         )
 
-    def collected_pickups(self, graph: WorldGraph) -> set[PickupIndex]:
-        return state_native.state_collected_pickups(self.resources, graph)
-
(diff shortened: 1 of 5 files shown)


randovania/graph/state.py
```diff
@@ -84,6 +84,9 @@ def copy(self) -> Self:
             copy.copy(self.hint_state),
         )
 
+    def collected_pickups(self, graph: WorldGraph) -> set[PickupIndex]:
+        return state_native.state_collected_pickups(self.resources, graph)
```
Miepee: i dont quite get right now, why use this over `collected_pickup_indices`?

Also how does this function work,when you're running in pure mode?

henriquegemignani: Good question. Should've edited that one instead of blindly creating this one.

As for pure mode, all the "native" files use cython pure mode so they're compatible.

Miepee (approved):

PR #8913, opened 2026-01-16
Fix as_set for NodeRequirement

henriquegemignani:


Miepee:
If you figured out how to trigger this, can you add a test case too?

henriquegemignani:
It was "open touch details popup for dread", which isn't the best test case...
I actually opened the test file and couldn't think some decent case that I was willing to make right now

Miepee:
Hmm. Ok.

PR #8917, opened 2026-01-17
Add Dread preset setting to replace item popups with async messages

ArcanoxDragon:
This PR is directly related to (and depends on) randovania/open-dread-rando-exlaunch#12 and randovania/open-dread-rando#411.

This adds a new preset setting for Metroid Dread in the "Other" tab that allows replacing "Item Acquired" popups with an async message near the top of the screen. The primary motivation for this setting is to make the game feel more fluid and to avoid the several-second-long interruption while waiting for the game to allow dismissing the popup.

The setting defaults to disabled, meaning the vanilla item behavior is left intact without explicit preset changes.

When the setting is enabled, any games generated with the preset will use the new feature in `open-dread-rando-exlaunch` to skip displaying a popup when collecting items. Instead, collecting an item will display a message at the top of the screen for 5 seconds, just like when receiving an item in a Multiworld session. All other behavior pertaining to collecting pickups is left intact. This includes pickup sounds, checkpoints (which normal pickups don't have, but just in case), and incrementing the "current" count of resources like missiles or PB ammunition along with the maximum.

Video of this feature in action:

https://github.com/user-attachments/assets/77f5759d-d239-4263-8c0e-f6db774a3226

I have played two full seeds with this new setting enabled and did not encounter any odd behavior or problems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed: Energy Parts are now logical.
 - Added: An option to enable Lua debug logging is available in the Cosmetic Options dialog when "Show Experimental Settings" is enabled (intended for game patcher developers).
+- Added: An option is now available in the "Other" preset settings tab to skip item acquisition popups, replacing them with a short on-screen message instead (similar to receiving Multiworld items).
 
 #### Logic Database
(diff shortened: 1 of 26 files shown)


ArcanoxDragon:
Tests will fail until a new ODR is released with randovania/open-dread-rando#411 and the dependency is updated

PR #8918, opened 2026-01-18
Allow cosmetic option dialogs to access Options

ArcanoxDragon:
This was originally just part of #8917, but it was suggested to make this its own PR so that other games can potentially benefit from it before that PR is merged.

The only change this makes is to pass an instance of `Options` to all Cosmetic Patch Options dialogs when they are constructed. This allows, for instance, showing or hiding specific cosmetic options depending on the "Show Experimental Settings" option.

### DIFF
--- a/test/games/am2r/gui/test_am2r_cosmetic_patches_dialog.py
+++ b/test/games/am2r/gui/test_am2r_cosmetic_patches_dialog.py
@@ -13,6 +13,8 @@ from randovania.games.am2r.layout.am2r_cosmetic_patches import AM2RCosmeticPatch
 if TYPE_CHECKING:
     from pytestqt.qtbot import QtBot
 
+    from randovania.interface_common.options import Options
+
 
 @pytest.mark.parametrize(
     ("music_start_value", "option_to_click", "music_end_value"),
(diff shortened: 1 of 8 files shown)


randovania/gui/dialog/base_cosmetic_patches_dialog.py
```diff
@@ -8,19 +8,22 @@
 from randovania.gui.lib import signal_handling
 
 if TYPE_CHECKING:
+    from randovania.interface_common.options import Options
     from randovania.layout.base.cosmetic_patches import BaseCosmeticPatches
 
 
 class BaseCosmeticPatchesDialog[CosmeticPatches: BaseCosmeticPatches](QtWidgets.QDialog):
     _cosmetic_patches: CosmeticPatches
+    _options: Options
```
Miepee, line 17: What's the reasoning as to why we need to save this as an attribute? I don't see that used anywhere here, nor do I see why, if for some reason you do need some attributes later, why you couldn't save them as game-specific cosmetic attributes

ArcanoxDragon: First, to be consistent with all of the other dialog code that accesses `Options` (there are numerous dialogs that store it in an attribute named `_options`). Second, if we're passing it into cosmetic dialogs just "so they can have it" later down the road, it makes sense that it would be easy to access. It seems silly to pass it into all the constructors but not actually do anything with it. Storing it in the new attribute makes it easier for dialogs to do something with it later down the road.

PR #8919, opened 2026-01-18
Add hidden cosmetic toggle for enabling Lua debug logs

ArcanoxDragon:
This depends on randovania/open-dread-rando-exlaunch#13 and randovania/open-dread-rando#411.

All this does is add a hidden option in the Dread cosmetic options dialog that enables Lua debug logging. The option is only visible when "Show Experimental Settings" is enabled in the main window. When the setting is turned on, it activates the exlaunch hook that forwards `Game.LogWarn` calls to the debug syscall, allowing them to show up in emulator consoles.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,7 +81,7 @@ exporters = [
     "factorio-randovania-mod==0.8.1",
 
     # Metroid Dread
-    "open-dread-rando==2.18.0",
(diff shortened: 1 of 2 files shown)


ArcanoxDragon:
Tests will fail until ODR is updated - need new config schema

PR #8927, opened 2026-01-22
Improve changelog regarding how faster it got

henriquegemignani:


Miepee (approved):
let's go, 100 world mention

PR #8930, opened 2026-01-25
ResolverReach: use ProcessNodesResponse in constructor

henriquegemignani:


randovania/resolver/resolver_reach.py
```diff
@@ -44,39 +38,45 @@ def path_to_node(self, node: WorldGraphNode) -> tuple[WorldGraphNode, ...]:
...
 
-        if requirements_excluding_leaving_by_node:
-            satisfiable_requirements_for_additionals = resolver_native.build_satisfiable_requirements(
-                logic,
-                requirements_excluding_leaving_by_node,
-            )
-        else:
-            satisfiable_requirements_for_additionals = frozenset()
+    def node_in_reach(self, node: WorldGraphNode) -> bool:
+        return node.node_index in self._health_at_node
```
Tyranisaur: How about calling this `is_node_in_reach`. And what about taking the node index as the argument directly, instead of the node?

henriquegemignani: As much as taking index is faster, I do prefer the API getting a node. The use cases have node objects so I'll leave it like that for now.

PR #8947, opened 2026-02-06
Dread/shutter gate overrides

### LINKED ISSUE #7067: Door lock overrides for Shutter Gates
Not all docks with Shutter Gates on them are created equal. Some can be traversed in more ways than what is currently logical for the door type. Using lock overrides, we can enable those other ways to be logical.

Tyranisaur:
Closes #7067


randovania/games/dread/logic_database/Burenia.txt
```diff
@@ -929,6 +929,15 @@ Extra - asset_id: collision_camera_010
...
   * Extra - right_shield_def: None
+  * Override default lock requirement:
+    Any of the following:
+        Traverse any Shutter Gate
+        All of the following:
+            Speed Booster
+            Any of the following:
+                Speed Booster Conservation (Intermediate)
+                Gravity Suit and Speed Booster Conservation (Beginner)
+
```
Logarhythmic, line 940: How does gravity lower the Speed Booster Conservation? Are you coming from below and hopping in the water?

randovania/games/dread/logic_database/header.txt
```diff
@@ -554,6 +554,12 @@ Templates
               Bomb or Cross Bomb
               Main Power Bomb and Power Bombs
 
+* Traverse any Shutter Gate:
```
Logarhythmic: A capitial "A" is used for "Lay Any Bomb", maybe we should be consistent here and call it "Traverse Any Shutter Gate"

Logarhythmic (approved):

randovania/games/dread/logic_database/Burenia.txt
Tyranisaur: The idea is that getting over this puddle without gravity warrants a higher difficulty 
image

PR #8965, opened 2026-02-14
OSH Preset Typo

StalledStorm:


Miepee (commented):
this is a user visible gui change  and thus requires a changelog entry

Miepee (approved):

PR #8969, opened 2026-02-17
Fix some old rdvgames with nothing hints not migrating

Miepee:
Tested with May's Super Metroid Dread
https://drive.google.com/drive/folders/1RCvTH_hznZCzrJhUkRdBnbbdoeekl9bx

I'm unsure if there are any "valid" rdvgames out there that hit that case, but wanted to at least fix that plando.


henriquegemignani:
Would be nice to have a test file for the migration.

Miepee:
After comparing SMD with an rdvgame generated on an old version, I noticed that May's plando was actually just using invalid syntax that just happened to work somehow. Seems like previously it would just default to `detailed` if there was any invalid hint item precision. So probably better for May to just update the one line in her plando.

Miepee:
Actually nevermind the above. Looking slightly further into it: "Nothing" was actually valid: https://github.com/randovania/randovania/blob/1ec73bc5b988ca1d079a2ad39a7c2c1c0da83ec6/randovania/game_description/hint.py#L36 . From what I can tell, the literal only way for it to generate validly was to have an ancient multiworld rdvgame from 2.2.0 back when there was still an "owner" precision.

Thus, I added a test rdvgame for it.

randovania/game_description/hint.py
```diff
@@ -34,6 +34,9 @@ class HintItemPrecision(Enum):
     # The exact item
     DETAILED = "detailed"
 
+    # DEPRECATED: vague/useless hint
+    NOTHING = "nothing"
```
Miepee, line 38: I added this back as it would otherwise broken the intent of May's plando. The old text was literally `An item can be found in Artaria`. Replacing it with the pickup's name would've destroyed the joke behind it.

henriquegemignani: Maybe we could have a hidden feature that is 0% precise?

Miepee: Not quite sure what you mean here. You mean for dread specifically? If yes, that'd be a bit weird.

"Detailed" is essentially 100%, so for me it just made sense to have "Nothing" as the 0% precision right underneath

henriquegemignani: Detailed has the special property of using the actual pickup's name. A 0% precise wouldn't. I don't like the idea of adding a new case `HintItemPrecision` explicitly labeled as deprecated and used nothing but migrating ancient rdvgames.

Miepee: From discord: i'll try adding "custom" item precision that can be used for these migrations / plandos

PR #8970, opened 2026-02-19
Fusion: Rename more rooms with unofficial names/sector reference

Miepee:
More of #8931


randovania/games/fusion/hash_words.py
```diff
@@ -49,7 +50,6 @@
     "Zeela",
     "Zero",
     "Zombie",
-    "Zoro",
```
M4RKERSM4SK, line 52: Did you want zero in as  hash word instead? or just remove zoro

randovania/games/fusion/gui/generated/games_tab_fusion_widget_ui.py
```diff
@@ -394,7 +394,7 @@ def retranslateUi(self, FusionGameTabWidget):
 "\n"
 "- Main Deck - Operations Ventilation - repaired the connection between Crew Quarters East and West.\n"
 "\n"
-"- Main Deck - Silo Entry - the Zoro cocoon is moved to allow traversal.\n"
+"- Main Deck - Silo Entry - the Zero cocoon is moved to allow traversal.\n"
```
M4RKERSM4SK, line 397: I havent seen any commits since all the UIC changes and stuff. Previously the generated files would not change or be commited. 

Are they supossed to be commited now?

M4RKERSM4SK (changes requested):
Looks like tests are failing, though not entirely sure why. Looks like maybe something with pretty print on the windows jobs, but somehting else on the mac and linux side?

randovania/games/fusion/gui/generated/games_tab_fusion_widget_ui.py
Miepee: Yes.

randovania/games/fusion/hash_words.py
Miepee: We already have Zero as a hash word. 2 lines right above. It's even visible in the diff :P
grafik

Miepee:
> Looks like tests are failing

Hmm I thought I had fixed those locally but apparently not. Will double check later again.

JeffGainsNGames (approved):

PR #9067, opened 2026-04-06
Allow PBs to open passage to xbox garage upper

StalledStorm:


randovania/games/fusion/logic_database/Sector 6 NOC.txt
```diff
@@ -905,6 +905,12 @@ Hint Features - Multiple Pickups, 1-way Shutter
               Movement (Expert)
               # Spam bombs to hover across: https://youtu.be/4y7ohyL0zkE&t=37
               Movement (Advanced) and Can Use Bombs
+              All of the following:
+                  # Wait for an Owtch to get far enough away and then open the passage with a Power Bomb: https://youtu.be/6ESqvHw7Qyk
+                  Screw Attack and Can Use Power Bombs and Have Any Jump Upgrade
+                  Any of the following:
+                      Can Freeze Enemies With Diffusion
+                      Wave Beam and Can Freeze Enemies With Ice Beam
```
cluekitty: @randovania/metroid-fusion-db Do we think that this type of trick would do well paired with a Knowledge Trick? While the movement doesn't seem too difficult, this is something I don't see a lot of people considering as a possibility. It's certainly not something I would have considered!

M4RKERSM4SK: I did something simular here https://youtu.be/U1LHQA0NoFM, but it was permited in just as a stand on frozen enemies expert (to account for the timing, wait, positioning of the owtch etc)

Should we keep the trend and just gate behind a stand on frozen enemies? Since its vanilla to freeze there, I could see the argument for just a knowledge check.

I do think it should be behind a tirck regardless

CHANGELOG.md
```diff
@@ -15,6 +15,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added: Door Lock Randomizer is now available.
 
+#### Logic Database
+
+##### Sector 6 (NOC)
+
+- Added: X-B.O.X. Garage: The passage to the upper check can now logically be opened with Power Bombs.
```
M4RKERSM4SK: I do not think this is very clear.  "Can use any bombs" is already in logic to open the passage. I think more so this is a trick. Once decided how the trick should defined, I beleive this should be cleared up.

randovania/games/fusion/logic_database/Sector 6 NOC.txt
M4RKERSM4SK: While you are in this section would you be willing to do some cleanup and place in the correct punctuation to match the other comments. Example, Jump off the Crumble Blocks morphed (frame perfect): https://youtu.be/4y7ohyL0zkE&t=14

M4RKERSM4SK: Same cleanup question. as R902. Optional but would be apprecated

randovania/games/fusion/logic_database/Sector 6 NOC.txt
```diff
@@ -905,6 +905,12 @@ Hint Features - Multiple Pickups, 1-way Shutter
               Movement (Expert)
               # Spam bombs to hover across: https://youtu.be/4y7ohyL0zkE&t=37
               Movement (Advanced) and Can Use Bombs
+              All of the following:
```
M4RKERSM4SK: With this logic section added to the bottom, it feel it doesnt flow as nice as if this section would be up by the other "All of the following" section.

However I know its a mega pain in the butt to remove logic and add it back in just for a positioning and readability look.

Do we have a guides or preference on how we want the readability to flow with new tricks added?

CHANGELOG.md
M4RKERSM4SK: The passage can already be logically opened with powerbombs and then the pickup can be obtained with the in place other methods (unless I am reading the line ups wrong):

```
- Logically open passage with any bombs
        - Freeze and bombs
        - movement and springball
        - springball movement
        - movement
        - (now yours of just pbs)
```

So there is already logic in place to open the passage and not require logically to get the pickup. So the changelog of "The passage to the upper check can now logically be opened" doesnt make sense to me. It was already there.

I do understand that while this is not behind a trick yet, it does make more sense on what you have written, but it would be more accurate and clear to me with something along the lines of:

- Added: X-B.O.X. Garage:  <difficulty of trick> <type of trick> to obtain the pickup in upper xbox garage by utlizing a frozen owtch while not requiring bombs.

or if no trick

- Added: X-B.O.X. Garage: Logic added to obtain the pickup in upper xbox garage but utlilizing a frozen owtch while not requiring bombs.

JeffGainsNGames (approved):

PR #9073, opened 2026-04-09
add quantity to DamageReduction

JeffGainsNGames:


henriquegemignani (changes requested):
Add a test case. I think the simplest way is to have a new damage resource in Blank game and add a new test function to `test_graph_requirement_list.py`

JeffGainsNGames:
first time editing blank, does this also require a new set of nodes/connections?
Or is it just a new dmg resource with it's own reduction that only gets hit via the new test function?

henriquegemignani:
yeah it's fine to just create the resource directly in the test. You could add some test for ResourceDatabase and test the `get_damage_reduction` method directly too

henriquegemignani (approved):

PR #9077, opened 2026-04-12
Hunters - Add Exporter

dyceron:
Seeds can now be generated and exported. The PDF was also updated to include some missing fields, as well as adding in the cosmetic window.
image
image


randovania/games/prime_hunters/gui/dialog/game_export_dialog.py
```diff
@@ -1,21 +1,62 @@
...
-    from pathlib import Path
-
     from randovania.interface_common.options import PerGameOptions
 
+# Expected MD5 hashes for vanilla "Metroid Prime Hunters.NDS" based on region
+EXPECTED_MD5_HASHES: list[str] = [
+    "b4c8a9398866b49c7be17d75736a223b",  # USA
+    "d6ac284bc69a8a2119fd9057743b6d89",  # Europe
+    "42850a19d7be2ee5e067df6984aa900e",  # Japan
+    "4f04529f79564020a56e187b8f9865c3",  # Korea
```
henriquegemignani: Unless you're actively testing all these versions and making sure it works, include only the ones that will work.
It's fine to leave the hashes commented out for later imo.

randovania/games/prime_hunters/gui/dialog/game_export_dialog.py
```diff
@@ -52,6 +145,7 @@ def get_game_export_params(self) -> HuntersGameExportParams:
 
         return HuntersGameExportParams(
             spoiler_output=spoiler_output,
-            input_path=self.input_file,
+            input_file=self.input_file,
             output_path=self.output_file,
+            post_export=None,
```
henriquegemignani: Why did you add (copy :P) all the `post_export` logic if it's not used?

henriquegemignani (changes requested):

dyceron:
Addressed

henriquegemignani (approved):

PR #9099, opened 2026-04-23
Dread: Add probability offset for placing Energy

Tyranisaur:
A probability offset it being added for Energy Tanks and Energy Parts for the purpose of increasing the likelihood of boss fights being collected by the generator in an order more similar to what a player might intuitively choose to.

The energy requirements for bosses has a progressive nature, where the first ones must be satisfied first, and the Raven Beak fight is typically the last requirement to beat the game.

With there typically being more energy in the pool than necessary, there are good chances that the player will find energy that was backfilled anyway. Thus, increasing the likelihood of placing energy logically is an attempt at having the logic chain more closely track what a player can reason about being logical.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Metroid Dread
 
 - Changed: Increased likelihood that doors that are logically challenging to reach get locks when using Individual Doors mode to randomize Door Locks.
+- Changed: Increased likelihood of the generator placing Energy Tanks or Energy Parts sooner.
 
 ### Metroid Fusion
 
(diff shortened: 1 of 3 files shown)


duncathan:
Have you done any testing to see if a difference can be felt yet, or is this just an arbitrary initial guess? (The latter would be perfectly fine, I just want a clear picture)

Tyranisaur:
> Have you done any testing to see if a difference can be felt yet, or is this just an arbitrary initial guess? (The latter would be perfectly fine, I just want a clear picture)

I've mostly just looked at the gen order in rdvgames and looked at the generation debug. I tried 1.0 at first, then upped it to 1.5. I can't tell from this how much it does. It's mostly just a vibes thing where I'm party using a ballpark number to try to counterbalance some of the weight these actions conceptually are missing or competing with. I can try generating some batches to get some more concrete data. However, from the outset I'm not sure what results would be considered hard, or how hard I want to push this.

duncathan (approved):

PR #9101, opened 2026-04-27
Dread: Randomizer specific save slots

steven11sjf:
- Added cosmetic option to use different save slots for each ranomizer
- Added short description to cosmetic dialogue
- Implemented in Dread's pdf
- Requires ODR bump

image


ThanatosGit:
I've created a new version of OSRR. ODR, MEDS, and OSRR can now be updated in this PR.

henriquegemignani (approved):
@ThanatosGit @dyceron Given changelog the Samus Returns rando update has relevant changelog entries, but confirmation is nice

ThanatosGit:
OSRR had only one last minute fix where only one edge case is relevant for RDV: When you changed the final boss setting to not being Ridley and the pickup Ridley drops after defeat was an item which was exclusively on Metroids, it would not have spawned.

E.g. if all progressive beams are on Metroids and one on Ridley it would have failed before. But if there is at least one progressive beam as pickup (besides Ridley) it worked.
If Ridley had a unique item like Power Bomb Launcher, it was also no issue.

steven11sjf:
TY, will add to the PR after work

steven11sjf:
> TY, will add to the PR after work

Wow I can't believe I worked an 80 hour shift :3

Added that line to the changelog, should be ready for review.

ThanatosGit (approved):

PR #9102, opened 2026-04-27
Hunters - Bugfixes, Updates, and Bump to OPHR v0.7.0

dyceron:
A lot of issues have been fixed as of this PR, namely:
- Fixed the portal bug when returning to High Ground Magma Drop Entrance in Portal Shuffle
- Fixed many softlocks regarding boss layers in a few rooms
- Missile Launcher now properly is handled during exporting
- Improved formatting and readability of hints
- More revisions are supported including USA 1.1 and EU 1.0/1.1
- Added seed hash to title screen

Hints with Region only:
image
image

Hints with Region and Area:
image
image

Title Screen Hash:
image

### DIFF
--- a/randovania/games/prime_hunters/gui/dialog/game_export_dialog.py
+++ b/randovania/games/prime_hunters/gui/dialog/game_export_dialog.py
@@ -19,17 +19,18 @@ from randovania.gui.dialog.game_export_dialog import (
     prompt_for_output_file,
     spoiler_path_for,
 )
-from randovania.interface_common.options import Options
 
 if TYPE_CHECKING:
-    from randovania.interface_common.options import PerGameOptions
+    from randovania.interface_common.options import Options, PerGameOptions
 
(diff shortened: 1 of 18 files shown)


randovania/games/prime_hunters/gui/dialog/game_export_dialog.py
```diff
@@ -134,7 +135,7 @@ def update_per_game_options(self, hunters_options: PerGameOptions) -> HuntersPer
         return dataclasses.replace(
             hunters_options,
             input_path=self.input_file,
-            output_path=self.output_file.parent,
+            output_path=Path(self.output_file).parent,
```
henriquegemignani: Seems redundant because `self.output_file` is already a Path

dyceron: Oops I could remove it then

henriquegemignani (approved):

PR #9107, opened 2026-04-30
Fixed logic in MSR Area 6 Poisonous Tunnel accidentally using "or" instead of "and"

ThanatosGit:
Also: Is it intentional that they use different `Hazard Damage`?


dyceron:
> Also: Is it intentional that they use different `Hazard Damage`?

I think I was being a little too specific here. They can be the same.

ThanatosGit:
Ok, rebased the branch and made them both use 95.

dyceron (approved):

PR #9111, opened 2026-05-02
Echoes: Add Dark and Light Ammo Expansions to dark- and light-aligned

Naii-the-Baf:
It doesn't make sense for the ammo expansions to not be part of these pickup features.


JeffGainsNGames:
Likely would want to include regular beam ammo as both dark/light based on the same thinking

Naii-the-Baf:
> Likely would want to include regular beam ammo as both dark/light based on the same thinking

Done.

duncathan (approved):
This has been on my to-do list for a while.

JeffGainsNGames:
@duncathan was there anything blocking merge when you approved at the time? (now there's a quick conflict with changelog)

duncathan:
No, nothing blocking

PR #9114, opened 2026-05-05
Fix guest user login

ImperatorPrime:
When playing around with the guest user for testing some multi-world stuff, I was unable to create a session after creating the user. 

The error message was in the form 
```
  File "/home/ImperatorPrime/Projects/randovania/randovania/gui/main_online_interaction.py", line 200, in _host_game_session
    new_session = await self.network_client.create_new_session(session_name)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ImperatorPrime/Projects/randovania/randovania/network_client/network_client.py", line 689, in create_new_session
    assert self.http.headers["X-Randovania-Sid"] is not None
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
KeyError: 'X-Randovania-Sid'
```

I tracked this down to the on_user_session_updated not adding the X-Randovania-Sid header to the session returned when the user is first created. I changed the guest login to connect first (which would get an sid) and then post to the guest login, properly creating the client session with the sid, which on_user_session_updated uses to correctly set the  X-Randovania-Sid http header. After this fix, I encountered no issues with the guest login. I also tweaked the guest user tests to now support passing in an sid.

I had browsing not work after the first guest user was created, but it seems to go away without a restart. Not sure exactly what's going on there.

### DIFF
--- a/randovania/server/user_session.py
+++ b/randovania/server/user_session.py
@@ -49,12 +49,7 @@ def _log_session_info(sa: ServerApp, sid: str, user: User) -> None:
     sa.logger.info(f"Client at {sa.current_client_ip(sid)} is user {user.name} ({user.id}).")
 
 
-async def _create_client_side_session(
-    sa: ServerApp, sid: str | None, user: User | None, session: dict | None = None
-) -> dict:
-
-    # The sid should never be None here
-    assert sid is not None
+async def _create_client_side_session(sa: ServerApp, sid: str, user: User | None, session: dict | None = None) -> dict:
     """
 
     :param user: If the session's user was already retrieved, pass it along to avoid an extra query.
@@ -346,10 +341,8 @@ async def guest_login_post(
         discord_id=None,
     )
 
-    session = {}
-
-    session["user-id"] = user.id
-    await sa.sio.save_session(sid, session)
+    async with sa.sio.session(sid) as session:
+        session["user-id"] = user.id
 
     if sa.is_api_request(request):
         return JSONResponse(await _create_client_side_session(sa, sid, user, session))
--- a/randovania/server/user_session.py
+++ b/randovania/server/user_session.py
@@ -38,12 +38,16 @@ def _encrypt_session_for_user(sa: ServerApp, session: dict) -> str:
 
 def _create_client_side_session_raw(sa: ServerApp, sid: str | None, user: User) -> dict:
     if sid is not None:
-        sa.logger.info(f"Client at {sa.current_client_ip(sid)} is user {user.name} ({user.id}).")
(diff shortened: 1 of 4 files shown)


ThanatosGit:
> I had browsing not work after the first guest user was created, but it seems to go away without a restart. Not sure exactly what's going on there.

Then it is missing something in the POST "/guest_login" login path in the server but it gets fixed by `restore_user_session`. Should be the `user-id` not properly written during guest login.

ThanatosGit:
lgtm so far but please fix the mypy errors.

ImperatorPrime:
I've solved the session being saved on initial guest login - everything seems to work now.

randovania/server/user_session.py
```diff
@@ -36,26 +36,31 @@ def _encrypt_session_for_user(sa: ServerApp, session: dict) -> str:
...
+def _log_session_info(sa: ServerApp, sid: str, user: User) -> None:
+    sa.logger.info(f"Client at {sa.current_client_ip(sid)} is user {user.name} ({user.id}).")
+
+
 async def _create_client_side_session(
     sa: ServerApp, sid: str | None, user: User | None, session: dict | None = None
 ) -> dict:
+
+    # The sid should never be None here
+    assert sid is not None
```
ThanatosGit: With your code changes, I think the `sid` really can never be `None` here.
If I see it correctly, all code paths leading to this function really use `sid: str` which means you can just use `sid: str` as type in the function.

randovania/server/user_session.py
```diff
@@ -339,11 +346,13 @@ async def guest_login_post(sa: ServerAppDep, request: Request, name: typing.Anno
         discord_id=None,
     )
 
-    request.session["user_id"] = user.id
+    session = {}
+
+    session["user-id"] = user.id
+    await sa.sio.save_session(sid, session)
```
ThanatosGit: The more idomatic would be to use the contextmanager here.


```suggestion
   async with sa.sio.session(sid) as session:
       session["user-id"] = user.id
```

ThanatosGit (changes requested):

randovania/server/user_session.py
ImperatorPrime: Fixed in next commit.

randovania/server/user_session.py
ImperatorPrime: Fixed in next commit.

ThanatosGit (approved):
Alright, thanks for the PR

PR #9117, opened 2026-05-08
Logic updates 5-'26: SRX and NOC

JaggerTSG:
SRX:
- Yameba Corridor and first Stabilizer: backdoor spark first stabilizer and previous geron

NOC:
- XBOX Garage: kick off the stalactite to ascend
- Zozoro Cellar: No need to time freezing Zozoros mid-fall. Just the one on the floor.
- Zozoro Cellar: Tight double d-boost jump extend to cross from door to ledge.


randovania/games/fusion/logic_database/Sector 6 NOC.txt
StalledStorm, line 907: This "Have Any Jump Upgrade" can be changed to match lines 885-888, as it's the same requirement set being used to travel from lower to upper garage

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -117,7 +117,15 @@ Extra - minimap_coordinates: [{'x': 7, 'y': 2}, {'x': 8, 'y': 2}, {'x': 9, 'y':
...
-      Can Kill Missile Geron
+      Any of the following:
+          Can Kill Missile Geron
+          All of the following:
+              # Spark stored from Motos, restored in Stabilizer NE and each room after
+              Level 0 Keycard and Speed Booster and After Stabilizer Northwest Restored and After Stabilizer Northeast Restored and Knowledge (Intermediate) and Shinespark Tricks (Expert) and Disabled Door Lock Rando and Disabled Entrance Rando
+              All of the following:
+                  # Accounting for Hornoad Hole post stabilizers roomstate: https://youtu.be/aq0PQxbFJQk
+                  Combat (Advanced)
+                  Charge Beam or Plasma Beam or Wave Beam or Wide Beam
```
cluekitty, line 128: A couple of things here:
1. Is this multi-room shinespark trick logic worth maintaining?
2. If Yes to 1: Since the trick is accounting for everything below, could the comment with the video be placed at the top level of the trick? (also See comment below this list regarding the video)
3. If Yes to 1: Is this a good use of "Knowledge" Tricks? If so, is intermediate the appropriate tag here?
4. The trick logic for the beams here differs from the logic in Atmospheric Stabilizer Northwest room, however they reference the same trick video. Please adjust the logic for consistency, or preferably simplify it so that you are not duplicating logic in multiple rooms. If you need to defeat the NW Stablizer (Stabilizer 1), the logic for defeating this geron should probably only use the "After Event Northwest Stabilizer Restored" piece of logic. Having the different beam requirements here feels redundant.
5. The only way I see that this trick would be required is that the shinespark for defeating the NorthEast stabilizer is also required, because otherwise defeating this missile geron would be more trivial with the appropriate weaponry. This is an observation and not necessarily anything that would need to change.

For other experienced RDV DB maintainers, because I haven't encountered this before, is this style of trick video OK to include? It features several cuts showcasing the different iterations of beam requirements at the room transition where they are needed and includes text that kind of indicates which beam it is for. I feel the 'nickname' of the beams being included instead of the actual in-game name might call for either an edit to the video or potentially documenting each beam type as its own video as different comments. If individual videos for each beam-type is preferred, then the comments should go above each beam type instead of at the root of the trick.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -146,6 +154,16 @@ Extra - minimap_coordinates: [{'x': 10, 'y': 1}, {'x': 10, 'y': 2}, {'x': 11, 'y
...
+  > Event - Stabilizer 1 Online
+      All of the following:
+          # Spark stored from Motos, restored in Stabilizer NE and Hornoad Hole
+          Level 0 Keycard and Speed Booster and After Stabilizer Northeast Restored and Knowledge (Intermediate) and Shinespark Tricks (Expert) and Disabled Door Lock Rando and Disabled Entrance Rando
+          Any of the following:
+              # Any Beam but Ice: https://youtu.be/aq0PQxbFJQk
+              All of the following:
+                  Combat (Intermediate)
+                  Wave Beam or Wide Beam
+              Plasma Beam and Combat (Advanced)
```
cluekitty, line 166: I feel like this bit of logic would be better suited placed at Room Center -> Event instead of Door to Hornoad Hole -> Event. The logic to get to room center from the door is already trivial.

randovania/games/fusion/logic_database/Sector 6 NOC.txt
```diff
@@ -700,14 +700,12 @@ Extra - minimap_coordinates: [{'x': 5, 'y': 11}]
...
               # Get up to the ledge
               Have Any Jump Upgrade
-              All of the following:
-                  # Freeze Zozoro: https://youtu.be/wu8pKLvrAMw
-                  # Freeze without Diffusion
-                  Stand On Frozen Enemies (Intermediate) or Can Freeze Enemies With Ice Beam or Can Freeze Enemies With Ice Missiles
-                  # Freeze with Diffusion
-                  Stand On Frozen Enemies (Advanced) and Can Freeze Enemies With Diffusion
+              # Freeze Zozoro: https://youtu.be/wu8pKLvrAMw (can be on floor)
+              Stand On Frozen Enemies (Beginner) and Can Freeze Enemies With Any Weapon
```
cluekitty, line 713: Was there any reason this trick difficulty was reduced to Beginner? I would still consider this to be Intermediate at a minimum without diffusion. And for diffusion backwards jump from Advanced -> Beginner is quite the leap and we shouldn't assume that new players will get this with any consistency.

randovania/games/fusion/logic_database/Sector 6 NOC.txt
cluekitty: I could agree with this, but some discussion in discord suggests that the logic here might have multiple ways previously unaccounted for. I would prefer that such logic change would be in a separate PR if we modify this to break a 1x2 section and grab the ledge to let the owtch into the bottom half instead of having samus traverse into the bottom half.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
cluekitty: I asked for a bit of feedback regarding the editing style of the video and it sounds like it's considered okay. No issues including the video as-is.

randovania/games/fusion/logic_database/Sector 6 NOC.txt
JaggerTSG: Is it simpler if I add that specific trick in another PR, or remove this room's changes from this pr and put both tricks in a separate PR?

randovania/games/fusion/logic_database/Sector 1 SRX.txt
JaggerTSG: 1. Yeah, it really helps the junctions save room be a more viable starting location for non dlr/er.
2. Ah, yes, that's likely better.
3. That's generally used for destroying gerons using shinesparks/speed.
4. There's a complicated issue there, because one could theoretically have all stabilizers online, causing Hornoad Hole's enemy arrangement to change, necessitating dodging an overhead enemy at the end. The geron next door, of course, cannot be destroy before all stabilizers are online, so it cannot have that enemy to dodge before using the spark.
5. Charge destroys stabilizer gerons, in either geron setting, but never missile gerons.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
JaggerTSG: I almost did, but I'm not so sure that's a good idea, because you can reach room center from either door, but since you can't reach the right door until after the stabilizer's online, putting the starting node at the right door makes it require entering from the right. Honestly, it might actually not matter, that had just been my cautious thought process.

randovania/games/fusion/logic_database/Sector 6 NOC.txt
JaggerTSG: I meant you can just freeze an enemy that's on the floor. You don't need to catch any in mid-air. I can see this being unclear to readers, though, as the video makes it look very precise. Maybe the video should be replaced?

cluekitty: If all you need to do is stand on a frozen enemy on the floor, then I think the video should be replaced so that the video reflects the difficulty in the logic.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
cluekitty: That's a good point. I had not considered the trivial left-door connection as well so you can leave this directly from the right door.

randovania/games/fusion/logic_database/Sector 6 NOC.txt
cluekitty: I think any of the following would be OK:
* match this section to the other changes in this room, but no additional logic changes in this PR
* move all changes for this room into a separate PR
* leave this section alone and merge in the other room change, but come back later with a separate PR to address this feedback and add other potential logic. ideally this would be sooner than later, but this one I like least of the options.

JaggerTSG: Opted for option 2.

randovania/games/fusion/logic_database/Sector 6 NOC.txt
JaggerTSG: vid replaced

cluekitty (approved):
Noting in this approval that XBOX Garage changes will be in a follow-up PR, per Jagger.

JeffGainsNGames (approved):

PR #9119, opened 2026-05-08
Dread: Add trick to get Space Jump Room Missile Pickup with Normal Bomb

Tyranisaur:


duncathan (approved):
db maintainers can have a day or two to bikeshed about the difficulty if they want. I don't feel very strongly so long as it's out of trickless :+1:

duncathan (approved):
db maintainers can have a day or two to bikeshed about the difficulty if they want. I don't feel very strongly so long as it's out of trickless :+1:

PR #9126, opened 2026-05-10
Added short boosts and ADC tricks

preston-mouw:
Added Short Boost tricks in Dairon and Cataris.

Added Floor Clip tricks in Artaria, Cataris, and Hanubia. The less straightforward floor clips have videos attached in the comment field.


randovania/games/dread/logic_database/Ghavoran.txt
```diff
@@ -1476,6 +1476,8 @@ Extra - asset_id: collision_camera_021
   * Open Passage to Flipper Room/Dock to Spin Boost Tower
   > Door to Save Station Center
       Trivial
+  > Dock to Elun Transport Access
+      Floor Clip (Intermediate) and Can SSC and Destroy Enky
```
Tyranisaur: Seems more natural to have this require the concrete enky event rather than the template for destroying them.

randovania/games/dread/logic_database/Dairon.txt
```diff
@@ -167,6 +167,10 @@ Extra - asset_id: collision_camera_001
       Trivial
   > Wide Beam Block - Above
       After Dairon - Teleport To Artaria Speed Blocks Destroyed
+  > Event - Speed Booster blocks Destroyed
+      Any of the following:
+          Speed Booster Upgrade ≥ 2
+          Speed Booster Upgrade and Short Boost (Intermediate) and Can Short Boost
```
Tyranisaur: This connection is redundant, because we already have this from the pickup node. And having it from the pickup node was done because you can't really do this without collecting the pickup in the process. This is valuable correctness in cases where the pickup has a dangerous item. Although, these days we also have the option to write in requirements for node resources, which could achieve the same thing.

Also note that the way this is expressed, you missed a requirement for speed booster itself when using 2 upgrades.

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -3946,8 +3964,12 @@ Extra - asset_id: collision_camera_053
...
       Morph Ball and After Cataris - Underlava Speed Blocks Destroyed
   > Event - Speed Blocks Destroyed
-      # https://youtu.be/syNED-7J8HM
-      Gravity Suit and Morph Ball and Speed Booster and Speed Booster Upgrade and After Cataris - Lava Grapple Block and Speed Booster Conservation (Beginner)
+      All of the following:
+          # https://youtu.be/syNED-7J8HM
+          Gravity Suit and Morph Ball and Speed Booster and After Cataris - Lava Grapple Block and Speed Booster Conservation (Beginner)
+          Any of the following:
+              Speed Booster Upgrade
+              Short Boost (Intermediate) and Can Short Boost
```
Tyranisaur, line 3986: Could you move the video link comment to the speed booster upgrade path?

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -3460,7 +3473,12 @@ Extra - asset_id: collision_camera_046
   > Tunnel to Green EMMI Introduction (Upper)
       Trivial
   > Tunnel to EMMI Zone Tower East (Upper)
-      Morph Ball and After Cataris - Central Unit
+      All of the following:
+          After Cataris - Central Unit
+          Any of the following:
+              Morph Ball
+              # https://youtu.be/oA9qZrd7FV0
+              Floor Clip (Intermediate) and Knowledge (Ludicrous) and Speed Booster Conservation (Intermediate) and Can SSC
```
Tyranisaur: This shouldn't have knowledge on it, but it should have a slide requirement. Also, you could add an option where you require the misc resource, door lock rando,  disabled, because running through the doorway clearly should be an option.

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -3072,6 +3081,10 @@ Extra - asset_id: collision_camera_042
 > Dock to Lava Button West (Mid); Heals? False
   * Layers: default
   * Open Passage to Lava Button West/Dock to Thermal Device Room North (Mid)
+  > Dock to Lava Button West (Bottom)
+      All of the following:
+          Floor Clip (Intermediate) and Can SSC
+          Gravity Suit or After Cataris - Lower Lava Button
```
Tyranisaur: You can remove the gravity suit or lava button requirement here, because it's already covered on the requirements going out of the target dock, and this isn't a requirement to do the trick at hand.

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -3056,6 +3063,8 @@ Extra - asset_id: collision_camera_042
       Any of the following:
           Varia Suit or After Cataris - Lower Lava Button
           Heat/Cold Runs (Beginner) and Heat Damage ≥ 60
+  > Dock to Lava Button West (Mid)
+      Floor Clip (Intermediate) and Can SSC
```
Tyranisaur, line 3079: Are you sinking on the left or right part of this tiny room? Should this require that you pick up the energy part pickup?

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -2893,7 +2897,10 @@ Extra - asset_id: collision_camera_040
   > Door to Transport to Dairon
       Trivial
   > Dock to EMMI Zone Tower West
-      After Cataris - Dairon Transport Wide Beam Block
+      Any of the following:
+          After Cataris - Dairon Transport Wide Beam Block
+          # https://youtu.be/tmzFKxS-O4g
+          Floor Clip (Advanced) and Can SSC
```
Tyranisaur: This should have a Can Slide requirement on it.

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -1083,11 +1083,15 @@ Extra - asset_id: collision_camera_015
...
-              Heat/Cold Runs (Intermediate) and Heat Damage ≥ 30
+      Any of the following:
+          Varia Suit or Lay Power Bomb
+          Heat/Cold Runs (Intermediate) and Heat Damage ≥ 30
+          All of the following:
+              Floor Clip (Advanced) and Can SSC
+              Any of the following:
+                  Speed Booster Conservation (Beginner) and Disabled Door Lock Randomizer
+                  # https://youtu.be/gdLkSVBt1uc
+                  Screw Attack and Short Boost (Advanced) and Can Short Boost
```
Tyranisaur, line 1099: Changing the top level of this connection from `All of` to `Any of` is wrong. You're meant to wrap the power bomb alternative inside an `Any of`.

I think if you're using the power beam door in the Moving Magnets Walls room for your "Disabled Door Lock Randomizer" condition, then I think you need a morph ball requirement to carry the speed through to here. Note that you also could use the phantom cloak door, due to how that door is one sided. In that case you would need cloak and slide.

randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -5069,7 +5114,9 @@ Extra - asset_id: collision_camera_081
   > Total Recharge
       After Artaria - Screw Attack Rotatable
   > Start Point - Screw Attack
-      Screw Attack
+      Any of the following:
+          Screw Attack
+          Before Artaria - Screw Attack Rotatable and Floor Clip (Advanced) and Can SSC
```
Tyranisaur: In addition to how doing this should require the missile tank pickup, I'm going to ask that you add `Highly Dangerous Logic` on this. You also most likely should have this connection come from the door instead.

randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -3045,6 +3071,9 @@ Extra - asset_id: collision_camera_049
       Trivial
   > Dock to Central Unit
       After Artaria - Central Unit
+  > Dock to Navigation Station South
+      # https://youtu.be/639KYWvNooA
+      Floor Clip (Intermediate) and Can SSC
```
Tyranisaur, line 3081: If you want to do this in the way that is depicted, you should add "Disabled Door Lock Randomizer" otherwise, can it be done through charging speed from the central unit arena or with various ways of shorter boost distance?

randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -2159,8 +2180,13 @@ Extra - asset_id: collision_camera_022
...
           Any of the following:
               Simple IBJ
-              Speed Booster and Disabled Door Lock Randomizer and Open Charge Door
+              All of the following:
+                  Speed Booster
+                  Any of the following:
+                      Disabled Door Lock Randomizer and Open Charge Door
+                      Flash Shift and Speed Booster Upgrade and Short Boost (Intermediate)
               Climb Sloped Tunnels (Intermediate) and Use Spin Boost
+              Flash Shift and Speed Booster and Speed Booster Upgrade and Short Boost (Intermediate)
```
Tyranisaur: There doesn't seem to be any difference between the two new branches.

You could use the "Can Short Boost" template to make this less verbose. And you probably should have an option for using more upgrades to avoid needing to short boost.

randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -765,6 +771,8 @@ Extra - asset_id: collision_camera_008
   * Extra - right_shield_def: None
   > Door to EMMI Zone Exit North (Bottom)
       Trivial
+  > Door to White EMMI Introduction (Lower)
+      Floor Clip (Expert) and Can SSC
```
Tyranisaur, line 775: Does this only work before activating the thermal device? If so, can you use the door on the right side of the same corridor?

randovania/games/dread/logic_database/Artaria.txt
```diff
@@ -671,6 +671,8 @@ Extra - asset_id: collision_camera_006
       Trivial
   > Door to EMMI Zone First Entrance
       Speed Booster
+  > Dock to White EMMI Arena
+      Morph Ball and After Artaria - Thermal Device and After Artaria - First Chase Blob and Floor Clip (Expert) and Speed Booster Conservation (Advanced) and Can SSC
```
Tyranisaur, line 675: If this is what I think it is then I can't tell what you need the thermal device for.

Tyranisaur:
Missing changelog

Tyranisaur (changes requested):

randovania/games/dread/logic_database/Cataris.txt
preston-mouw: You can sink on the left side of the room, near the camera transition, to avoid the item pickup.

randovania/games/dread/logic_database/Artaria.txt
preston-mouw: Understood, will update that.

randovania/games/dread/logic_database/Artaria.txt
preston-mouw: This can be done with the door still locked, as there is enough space with the door closed to charge a vanilla length speed booster. I have updated the video to more clearly show that.

randovania/games/dread/logic_database/Artaria.txt
preston-mouw: Yes, this works on either side of the room, and with the Thermal Door still closed. Should I change the connection?

randovania/games/dread/logic_database/Artaria.txt
preston-mouw: Thermal door needs to be open or the Grapple Beam will contact it early and Samus will then die as she is OOB without Grapple being active.

randovania/games/dread/logic_database/Artaria.txt
preston-mouw: I'll add the template and the alternate extra speed booster upgrade logic.

randovania/games/dread/logic_database/Cataris.txt
preston-mouw: Will do.

randovania/games/dread/logic_database/Cataris.txt
preston-mouw: Understood, will do.

randovania/games/dread/logic_database/Cataris.txt
preston-mouw: I thought about adding the knowledge requirement because the slide has to be buffered. If the player holds left without timing the slide input correctly, Samus will mantle out of bounds and die. I'll remove the knowledge requirement on it, I can instead bump the Floor Clip level to expert or advanced.

I will add the alternate doorway logic.

randovania/games/dread/logic_database/Cataris.txt
preston-mouw: Will do.

randovania/games/dread/logic_database/Cataris.txt
preston-mouw: I'll fix the Any of to All of.

I did test, it is possible to start a speed booster from the Magnet Elevator room and carry the spark without morph and use it to clip. I'll create a few alternative logic branches with differing levels of Speed Booster Conservation with morph or phantom cloak.

randovania/games/dread/logic_database/Ghavoran.txt
preston-mouw: Will do.

randovania/games/dread/logic_database/Dairon.txt
preston-mouw: Didn't see the connection from the pickup node. I will remove this.

randovania/games/dread/logic_database/Artaria.txt
Tyranisaur: Is there enough space when there's a door lock cover, which builds a bit of distance?

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -1084,10 +1084,23 @@ Extra - asset_id: collision_camera_015
           Heat/Cold Runs (Intermediate) and Heat Damage ≥ 110
   > Right Room
       All of the following:
-          Lay Power Bomb
+          Varia Suit and Heat/Cold Runs (Intermediate) and Heat Damage ≥ 30
```
Tyranisaur: There's not supposed to be an `and` between Varia Suit and this heat run. As an aside, the heat run should probably require more health in the path where you short boost inside the room., but that's kind of hypothetical since heat runs can't be used.

CHANGELOG.md
```diff
@@ -16,6 +16,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Changed: Thermal Device Room North: Connection between Tunnel to Lava Button West and Dock to Lava Button West (Mid) now has Floor Clip logic.
+- Changed: Thermal Device Room North: Connection between Dock to Lava Button West (Mid) and Dock to Lava Button West (Bottom) now has Floor Clip logic.
+- Changed: Teleport to Artaria (Blue): Connection between Center Room and Right Room now has Floor Clip logic with varying levels of trick and item requirements to get the speed booster store for the clip.
+- Changed: Underlava Puzzle Room 2: Connection between Below Lava and Event - Speed Blocks Destroyed now has Short Boost logic.
+
+##### Ghavoran
+
+- Added: Dock between Spin Boost Tower and Elun Transport Access.
+- Added: Spin Boost Tower: New connection between Dock to Flipper Room and Dock to Elun Transport Access which uses a Floor Clip.
+- Added: Elun Transport Access: New connection between Dock from Spin Boost Tower to Center Platform (trivial).
```
Tyranisaur: You don't need to list out all the new nodes and connections. Just mentioning the effective new logic should be enough.

Tyranisaur (changes requested):
So the highly dangerous logic to obtain the pickup in artaria flipper has a requirement that door rando is enabled. This is kind of redundant, but was meant as a way to stop making flipping the spinner be dangerous in cases where it isn't even possible not to be exposed to that logic. However, with sinking into there, you're now making it possible to see that event as dangerous. So you could remove that condition. However, I'm not entirerly sure if it matter, since if you don't randomize the doors, then the logic to sink into there seems like it doesn't achieve anything you couldn't already do, given the items you need to do that sink.

randovania/games/dread/logic_database/Artaria.txt
preston-mouw: I just tested in a seed where that door was randomized to be a plasma cover. Assuming all the covers are the same size, you can do this regardless of the door type.

CHANGELOG.md
preston-mouw: Hopefully this upcoming change is more readable.

randovania/games/dread/logic_database/Cataris.txt
preston-mouw: Fixed, and added new logic to future proof this clip in case of heat/cold runs in the future

preston-mouw: Fixed.

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -1083,11 +1083,27 @@ Extra - asset_id: collision_camera_015
...
+          All of the following:
+              Lay Power Bomb
+              Any of the following:
+                  Varia Suit
+                  Heat/Cold Runs (Intermediate) and Heat Damage ≥ 30
+          All of the following:
+              Floor Clip (Advanced) and Can SSC
+              Any of the following:
+                  Varia Suit
+                  Heat/Cold Runs (Ludicrous) and Heat Damage ≥ 60
```
Tyranisaur: There's no way this heat run is ludicrous.

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -3460,7 +3483,14 @@ Extra - asset_id: collision_camera_046
...
   > Tunnel to EMMI Zone Tower East (Upper)
-      Morph Ball and After Cataris - Central Unit
+      All of the following:
+          After Cataris - Central Unit
+          Any of the following:
+              Morph Ball
+              All of the following:
+                  # https://youtu.be/oA9qZrd7FV0
+                  Floor Clip (Expert) and Can SSC
+                  Speed Booster Conservation (Intermediate) or Disabled Door Lock Randomizer
```
Tyranisaur, line 3493: This currently doesn't do anything, because to get anywhere from this node, after the trick,, all paths require morph ball and the cataris central unit. The tunnel to the other room has a type that also has those requirements on it. To fix that I think you would need to go edit the header.json file manually to add a cataris emmi slide tunnel, load the data editor and save. And then make the tunnel be that new tunnel type on this side and keep the old type on the other side. And you also want to adjust the requirement on the connection from the runnel to the door below, so you can do it with either a slide or a ball.

Tyranisaur (changes requested):

randovania/games/dread/logic_database/Cataris.txt
preston-mouw: I was considering the logic here to consider the player needing to do a short boost in the middle of the heat run. That seemed quite on the high end of heat/cold runs for me. What would recommend this trick level to be?

duncathan: Have you reviewed our documentation? We do have some guidelines for trick difficulty, including an explanation of why Ludicrous is a special case: https://github.com/randovania/randovania/blob/main/docs/Database%20Format.md#trick-difficulty

Tyranisaur: Maybe advanced to match it up with the difficulty rating of floor clip trick. Considering that this isn't a particularly deep heat run it  seems like the load bearing factor for it should be floor clip trick rather than the heat run itself. This is also a bit of an academic decision, since heat runs can't be used in dread for the any foreseeable future, and all of them probably need to be revisited if it becomes relevant.

randovania/games/dread/logic_database/Cataris.txt
preston-mouw: Fixed.

randovania/games/dread/logic_database/header.txt
```diff
@@ -790,6 +790,14 @@ Dock Weaknesses
...
+      Open:
+          All of the following:
+              After Cataris - Central Unit
+              Morph Ball or Can Slide
+      No lock
+
+
+  * Cataris EMMI Slide Tunnel
       Open:
           Morph Ball and After Cataris - Central Unit
```
Tyranisaur, line 802: First off, this looks backwards. The slide tunnel should require Can Slide instead of Morph, and the other one should require Morph. Second, the "Can Slide" template already is "Slide or Morph".

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -3482,9 +3512,11 @@ Extra - asset_id: collision_camera_046
 
 > Tunnel to EMMI Zone Tower East (Upper); Heals? False
   * Layers: default
-  * Cataris EMMI Tunnel to EMMI Zone Tower East/Tunnel to EMMI Zone East Tower Access (Upper)
+  * Cataris EMMI Slide Tunnel to EMMI Zone Tower East/Tunnel to EMMI Zone East Tower Access (Upper)
   > Door to EMMI Zone Tower East (Middle)
-      Morph Ball and After Cataris - Central Unit
+      All of the following:
+          After Cataris - Central Unit
+          Morph Ball or Can Slide
```
Tyranisaur, line 3519: I usually don't repeatedly point out the same thing multiple times, but the "Can Slide" template already is shorthand for "Slide or Morph", so it's redundant to phrase it like this.

I started this review adding comments from the bottom, and I don't know which order these will appear in.

Tyranisaur (changes requested):

randovania/games/dread/logic_database/Cataris.txt
```diff
@@ -1083,11 +1083,27 @@ Extra - asset_id: collision_camera_015
...
+              Any of the following:
+                  # https://youtu.be/gdLkSVBt1uc
+                  Screw Attack and Short Boost (Advanced) and Can Short Boost
+                  All of the following:
+                      Disabled Door Lock Randomizer
+                      Any of the following:
+                          Speed Booster Conservation (Intermediate)
+                          All of the following:
+                              Speed Booster Conservation (Beginner)
+                              Phantom Cloak or Morph Ball or Can Slide
```
XenoWars, line 1106: As Tyranisaur said in other parts, Morph Ball is already included in the "Can Slide" template.

XenoWars (changes requested):

PR #9129, opened 2026-05-11
bump mars to 0.14.1

JeffGainsNGames:
comes with a major change to the schema casing, require pdf, exporter, database, etc changes
tested various exports in-game and with as many combinations of settings as I could to ensure things were working the same as before


henriquegemignani:
Assuming it's purely internal stuff and no changelog needed?

JeffGainsNGames:
that is correct. no in-game changes, only internal

henriquegemignani (approved):

PR #9139, opened 2026-05-13
fix damage reductions multipliers above 1.0

JeffGainsNGames:
fixes an issue where multipliers above 1.0 were being incorrectly handled, wasn't sure how to handle changelog heading lmk if this is right
related:
https://discordapp.com/channels/914291389293027329/914294272579227668/1504150059070132455

also fixes a DR multiplier for fusion, not in stable yet so no changelog
related:
https://discordapp.com/channels/914291389293027329/1503997036536533082

### DIFF
--- a/randovania/games/fusion/generator/bootstrap.py
+++ b/randovania/games/fusion/generator/bootstrap.py
@@ -89,8 +89,9 @@ class FusionBootstrap(Bootstrap[FusionConfiguration]):
         default_cold_dmg = 15
         damage_reductions[db.get_damage("LavaDamage")] = [
             DamageReduction(None, 0, configuration.lava_damage / default_lava_dmg),
+            # FIXME 1.1x is required to offset the base_reduction of 0.9x until base_reduction can be ignored
             DamageReduction(
-                db.get_item_by_display_name("Varia Suit"), 1, 0.4 * (configuration.lava_damage / default_lava_dmg)
+                db.get_item_by_display_name("Varia Suit"), 1, (configuration.lava_damage / default_lava_dmg) * 1.1
             ),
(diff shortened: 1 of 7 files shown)


henriquegemignani (changes requested):
Add a new rdvgame to use a test case.

randovania/games/fusion/generator/bootstrap.py
```diff
@@ -89,8 +89,9 @@ def patch_resource_database(self, db: ResourceDatabase, configuration: FusionCon
         default_cold_dmg = 15
         damage_reductions[db.get_damage("LavaDamage")] = [
             DamageReduction(None, 0, configuration.lava_damage / default_lava_dmg),
+            # FIXME 1.1x is required to offset the base_reduction of 0.9x until base_reduction can be ignored
             DamageReduction(
-                db.get_item_by_display_name("Varia Suit"), 1, 0.4 * (configuration.lava_damage / default_lava_dmg)
+                db.get_item_by_display_name("Varia Suit"), 1, (configuration.lava_damage / default_lava_dmg) * 1.1
```
duncathan: why not `/ 0.9` to be more precise?

JeffGainsNGames: that would be more accurate, makes the test asserts a bit more ugly but I guess that's fine
assert 2.2222222222222223 == 2.2
assert 0.5555555555555556 == 0.55

duncathan: you could also use `2.0 / 0.9` as the value, if you want

henriquegemignani (approved):

PR #9140, opened 2026-05-14
Dread Cold Room Intro Logic

preston-mouw:
The Cold introduction room in Dread is a room I would like to include in logic without Gravity. Since Heat/Cold runs do not work in logic, I would like to offer that this room be given logic with a movement trick. At beginner with extra energy, or intermediate with no energy. There is no danger to softlocks thanks to respawning enemies on the left, and an energy refill on the right.

### DIFF
--- a/randovania/games/dread/logic_database/Artaria.json
+++ b/randovania/games/dread/logic_database/Artaria.json
@@ -21183,6 +21183,58 @@
                                                 }
                                             ]
                                         }
+                                    },
+                                    {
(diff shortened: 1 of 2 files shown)


duncathan (changes requested):
while I'm not a database maintainer, as a lead for dread rando I want to make it clear that I'm firmly against any damage runs being in logic. either the PR author or the DB maintainers will need to convince mii otherwise if they want approval for this PR

PR #9145, opened 2026-05-15
Dread: Fix leaving Ghav EMMI Zone without killing Ice EMMI

Tyranisaur:


Logarhythmic (approved):
I see, requiring that pickup event so the doors open back up to leave the emmi zone. I don't see any issues here.

ThanatosGit (approved):
approved by maintainer

PR #9149, opened 2026-05-17
Dread: Improve resolver safety of Ghavoran Chozo-X

Tyranisaur:
Make sure the fight requirements are only paid once.


Logarhythmic (approved):
This fight is always a surprise to me because I forget the trigger is in the spin boost room. ~~Seems OK to me.~~ Wait, does this mean you can't get the item and leave without having dangerous logic or a way to fight the chozo? Would this prevent just dropping in to get the item and leaving via green teleporter without the need to fight it? Because the only way to leave this room now is through the chozo.

image

Tyranisaur:
> This fight is always a surprise to me because I forget the trigger is in the spin boost room. ~Seems OK to me.~ Wait, does this mean you can't get the item and leave without having dangerous logic or a way to fight the chozo? Would this prevent just dropping in to get the item and leaving via green teleporter without the need to fight it? Because the only way to leave this room now is through the chozo.
> 

I just want to state up front that it has been like this for a long time. The way this works, is that the option to get that pickup and the leave without having a kit to do the fight is cordoned off to the "highly dangerous logic" option. Otherwise you are logically required to be able to finish the fight before you leave this room. That means the pickup in there could be part of how you're meant to satisfy the fight requirements. The last, and new, option is a safety mechanism that tells the logic that going back in through here multiple times is allowed without potentially needing to satisfy the fight requirements multiple times in a row.

Technically, dropping into the spin boost room is dangerous in this logic, because in certain situations it's a PONR. While the original logic, the one that applies with high dangerous logic enabled, is that it's a dangerous action because it cuts off the possibility to cross through the chozo x arena before satisfying the fight requirements, and it's much harder to reason about when it's correct to do that.

An alternative would be to logically always have the fight trigger on first pass unconditionally, at least without highly dangers logic, or maybe even patch the game to behave that way. Or maybe having a knowledge trick to enable going through the arena before the trigger.

Logarhythmic (approved):
I think an option to have the fight always active might make the most sense. It is what new players would expect and gets rid of the confusion of "why didn't this fight activate earlier?" But I know that that's beyond a logic PR. This is clearly better than what it was before.

henriquegemignani (approved):

PR #9151, opened 2026-05-18
Make mypy mandatory for the whole games folder

henriquegemignani:


henriquegemignani:
```
    # _ConnectionBase is the common base class of Connection and PipeConnection
    # and can be used in cross-platform code.
```
Guess that's what we use

randovania/games/common/prime_family/gui/export_validator.py
```diff
@@ -40,6 +40,8 @@ def is_prime1_iso_validator(file: Path | None, *, iso_required: bool = False) ->
     if is_file_validator(file):
         return True
 
+    assert file is not None  # comes from is_file_validator
```
duncathan, line 43: I don't love this, but it's awkward to try and make `is_file_validator` a typeguard so I guess it's whatever

duncathan (approved):

PR #9162, opened 2026-05-19
Prime1: Transport to Magmoor South climb rework

JeffGainsNGames:


Samuel6710 (commented):
If you want to go the extra mile and model the L-Jump from the pole to the spider track standable (since with the logic, it currently makes it look like the only way onto it is by doing the CBJ), it could improve readability.

JeffGainsNGames:
image
updated the comments to improve readability, I don't think it needs to be modelled any differently

Samuel6710 (approved):

toasterparty (approved):

PR #9180, opened 2026-05-25
Dark Stairwell <-> Central Nexus Connection Fixes

StalledStorm:
Central Nexus Connection from "Door to Station Entrance" to "Other to Dark Stairwell" was not requiring Morph Ball.

Dark Stairwell Connection from "Other to Central Nexus" to "Door to Nexus Navigation Room" was not requiring the ability to Mid-Air Morph or Bounce in Ball.

Dark Stairwell Connection from "Door to Nexus Navigation Room" to "Other to Central Nexus" was not requiring a Jump Upgrade, and was not allowing for Mid-Air Morph. 
While a shinespark trick is possible here to negate the Jump Upgrade necessity, it would require coming from one of the other rooms. This would mean disabling door lock/entrance rando, which would remove the requirement for the shinespark, since you would then be able to just go through Central Nexus Navigation Room.


randovania/games/fusion/logic_database/Main Deck.txt
```diff
@@ -694,7 +698,9 @@ Extra - minimap_coordinates: [{'x': 11, 'y': 8}, {'x': 11, 'y': 9}, {'x': 11, 'y
   * Tunnel to Central Nexus/Other to Dark Stairwell
   * Extra - door_idx: (202,)
   > Door to Nexus Navigation Room
-      Morph Ball and Screw Attack and Knowledge (Beginner)
+      All of the following:
```
cluekitty, line 701: I think this should be "Any of the following", aka an "OR" statement. Otherwise this path will always require Advanced logic.

randovania/games/fusion/logic_database/Main Deck.txt
```diff
@@ -666,7 +666,11 @@ Extra - minimap_coordinates: [{'x': 11, 'y': 8}, {'x': 11, 'y': 9}, {'x': 11, 'y
   > Other to Cubby Hole
       Morph Ball
   > Other to Central Nexus
-      Screw Attack and Knowledge (Beginner) and Can Bounce in Ball
+      All of the following:
```
cluekitty, line 669: I think this should be "Any of the following", aka an "OR" statement. Otherwise this path will always require Advanced logic.

cluekitty (changes requested):

randovania/games/fusion/logic_database/Main Deck.txt
cluekitty: I read this wrong, the logic here is correct as-is.

randovania/games/fusion/logic_database/Main Deck.txt
cluekitty: I read this wrong, the logic here is correct as-is.

cluekitty (approved):

ThanatosGit (approved):
approved by db maintainer

PR #9181, opened 2026-05-25
Fix typos in Knowledge description and Twin Caverns Save Room name

StalledStorm:


randovania/layout/preset_migration.py
```diff
@@ -1271,6 +1271,24 @@ def _migrate_v117(preset: dict, game: RandovaniaGame, *, from_layout_description
         preset["configuration"]["skip_opening"] = False
 
 
+def _migrate_v118(preset: dict, game: RandovaniaGame, *, from_layout_description: bool) -> None:
+    if game == RandovaniaGame.FUSION:
+        rename = {
+            "Sector 6 (NOC)/Twin Caverns East/Door to Twin Cavern Save Room": "Sector 6 (NOC)/Twin Caverns East/Door to Twin Caverns Save Room"
+        }
```
cluekitty: Should this information live and be referenced from `randovania/games/fusion/assets/migration_data.json`? Some migrations do not seem to specify the data directly in the code as you have here, but I do see that smaller changes don't seem to be in this migration data file.

Or, if it's OK for data to live outside of the migration data file then your indentation here means that the `rename` dict will be scoped to the `if` statement above and won't be available for anything below it. Either the other code below this `if` needs to be contained within the `if`, or `rename` should be defined on the same level as the `def fix()` function.

It might be better to write it like below no matter which option is acceptable.
```
def _migrate_v118(preset: dict, game: RandovaniaGame, *, from_layout_description: bool) -> None:
    if game != RandovaniaGame.FUSION:
        return

    rename = ... # Whichever option works best for including migration data

    def fix(identifier: dict) -> None:
        ...
```

Miepee: Generally, we only put data into the migration json if 1) they affect multiple games and 2) the data is "big enough" where it warrants not being "inline" so to speak.
For a few small lines for only one game its fine to just have them directly in code.

StalledStorm: Implemented in newest commit

duncathan:
For the record: I strongly dislike area renames, and the mess they leave in our migrations. I don't think fixing a single typo is worth this much churn. If other core disagree that's fine but I wanted to voice my disapproval on the PR itself

Miepee:
I mostly agree with you. 
I do like seeing typos fixed, but it is definitely annoying how much effort those require.

My main sympathy with this PR comes from the fact that it has been carried through now for multiple months. But we do probably want to have something like Henrique suggested, where we have display names for areas/regions, so that we can easily rename those without the huge messes. It would help with the renames that May had planned for Dread too.

randovania/games/fusion/logic_database/header.json
```diff
@@ -719,7 +719,7 @@
             },
             "Knowledge": {
                 "long_name": "Knowledge",
-                "description": "The player may be expected to know alternate enemy/block weaknesses or be able to identify potential softlocks risks.",
+                "description": "The player may be expected to know alternate enemy/block weaknesses or be able to identify potential softlock risks.",
```
Miepee, line 722: This would have been perfect as its own PR so that you dont block a single typo on all the other complicated stuff :)

StalledStorm: Done now in #9434 :)

StalledStorm:
Knowledge typo corrected in #9434, other typo to be addressed down the line with a better renaming system

PR #9185, opened 2026-05-26
Added Zipline Activation item and logic for Kraid's attic

steven11sjf:
Added an item for Zipline Activation and replaced uses of the event with it. Added logic for all remaining rooms in the northeast of Kraid.


randovania/games/zero_mission/logic_database/Kraid.txt
```diff
@@ -232,22 +232,24 @@ Extra - room_id: [5]
   * Blue Door to Zeela-Infested Room/Door to Acid Worm Arena
   * Extra - door_idx: (19,)
   > Door to Acid Worm Arena Access
-      Trivial
+      No Zipline Activation or After Boss Acid Worm Defeated
```
JeffGainsNGames: we should likely make acid worm always spawn in order to remove this before statement

randovania/games/zero_mission/logic_database/header.txt
```diff
@@ -70,6 +70,14 @@ Templates
           Can Use Any Bombs
           Screw Attack and Knowledge (Beginner)
 
+* Defeat Acid Worm:
+      All of the following:
+          Zipline Activation
+          # Health Requirements
+          Energy Tank ≥ 2 or Varia Suit
+          # Damage Requirements
+          Missiles ≥ 45 or Super Missiles ≥ 9
```
JeffGainsNGames: this does not need to be a template since it only has one use, it's also easier for majority of users to read quickly rather than going to the templates tab

randovania/games/zero_mission/logic_database/Kraid.txt
```diff
@@ -169,14 +169,16 @@ Extra - room_id: [3]
...
+          Morph Ball
+          Zipline Activation or Climb High Ledges
 
 > Door to Apex Launcher Hub; Heals? False
   * Layers: default
   * Blue Door to Apex Launcher Hub/Door to Upper Corridor
   * Extra - door_idx: (16,)
   > Door to Zeela-Infested Room
-      Power Grip
+      Morph Ball and Power Grip
```
JeffGainsNGames: tunnel climb template

randovania/games/zero_mission/logic_database/Kraid.txt
JeffGainsNGames, line 227: Image

whole room should likely be modeled like this
T = trivial, A = after acid worm, R = fight requirements

randovania/games/zero_mission/logic_database/Kraid.txt
```diff
@@ -290,10 +292,10 @@ Extra - room_id: [6]
 > Ball Launcher; Heals? False
   * Layers: default
   > Door to Upper Corridor
-      Space Jump (Unknown Item 2) or Can Use Bombs
+      Space Jump (Unknown Item 2) or Can Use Ball Launcher
   > Door to Zipline Activation Chamber
       Any of the following:
-          Can Use Bombs
+          Can Use Ball Launcher
           Morph Ball and Mid-Air Morphs (Beginner) and Use Space Jump
```
JeffGainsNGames: can be trickless using `tunnel climb` instead of mid-air morph

randovania/games/zero_mission/logic_database/Kraid.txt
```diff
@@ -362,7 +364,7 @@ Extra - room_id: [8]
   > Pickup (Missile Tank)
       Gravity Suit and Morph Ball and Speed Booster
```
JeffGainsNGames, line 381: I think this needs `can ballspark`?

randovania/games/zero_mission/pickup_database/pickup-database.json
```diff
@@ -497,6 +497,28 @@
...
+        "Zipline Activation": {
+            "gui_category": "movement",
+            "hint_features": [
+                "major",
+                "movement"
+            ],
+            "model_name": "ZIPLINES",
+            "offworld_models": {},
+            "preferred_location_category": "major",
+            "probability_multiplier": 0.5,
```
JeffGainsNGames: can remove probability modifier line

randovania/games/zero_mission/logic_database/header.json
```diff
@@ -277,6 +277,17 @@
...
                     "MessageID": 55
                 }
+            },
+            "Ziplines": {
+                "long_name": "Zipline Activation",
+                "max_capacity": 1,
+                "extra": {
+                    "item": "ZIPLINES",
+                    "StartingItemCategory": "Abilities",
+                    "StartingItemName": "ZIPLINES",
```
JeffGainsNGames: StartingItemCategory will be need to not be `Abilities` per the [patcher schema](https://github.com/MetroidAdvRandomizerSystem/mars-patcher-py/blob/main/src/mars_patcher/zm/data/schema.json#L136-L202). the [PDF](https://github.com/randovania/randovania/blob/main/randovania/games/zero_mission/exporter/patch_data_factory.py#L46-L76)/configuration/preset side will need a rework for this as a starting item
I can grab this later unless you are interested in doing that process (but too irl tired to explain further here/now)

JeffGainsNGames (changes requested):

randovania/games/zero_mission/logic_database/header.txt
steven11sjf: I did this to make it so either door can start the boss, since the player shouldn't be able to cross the room as it will lock them in.

randovania/games/zero_mission/logic_database/Kraid.txt
steven11sjf: Out of scope, the only reason I touched this room here was to replace the event with the item.

randovania/games/zero_mission/logic_database/Kraid.txt
steven11sjf: I think you're saying to drop down from above by using tunnel climb? This is from the ball launcher to there. I would also put it under an intermediate trick regardless since you have to shoot out the block and fail = reset.

JeffGainsNGames: Image

```
any of 
  can use ball launcher
  all of
     use spacejump
     any of
         tunnel climb
         mid-air morph and morph
``` 
was my thinking, so that this path can be done trickless with SJ but without using the launcher (bombs)

randovania/games/zero_mission/logic_database/Kraid.txt
```diff
@@ -231,38 +231,48 @@ Extra - room_id: [5]
...
+      All of the following:
+          After Boss Acid Worm Defeated and Can Use Any Bombs
+          Any of the following:
+              Tunnel Climb
+              Movement (Beginner) and Can Bounce in Ball
+  > Event - Acid Worm Defeated
+      All of the following:
+          Zipline Activation
+          # Health requirements
+          Energy Tank ≥ 2 or Varia Suit
```
JeffGainsNGames: `x etanks` should be replaced by `X normal damage`, since if we have custom etank increments a player could be expected to do this with 2 x 10hp etanks instead of taking 299 normal dmg (2x100hp + 99)
all bosses will need a secondary pass anyways we can capture this in that pass as well

randovania/games/zero_mission/logic_database/Kraid.txt
```diff
@@ -231,38 +231,48 @@ Extra - room_id: [5]
...
   * Extra - door_idx: (91,)
-  > Door to Zeela-Infested Room
+  > Acid Worm Arena
       After Boss Acid Worm Defeated and Can Use Any Bombs
 
 > Event - Acid Worm Defeated; Heals? False
   * Layers: default
   * Event Boss Acid Worm Defeated
+  > Acid Worm Arena
+      After Boss Acid Worm Defeated
```
JeffGainsNGames: this can just be  left as trivial

randovania/games/zero_mission/logic_database/header.json
JeffGainsNGames: if you want to just set this to `StartingItemCategory` to Ziplines we can handle the rest outside of this logic PR

JeffGainsNGames (approved):

PR #9191, opened 2026-05-27
Prime 1 MW: Don't read invalid CPlayer at title screen

Miepee:
The `offset=0` causes nintendont to treat the given address as a pointer, dereference it, and read the 4 bytes at the new address. There are a few situations where CPlayer can be null, primarily the title screen and elevators, so in order to not force a disconnect there, let's be safe and manually throw another request to read the vtable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,8 +58,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Prime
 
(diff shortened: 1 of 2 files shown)


duncathan:
what are the odds we can get this in and tested in time for CGC?

henriquegemignani (approved):
@toasterparty 's call to merge or not

toasterparty (approved):

PR #9195, opened 2026-05-28
Also read tracker layouts from local appdata

henriquegemignani:
Absolutely no GUI or public documentation, only works for these that really want it. Realised I could implement this as a very low effort PR.


ThanatosGit (approved):
I leave it to you if you want to fix the typing.

PR #9199, opened 2026-05-29
refactor PrimeTrilogyTeleporterConfiguration

duncathan:
divides it into separate classes for prime and echoes, rather than checking a game enum


henriquegemignani (approved):
Let's merge after the release. No need to add this right now.

PR #9200, opened 2026-05-29
fix render regions

### LINKED ISSUE #8785: CLI: render regions command stopped working
First, it stopped properly recognizing which game to render and tries to interpret `development` as a game.
And if you hardcode echoes, then it'll crash trying to get colors for the now split regions (dark agon, ing hive etc)

duncathan:
fixes #8785

### DIFF
--- a/randovania/cli/commands/render_regions.py
+++ b/randovania/cli/commands/render_regions.py
@@ -4,6 +4,8 @@ import argparse
 from argparse import ArgumentParser
 from typing import TYPE_CHECKING
 
+from randovania.game.game_enum import RandovaniaGame
+
 if TYPE_CHECKING:
     from argparse import Namespace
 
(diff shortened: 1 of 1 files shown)


randovania/cli/commands/render_regions.py
```diff
@@ -175,6 +178,7 @@ def render_regions_graph(sub_parsers: argparse._SubParsersAction) -> None:
         help="Renders an image with all area connections",
         formatter_class=argparse.MetavarTypeHelpFormatter,
     )
+    parser.add_argument("game", type=str, choices=[game.value for game in RandovaniaGame.all_games()])
```
henriquegemignani, line 181: can't believe we don't have a helper for adding this argument.

henriquegemignani (approved):

PR #9202, opened 2026-05-29
refactor map tracker's handling of configurable nodes

duncathan:
removes references to the echoes game enum, and as a bonus makes configurable nodes work in the tracker for hunters!

image


randovania/resolver/bootstrap.py
```diff
@@ -369,3 +386,86 @@ def pre_place_pickups(
...
+        return ConfigurableNodeBootstrap[Configuration, Any]
+
+
+class ConfigurableNodeBootstrap[Configuration: BaseConfiguration, ConfigNodeData]:
+    @property
+    def category_name(self) -> str:
+        """
+        The name used to refer to this game's configurable nodes.
+        """
+        return "Configurable Nodes"
```
henriquegemignani: Maybe this should just be `raise NotImplementedError`?

henriquegemignani (changes requested):

randovania/resolver/bootstrap.py
duncathan: sure

henriquegemignani (approved):

PR #9211, opened 2026-05-30
move prime's dangerous gravity out of tricks tab and into generation tab

duncathan:
image

Part of my ongoing effort to remove game enum checks from shared code. I moved this setting to another tab for two reasons: one, the Generation tab already has the infrastructure set up for game-specific UI elements like this so it was much easier than making a custom trick level tab for prime. two, it's consistent with Dread's related highly dangerous logic setting

also removes a no-longer-necessary check for prime 1 in the pickup pool tab


toasterparty (approved):
Might clear up some common confusion about this setting you add some text like:

> Only has effect if *Gravityless Underwater Movement* tricks are enabled

duncathan:
> Might clear up some common confusion about this setting you add some text like:
> 
> > Only has effect if _Gravityless Underwater Movement_ tricks are enabled

sure, can do

duncathan:
Added a line to both the Generation tab and to the trick description.

image

image

JeffGainsNGames:
image

I think this line can be removed, since we do have cases where the `enabled dangerous gravity suit logic` doesn't impact `gravityless underwater movement`
image

the line in generation tab is good

Miepee:
With #8865 would it make sense to already rename the option/description to be more general about dangerous logic instead of just gravity stuff?

toasterparty (changes requested):
The core intent of this PR is made obsolete by:
- #9486

PR #9216, opened 2026-05-31
Fix pickle issue with 3.12+ generics

ThanatosGit:


henriquegemignani (commented):
This really needs to have a test case that would fail before.

henriquegemignani (approved):
It's a problem with pickling the LayoutDescription. Which is actually a bit scary that happens, so I tried avoiding that and it broke badly. Ugh.

duncathan:
oh boy. always fun to encounter a *python* bug

PR #9223, opened 2026-06-01
Dolphin Executor: Treat invalid address operation as fatal

Miepee:
Gives some parity with the Nintendont executor. Makes it easier to find bugs/wrong behaviour without having them run on nintendont first and should also guarantee that the result dict will always have as many entries as the operations that were requested

### DIFF
--- a/randovania/game_connection/executor/dolphin_executor.py
+++ b/randovania/game_connection/executor/dolphin_executor.py
@@ -123,13 +123,9 @@ class DolphinExecutor(MemoryOperationExecutor):
 
             try:
                 pointers[pointer] = self.dolphin.follow_pointers(pointer, [0])
-            except RuntimeError:
-                pointers[pointer] = None
+            except RuntimeError as e:
                 self.logger.debug(f"Failed to read a valid pointer from {pointer:x}")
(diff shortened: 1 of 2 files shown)


henriquegemignani (commented):
I'd also make `_memory_operation` raise here instead of returning None:
```python
            new_address = pointers[address]
            if new_address is None:
                return None
```

Miepee:
> I'd also make _memory_operation raise here

If I see things correctly, we don't actually need to raise at all.
Since the loop that populates the pointers dict now errors on invalid addresses, the pointers dict never gets None in the first place and becomes a dict[int, int]. And since we also check above whether the address is in the pointers dict, the `if new_address is None` would always be False.

henriquegemignani:
> > I'd also make _memory_operation raise here
> 
> If I see things correctly, we don't actually need to raise at all. Since the loop that populates the pointers dict now errors on invalid addresses, the pointers dict never gets None in the first place and becomes a dict[int, int]. And since we also check above whether the address is in the pointers dict, the `if new_address is None` would always be False.

I hallucinated a `.get` call there. Go me.

henriquegemignani (approved):

PR #9225, opened 2026-06-01
Prime 1: Don't read invalid Game State while booting up

### LINKED ISSUE #9224: Prime 1 Connector: game_state_pointer is invalid
When we're doing this memory operation:
`MemoryOperation(self.version.game_state_pointer, offset=mlvl_offset, read_byte_count=asset_id_size)`
it caused an error because apparently the game_state_pointer was invalid. Easy fix would be to also put that behind another check, but ideally it should investigated where/when that happens to be the case so that we can document it

Sentry Issue: [RANDOVANIA-CLIENT-32C](https://randovania.sentry.io/issues/7517855342/?referrer=github_integration)

```
ValueError: not enough values to unpack (expected 3, got 2)
  File "randovania\lib\infinite_timer.py", line 43, in _target_wrap
  File "randovania\game_connection\connector\prime_remote_connector.py", line 400, in update
  File "randovania\game_connection\connector\prime1_remote_connector.py", line 77, in current_game_status
```

Miepee:
Turns out the easiest way to trigger this is to be in-game already and then either soft reset or use quit the game from the menu.

Fixes #9224

### DIFF
--- a/randovania/game_connection/connector/prime1_remote_connector.py
+++ b/randovania/game_connection/connector/prime1_remote_connector.py
@@ -71,22 +71,20 @@ class Prime1RemoteConnector(PrimeRemoteConnector):
         mlvl_offset = 0x84
         cplayer_offset = 0x84C
 
-        pending_byte_op = MemoryOperation(cstate_manager_global + self._pending_op_offset, read_byte_count=1)
-        pending_byte_result = await self.executor.perform_single_memory_operation(pending_byte_op)
-
         # Both of these can be a nullpointer. The first one while the game is booting up, the second at
-        # title screen/elevators. In both cases we can just say that they're in an invalid World.
+        # title screen/elevators. In both cases we can just say that they're in an invalid World / can't be acted on.
         world_status_ops = [
             MemoryOperation(self.version.game_state_pointer, offset=mlvl_offset, read_byte_count=asset_id_size),
-            MemoryOperation(cstate_manager_global + cplayer_offset, read_byte_count=4),
+            MemoryOperation(cstate_manager_global + cplayer_offset, offset=0, read_byte_count=4),
         ]
         try:
             world_status_results = await self.executor.perform_memory_operations(world_status_ops)
             world_asset_id, cplayer_vtable = world_status_results.values()
         except MemoryOperationException:
-            world_asset_id = None
-            cplayer_vtable = None
+            return True, None
 
+        pending_byte_op = MemoryOperation(cstate_manager_global + self._pending_op_offset, read_byte_count=1)
+        pending_byte_result = await self.executor.perform_single_memory_operation(pending_byte_op)
         has_pending_op = pending_byte_result != b"\x00"
 
         current_world = self._current_status_world(world_asset_id, cplayer_vtable)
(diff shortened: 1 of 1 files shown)


randovania/game_connection/connector/prime1_remote_connector.py
```diff
@@ -66,22 +66,41 @@ async def current_game_status(self) -> tuple[bool, Region | None]:
         asset_id_size = struct.calcsize(self._asset_id_format())
         mlvl_offset = 0x84
         cplayer_offset = 0x84C
+        empty_pointer = b"\x00" * 4
 
-        memory_ops = [
-            MemoryOperation(self.version.game_state_pointer, offset=mlvl_offset, read_byte_count=asset_id_size),
+        # Game State can be nullpointer while game is booting up
+        # Cplayer can be nullpointer on title screen/during elevators
+        # So let's do a first reading pass just normally reading the values
```
Miepee: instead of doing two reading passes like this, we could also seperate them by offset and non-offset.
E.g.
```py
pending_byte_operation  = MemoryOperation(cstate_manager_global + self._pending_op_offset, read_byte_count=1)
pending_byte = await self.executor.perform_single_memory_operation(pending_byte_op)
offset_ops = [
    MemoryOperation(self.version.game_state_pointer, offset=mlvl_offset, read_byte_count=asset_id_size),
    MemoryOperation(cstate_manager_global + cplayer_offset, read_byte_count=4)
]
try:
    offset_results = await self.executor.perform_memory_operations(offset_ops)
    world_asset_id, cplayer_vtable = offset_results.values()
except MemoryOperationException:
    world_asset_id = cplayer_vtable = None
```

but I'm not sure whether that's cleaner / more pythonic.

randovania/game_connection/connector/prime1_remote_connector.py
```diff
@@ -67,21 +71,23 @@ async def current_game_status(self) -> tuple[bool, Region | None]:
...
-
-        has_pending_op = pending_op_byte != b"\x00"
-
-        cplayer_vtable = None
-        if cplayer_pointer != b"\x00" * 4:
-            cplayer_memory_op = MemoryOperation(int.from_bytes(cplayer_pointer, "big"), read_byte_count=4)
-            cplayer_vtable = await self.executor.perform_single_memory_operation(cplayer_memory_op)
+        try:
+            world_status_results = await self.executor.perform_memory_operations(world_status_ops)
+            world_asset_id, cplayer_vtable = world_status_results.values()
```
henriquegemignani, line 82: This returns `cplayer_pointer`, not cplayer_vtable

randovania/game_connection/connector/prime1_remote_connector.py
```diff
@@ -67,21 +71,23 @@ async def current_game_status(self) -> tuple[bool, Region | None]:
         mlvl_offset = 0x84
         cplayer_offset = 0x84C
 
-        memory_ops = [
+        pending_byte_op = MemoryOperation(cstate_manager_global + self._pending_op_offset, read_byte_count=1)
```
henriquegemignani: eh, if the other two requests are failing we don't really care for reading the pending byte op.

Does make me wonder if it's properly cleared when cstatemanager is recreated.

henriquegemignani (changes requested):

randovania/game_connection/connector/prime1_remote_connector.py
Miepee: Oh right. I thought the pending byte would be used for something, but forgot we can only do things with it when we know they're in a valid. region.

henriquegemignani: More precisely, we only do anything with the game when we're ingame. The only purpose of the CPlayer vtable check is to be sure we're ingame.

toasterparty (approved):

henriquegemignani (approved):

PR #9226, opened 2026-06-01
Add a test to ensure that Bootstrap is serializable

ThanatosGit:
I thought about adding a good test to prevent regressions, but this is the only thing I could come up with.

It fails for me with the PEP-695 syntax and the following error, so it should be sufficient as a safeguard.
```
================================== FAILURES ===================================
____________________________ test_pickle_roundtrip ____________________________

    def test_pickle_roundtrip():
        bootstrap_obj = bootstrap.Bootstrap()
>       serialised = pickle.dumps(bootstrap_obj)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       _pickle.PicklingError: Can't pickle Configuration: attribute lookup Configuration on typing failed
```

### DIFF
--- a/test/resolver/test_bootstrap.py
+++ b/test/resolver/test_bootstrap.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import pickle
 from random import Random
 
 import pytest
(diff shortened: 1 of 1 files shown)


henriquegemignani (approved):
Hm... this is fine. Maybe the test should be pickle loop a LayoutDescription for every game?

ThanatosGit:
> Maybe the test should be pickle loop a LayoutDescription for every game?

IIRC it did not trigger the error and don't want so search through discord again. I merge this tiny safeguard now. Happy to remove or replace it if we find something better.

PR #9233, opened 2026-06-03
Reworked read and write memory operations

Miepee:
Merge conflicting on purpose atm, because I dont think this will get merged before #9255

Tests fail on purpose atm because I'm not sure yet whether this abstraction is final. I do find it neat that we now explicitly allow only reading *or* writing, because it makes checking for these things slightly more convenient (now you dont have to check both and deal with potential None errors if you didnt). I was kinda hoping to simplify the weird add_op in the nintendont connector more but eh.


randovania/game_connection/executor/memory_operation.py
```diff
@@ -2,65 +2,126 @@
...
-            self.write_bytes is not None
-            and self.read_byte_count is not None
-            and len(self.write_bytes) != self.read_byte_count
-        ):
-            raise ValueError(f"Attempting to read {self.read_byte_count} bytes while writing {len(self.write_bytes)}.")
+        """Return the number of bytes involved in the operation."""
+        raise NotImplementedError
+
+    @staticmethod
+    def is_read_op(instance: MemoryOperation) -> TypeGuard[MemoryReadOperation]:
```
henriquegemignani, line 31: I'd prefer this to be a regular method of the class, so you just do `op.is_read()` but I know this doesn't match the TypeGuard API.

Which is why the TypeGuard API sucks >.>

randovania/game_connection/connector/echoes_remote_connector.py
```diff
@@ -73,9 +76,9 @@ async def current_game_status(self) -> tuple[bool, Region | None]:
         cplayer_offset = 0x14FC
 
         memory_ops = [
-            MemoryOperation(self.version.game_state_pointer, offset=mlvl_offset, read_byte_count=asset_id_size),
-            MemoryOperation(cstate_manager_global + self._pending_op_offset, read_byte_count=1),
-            MemoryOperation(cstate_manager_global + cplayer_offset, offset=0, read_byte_count=4),
+            MemoryReadOperation(self.version.game_state_pointer, offset=mlvl_offset, read_byte_count=asset_id_size),
```
henriquegemignani: I don't like the `read_` redundancy given it's a `MemoryReadOperation` already

randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -289,8 +293,8 @@ async def execute_remote_patches(self, patches: list[DolRemotePatch]) -> None:
         )
         memory_operations.extend(
             [
-                MemoryOperation(patch_address, write_bytes=patch_bytes),
-                MemoryOperation(self.version.cstate_manager_global + self._pending_op_offset, write_bytes=b"\x01"),
+                MemoryWriteOperation(patch_address, write_bytes=patch_bytes),
```
henriquegemignani: Less bad but double `write` is also ehh for me.

henriquegemignani (changes requested):

randovania/game_connection/executor/memory_operation.py
Miepee: Yeaaaah, but I couldnt find anything with similar capabilities. But i can search later again

randovania/game_connection/connector/echoes_remote_connector.py
Miepee: Whatd you prefer, just calling it `byte_count`  and `bytes`  for read/write respectively?

henriquegemignani: `byte_count` is decent for read.
`bytes` is bad for writing because it shadows the type. Could be `data` I guess

PR #9238, opened 2026-06-03
Nintendont: Use smaller requests for timeouts

Miepee:
Technically user facing but i dont think this needs a changelog?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [10.9.0] - 2026-07-0?
 
 - Waiting changes to be added.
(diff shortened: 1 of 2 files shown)


henriquegemignani (approved):
I would add a changelog entry still

duncathan:
yeah this is going to result in some users who previously believed they could connect to their Wii to suddenly be unable to. the changelog entry is important I think

Miepee:
Done. Would probably also be a good candidate for an eventual patch release.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [10.9.0] - 2026-07-0?
 
 - Waiting changes to be added.
+- Changed: Nintendont: The time on how long Randovania waits for requests from the Wii before it disconnects has been made consistent and reduced to 5 seconds.
```
henriquegemignani: `waits for the Wii`

PR #9246, opened 2026-06-05
Take 2: Split HintNode into multiple subclasses

duncathan:
Closes #8107 

I'm better equipped to explain the tangible benefits this time:
1. Improved UX for all types of hint nodes in the db viewer
2. The Specific Pickup nodes, while not actually used during hint placement or exporting, are a much cleaner way to ensure that the generator assigns a weight to unlocking the hints than we've previously had. Other games have either not included a hint node, meaning their specific pickup hints don't have a generation weight (such as in Echoes); or they've abused generic hint nodes and assigned dummy hints to them (such as Dread and Fusion)
3. This opens the door to future improvements. The Specific Pickup nodes are tied to a specific pickup hint mode (as defined in game data) so we could theoretically make a change in the generator so that hints only provide weight when they're enabled in the preset. This wasn't possible to do with the old workarounds

While I think at least a few games could benefit from using the new Specific Pickup hint nodes in their databases, that's been left out of this PR for future work, as it starts to get into generation design stuff


randovania/gui/dialog/node_details_popup.py
```diff
@@ -338,6 +368,13 @@ def on_dock_update_name_button(self) -> None:
         self.name_edit.setText(expected_name)
         self.on_name_edit(self.name_edit.text())
 
+    def on_hint_kind_combo(self, _) -> None:
+        self._update_hint_target_visibility(self.hint_kind_combo.currentData())
+
+    def on_specific_location_target_spin(self, value: int) -> None:
+        has_error = not integrity_check.pickup_index_exists(PickupIndex(value), self.game)
```
henriquegemignani, line 375: mypy

randovania/gui/dialog/node_details_popup.py
```diff
@@ -123,9 +129,17 @@ def __init__(self, game: GameDescription, node: Node):
...
+        # Hint
         for hint_node_kind in enum_lib.iterate_enum(HintNodeKind):
             self.hint_kind_combo.addItem(hint_node_kind.long_name, hint_node_kind)
 
+        specific_pickup_hints = game.game.hints.specific_pickup_hints
+        if specific_pickup_hints:
+            for specific_hint, details in specific_pickup_hints.items():
+                self.specific_pickup_target_combo.addItem(details.long_name, specific_hint)
+        else:
+            self.hint_kind_combo.removeItem(2)
```
henriquegemignani: let's make a helper method can remove an item based on the data. This 2 is too magical here

randovania/gui/dialog/node_details_popup.py
```diff
@@ -338,6 +368,13 @@ def on_dock_update_name_button(self) -> None:
         self.name_edit.setText(expected_name)
         self.on_name_edit(self.name_edit.text())
 
+    def on_hint_kind_combo(self, _) -> None:
```
henriquegemignani: also mypy

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #9248, opened 2026-06-05
Better pretty-print nested structures in extras, alternative way

henriquegemignani:
Different way of doing the same as #9247.


duncathan (commented):
is there a way put the first list/dict entry on a new line? I prefer this one but the lack of newline is a little unfortunate

henriquegemignani:
> is there a way put the first list/dict entry on a new line? I prefer this one but the lack of newline is a little unfortunate

Added an extra line break if the formatted dict/list has line breaks

duncathan (approved):
cool, I like this one. I'll leave it up for a bit in case another core reviewer disagrees

JeffGainsNGames (approved):

henriquegemignani:
Updated the PR to also format the nodes

duncathan (approved):

PR #9253, opened 2026-06-05
Updates for new OSRR patcher (Cosmetics, beam fix, skip intro)

ThanatosGit:
image
image

ToDo:
- [x] requires patcher update

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added: Cosmetic option to reveal the map at the start (default: on)
 - Added: Cosmetic option to use Fusion models (default: off)
 - Added: Preset option to skip the opening storyboard cutscene (default: off)
-- Fixed: Spazer and Plasma beam projectiles no longer shoot through blocks
+- Fixed: Spazer and Plasma beam projectiles no longer shoot through blocks without Wave
 
 
 ## [10.8.0] - 2026-06-01
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Metroid: Samus Returns
 
 - Added: Cosmetic option to reveal the map at the start (default: on)
-- Added: Cosmetic option to use Fusion models (default: off)
+- Added: Cosmetic option to use Fusion suits (default: off)
 - Added: Preset option to skip the opening storyboard cutscene (default: off)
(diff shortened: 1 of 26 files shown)


dyceron:
I feel a toggle defaulted to on is fine

dyceron (commented):
I think that intro skip should be a preset option since it technically "affects" the seed.

duncathan:
yeah. that would be an obvious advantage in a race context

ThanatosGit:
Well, it was unanswered here: https://github.com/randovania/open-samus-returns-rando/issues/20#issuecomment-4017034244

It still applies: When the race timer starts as soon as you gain control, then it doesn't matter.

dyceron:
I totally forgot about that comment. Races start from file select

dyceron:
I also didn't even think about it until I merged the patcher pr. Probably should be a game patch instead

ThanatosGit:
Meh, then I have to move it.

Loading times are really bad for races. Like emulator already gives you an advantages becasue it loads often much faster. I think for dread it was like more than 10 seconds just for the initial Artaria load from ryujinx compared to my switch after file select.

ThanatosGit:
Updated the PR.

randovania/games/samus_returns/gui/ui_files/preset_msr_patches.ui
dyceron, line 1: Could you try to remove that extra space at the end of the `Item Patches` group?

CHANGELOG.md
```diff
@@ -19,6 +19,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed: Dark Stairwell: Using the hidden connection to/from Central Nexus can now expect Mid-Air Morphs instead of Bombs/Hi-Jump.
 
+### Metroid: Samus Returns
+
+- Added: Cosmetic option to reveal the map at the start (default: on)
+- Added: Cosmetic option to use Fusion models (default: off)
+- Added: Preset option to skip the opening storyboard cutscene (default: off)
+- Fixed: Spazer and Plasma beam projectiles no longer shoot through blocks
```
dyceron: I'd include `without Wave`

dyceron (changes requested):

randovania/games/samus_returns/gui/ui_files/preset_msr_patches.ui
ThanatosGit: Oh, I guess it's there to have some extra space for scrolling, right?
Then I could move it to the new last element:
image

dyceron:
Probably should've said 'Use Fusion Suit` instead of models to be more clear as to what the option does. Can just change the cosmetic window and that'll be fine.

ThanatosGit:
image

dyceron (approved):

PR #9256, opened 2026-06-06
Add reject_undocumented_tricks_in_database to GameData

henriquegemignani:


randovania/game/data.py
```diff
@@ -68,6 +68,9 @@ class GameData:
     test_data: Callable[[], GameTestData]
     """Contains data and configuration parameters to allow the test suit to assert things properly for this game."""
 
+    reject_undocumented_tricks_in_database: bool = True
```
Miepee, line 71: instead of adding a new game data param, why not just call a function for all games that want it in `db_integrity`? that whole file's purpose is for game-specific tests

henriquegemignani: I want it to be opt-out :)

Miepee: ah. i would like to avoid stuffing one use stuff into game data but also cant think of anything quickly of the top of my head to avoid that so i guess its fine then.

duncathan: why do you say it's one-use?

Miepee: because its only used once in integrity check (i dont count "game configures the value" as a use in this case)

duncathan: "game configures the value" is the entire point of GameData, though?

henriquegemignani: `permalink_reference_preset` is only used in one place. defaults_available_in_game_sessions is almost that too. `racetime_url` is in one place. `hash_words` is also one place. Most of these fields are used in few places, but they're here because it's per-game fields.

henriquegemignani:
@Miepee @duncathan ping

JeffGainsNGames (approved):

PR #9263, opened 2026-06-08
Echoes: dont disconnect due to invalid world status

Miepee:
Same as the other PR but for Echoes

### DIFF
--- a/randovania/game_connection/connector/echoes_remote_connector.py
+++ b/randovania/game_connection/connector/echoes_remote_connector.py
@@ -7,7 +7,11 @@ from typing import TYPE_CHECKING, TypeGuard, override
 from open_prime_rando.dol_patching import all_prime_dol_patches
 
 from randovania.game_connection.connector.prime_remote_connector import PrimeRemoteConnector
-from randovania.game_connection.executor.memory_operation import MemoryOperation, MemoryOperationExecutor
+from randovania.game_connection.executor.memory_operation import (
+    MemoryOperation,
+    MemoryOperationException,
+    MemoryOperationExecutor,
+)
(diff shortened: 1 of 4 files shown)


randovania/game_connection/connector/echoes_remote_connector.py
```diff
@@ -72,16 +76,24 @@ async def current_game_status(self) -> tuple[bool, Region | None]:
         mlvl_offset = 4
         cplayer_offset = 0x14FC
 
-        memory_ops = [
+        # Both of these can be a nullpointer. The first one while the game is booting up, the second at
+        # title screen/elevators. In both cases we can just say that they're in an invalid World / can't be acted on.
```
henriquegemignani: Title screen is wrong for Echoes, because the front end is a world

henriquegemignani (changes requested):

randovania/game_connection/connector/echoes_remote_connector.py
Miepee: Thats strange. From looking at naii's log, it seems like title screen was part of failing too.
Or maybe it were the attract videos / title-screen-before-in-game-menu. Are those part of front_end too?

henriquegemignani: During the attract is likely already loaded (but it's not like we have attract anymore).

But right at the start is the "game booting" part. I personally consider everything before frontend loaded as game booting.

Miepee: double checked; you can sit on the health and warning safety screen for quite a while, so maybe it was just that.
As soon as the dino tail (to this day i dont know what it is) starts showing it already seems to have a world

henriquegemignani (approved):

PR #9264, opened 2026-06-08
Add logging to game connection

Miepee:
Ive heard multiple reports in the past about randovania not connecting to certain games, I now have gotten an actual log file from sam where that happened and i cant decipher anything useful from it, so adding more log statements it is.


henriquegemignani:
Some of these should be debug to go only to the log file and not the console.

henriquegemignani (approved):

PR #9268, opened 2026-06-10
Primes: Don't disconnect when trying to read inventory

Miepee:
As is expected by now, reading the items can return an invalid address because we're reading them based off of a cplayer_state pointer which can be null. If that is the case, handle that gracefully so that we don't need to do the whole dance of reconnecting again.


henriquegemignani (approved):

Miepee:
Redid the place where I check for it, because I remembered that Echoes also does object count reading via offsets, and those should also not disconnect....

henriquegemignani (approved):

henriquegemignani:
Enjoy your test failures

PR #9280, opened 2026-06-13
support projecting 3d world-space nodes onto screenshots, for the data editor

duncathan:
This allows us to use in-game coordinates for 3D games, while still having useful maps in the Data Editor. Hydrodynamo Station, for example:
image

If a better viewing angle is desired, a new screenshot can be taken and the camera data updated, and all the nodes will re-transform accordingly:
image

For Echoes, I made a custom build of PWE that allows taking a screenshot of the viewport, and saves a json file next to the screenshot with the camera data formatted for RDV that can be copied into area extra. I'll leave that build here, too, for anyone interested: https://www.dropbox.com/scl/fi/bv962d7tqueov4xp86c8u/pwe-screenshot-build.zip?rlkey=e6ap0qqvxjnfbhfr8sn0n7400&st=3hvzd739&dl=0


henriquegemignani (changes requested):
Add `numpy` as a direct dependency of rdv, at least for the `gui` extra.

And add some tests for `game_loc_to_qt_local`

duncathan:
> Add numpy as a direct dependency of rdv, at least for the gui extra.

Kk. To be clear, it's already an indirect dependency of the gui extra via matplotlib so it isn't a *new* dependency here

henriquegemignani (approved):
Maybe an extra test to get codecov patch coverage pass, but up to you.

PR #9287, opened 2026-06-14
Better safeguards if we can't read the inventory

Miepee:
A fix because *someone* put in 99 missiles per expansion (see #9286 )
I think under normal circumstances, the worst case is that we fail when reading the inventory during an elevator transition or whatever, and then duplicate some ammo for the player.

### DIFF
--- a/randovania/game_connection/connector/prime_remote_connector.py
+++ b/randovania/game_connection/connector/prime_remote_connector.py
@@ -407,6 +407,9 @@ class PrimeRemoteConnector(RemoteConnector):
             if region is not None:
                 await self.update_current_inventory()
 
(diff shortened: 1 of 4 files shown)


randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -407,6 +407,9 @@ async def update(self) -> None:
             if region is not None:
                 await self.update_current_inventory()
 
+                if self.last_inventory == Inventory.empty():
```
henriquegemignani: Rather have the inventory be `None`

Miepee: Done

henriquegemignani (approved):

PR #9296, opened 2026-06-15
OPR: add temple grounds area images

JeffGainsNGames:
also includes node data for generic, translator gates, and event nodes
all images compressed to ~50% original size

will be best to review in app but below is an example of what most rooms look like perspective-wise
image

note there's no image for credits and there's currently an issue when going from an area with a map -> no map


randovania/games/prime2_opr/assets/maps/Agon Gate.png
duncathan, line 1: The transport rooms are all going to be non-unique. Let's use their GUI names (e.g. Agon Gate) as their map names

randovania/games/prime2_opr/assets/maps/Temple Assembly Site.png
duncathan, line 1: too bad that ceiling is blocking the portal. doubt we can do too much about it, though

duncathan (changes requested):

randovania/games/prime2_opr/assets/maps/Temple Assembly Site.png
JeffGainsNGames: yeah there's a few rooms like this where there's also backfacing collision that prevents possibly better angles

Image

duncathan: shame. maybe that's improvable in the future but feels like that's gonna get real complicated real fast lol

duncathan:
you forgot the temple transports!

duncathan (approved):

PR #9304, opened 2026-06-18
Primes: Read inventory array in bulk

Miepee:
Havent checked yet if tests fail (they probably do), just wanted to have this opened so it can get a look already.
Have not tested yet whether this fatally crashes somewhere in the current state as ive rewritten it a bit, but the general gist was that it was ~2ms faster for prime1 and ~2x as fast for echoes (my tests in debug went from 6.5 seconds to 3 seconds)

### DIFF
--- a/randovania/game_connection/connector/prime_remote_connector.py
+++ b/randovania/game_connection/connector/prime_remote_connector.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import dataclasses
+import itertools
 import logging
 import struct
 import uuid
@@ -184,17 +185,14 @@ class PrimeRemoteConnector(RemoteConnector):
             splits.append(total_bytes)
 
         # Create memory ops to read each chunk.
-        mem_ops = []
-        for index, element in enumerate(splits[:-1]):
-            mem_ops.append(
-                MemoryOperation(
-                    address=player_state_pointer,
-                    offset=self.powerup_offset(0) + element,
-                    read_byte_count=splits[index + 1] - element,
-                )
+        return [
+            MemoryOperation(
+                address=player_state_pointer,
+                offset=self.powerup_offset(0) + element,
+                read_byte_count=next_element - element,
             )
-
-        return mem_ops
+            for element, next_element in itertools.pairwise(splits)
+        ]
 
     @property
     def multiworld_magic_item(self) -> ItemResourceInfo:
--- a/randovania/game_connection/connector/prime_remote_connector.py
+++ b/randovania/game_connection/connector/prime_remote_connector.py
@@ -226,13 +226,20 @@ class PrimeRemoteConnector(RemoteConnector):
         item_responses = [
             joined_result[i : i + self.powerup_size] for i in range(0, len(joined_result), self.powerup_size)
         ]
+        if len(item_responses) != self.total_item_length:
+            raise ValueError(
+                f"Should have read {self.total_item_length} items from the game, instead read {len(item_responses)}"
+            )
 
         inventory = {}
         for item in resource_db.get_all_items():
             if item.extra["item_id"] >= 1000:
                 continue
             item_response = item_responses[item.extra["item_id"]]
-            inv = InventoryItem(*struct.unpack(">II", item_response))
+            struct_format = ">II"
+            if self.powerup_size > 8:
+                struct_format += f"{self.powerup_size - 8}x"
+            inv = InventoryItem(*struct.unpack(struct_format, item_response))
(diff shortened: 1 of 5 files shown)


randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -81,6 +81,20 @@ def __init__(self, version: all_prime_dol_patches.BasePrimeDolVersion, executor:
 
         self._timer = InfiniteTimer(self.update, self._dt)
 
+    @property
+    def total_item_length(self) -> int:
+        """Returns the length of the inventory array for this game."""
+        raise NotImplementedError
+
+    @property
+    def powerup_size(self) -> int:
```
henriquegemignani, line 91: how is this remotely working, I don't see any implementation for these methods

Miepee: ....i forgot to paste them back in oops

randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -181,13 +219,23 @@ async def get_inventory(self) -> Inventory:
...
+            joined_result[i : i + self.powerup_size] for i in range(0, len(joined_result), self.powerup_size)
+        ]
 
         inventory = {}
-        for item, memory_op in zip(all_items, memory_ops, strict=True):
-            inv = InventoryItem(*struct.unpack(">II", ops_result[memory_op]))
+        items_by_id = {
+            item.extra["item_id"]: item for item in resource_db.get_all_items() if item.extra["item_id"] < 1000
+        }
+        for index, item_response in enumerate(item_responses):
```
henriquegemignani: Why not:
```python
for item in resource_db.get_all_items():
    if item.extra["item_id"] >= 1000:
        continue
    item_response = item_responses[item.extra["item_id"]]
```

Miepee: because my initial PoCs for this were hackily implemented and i didnt think of it when refactoring this to support both games.

randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -181,13 +219,26 @@ async def get_inventory(self) -> Inventory:
...
+            )
 
         inventory = {}
-        for item, memory_op in zip(all_items, memory_ops, strict=True):
-            inv = InventoryItem(*struct.unpack(">II", ops_result[memory_op]))
+        for item in resource_db.get_all_items():
+            if item.extra["item_id"] >= 1000:
+                continue
+            item_response = item_responses[item.extra["item_id"]]
+            # using "unpack_from", because Prime has amount/capacity, while
```
henriquegemignani: actually, thinking about it this way maybe should just use the `joined_result`  directly and `unpack_from`, with the offset being `self.powerup_size * item.extra["item_id"]`

Miepee: didn't think of that, good catch.

Miepee: done

randovania/game_connection/connector/prime_remote_connector.py
```diff
@@ -151,12 +165,36 @@ async def current_game_status(self) -> tuple[bool, Region | None]:
...
+        # being aligned to the power ups and not surpassing the executor's max output.
+        total_bytes = self.total_item_length * self.powerup_size
+        chunk_size = self.executor.max_output - (self.executor.max_output % self.powerup_size)
+        splits = [i * chunk_size for i in range((total_bytes // chunk_size) + 1)]
+        if total_bytes % chunk_size:
+            splits.append(total_bytes)
+
+        # Create memory ops to read each chunk.
+        mem_ops = []
+        for index, element in enumerate(splits[:-1]):
```
henriquegemignani: Maybe this?
```python
mem_ops = [
    MemoryOperation(
        address=player_state_pointer,
        offset=self.powerup_offset(0) + element,
        read_byte_count=next_element - element,
    )
    for element, next_element in itertools.pairwise(splits)
]
```

henriquegemignani (approved):

randovania/game_connection/connector/prime_remote_connector.py
Miepee: oooh, wasn't aware that that existed. that make things nicer.

henriquegemignani: I opened an editor to write a suggestion with `zip` and ruff told me to use pairwise instead lmao

PR #9305, opened 2026-06-18
Refactor socketio event handlers, and typing for network_client and network_common

duncathan:
The big stuff is in [server/socketio.py](https://github.com/randovania/randovania/blob/38f1295393a973df21ac1cfae256733ebf7d7fb2/randovania/server/socketio.py)


duncathan:
here are the docstrings from the two main things added, to give a better idea what's happening in this pr:

### `server_event_handler()`
Transforms a function into a `ServerEventHandler` so that it can be registered via
`ServerApp.on()` or called from the client using `fn.server_call()`.

Example usage:

```py
@server_event_handler("multiplayer_list_sessions")
async def list_sessions(sa: ServerApp, sid: str, limit: int | None) -> list[dict]:
    return [{"number": i} for i in range(limit if limit is not None else 100)]

result = await list_sessions.call_server(NetworkClient())(2)

# prints "[{'number': 0}, {'number': 1}, {'number': 2}]"
print(result)
```

### `client_signal()`
Transforms the callable into a `ClientSignal` for fully type-checked signal emission from the server.

Example usage:
```py
@client_signal("multiplayer_binary_inventory")
async def WORLD_BINARY_INVENTORY(entry_id: str, user_id: int, raw_inventory: bytes) -> None: ...

class NetworkClient:
    def __init__(self, sio: AsyncClient):
        self.sio = sio

        WORLD_BINARY_INVENTORY.register(self.sio, self._on_world_user_inventory_raw)

    def _on_world_user_inventory_raw(self, entry_id: str, user_id: int, raw_inventory: bytes) -> None:
        print(entry_id, user_id, raw_inventory)

# prints "'entry', 1234, b'4321'"
await WORLD_BINARY_INVENTORY.emit(ServerApp())("entry", 1234, b"4321")
```

duncathan:
bleh, I see an issue with the current setup. gonna have to think about how to avoid a server import in the client, for this

duncathan:
okay now it's working. would appreciate feedback, and I think I still need to update a docstring that I wrote for the old way I had it set up. it's definitely more verbose now but I think it's worth it to be able to have full static typing across the client/server barrier

randovania/network_common/signals/client_signals.py
```diff
@@ -0,0 +1,119 @@
...
+    else:
+        return typing.cast("tuple[SioDataType, ...]", args)
+
+
+class ClientSignal[**P]:
+    def __init__(self, fn: AsyncCallable[P, None], message: str):
+        self.fn = fn
+        self.message = message
+
+    async def __call__(self, *args: P.args, **kwargs: P.kwargs) -> None:
```
henriquegemignani: Just reading code I don't know what's the purpose of this one

randovania/server/multiplayer/session_admin.py
```diff
@@ -396,6 +397,8 @@ async def _get_permalink(sa: ServerApp, sid: str, session: MultiplayerSession) -
 
 
 async def admin_session(sa: ServerApp, sid: str, session_id: int, action: str, *args: typing.Any) -> typing.Any:
+    # FIXME: break this out into separate event handlers for each action
```
henriquegemignani: nah, should just swap these to REST api instead

randovania/server/multiplayer/session_admin.py
```diff
@@ -649,6 +652,8 @@ async def _create_patcher_file(
 
 
 async def admin_player(sa: ServerApp, sid: str, session_id: int, user_id: int, action: str, *args: typing.Any) -> None:
+    # FIXME: break this out into separate event handlers for each action
```
henriquegemignani: same

henriquegemignani (changes requested):

randovania/server/multiplayer/session_admin.py
duncathan: either way, should still break out into separate endpoints rather than a single one

henriquegemignani: oh yeah, just want the comment to be clear. It wouldn't really be event handler in that case!

randovania/network_common/signals/client_signals.py
duncathan: I guess it's not strictly necessary, since this decorator just gets applied to stubs, but it felt weird for a decorated function to not actually be callable. maybe I'll make this raise an error with a descriptive message

henriquegemignani (approved):

PR #9310, opened 2026-06-19
Dread: Remove knowledge trick from pb limitations disabled logic

Tyranisaur:


randovania/games/dread/logic_database/Ferenia.txt
```diff
@@ -1536,7 +1536,7 @@ Extra - asset_id: collision_camera_021
           Any of the following:
               # Open the door from below
```
Miepee, line 1537: these things would benefit from a small comment explaining why the template cant be used. Makes it easier to grep for in the future
Or maybe that comment should be on the template's side instead...

Logarhythmic (approved):
We're hiding the pb limitations setting in the UI, yeah? No need to keep these.

Miepee:
AFAIK the PB limitations aren't hidden - its the main PB setting that is hidden. Not sure if that changes your opinion of this or not.

Miepee (approved):
Approved by logic maintainer

PR #9316, opened 2026-06-20
Prime 1: Make charge combos missile systems and missile related

Miepee:
Small part of #7930 I presume.

Since this change is during a CGC tournament match, I'd like to get explicit approval from the majority of current TAs whether they would want to have this change. 
Although if the prime leads don't want to ever have prime and echoes match in this, then that holds more power for me.


duncathan:
I don't think we should be merging a balance change like this in the middle of CGC, regardless of what the TAs say lol.

Miepee:
Maybe. My thought was that it'd make the rest of the matches line more up with player expectations because from what I understood, most players already assume that the prime features for charge combos are the same as the choes ones.

Miepee:
With the tournament coming to a close, this would be nice to get reviewed

Samuel6710 (approved):

ThanatosGit (approved):

PR #9321, opened 2026-06-22
typing for GUI module

duncathan:
wip


randovania/game_connection/game_connection.py
```diff
@@ -112,7 +112,9 @@ def remove_connection_builder(self, builder: ConnectorBuilder) -> None:
         self.connection_builders.remove(builder)
         self._on_builders_changed()
 
-    def get_connector_for_builder(self, builder: ConnectorBuilder) -> RemoteConnector | None:
+    def get_connector_for_builder(self, builder: ConnectorBuilder | None) -> RemoteConnector | None:
+        if builder is None:
```
henriquegemignani, line 116: bleh I hate this annoyance with typing. `dict.get` returns None if you pass unrelated things, such as None >.>

randovania/gui/dialog/async_race/async_race_proof_popup.py
```diff
@@ -34,7 +34,7 @@ def _on_proof_updated(self) -> None:
         # Mark proof url as error only if one was given that does not look like a url
         common_qt_lib.set_error_border_stylesheet(
             self.ui.proof_edit,
-            self.proof_url and self._proof_re.match(self.proof_url) is None,
+            bool(self.proof_url and self._proof_re.match(self.proof_url) is None),
```
henriquegemignani: `bool(self.proof_url) and self._proof_re.match(self.proof_url) is None`

randovania/game_connection/game_connection.py
duncathan: yeah :/ not much to be done about it though

duncathan:
ok i think it's ready. that was really annoying to do lol

randovania/game_description/db/region_list.py
```diff
@@ -168,12 +168,12 @@ def area_by_area_location(self, location: AreaIdentifier) -> Area:
     def region_by_area_location(self, location: AreaIdentifier) -> Region:
         return self.region_with_name(location.region)
 
-    def region_and_area_by_area_identifier(self, identifier: AreaIdentifier) -> tuple[Region, Area]:
+    def region_and_area_by_area_identifier(self, identifier: AreaIdentifier | NodeIdentifier) -> tuple[Region, Area]:
```
henriquegemignani, line 171: Kinda dislike needing to be explicit like this :/

henriquegemignani (changes requested):

randovania/game_description/db/region_list.py
duncathan: the alternatives were:
- change every site that calls it with a NodeIdentifier to do `identifier.area_identifier`
- make a whole protocol just for this one tiny function
- make `NodeIdentifier` inherit from `AreaIdentifier`

they all felt way worse than doing it this way

randovania/gui/preset_settings/preset_teleporter_tab.py
```diff
@@ -24,13 +23,14 @@
...
     from randovania.interface_common.preset_editor import PresetEditor
+    from randovania.layout.base.base_configuration import BaseConfiguration
 
 
-class PresetTeleporterTab(PresetTab[ConfigurationT_co], NodeListHelper):
-    _teleporters_source_for_location: dict[NodeIdentifier, QtWidgets.QCheckBox]
-    _teleporters_target_for_region: dict[str, QtWidgets.QCheckBox]
-    _teleporters_target_for_area: dict[AreaIdentifier, QtWidgets.QCheckBox]
-    _teleporters_target_for_node: dict[NodeIdentifier, QtWidgets.QCheckBox]
+class PresetTeleporterTab[ConfigurationT: BaseConfiguration](NodeListHelper, PresetTab[ConfigurationT]):
```
henriquegemignani, line 29: why in this one you swapped the NodeListHelper to be before, but in StartingAreaTab it's still later?

henriquegemignani (changes requested):

randovania/gui/preset_settings/preset_teleporter_tab.py
duncathan: this is a bit of a leftover. there's a commit I left out of this PR that changed the mixin order for all of these classes, because they're not really following the MRO they should be, but I decided it was unnecessary for this PR. when I undid it this line conflicted and I just kept the changed version

duncathan: it's related to the FIXME down below. one idea I had was to just define an empty `setupUi` function on this class, and then the children would just override it when they inherit from the generated UI class. but for that to work correctly, I needed to fix the MRO, and it just felt like too much hassle for right now

randovania/gui/dialog/game_export_dialog.py
```diff
@@ -167,36 +168,89 @@ def spoiler_path_for_directory(save_spoiler: bool, output_dir: Path) -> Path | N
...
+        accept_button,
+        fields,
+        tab_widget,
+    )
+
+
+def add_field_validation(
+    accept_button: QtWidgets.QPushButton,
+    fields: dict[QtWidgets.QWidget, Callable[[], bool]],
+) -> None:
```
henriquegemignani, line 245: add some docstrings to both this and add_tabbed_field_validation

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #9322, opened 2026-06-22
Planets (Zebeth) DB: Allow Downward Shots are now in logic

UltiNaruto:
Makes the option `Allow Downward Shots` logical when enabled.


steven11sjf (changes requested):
I would make a new template `Can Break Floor Blocks = OR( Can Use Bombs, Allow Downward Shots )`. This would also allow us to add other changes easier (i.e. screw attack breaking blocks).

UltiNaruto:
I'll do that tomorrow when I wake up then

steven11sjf (approved):
LGTM!

ThanatosGit (approved):
Just the approval for merge.

PR #9328, opened 2026-06-22
Prepare Dock Weakness Distributor to better support multiple dock types

henriquegemignani:
It currently acts as if all dock types being shuffled are different kinds of "Door", not unrelated dock types.

This would be it, but I don't want to touch the GUI for setting it up, so I'm leaving that for later.


randovania/gui/game_details/door_lock_rando_details_tab.py
```diff
@@ -0,0 +1,22 @@
...
+
+from randovania.game_description.db.dock import DockType
+from randovania.gui.game_details.dock_weakness_distribution_details_tab import DockWeaknessDistributionDetailsTab
+
+if TYPE_CHECKING:
+    from randovania.game_description.game_database_view import GameDatabaseView
+
+
+class DoorLockRandoDetailsTab(DockWeaknessDistributionDetailsTab):
+    """A special case for Dock Weakness Distribution, when dealing with Doors."""
```
duncathan: this feels really ugly. I quite like that the details tab is generic and games don't need to worry about it. I know you didn't want to rework the GUI in this PR but I don't love that it needs a downgrade like this in the meantime

randovania/layout/base/dock_rando_configuration.py
```diff
@@ -204,11 +204,19 @@ def bit_pack_unpack(cls, decoder: BitPackDecoder, metadata: dict) -> Self:
             types_state=types_state,
         )
 
-    def is_enabled(self) -> bool:
+    def is_enabled_for(self, dock_type: DockType) -> bool:
+        # FIXME: this really should be per type :)
```
duncathan: I think this needs to be done before I'd wanna merge this

duncathan (dismissed):

randovania/gui/game_details/door_lock_rando_details_tab.py
henriquegemignani: I did think about having a method that creates a new class based on arguments. 
But it needs some input from the game for what type should be used. 

Alternatively, we go and introduce the concept of "door", which games can configure what dock type it is and a lot of the UI handles it automatically. We'd expand this to a type for "teleporter" for the elevator rando tooling.

duncathan: Why not just display the tab for any dock type that has DLR enabled? (Besides the fact that you haven't split that up yet)

henriquegemignani: Create one tab for each? I still would need to create classes for it. And names, because I sure don't want "Portal Locks" or "translator gate locks"

henriquegemignani: Guess it could be a field of the dock type. Also probably could move the dlr config to inside dock type directly 🤔

duncathan: Yeah, one tab for each. Put the tab name in the distributor params, and then create the classes dynamically like you mentioned

duncathan: > Also probably could move the dlr config to inside dock type directly

I like it being its own class so that you can check `dock_type.distributor_params is None` as a canonical way to know if the dock type supports weakness randomization

henriquegemignani: Yes I do mean that. Just nested in dock type, instead of in a dict mapping dock type to that structure

randovania/game_description/pretty_print.py
```diff
@@ -258,15 +259,20 @@ def write_human_readable_meta(game: GameDescription, output: TextIO) -> None:
...
 
-        for weakness in game.dock_weakness_database.get_by_type(dock_type):
+        for weakness in game.dock_type_database.get_by_type(dock_type):
             output.write(f"\n  * {weakness.name}\n")
             for extra_name, extra_field in weakness.extra.items():
                 output.write(f"      Extra - {extra_name}: {extra_field}\n")
+            if weakness.unsafe_target_in_distributor_wtw:
+                output.write(
+                    f"      Considered unsafe as target for "
+                    f"{DockWeaknessDistributorMode.WEAKNESS_TO_WEAKNESS.long_name}\n"
```
ThanatosGit: capitalization of "By Type" looks odd in the text files but dunno if we want to change it here

randovania/layout/base/dock_weakness_distributor_configuration.py
```diff
@@ -0,0 +1,288 @@
...
+        """
+        return (
+            dock_type in self.types_state and self.types_state[dock_type].mode != DockWeaknessDistributorMode.ORIGINAL
+        )
+
+    def is_any_type_mode(self, mode: DockWeaknessDistributorMode) -> bool:
+        """
+        True, if at least one type is configured as the given mode.
+        """
+        return any(type_state == mode for type_state in self.types_state.values())
```
ThanatosGit: Such large PRs I always check out to have something local to easily check types.
This check is always `False` because it compares two different things? It's missing a `.mode`

(variable) type_state: WeaknessDistributorTypeState
(parameter) mode: DockWeaknessDistributorMode

randovania/bitpacking/bitpacking.py
```diff
@@ -229,6 +229,8 @@ def bit_pack_encode(self: DataclassInstance, metadata: dict) -> Iterator[tuple[i
 
             field_meta = dict(**field.metadata)
             field_meta["reference"] = reference_item
+            field_meta["parent_object"] = self
```
ThanatosGit, line 232: I don't really understand what I'm looking at here but this never gets `decoded` because?

ThanatosGit (changes requested):

randovania/bitpacking/bitpacking.py
henriquegemignani: This is for `bit_pack_encode` so `decoded` doesn't even exist in this context.

The changes are so that when encoding `TrickLevelConfiguration`, it gets access to the `BaseConfiguration` object. Main purpose was just so I could grab the `RandovaniaGame`, but a few of these changes ended up not needed but I left it in.

randovania/game_description/pretty_print.py
henriquegemignani: I put some quotes around it to make it more clear it's the mode name

ThanatosGit (approved):

JeffGainsNGames (approved):
Reviewed all files, no comments

PR #9333, opened 2026-06-24
Add Invalid pickup index status message + client signal  for status messages

Miepee:
Inspired by this lovely tournament match:
grafik

Looked at the prime side and turns out that the game description view does not have the "items in every room" pickups when they're disabled, which helps a bit too.
grafik


henriquegemignani (approved):

Miepee:
Resolved conflicts. would appreciate a quick glance just to check that i didnt mess up anything majorly

JeffGainsNGames:
".. did you play while disconnected?" feels like odd phrasing to me for this memo

"Please check your connection to Randovania" ?

toasterparty:
You for sure need a strong call-to-action. "An error occurred. Reset your game and check your connection to avoid permanently loosing items." or something

Miepee:
> feels like odd phrasing to me for this memo 

I was phrasing it that way specifically to 1) avoid them opening a bug 2) suggest a possible cause on what could have caused it, so they know to pay attention for it to not happen in the future

> "An error occurred. Reset your game and check your connection to avoid permanently loosing items." or something

That is too long. We only have about... 70 characters to use. That one is almost 100.

Shaleblade:
> That is too long. We only have about... 70 characters to use. That one is almost 100.

How about this? 

> Error sending item. Reset the game and check Randovania's connection.

Miepee:
> check Randovania's connection

As is, that part is useless. The message can only appear if youre connected both to the game and to the server.   
It also imo still doesn't fully convey *what* caused it.

Miepee:
The PR has already been approved by Henrique. The only thing that seems to be blocking is the prime/echoes specific flavourtext I added to avoid us getting bug reports about a well known occurence.
It would be nice to have some consensus reached so that I can resolve the conflicts and have it merged.

toasterparty (approved):

Miepee:
Fixed conflicts and rephrased slightly to:
> Invalid pickup location. Reset game & check Randovania for details.

Keeps toaster's original call to action + encourages them to see which items have been lost

PR #9344, opened 2026-06-29
Update to new MSR version

ThanatosGit:
For August release
image


randovania/games/samus_returns/logic_database/header.txt
```diff
@@ -375,7 +375,7 @@ Dock Weaknesses
               All of the following:
                   Knowledge (Beginner)
                   Any of the following:
-                      Shoot Beam Burst
+                      Disabled Buff Charge Doors and Shoot Beam Burst
```
dyceron: I think this whole block could redone entirely. As with Dread, there's no need to have a Knowledge check and the option accounted for here. So I'd change this whole thing by replacing the knowledge line with disabled buff and then any of the following becomes beam burst or power bomb

dyceron (changes requested):

randovania/games/samus_returns/logic_database/header.txt
ThanatosGit: Changed it (and hopefully it's correct)

dyceron: Yes that's good. Don't forget to update the changelog.

ThanatosGit:
Should be good now?!

dyceron:
Seems correct. Need to fix the tests though.

ThanatosGit:
It failed on `test/server/async_race/test_room_api.py::test_livesplit_socket[True]`. There is nothing to fix besides pressing the "Re-run job" button.

dyceron (approved):

dyceron (changes requested):
Oh you need to fix the changelog now.

dyceron (approved):

PR #9358, opened 2026-07-02
Move some game connection logging to debug

Miepee:
We've received some reports about users being confused seeing these on the terminal, so I moved them all to debug for now.
PR is targeting stable, not main.


henriquegemignani (approved):

henriquegemignani:
can't merge without updating the pr

PR #9359, opened 2026-07-02
Logs about nintendont uploading

Miepee:
This is 1) to catch weird errors sometimes appearing despite the upload being successful, and 2) due to a rando match that happened in the CGC tournament 

PR is targeting stable, not main.

### DIFF
--- a/randovania/gui/widgets/game_connection_window.py
+++ b/randovania/gui/widgets/game_connection_window.py
@@ -397,9 +397,12 @@ class GameConnectionWindow(QtWidgets.QMainWindow, Ui_GameConnectionWindow):
         box.show()
 
         try:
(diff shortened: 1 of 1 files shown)


randovania/gui/widgets/game_connection_window.py
```diff
@@ -397,9 +397,12 @@ async def on_upload_nintendont_action(self, builder: NintendontConnectorBuilder)
         box.show()
 
         try:
+            builder.logger.debug("Uploading Nintendont to ther Wii...")
```
JeffGainsNGames: typo: ther

Miepee: Fixes

henriquegemignani (approved):

PR #9361, opened 2026-07-03
Dread: Have the Ferenia Twin Robot fight live from the start

Tyranisaur:
Confine dangerous storm missile logic to Highly Dangerous Logic.


Logarhythmic (approved):
I've never run into this logic personally but it seems OK to flag it as dangerous.

While I was looking at the room, I noticed this `Start Point 2` node that is trivially connected to a door and nothing else. Does it have a purpose? 

Image

duncathan:
> I've never run into this logic personally

Actually, everyone who's ever played a dread seed has been affected by this logic, because of the fact it made storms into a dangerous resource.

Logarhythmic:
> > I've never run into this logic personally
> 
> Actually, everyone who's ever played a dread seed has been affected by this logic, because of the fact it made storms into a dangerous resource.

I understand that. What I mean to say is that I've never run into the case in which I couldn't make a return trip through the room because I was inconvenienced by having the robots spawn.

ThanatosGit (approved):
Approved by DB maintainer. Just needs to resolve the changelog conflict.

PR #9372, opened 2026-07-05
Add pydantic_util

henriquegemignani:
Going to use this to shove pydantic models as compressed json in the server db.


henriquegemignani:
Ooops can't merge without an use. Uhh

JeffGainsNGames (approved):
learned a bit of about pydantic

PR #9376, opened 2026-07-06
test to ensure that all signals have at least one handler registered

duncathan:


henriquegemignani:
A bit big of change that doesn't exactly fix bugs, I'd rather have it on `main`.

henriquegemignani (changes requested):
Please change the PR to be against `main` (this time via a Request changes)

PR #9379, opened 2026-07-06
Various improvements to our custom version schema

henriquegemignani:
- Embed the version schema, instead of separate package
- Update to latest setuptools-scm
- Adjust git describe so setuptools-scm provides us the git hash, avoiding the need for fetching ourselves


setup.py
```diff
@@ -165,11 +172,46 @@ class CustomBuild(build):
...
+
+    version_file = Path(version.config.absolute_root).joinpath(version.config.version_file)
+
+    # We want to save the git hash to a file, so we can easily access it in frozen runtime, where git is not available.
+    #
+    version_file.with_name("version_hash.py").write_text(
+        f"""# coding: utf-8
+# file generated by setuptools_scm + setup.py
+# don't change, don't track in version control
+git_hash = {version.short_node!r}
```
ThanatosGit: According to https://github.com/pypa/setuptools-scm/blob/main/vcs-versioning/src/vcs_versioning/_node_utils.py, `short_node` is simply truncated to 10 characters, including the prefix `g`. Now you have 10 bytes.

If you don't want to change the entire version thing (including the permalink), you still need to use `git_hash = repr(bytes.fromhex(version.node[1:9]))` as in `randovania-scm-version-schema` to get an 8-character string that fits into 4 bytes.

ThanatosGit (changes requested):

setup.py
henriquegemignani: Good catch. So focused on getting the surroundings working, I forgot to pay attention on what the short hash was.

ThanatosGit (approved):

PR #9383, opened 2026-07-07
Don't stretch the region image if it's aspect ratio doesn't match the region's bounds.

henriquegemignani:


ThanatosGit:
Since you touched this code part, is there a way I can convice you to also fix the zoom + draggable thing? 
How much much you can drag from the region image depends on the zoom. It just cuts off.
See the screenshot of same region, same marked area with different zoom.

Tbh: I don't really know what is the issue here. I thought that it is maybe just the image bounds being too small and that it maybe fits into this PR touching the bounds.



image

----
image

henriquegemignani:
> Since you touched this code part, is there a way I can convice you to also fix the zoom + draggable thing? How much much you can drag from the region image depends on the zoom. It just cuts off. See the screenshot of same region, same marked area with different zoom.

Oh no you've predicted my follow-up PR. Yeah I already have that fixed was gonna PR it up after this to avoid conflicts.

ThanatosGit (approved):

PR #9393, opened 2026-07-08
Connection Editor rework

### LINKED ISSUE #7871: Easier way to change individual logic blocks
Sometimes when updating the logic database it is relevant to take existing conditions on a connection and wrap them inside an AND or an OR, to add other branches of requirements. When there's a sufficiently large combination of requirements, it quickly gets tempting to modify the json file manually instead of recreating the logic one level deeper manually.

JustinDMS:
# Context

My interest in working on the connection editor began after reading through this issue: #7871. Having worked with the editor on occasion, a lot of the issues resonated with me. I took this as an opportunity to learn more about Randovania and set out to add one or two useful features. I quickly found that it was awkward to manipulate the requirement tree because it existed _as the widget tree_, that is, there was no separation between the data and the interface. This prompted a month-long deep dive into Model-View-Controller architecture, which brings us to the new editor.

# Goal

- Modernize the connection editor's workflow and facilitate a core which can more easily be expanded upon
- Fix #7871 

# Scope

- Original scope rundown out of date, please see https://github.com/randovania/randovania/pull/9393#issuecomment-5060213718

---

I opened this PR as a draft to get eyes/hands on the rework and spark some discussion. There are undoubtedly things that will need to be changed, so I invite those interested to check it out!

image


ThanatosGit:
I don't really deal with database maintenance that much, but I really like the result of your work.
Especially because it much more closely matches what you see in the user interface.

I haven’t looked at the code in detail yet, but maybe it’s time to split up the file, since I’m not a big fan of a single file with 1,000 LOC, especially since you’ve already mentioned that it contains several logically distinct parts with the MVC pattern

JustinDMS:
> I’m not a big fan of a single file with 1,000 LOC, especially since you’ve already mentioned that it contains several logically distinct parts with the MVC pattern

I could realistically see the model being extracted and used for any place that works with visualizing requirements (`resource_database_editor.TemplateEditor`, `connections_visualizer.py`). However, those places use `QTreeWidget` and `QTreeWidgetItem`, a level of abstraction above what I used (`QStandardItemModel / QStandardItem` + `QTreeView`). The increase in verbosity can be partially attributed to that as well. 

It's hard for me to estimate how much work it would be to replace. I'm not against it, but it would be a pretty drastic increase in scope.

ThanatosGit:
I did not mean a generalizing of the model into a shared abstraction. That would be out of scope for this here.

I mean just a mechanical split of this big file since you mentioned it already contains different classes in a MVC pattern.
At least for me, it is much easier to review a couple of small files compared to one very big file.

I tooke the current state of this PR and ask an LLM if it could be split easily, it answered with the following. 
As always, take this with a grain of salt.

-----

## Does it have proper cut points?

Yes — unusually clean ones. The file is already partitioned into cohesive, single-responsibility classes with narrow interfaces (signals + a couple of public methods):

| Unit | Depends on |
|---|---|
| `ConnectionsEditor` (dialog, the public API) | model, view, controller |
| `Command` (undo) | model, view |
| `RequirementModel` + `_ROLE` | — |
| `RequirementView` | model |
| `RequirementController` | everything (the hub) |
| `Editor` + 8 subclasses | `_ROLE`, db |
| `signals_blocked` | — (generic util) |

The coupling is a clean DAG, not a tangle. The only widely-shared things are the `_ROLE` constant and the `Command`/model/view trio. Nothing here fights a split.

## Suggested layout

There's already precedent one level up: `dialog/async_race/` is a sub-package of exactly this kind. Mirror it:

```
randovania/gui/dialog/connections_editor/
    __init__.py      # from .dialog import ConnectionsEditor  → import path stays stable
    dialog.py        # ConnectionsEditor
    model.py         # RequirementModel, Command, _ROLE
    view.py          # RequirementView
    controller.py    # RequirementController  (still the big one, ~350 LOC — fine, it's the hub)
    editors.py       # Editor + the 8 editor subclasses (~340 LOC, one cohesive family)
```

Using a package with an `__init__.py` re-export keeps `from ...dialog.connections_editor import ConnectionsEditor` working for every existing caller, so it's zero-churn elsewhere.

JustinDMS:
The file has been split as suggested

JustinDMS:
Added general keyboard support and a few shortcuts for basic operations. Looking for feedback on the keybinds used:

`Insert` - Add
`Delete` - Delete
`Alt+Up` - Shift Up
`Alt+Down` - Shift Down

JustinDMS:
I've refined the PR to a comfortable-enough point to graduate it from being a draft. I took a lot of care to make it as architecturally correct as possible.

Here's a more up-to-date look into the scope of the changes:
- Full replacement of `connections_editor.py`
- New module structure: `model.py`, `view.py`, `controller.py`, `command.py`, `editor.py`, `path.py`, `requirement_tree.py`
- Tree operations: add (as sibling or child, context-sensitive), delete (with smart selection fallback), and outliner-style move up/down (sibling swap, ascend/descend across array boundaries)
- Full undo/redo via `QUndoCommand`, including correct selection restoration in both directions
- Keyboard shortcuts (Insert, Delete, Alt+Up, Alt+Down), scoped to the tree view via `WidgetShortcut` context so they don't interfere with property-panel field editing
- Per-type property editors (Item, Event, Trick, Damage, Version, Misc, Template, Node, Array) driven by a single type-selector combo box, built only for resource types the target game's database actually has
- `Path`, a small immutable value type wrapping `tuple[int, ...]`, used for expressing a location within the tree
- A dedicated requirement_tree module of pure (Qt-free) functions (`insert_at_path`, `remove_at_path`, `replace_at_path`, `default_from_type`, `change_to_type`) implementing structural-sharing immutable tree edits
- Test coverage across all of the above

Also, here's a collapsed call graph (courtesy of Crabviz) for a more overhead look into how everything works together:

image

test/conftest.py
```diff
@@ -194,6 +203,41 @@ def echoes_game_description() -> GameDescription:
     return default_database.game_description_for(RandovaniaGame.METROID_PRIME_ECHOES)
 
 
+@pytest.fixture
+def echoes_simple_resource(echoes_resource_database) -> Callable[[str], ResourceRequirement]:
```
henriquegemignani: Place these two fixtures in a new `conftest.py` on a folder closer to where it's used instead of the root one.

(yes a lot of the stuff in this file could be moved elsewhere)

randovania/gui/dialog/connections_editor/path.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+
+class Path(tuple[int, ...]):
```
henriquegemignani: Add some docstrings to this class to explain it's purpose and maybe what these methods are too.

randovania/gui/dialog/connections_editor/requirement_tree.py
```diff
@@ -0,0 +1,103 @@
...
+from randovania.game_description.requirements.requirement_template import RequirementTemplate
+from randovania.game_description.requirements.resource_requirement import ResourceRequirement
+from randovania.game_description.resources.resource_database import ResourceDatabase
+from randovania.game_description.resources.resource_info import ResourceInfo
+from randovania.game_description.resources.resource_type import ResourceType
+
+from .path import Path
+
+
+def tuple_insert(items: tuple, idx: int, value: Any) -> tuple:
```
henriquegemignani: I think this, and the other two tuple methods, could be typed like this:
```python
def tuple_insert[T](items: tuple[T, ...], idx: int, value: T) -> tuple[T, ...]:
```

randovania/gui/dialog/connections_editor/requirement_tree.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def remove_at_path(root: RequirementArrayBase, path: Path) -> Requirement:
+    idx = path.head()
+
+    if len(path) == 1:
+        new_items = tuple_remove(root.items, idx)
+        return type(root)(new_items, root.comment)
+
+    next_root = cast(RequirementArrayBase, root.items[idx])
```
henriquegemignani: I would say this is a case of doing `assert isinstance(next_root, RequirementArrayBase)`, or maybe even an `if not isinstance(...): raise CustomException("Invalid path ...")`

randovania/gui/dialog/connections_editor/requirement_tree.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def replace_at_path(root: RequirementArrayBase, path: Path, requirement: Requirement) -> Requirement:
+    if len(path) == 0:
+        return requirement
+
+    next_root = root.items[path.head()]
+    child = (
+        requirement
+        if not isinstance(next_root, RequirementArrayBase)
```
henriquegemignani: Why does `replace_at_path` ignore part of the `Path` if the requirement doesn't match, but `remove_at_path` would error?

Does it even make sense ignore part of the Path?

randovania/gui/dialog/connections_editor/requirement_tree.py
```diff
@@ -0,0 +1,103 @@
...
+
+    new_items = tuple_replace(root.items, path.head(), child)
+    return type(root)(new_items, root.comment)
+
+
+def default_from_type(
+    from_type: ResourceType | type[Requirement], db: ResourceDatabase, region_list: RegionList
+) -> Requirement:
+    if isinstance(from_type, ResourceType):
+        resource_info = cast(ResourceInfo, next(iter(db.get_by_type(from_type))))
```
henriquegemignani: Given `get_by_type` just returns a list, I'd just do `db.get_by_type(from_type)[0]`. Also not sure why this needs a cast?

randovania/gui/dialog/connections_editor/requirement_tree.py
```diff
@@ -0,0 +1,103 @@
...
+    if from_type == NodeRequirement:
+        node_id: NodeIdentifier = next(region_list.iterate_nodes()).identifier
+        return NodeRequirement(node_id)
+
+    raise RuntimeError(f"Unknown requirement type: {from_type}")
+
+
+def change_to_type(current: Requirement, to_type: type, db: ResourceDatabase, region_list: RegionList) -> Requirement:
+    """Wrapper to retain data when changing between array types"""
+    if isinstance(current, RequirementArrayBase) and issubclass(type(to_type), RequirementArrayBase):
```
henriquegemignani: `type(to_type)` given that `to_type` is already a `to_type` seems very wrong.

randovania/gui/dialog/connections_editor/requirement_tree.py
```diff
@@ -0,0 +1,103 @@
...
+        return to_type(current.items, current.comment)
+    return default_from_type(to_type, db, region_list)
+
+
+def _at_path(requirement: Requirement, path: Path) -> Requirement:
```
henriquegemignani: This function seems to only be used by tests?

henriquegemignani (changes requested):
Initial things. Haven't looked much at the actual mvc yet...

More of the classes could use some docstrings I'd say.

randovania/gui/dialog/connections_editor/requirement_tree.py
JustinDMS: Correct. I felt that it made sense to define it as part of the module rather than locally, even if it is currently only used by the tests. That way the test is not concerning itself with implementation, just behavior.

randovania/gui/dialog/connections_editor/requirement_tree.py
JustinDMS: Fixed in commit e5b3bef4e523962ce0e9993f10f817fdd6a17840

For future reference, do we generally prefer generics over `typing.Any`?

henriquegemignani: generics is more precise than Any, so yes.

randovania/gui/dialog/connections_editor/requirement_tree.py
JustinDMS: TLDR: It is guaranteed that the recursive case will always be a `RequirementArrayBase`, so I opt to cast for brevity. I can still change it if you prefer, but I wanted to present my case first.

It did use an `isinstance` call back when `Path` was just a raw list of int managed by the controller. With `Path` and the `requirement_tree` module, a line in the sand was drawn with regards to who's responsibility it is that a valid `Path` is passed in. 

In the newly-added `Path` docstring:
> Every index except the last must point to a RequirementArrayBase, since only array requirements have children to descend into. The last index may point to any Requirement type, since that's the node the Path is actually identifying.

duncathan (dismissed):
some general notes from stuff I noticed at a glance:
- use absolute imports, not relative ones
- class names should be self descriptive rather than needing context to make any sense. stuff like `Command` and `Path` are far too broad and may conflict with other existing names

randovania/gui/dialog/connections_editor/command.py
```diff
@@ -0,0 +1,45 @@
...
+if TYPE_CHECKING:
+    from randovania.game_description.requirements.base import Requirement
+    from randovania.gui.dialog.connections_editor.model import RequirementModel
+    from randovania.gui.dialog.connections_editor.path import RequirementTreePath
+    from randovania.gui.dialog.connections_editor.view import RequirementView
+
+
+class RequirementEditCommand(QUndoCommand):
+    """
+    An undoable/redoable edit to a requirement tree.
```
henriquegemignani: reversible would be better than undoable/redoable

randovania/gui/dialog/connections_editor/model.py
```diff
@@ -0,0 +1,139 @@
...
+        return walk(self.invisibleRootItem().child(0))
+
+    def flags(self, index: QModelIndex | QPersistentModelIndex) -> Qt.ItemFlag:
+        """Overrides Qt's default item flags"""
+        if not index.isValid():
+            return Qt.ItemFlag.NoItemFlags
+
+        flags = Qt.ItemFlag.ItemIsEnabled | Qt.ItemFlag.ItemIsSelectable
+
+        # if index.parent().isValid():
```
henriquegemignani: why this commented code block?

randovania/gui/dialog/connections_editor/model.py
```diff
@@ -0,0 +1,139 @@
...
+            if isinstance(requirement, RequirementArrayBase):
+                children = [walk(item.child(idx)) for idx in range(item.rowCount())]
+                return type(requirement)(children, requirement.comment)
+
+            return requirement
+
+        return walk(self.invisibleRootItem().child(0))
+
+    def flags(self, index: QModelIndex | QPersistentModelIndex) -> Qt.ItemFlag:
+        """Overrides Qt's default item flags"""
```
henriquegemignani: Maybe list why the flags you're setting? But just nitpicking at this point, feel free not to.

randovania/gui/dialog/connections_editor/editor.py
```diff
@@ -0,0 +1,410 @@
...
+
+    def widget(self) -> QWidget:
+        """Returns the root widget used by the Editor"""
+        return self._widget
+
+    def name(self) -> str:
+        """Returns the user-facing name for the Editor"""
+        return self._display_name
+
+    def type(self) -> ResourceType | type[Requirement]:
```
henriquegemignani: Don't create a method named `type` as that overrides the global `type` due to really weird python scope rules. In fact, it can cause issues like the type annotation on this very method not being the `type` you expect 😅

randovania/gui/dialog/connections_editor/editor.py
```diff
@@ -0,0 +1,410 @@
...
+
+        self._add_to_layout([self._combo_name])
+
+    def type(self) -> ResourceType:
+        return self._resource_type
+
+    def to_string(self, resource_info: ResourceInfo) -> str:
+        return resource_info.long_name
+
+    def populate(self, requirement: Requirement) -> None:
```
henriquegemignani, line 153: Add `@override` to all methods that are overriding one.

randovania/gui/dialog/connections_editor/editor.py
```diff
@@ -0,0 +1,410 @@
...
+            self._set_negate_combo(requirement.negate)
+        with signals_blocked(self._spinbox_amount):
+            self._set_spinbox(requirement)
+
+    def _set_negate_combo(self, value: bool) -> None:
+        set_combo_with_value(self._combo_negate, value)
+
+    def _set_spinbox(self, requirement: ResourceRequirement) -> None:
+        resource_info: ResourceInfo = requirement.resource
+        max: int = 10_000  # NOTE - Arbitrary value for damage
```
henriquegemignani: I don't mind the arbitrary max value, but it needs to come from somewhere else.

randovania/gui/dialog/connections_editor/editor.py
```diff
@@ -0,0 +1,410 @@
...
+        super().populate(requirement)
+        requirement = cast("ResourceRequirement", requirement)
+        with signals_blocked(self._checkbox_negate):
+            self._set_negate_combo(requirement)
+
+    def _set_negate_combo(self, requirement: ResourceRequirement) -> None:
+        self._checkbox_negate.setChecked(requirement.negate)
+
+        if requirement.resource.resource_type == ResourceType.EVENT:
+            self._checkbox_negate.setText(
```
henriquegemignani: Hm... feels like it should just always use the type's `negated_prefix`? At the very least, this special casing for Event is weird. The old code was bad but could've been just `resource_type.negated_prefix.strip()` (and probably was older than that attribute...)

randovania/gui/dialog/connections_editor/editor.py
```diff
@@ -0,0 +1,410 @@
...
+            self._repopulate_combo(self._combo_area, region.areas)
+        area: Area = next(iter(region.areas))
+        set_combo_with_value(self._combo_area, area)
+        self._area_changed()
+
+    def _area_changed(self, text: str = "") -> None:
+        area: Area = self._combo_area.currentData(ROLE)
+        with signals_blocked(self._combo_node):
+            self._repopulate_combo(self._combo_node, area.nodes)
+        node: Node = next(iter(area.nodes))
```
henriquegemignani, line 370: This and `_region_changed` both have issues with regions/areas with no children.

I do consider these cases as very niche that doesn't deserve a big re-organisation, but it's nice to not raise exceptions either.
Maybe just filter out all areas with no nodes, and then all regions without areas?

I think this is the case why you could get `None` as the result of a requirement in the existing editor.

randovania/gui/dialog/connections_editor/editor.py
```diff
@@ -0,0 +1,410 @@
...
+    def keyPressEvent(self, event: QKeyEvent) -> None:
+        if event.key() in (Qt.Key.Key_Enter, Qt.Key.Key_Return):
+            self.clearFocus()
+            event.accept()
+            return
+        super().keyPressEvent(event)
+
+
+@contextmanager
+def signals_blocked(widget: QWidget) -> Iterator[None]:
```
henriquegemignani: Move this to `signal_handling`

randovania/gui/dialog/connections_editor/view.py
```diff
@@ -0,0 +1,128 @@
...
+)
+
+from randovania.game_description.requirements.array_base import RequirementArrayBase
+from randovania.gui.dialog.connections_editor.model import ROLE, RequirementModel
+
+if TYPE_CHECKING:
+    from randovania.gui.dialog.connections_editor.path import RequirementTreePath
+
+
+class RequirementView(QObject):
```
henriquegemignani: I would say this is more of a `RequirementEditorView`, or if following the folder name, `ConnectorsEditorView`?

This might apply to more things. I would reserve names of the style `RequirementX` when it interacts directly with a `Requirement` itself.

randovania/gui/dialog/connections_editor/connections_editor.py
```diff
@@ -0,0 +1,119 @@
...
+            self._model,
+            self._view,
+            self._undo_stack,
+        )
+
+        # Misc actions
+        add_action = QAction("Add", self._view._tree)
+        add_action.setShortcut(QKeySequence("Insert"))
+        add_action.setShortcutContext(Qt.ShortcutContext.WidgetShortcut)
+        add_action.triggered.connect(self._controller._on_add_requirement_pressed)
```
henriquegemignani: This and in many other cases in this file needs to access what are being labeled as private methods. I would say these aren't that private after all.

randovania/gui/dialog/connections_editor/connections_editor.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @property
+    def final_requirement(self) -> Requirement:
+        result = self._model.build_requirement()
+        assert result is not None
+        return result
+
+    def accept(self) -> None:
+        result = self._model.build_requirement()
+        if result is None:
```
henriquegemignani: self._model.build_requirement is never None

henriquegemignani (changes requested):

.gitignore
```diff
@@ -97,3 +97,4 @@ exported_videos/*
 /*.rdvpreset
 /*.log
 /*.txt
+/.vscode
```
henriquegemignani: We actually commited quite a few files in it, so the entire directory is too much

JustinDMS: Oh whoops, thought that was only local. Will revert

randovania/gui/dialog/connections_editor/connections_editor.py
henriquegemignani: How was this resolved?

JustinDMS: https://github.com/randovania/randovania/pull/9393/changes/4a95529a7b82439b4322fd3f6742709c805c7e1f#diff-0e570980a0a43e8635d6e8dbdcbc64d4fda2d35786a5120fe1faaa090438c011R111

I had marked it as resolved before pushing the change.

henriquegemignani: That's a different method :)

JustinDMS: Blegh, reading hard. Fixed.

henriquegemignani (approved):

randovania/gui/dialog/connections_editor/controller.py
```diff
@@ -0,0 +1,272 @@
...
+        if has_resource(ResourceType.EVENT):
+            editors.append(SimpleResourceEditor(self._db, self._stacked_widget, "Event", ResourceType.EVENT))
+        if has_resource(ResourceType.TRICK):
+            editors.append(TrickResourceEditor(self._db, self._stacked_widget))
+        if has_resource(ResourceType.DAMAGE):
+            editors.append(CountedResourceEditor(self._db, self._stacked_widget, "Damage", ResourceType.DAMAGE))
+        if has_resource(ResourceType.VERSION):
+            editors.append(SimpleResourceEditor(self._db, self._stacked_widget, "Version", ResourceType.VERSION))
+        if has_resource(ResourceType.MISC):
+            editors.append(SimpleResourceEditor(self._db, self._stacked_widget, "Misc", ResourceType.MISC))
```
duncathan: I'd probably make this a loop:
```py
res_editors = {
    ResourceType.ITEM: CountedResourceEditor,
    ResourceType.EVENT: SimpleResourceEditor,
    # etc
}
for res_type, editor_cls in res_editors.items():
    if has_resource(res_type):
        editors.append(editor_cls(self._db, self._stacked_widget, res_type.name.title(), res_type)
```
Note this would require a change to the signature of `TrickResourceEditor`

randovania/gui/dialog/connections_editor/controller.py
```diff
@@ -0,0 +1,272 @@
...
+        if len(self._db.requirement_template) > 0:
```
duncathan: this can just be `if self._db.requirement_template`. python's built-in containers implement their `__bool__` as this length check and it's the pattern we use throughout the codebase

randovania/gui/dialog/connections_editor/controller.py
```diff
@@ -0,0 +1,272 @@
...
+            self._combo_type.addItem(editor.name(), editor.type_of())
+
+            editor.changed.connect(self._on_editor_field_changed)
+
+    def _show_editor(self, editor: Editor) -> None:
+        self._stacked_widget.setCurrentWidget(editor.widget())
+
+    def _get_editor_for(self, requirement: Requirement) -> Editor | None:
+        if requirement is None:
+            return None
```
duncathan, line 106: if this is a case you support, the signature of the function should accept `Requirement | None`

randovania/gui/dialog/connections_editor/controller.py
```diff
@@ -0,0 +1,272 @@
...
+        else:
+            return self._editors[type(requirement)]
+
+    def on_item_selected(self, index: QModelIndex) -> None:
+        """
+        Switch to the editor for the selected Requirement type and update fields
+        """
+        self._active_item_index = index
+        requirement: Requirement = index.data(ROLE)
+        editor: Editor | None = self._get_editor_for(requirement)
```
duncathan: the type annotation here shouldn't be necessary, since the function return type is annotated. though it's not a problem to have it either; we just typically only use annotations for local variables when it's actually necessary for correct type inference

randovania/gui/dialog/connections_editor/controller.py
```diff
@@ -0,0 +1,272 @@
...
+        elif len(path) > 1:
+            path = path.parent()
+        else:
+            return
+
+        after = requirement_tree.insert_at_path(after_remove, path, active)
+
+        self._push_command(before, after, path, "Move Up")
+
+    def on_shift_down_pressed(self) -> None:
```
duncathan, line 218: there's a lot of duplicated logic between shift up and shift down. surely there's a way to combine the logic here?

randovania/gui/dialog/connections_editor/controller.py
```diff
@@ -0,0 +1,272 @@
...
+        after = changed
+        if isinstance(before, RequirementArrayBase):
+            after = requirement_tree.replace_at_path(before, path, changed)
+
+        self._push_command(before, after, path, "Change Type")
+
+    def _on_editor_field_changed(self, requirement: Requirement) -> None:
+        if self._active_is_array() and isinstance(requirement, RequirementArrayBase):
+            # Preserve array items
+            requirement = type(requirement)(self._active_item_index.data(ROLE).items, requirement.comment)
```
duncathan, line 264: in what cases are `self._active_item_index.data(ROLE).items` and `requirement.items` different?

duncathan (commented):
posting my WIP review as-is without finishing, as I don't expect to be able to get back to this for the time being

randovania/gui/dialog/connections_editor/controller.py
JustinDMS: I tried. They're _just_ different enough (ascend into/out, descend out/into) that attempting to remove duplication wouldn't actually reduce the amount of code and just muddy things up.

randovania/gui/dialog/connections_editor/controller.py
JustinDMS: The `Editor` classes don't store requirements. When a change is detected, a requirement is constructed fresh from its widgets' fields. Therefore the `ArrayEditor` has no knowledge of `items`, and passes up an empty array requirement to be filled.

henriquegemignani (approved):

PR #9397, opened 2026-07-09
Reduce renovate PR frequency

henriquegemignani:


.github/renovate.json
```diff
@@ -8,5 +8,12 @@
     "extends": [
       "schedule:weekends"
     ]
-  }
+  },
+  "packageRules": [
+    {
+      "matchManagers": ["github-actions"],
+      "schedule": ["once a week"],
```
ThanatosGit: I'm pretty sure that's not valid.
It is an example text of what they offer but not a thing you can use literally.

```
Schedule presets¶
Renovate has built-in presets for common schedules, like "once a week", "outside office hours" and so on
```

ThanatosGit (changes requested):
RIP

ThanatosGit:
lmao. I was about to write "RIP my daily morning routine" but fat fingered.

henriquegemignani:
> lmao. I was about to write "RIP my daily morning routine" but fat fingered.

Oops! My intent is more to reduce the PR volume and partly to protect a bit against malicious versions. The later I've updated to add `minimumReleaseAge` that fits better too.

ThanatosGit (approved):

PR #9398, opened 2026-07-10
Abandoned world feature

ThanatosGit:
The current WIP state:
- implemented via client side RemoteConnector
- "abandoned" can be trigger by user action and the user can decide to just abandon or directly attach a remote connector to it
- the remote connector can later be added via the actions menu and it can be done by any session member
- Like for coop, collecting a local item (receiver and provider world are equal) by a bot is also a WorldAction
- The action is irreversible
- unclaimed world are in its own section

Open question in code:
```
        # TODO: Lengthy llm summary for discussion if it has to be changed for now:
        # two bots driving the same world (on different clients) don't fully coordinate.
        # A world's own collected locations come back here as coop_location entries, so both bots
        # share those and never re-collect them. But a location holding *another* world's item is
        # reported to the server with that other world as the receiver, so collecting it only
        # refreshes the receiver's pickups -- never this world's. This bot is therefore never told
        # the sibling bot already took such a location: it stays out of `_collected`, remains a valid
        # victory target, and gets collected again. That redundant collection is harmless because the
        # server de-duplicates it (WorldAction has a (provider, location) unique key, so the second
        # report is dropped and the item isn't granted twice), but the two bots still do wasted work.
```

Some screenshots:

image

image

image

image

image


henriquegemignani:
Our current server has barely no CPU available. My idea was to have some _client_ in the session do run the resolver.

ThanatosGit:
reworked and added some screenshots.

henriquegemignani:
I need to go over again and think a bit more, but I have the following:

- Never reference the GUI as "client". There's some places that say "this client".
- The resolver cap message is also very technical. That could fit as a log message, but I'd keep it much simpler for the GUI.
- I don't like calling it "Bot" in the code. I'm not sure I like it in the GUI, but that one I'm less picky.
- Maybe the world should stay associated with who's running the collect process?
- Ideally the "collect data about this world for abandon" would be a new REST api, since I'm trying to migrate all such APIs to REST.

duncathan:
> Maybe the world should stay associated with who's running the collect process?

I was thinking about this too. Some of the ambiguities get resolved if the world has to be claimed

ThanatosGit:
> * Never reference the GUI as "client". There's some places that say "this client".

Ok.

> * I don't like calling it "Bot" in the code. I'm not sure I like it in the GUI, but that one I'm less picky.

I'm fine with a better name. Even I thought it is best described as "a bot who plays by logic on his own" because that is very easy to understand from an unexperienced user view.

> * Maybe the world should stay associated with who's running the collect process?

That's what I had first but removed it. It is going into the same route which I commented as "Open question": If you allow multiple clients to run it (which is currently the case in this PR), there is no clear user association with the world. That's why I removed it.

> * Ideally the "collect data about this world for abandon" would be a new REST api, since I'm trying to migrate all such APIs to REST.

Ok, will change it.

randovania/game_connection/connector/abandoned_world_remote_connector.py
```diff
@@ -0,0 +1,305 @@
...
+        self._dirty = asyncio.Event()
+        self._dirty.set()
+        self._task = asyncio.create_task(self._resolve_loop())
+
+    @property
+    def game_enum(self) -> RandovaniaGame:
+        return self._preset.game
+
+    def description(self) -> str:
+        return f"Abandoned world bot - {self.game_enum.long_name}: {self._layout_uuid}"
```
henriquegemignani: The layout UUID should not show up in the UI. If anything, this should have Session+World names.

randovania/server/multiplayer/world_api.py
```diff
@@ -113,10 +117,18 @@ def log(msg: str, *args: object) -> None:
...
+        # we must properly collect a nothing item for ourself in an abandoned world for the connector
+        try:
+            WorldAction.create(provider=world, location=pickup_location, session=session, receiver=world)
+        except peewee.IntegrityError:
+            pass
         log("It's nothing.")
         return None
 
-    if pickup_target.player == world.order and not session.allow_coop:
+    if pickup_target.player == world.order and not session.allow_coop and not world.abandoned:
```
henriquegemignani: I'm fine with changing the behavior for the server to always collect every location, even your own. Just filter own locations from the `received-pickups` list when not co-op? Or maybe do that filtering on the client even?

randovania/server/multiplayer/session_admin.py
```diff
@@ -544,6 +549,24 @@ async def _unclaim_world(
...
+    _verify_world_has_session(world, session)
+    await verify_has_admin_or_claimed(sa, sid, world)
+
+    if not session.has_layout_description():
+        raise error.InvalidActionError("Session has no generated game")
+
+    if world.abandoned:
+        raise error.InvalidActionError("World is already abandoned")
+
+    world.abandoned = True
```
henriquegemignani, line 570: Maybe an additional room setting for allowing abandoning?

randovania/game_connection/connector/abandoned_world_remote_connector.py
```diff
@@ -0,0 +1,305 @@
...
+debug connector), this connector drives the world with the resolver. It works in *paced rounds*: each
+round finds one more reachable location, then waits ``ROUND_DELAY_SECONDS`` before the next round.
+Once a round finds nothing new, the connector goes idle and only re-checks when it receives something
+new through ``set_remote_pickups``.
+
+Each connector owns all of its mutable resolution state (collected locations, received pickups,
+world graph and starting state), so many abandoned worlds can be driven concurrently by fully
+independent connectors. Only the immutable game databases are shared, via ``filtered_database``'s
+cache.
+"""
```
henriquegemignani, line 10: I feel this description is too detailed and could use a simplification. But I'll leave that to your discretion.

randovania/game_connection/connector/abandoned_world_remote_connector.py
```diff
@@ -0,0 +1,305 @@
...
+
+    Collecting the already-collected nodes grants their resources (the permanent inventory the player
+    keeps) and marks them collected, so the resolver won't re-collect them.
+    """
+    state = starting_state.copy()
+    for index in collected:
+        node = graph.node_by_pickup_index.get(PickupIndex(index))
+        if node is not None:
+            state = state.collect_resource_node(node, state.damage_state)
+    if received_pickups:
```
henriquegemignani: I'd skip this if

henriquegemignani (changes requested):
What isn't clear: since the abandoned world isn't associated with anyone, is there any indicator that the abandon worker is being ran by someone?

Personally I'm fine with limiting it to only one person being allowed to run the abandon at a time, with the bot state being indicated in the session for all to see, including it's inventory.

randovania/game_connection/connector/abandoned_world_remote_connector.py
ThanatosGit: Ok. The screenshot already shows the session and world name in the second line anyways.

randovania/server/multiplayer/world_api.py
ThanatosGit: I was jumping back and forth if I want to do that.
The nice thing about it, would be that the abandoned world connector can start with all items the player had when they abandoned the world instead of recollecting all local pickups which gave a local item.
The "issue" is that it also means, that if the player collected out of logic pickups, the resolver also starts with out of logic items and is not following the "real" logic.

ThanatosGit:
> What isn't clear: since the abandoned world isn't associated with anyone, is there any indicator that the abandon worker is being ran by someone?

It emits a location and thus showing as In-game.
Though I think I missed setting it back to disconnected but maybe that's already in the common handler for the teardown / finish of a connector. (I don't want to look it up atm on mobile)

> 
> Personally I'm fine with limiting it to only one person being allowed to run the abandon at a time, with the bot state being indicated in the session for all to see, including it's inventory.

Ok, I will have a look on what I have to change. Very likely needs some sort of mechanism like "force kick" if for whatever reason the state got stuck in "someone is connected" without it being true.

ThanatosGit:
> > Personally I'm fine with limiting it to only one person being allowed to run the abandon at a time, with the bot state being indicated in the session for all to see, including it's inventory.
> 
> Ok, I will have a look on what I have to change. Very likely needs some sort of mechanism like "force kick" if for whatever reason the state got stuck in "someone is connected" without it being true.

Was working on it today and it basically gets back to use the WorldUserAssociation again.
I will change it such that an abandoned world must be claimed to use the AbandonedWorldConnector. That way it is ensured that only one can run it. If the one who was running it is offline and then an admin claimed it to continue running it, it will be auto removed from the game connection for the other when they open randovania again.

randovania/server/multiplayer/session_admin.py
ThanatosGit: Is there any reason to deactivate it? For races or something?

henriquegemignani: Mainly that, but I guess it can help people abandoning worlds too eagerly by having to at least ask for it to be enabled?

duncathan: I agree there should be a setting. I wouldn't mind it defaulting to on

randovania/server/multiplayer/world_api.py
ThanatosGit: Do you - or does anyone else - have a preference here?

henriquegemignani: Whatever is simpler I guess. 

I kind wouldn't trust the inventory that much, so maybe from blank state is better?

henriquegemignani: Nevermind, re-read the context more that I'm not half-asleep. 

Cotninuing from the collected locations from the world seems more appropriate.

ThanatosGit:
Following changes:
- Abandon feature is behind a flag for a Sessions (default is true = allowed)
- An abandoned world is only allowed to be played by one:
    - This is achieved by introducing back the WorldUserAssociation
    - If for whatever reason anyone (person 1) was not able to continue running the AbandonedWorldConnector, an admin can unclaim it / claim it for someone (person 2) else who can run the connector then
    - Next time person 1 starts, is connected to the server and it tries to sync, the whole mechanism throws an error, marks the connector as not usable anymore and removes it for them
- Also all locations are now tracked, even the locations for the own world like in coop (Requires an additional changelog)?
- locations for the own world are filtered server side when not coop or not abandoned
image

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [10.10.0] - 2026-08-02
 
+- Added: Multiworld worlds can now be abandoned; abandoning a world is irreversible. An abandoned world can be played automatically by your own Randovania instance like a bot, via a dedicated game connection: it collects one location that is in logic for it every few seconds, and keeps doing so as it receives more items from other players.
```
duncathan: think this should be marked major?

henriquegemignani: for sure

reedwhaley:
One request is that this feature be unavailable for race (no spoiler log) seeds if possible as an anti-cheat measure.

henriquegemignani:
> One request is that this feature be unavailable for race (no spoiler log) seeds if possible as an anti-cheat measure.

That is why you can disable it.

Tyranisaur:
> One request is that this feature be unavailable for race (no spoiler log) seeds if possible as an anti-cheat measure.

I was thinking this as well, but I also wonder if pairing up with an auto release world could be an interesting race format. Maybe it's worthwhile to put flags in the game generation that disables co-op and abandoning for the seed at generataion time.

reedwhaley:
> > One request is that this feature be unavailable for race (no spoiler log) seeds if possible as an anti-cheat measure.
> 
> That is why you can disable it.

Very fair, I didn’t see that setting. I apologize. Fantastic addition to the software!

randovania/server/multiplayer/session_admin.py
```diff
@@ -590,6 +630,16 @@ async def _set_allow_everyone_claim(sa: ServerApp, sid: str, session: Multiplaye
         session.save()
 
 
+async def _set_allow_abandon_worlds(sa: ServerApp, sid: str, session: MultiplayerSession, new_state: bool) -> None:
+    await verify_has_admin(sa, sid, session.id, None)
+
+    with database.db.atomic():
+        session.allow_abandon_worlds = new_state
```
henriquegemignani, line 642: Only allow disabling this setting when there isn't already an abandoned world.

henriquegemignani (changes requested):

henriquegemignani (changes requested):
Missing server tests for:
- admin request for SessionAdminGlobalAction.SET_ALLOW_ABANDON_WORLDS
- clearing the layout when there's an abandoned world
- trying to abandon a world when it's not allowed
- trying to export an abandoned world

henriquegemignani (approved):

PR #9406, opened 2026-07-16
get_abandoned_world_data: handle world not existing

henriquegemignani:


ThanatosGit:
Iirc the 403 this method can throw, has logic that clears the connector at the randovania instance.
It should probably do the same for the 404.

henriquegemignani:
> Iirc the 403 this method can throw, has logic that clears the connector at the randovania instance. It should probably do the same for the 404.

Given the security idea of "not leaking info" kinda want to return 403 to not tell the user which world ids are valid...

ThanatosGit:
Sounds fine. Then change the status to 403 and perhaps just add a comment explaining that a 404 is intentionally not being returned - even though a 404 would actually be more appropriate for a “Not Found” message.

henriquegemignani:
Changed.

ThanatosGit (approved):

PR #9409, opened 2026-07-18
Brighten Dread's Super Missile icon in Auto Tracker

gizmomike:
Updated the color of the Super Missile icon to match Plasma and Speed Boost icons.

SuperMissileRecolor


ThanatosGit (changes requested):
I like the brighter color. 
However, it looks like the new icon has worse quality.
At first, I thought it's maybe because the brighter color stands out more against the background but it is very noticeable when you look at the ring around the icon itself.

gizmomike:
Compare of Original and Rejected:
You can see where changing the levels removed some detail.
SuperMissileRecolor_Original and Rejected

Compare of Rejected and Updated:
Differences between MY old version (expanded) and MY new version (smaller)
SuperMissileRecolor_Rejected and Updated

Compare of Original and Updated:
Current production version and my newest version with the requested changes. (matches original shape better)
SuperMissileRecolor_Original and Updated

ThanatosGit (approved):

PR #9410, opened 2026-07-18
Dread: Better understand when Grav Suit Tower Break Floor is safe

Tyranisaur:
Merge break floor blobs events into the grapple block event. Add connection back to early gravity blob.


Logarhythmic (approved):
The Break Floor event is much easier to read now. Looks good to me.

ThanatosGit (approved):

PR #9415, opened 2026-07-19
Generator: Fix reachable nodes not pathing via lock nodes

Tyranisaur:
Ref https://discord.com/channels/914291389293027329/914294272579227668/1527103293413916823

### DIFF
--- a/test/generator/test_generator_reach.py
+++ b/test/generator/test_generator_reach.py
@@ -364,6 +364,6 @@ def test_graph_module(blank_world_graph, blank_game_description, empty_patches):
     state.node = blank_world_graph.nodes[1]
     costs = g.calculate_reachable_costs(blank_world_graph, state)
     costs_dict = {key: costs[key] for key in costs.keys()}
-    assert costs_dict == {5: 1}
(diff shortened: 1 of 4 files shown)


test/generator/test_generator_reach.py
```diff
@@ -364,6 +364,6 @@ def test_graph_module(blank_world_graph, blank_game_description, empty_patches):
     state.node = blank_world_graph.nodes[1]
     costs = g.calculate_reachable_costs(blank_world_graph, state)
     costs_dict = {key: costs[key] for key in costs.keys()}
-    assert costs_dict == {5: 1}
+    assert costs_dict == {5: 0}
```
henriquegemignani: Would be good to still have a test scenario with case 1

henriquegemignani (changes requested):

test/generator/test_generator_reach.py
Tyranisaur: Added node 3 (missile expansion pickup).

henriquegemignani (approved):

PR #9417, opened 2026-07-20
Planets (Zebeth): Fixed PDF error about pickup quantity

UltiNaruto:
pickup.conditional_resources[0].resources wasn't being checked when defining quantity which resulted in a raised error when doing pickup.conditional_resources[0].resources[0][1]


Miepee (approved):

Miepee:
How come this wasn't caught by any tests?

UltiNaruto:
No idea, I noticed it when I tried to work on the logic for SA breaking blocks

Miepee:
Ideally it would be nice to have a test case for it, but up to other core devs if they feel that thats nececssary.

henriquegemignani:
This kind of corner case if the kind of cases that are good to have in the test files.

UltiNaruto:
Anything else to do before it gets pushed?

Miepee:
Adding the test would be nice.

UltiNaruto:
Added a test for when creating pickups

PR #9420, opened 2026-07-20
Dread: Adjust dangerous actions without Gravity Suit in Burenia and Dairon

Tyranisaur:


CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed: Merged event nodes for breaking the floor in Gravity Suit Tower, decluttering the room and helping the generator see when it is safe to complete the sequence.
 - Changed: Added connection back to the Early Gravity Blob in the Gravity Suit Tower to help the generator see when it is safe to complete the breakable floor sequence.
+- Changed: Obtaining the Pickup underneath Drogyga with a Water Bomb Jump and Normal Bomb is now Intermediate, for consistency with other out of water bomb jumps.
+- Changed: The dangerous logic surrounding the water underneath Drogyga in terms of getting up the left left ledge and across to the right ledge is now waived. Logically by construction, if the water is there then Drogya has been defeated, whic requires Grapple Beam, which is a solutuion to the connections that are changed.
```
Miepee:
```suggestion
- Changed: The dangerous logic surrounding the water underneath Drogyga in terms of getting up the left left ledge and across to the right ledge is now waived. Logically by construction, if the water is there then Drogyga has been defeated, which requires Grapple Beam, which is a solution to the connections that are changed.
```

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed: Merged event nodes for breaking the floor in Gravity Suit Tower, decluttering the room and helping the generator see when it is safe to complete the sequence.
 - Changed: Added connection back to the Early Gravity Blob in the Gravity Suit Tower to help the generator see when it is safe to complete the breakable floor sequence.
+- Changed: Obtaining the Pickup underneath Drogyga with a Water Bomb Jump and Normal Bomb is now Intermediate, for consistency with other out of water bomb jumps.
```
XenoWars: I'm not sure about this change. The water here is low enough that the difficulty from out of water bomb jumps is not there. All other ones you are at the surface of the water while this one you are almost completely exiting the water. Moreover, the trick is mostly flavor since you can do it without setting up a water bomb jump especially for cross bombs.

Tyranisaur: This is the same 3 blocks high body of water that we have elsewhere. I know this is part is not technically a water bomb jump, but the funny timing and placement of the bomb relating to the surface of the water is why this tech is usually rated as an intermediate water bomb jump with normal bombs.

XenoWars: I do admit my initial thought was wrong, but it doesn't change my stance.

The main difference with this one compared to the one in Artaria's Screw Attack room (which has the only other instances of an out-of-water bomb jump) is that with Artaria's you also have to do a wall jump in that short window of time of being above water (Total Recharge > Under Left Side Blocks). The WBJ for the opposite direction in that room is just the normal WBJ template with IBJ beginner + Cross Bombs as an alternative. This one here in Burenia should mimic the latter, not the former.

Out-of-water bomb jumps are not any different from normal water bomb jumps; it's merely the wall jump that makes it a higher difficulty with normal bombs.

Tyranisaur: Alright I rolled this back, because a standard beginner WBJ gives you enough height to reach the ledge.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed: Merged event nodes for breaking the floor in Gravity Suit Tower, decluttering the room and helping the generator see when it is safe to complete the sequence.
 - Changed: Added connection back to the Early Gravity Blob in the Gravity Suit Tower to help the generator see when it is safe to complete the breakable floor sequence.
+- Changed: The dangerous logic surrounding the water underneath Drogyga in terms of getting up the left left ledge and across to the right ledge is now waived. Logically by construction, if the water is there then Drogyga has been defeated, which requires Grapple Beam, which is a solution to the connections that are changed.
```
XenoWars:
```suggestion
- Changed: The dangerous logic surrounding the water underneath Drogyga in terms of getting up the left ledge and across to the right ledge is now waived. Logically by construction, if the water is there then Drogyga has been defeated, which requires Grapple Beam, which is a solution to the connections that are changed.
```

XenoWars: This should be all then it's good.

XenoWars (changes requested):

XenoWars (approved):

Miepee (approved):
approved by db maintainer

PR #9421, opened 2026-07-20
Some Player->World renaming, swap `dict[int, T]` to `list[T]` in other places

henriquegemignani:
Might be better to review by commit? This is mostly cosmetic changes in code.


JeffGainsNGames (approved):
going commit by commit was indeed helpful

PR #9425, opened 2026-07-21
UV exclude packages newer than 7 days

ThanatosGit:
From my understanding, our `renovate.json` creates PR and adds a pending check until the updated packages are older than 7 days.
The problem is that it does that via a `uv lock --upgrade` and this pulls transitive dependencies which are newer.
This is an attempt to keep `uv` via `pyproject.toml` in sync with the renovate setting.
I excluded our patchers because we often update them way earlier.

In the following output, websockets 16.1.1 is now absent, which was one example of a dependcy updated in the renovate PR which is newer than 7 days.
```sh
 uv lock --upgrade --dry-run
Resolved 173 packages in 4.58s
warning: The package `python-socketio==5.16.3` does not have an extra named `asyncio`
Update coverage v7.15.0 -> v7.15.1
Update httpcore2 v2.5.0 -> v2.6.0
Update httpx2 v2.5.0 -> v2.6.0
Update sentry-sdk v2.64.0 -> v2.65.0
Update types-cachetools v7.0.0.20260518 -> v7.0.0.20260713
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -382,6 +382,7 @@ module = [
 ignore_missing_imports = true
 
 [tool.uv]
+exclude-newer = "7 days"
 cache-keys = [
     { git = { commit = true, tags = true } },
     { file = "pyproject.toml" },
@@ -389,3 +390,17 @@ cache-keys = [
     { file = "README.md" },
     { file = "randovania/enable-cython" },
 ]
+
+[tool.uv.exclude-newer-package]
+am2r-yams = false
+cave-story-randomizer = false
+factorio-randovania-mod = false
+open-dread-rando = false
+mars-patcher = false
+planets-yapr = false
+py_randomprime = false
+random-enemy-attributes = false
+retro-data-structures = false
+mp2hudcolor = false
+open-prime-hunters-rando = false
+open-samus-returns-rando = false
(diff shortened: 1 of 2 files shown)


henriquegemignani (approved):

Tyranisaur:
```$ tools/prepare_virtual_env.bat --full --native
warning: Failed to parse `pyproject.toml` during settings discovery:
  TOML parse error at line 385, column 17
      |
  385 | exclude-newer = "7 days"
      |                 ^^^^^^^^
  failed to parse year in date "7 days": failed to parse "7 da" as year (a four digit integer): invalid digit, expected 0-9 but got

warning: Failed to parse `pyproject.toml` during settings discovery:
  TOML parse error at line 385, column 17
      |
  385 | exclude-newer = "7 days"
      |                 ^^^^^^^^
  failed to parse year in date "7 days": failed to parse "7 da" as year (a four digit integer): invalid digit, expected 0-9 but got```

henriquegemignani:
If you try updating to latest`uv` (uv self update) does it work?

ThanatosGit:
Yeah, it requires a kinda newer version of uv.
Relative user friendly durations likes "7 days" were added in December 2025. The excludes were added later.

Tyranisaur:
I tried to look into that last nigh. Yes, updating uv fixes this. I was still on version 0.6.something Before updating, I tried to set a `required-version`, but I'm not sure it did anything until after I updated, and tried to set the next patch version.

PR #9426, opened 2026-07-21
Adjusting resolver structure to support multiworld

henriquegemignani:
The multiworld logic is not correct, but it should be ready for smaller improvements until it's there.


JeffGainsNGames (commented):
reviewed as best I can, no comments on the work itself. 
- since this is a structural thing is there any changes to test coverage we need at this point?
- only one "todo" left, this looks like its for adding the world name to the playthrough tab? definitely seems like a good thing to have in that view
- need to fix merge conflicts

henriquegemignani:
> * since this is a structural thing is there any changes to test coverage we need at this point?

Ideally since this is just a refactor that doesn't introduce new functionality, the existing test cases should be enough but I can take a look at these changes with low coverage.

> * only one "todo" left, this looks like its for adding the world name to the playthrough tab? definitely seems like a good thing to have in that view

I tried to leave TODOs in places where changes would be needed to properly handle multiworlds. The Playthrough tab will probably need some extra thinking to be usable with multiworld.

randovania/generator/generator.py
```diff
@@ -365,8 +369,10 @@ def actual_status_update(msg: str) -> None:
 
     if resolve_after_generation and generator_params.world_count == 1:
         final_state_async = resolver.resolve(
-            configuration=generator_params.get_preset(0).configuration,
-            patches=result.all_patches[0],
+            [
+                (preset.configuration, patches)
+                for preset, patches in zip(generator_params.presets, result.all_patches, strict=True)
+            ],
```
Tyranisaur, line 376: This part is inside an if that guards against `world_count` being anything other than 1. Probably something to fix.

randovania/generator/hint_distributor.py
```diff
@@ -774,7 +774,7 @@ async def get_resolver_hint_state(world: int, patches: GamePatches) -> ResolverH
         return None
     else:
         debug.debug_print(f">> World {world + 1} is solve-able after door placement. Beginning hint placement.")
-        return new_state.hint_state
+        return new_state[0].hint_state
```
Tyranisaur, line 777: There's something weird about this function. Surely if the resolver is being used for hint placement, then it needs to run on the multi as a whole?

randovania/resolver/resolver.py
```diff
@@ -181,131 +182,195 @@ def _progressive_chain_info(node: WorldGraphNode, resources: ResourceCollection)
...
+
+    return None
+
+
 async def _inner_advance_depth(
-    state: State,
     logic: Logic,
+    states: list[State],
+    reaches: list[ResolverReach],
+    last_action: int,
```
Tyranisaur: Maybe use a different name to make it clear that this is the world that had the previous action.

randovania/resolver/resolver.py
```diff
@@ -320,29 +385,36 @@ async def _inner_advance_depth(
 
     if has_action:
         additional_alts: set[GraphRequirementList] = set()
-        for resource_node in reach.collectable_resource_nodes(state.resources):
-            additional_alts |= set(logic.get_additional_requirements(resource_node).alternatives)
+        for resource_node in reaches[last_action].collectable_resource_nodes(states[last_action].resources):
+            additional_alts |= set(logic.get_additional_requirements(last_action, resource_node).alternatives)
```
Tyranisaur: I think you're meant to use additional requirements from all worlds both here and on the line where you start building them with `satisfiable_requirements_for_additionals`.

The collectable resource nodes represents the combination of resources you either tried or skipped, that at this point should have unmet additional requirements. These are from all the worlds.

The story is something along the lines of; You tried to collect the pickup in Storage D without your Dark Beam. It was the Speed Booster for the Dread world. However, at this time, that Speed Booster did not lead to your Dark Beam. In the grand scheme of things, it turns out that the Dread world needs both Speed Booster and Morph Ball to obtain your Dark Beam. You have to find their Morph Ball somewhere else before sending the Speed Booster locking yourself in Storage D. The unmet requirements for both Speed Booster and Morph Ball could either already be visible from the `satisfiable_requirements_for_additionals`, or you might need to find them from trying actions and then retrieve them from `collectable_resource_nodes`.

randovania/generator/generator.py
henriquegemignani: At best a fixme. We don't want to actually use the multiworld resolver right now.

randovania/generator/hint_distributor.py
henriquegemignani: Right now resolver-based hints are only single world compatible. Adjusting it to use the multiworld resolver would be a PR of it's own.

randovania/resolver/resolver.py
henriquegemignani: I do agree it needs to be something more. I thought about having the additional requirements be a list for every world, but then it's used in the resolver during the graph exploration so I wasn't entirely sure how to adjust that.

It for sure needs additional thinking to work properly, so I focused on a solution that keeps single-world working for this PR.

randovania/resolver/resolver.py
```diff
@@ -181,131 +182,197 @@ def _progressive_chain_info(node: WorldGraphNode, resources: ResourceCollection)
...
+                    actions_by_priority[ActionPriority.POINT_OF_NO_RETURN].append(
+                        (world_logic.world_index, action, damage_state)
+                    )
+                    continue
+
+            actions_by_priority[_priority_for_resource_action(action, state, world_logic)].append(
+                (world_logic.world_index, action, damage_state)
+            )
+
+    return None
```
Tyranisaur: Is there a missing return type from this function? I think we lost the early return up the recursive search if you tried a safe action and did not solve the seed.

henriquegemignani: Good catch. I'm using None as a sentinel, but None is a possible return value.

Tyranisaur (approved):

PR #9427, opened 2026-07-22
Add minimum version for UV

henriquegemignani:
I looked for the version that added what we need, but second to last version (right now) has fixes related to exclude-newer so I guess we're going with very very new.

Maybe we add an auto `uv self update` inside of `tools/prepare_virtual_env`.

### DIFF
--- a/.github/workflows/arm-python.yml
+++ b/.github/workflows/arm-python.yml
@@ -27,7 +27,7 @@ jobs:
           fetch-depth: 0
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v9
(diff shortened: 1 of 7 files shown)


ThanatosGit:
Looking at all those red icons: astral-sh/setup-uv@v8.3.2 has to be changed because it request 0.11.28.
Maybe the action itself has to be update to v9.0.0 because that was adding the checksums for 0.11.29 and 0.11.30

henriquegemignani:
Actually we explicitly request `0.11.28` in the workflows. Now, there's some job that was updating that for us but I can't remember what was between pre-commit, renovate and dependabot.

ThanatosGit (approved):

PR #9430, opened 2026-07-24
allow items to be placed in bulk

JeffGainsNGames:
currently doesn't work for generated pickups


randovania/generator/filler/retcon.py
```diff
@@ -318,16 +319,22 @@ def action_report(message: str) -> None:
...
         new_pickups.sort()
-        rng.shuffle(new_pickups)
 
         for new_resource in new_resources:
             debug_print_collect_event(new_resource)
             # This action is potentially dangerous. Use `act_on` to remove invalid paths
             current_player.reach.act_on(new_resource)
 
         if new_pickups:
-            if current_player.configuration.staggered_multi_pickup_placement:
```
henriquegemignani, line 370: Also delete the variable definition

henriquegemignani (changes requested):

randovania/generator/filler/retcon.py
```diff
@@ -318,16 +319,22 @@ def action_report(message: str) -> None:
...
-                new_pickups = [new_pickups[0]]
+            # Group all the like pickups
+            nested = [list(group) for _, group in groupby(new_pickups)]
+            # Shuffle them so we dont always pick the same progression first
+            rng.shuffle(nested)
+            # Check if we can place in batch, else only use one element
+            if nested[0][0].generator_params.batch_placement:
+                new_pickups = nested[0]
+            else:
+                new_pickups = [nested[0][0]]
```
toasterparty: I'm confused. I thought `staggered_multi_pickup_placement` was `true` for nearly all presets. The code path that gets used here when your new field is at it's default value is `new_pickups = [nested[0][0]]` which is different than the replaced `new_pickups = [new_pickups[0]]`.

It looks like you added a tweakable new feature that defaults off, but then also changed default behavior for all generated games?

toasterparty (commented):
At a high level, it looks like your change is complimentary to mine. I pushed back on the generator being over-eager to drop artifacts on ToL from the back end of the progression chain, you push back from the other end: making the generator less likely to be in that situation in the first place.

randovania/generator/filler/retcon.py
JeffGainsNGames: for your first comment, what I did was
1) start with `new_pickups` being a `list[PickupEntry]`
2) create list `nested` a new "grouped" `list[list[PickupEntry]` of similar `PickupEntry`s
3) check the first element in the first group if it's got the `batch_placement` property
  a) if yes, make `new_pickups` the first group from `nested` ie place the entire list
  b) if no, make `new_pickups` only the first element of the first group ie place only one element (staggered)

JeffGainsNGames: the second, I'd need to look into more because yeah I assume if it's default off there *should* be no changes to expected seed hashes but maybe it's because of how I moved the shuffle operation (to group the lists by likeness they need to be sorted still)?

henriquegemignani: The staggered flag should just be a generator setting, always set to true.

randovania/generator/filler/retcon.py
```diff
@@ -318,16 +319,22 @@ def action_report(message: str) -> None:
...
         for new_resource in new_resources:
             debug_print_collect_event(new_resource)
             # This action is potentially dangerous. Use `act_on` to remove invalid paths
             current_player.reach.act_on(new_resource)
 
         if new_pickups:
-            if current_player.configuration.staggered_multi_pickup_placement:
-                new_pickups = [new_pickups[0]]
+            # Group all the like pickups
+            nested = [list(group) for _, group in groupby(new_pickups)]
```
Tyranisaur: To make it work for generated pickups, what you could do is, instead of grouping by pickup identity, you put all batch_placement enabled pickups in the same bucket.

randovania/generator/filler/retcon.py
Tyranisaur: Is it possible that the hashes changed because calling shuffle on a list with 0 or 1 elements advances the seed in the random number generator?

Tyranisaur: No, what's different is that you're shuffling the grouped lists always, and only use the flag for choosing whether to use the whole first bucket or the first pickup in the first bucket.

henriquegemignani (approved):

PR #9431, opened 2026-07-24
Skip location recency bias for "inert" pickups

toasterparty:
I looked at this problem more wholistically than a single item type being over-placed at single location. I recall there being issues with artifact bias in other dead-end locations like PMT and I believe those were addressed with heuristics more than they were addressed with fundamental generator changes.

# Problem

The most popular preset used by the most vocal players of the community is the standard weekly preset. This preset is characterized by Landing Site start, vanilla elevators, near-default item pool and all tricks set to beginner/disabled. A pattern has been noticed by frequently players of this preset (or ones like it), and that pattern represents a valid critique of the randomizer's generation biases, not just merely a heuristic observation.

***Tower of Light places a Chozo Artifact about 30% of the time which is ~4x more than would be expected from a perfectly "fair" generation algorithm of ~8%.***

Tower of Light is a unique location in that it:
- Has the single highest Missile requirement of all progression locations
- Typically unlocks last or near-last in the progression "chain" (next to locations like PMT). Upon first glance it only requires `Missile Launcher + 7 Missile Expansions + Space Jump Boots + Wave Beam` to acquire. However, consider the fact that this is requires at *minimum* 10 item locations accessible to place those items. The game actually requires the vast majority of: `Morph Ball, Bombs + Varia Suit + Ice Beam + Power Bombs` to guarantee the remaining ToL requirements can actually fit in the preceding "sphere".

Artifacts are unique in that they are as close as possible to being useless while still being required for progression. In other words, they help the player to beat the game and nothing else. In this PR, I refer to this quality as "inertness".

Here is are visualizations which show placement considerations and weighting made by the generator. First here is a seed which shows a single seed being generated where an artifact ends up in Tower of Light:

seed2007_before_a0

Here is another which shows the average across 300 seeds:

AVG_before_a0

# Diagnosis

The generator implementation has the function `calculate_uncollected_index_weights` which calculates weights for each of the uncollected pickups. These weights are used to skew the probability of that location having a pickup placed at it next:

```py
for indices in indices_groups:
    weight_from_collected_indices = math.sqrt(len(indices) / ((1 + len(assigned_indices & indices)) ** 2))

    for index in sorted(uncollected_indices & indices):
        weight_from_index_age = min(10.0, index_age[index] + 1.0) ** -2
        result[index] = weight_from_collected_indices * weight_from_index_age

return result
```

These weights are used to define an intentional deviation from a pure "fair" generation algorithm:

***On average, prefer to place pickups in locations which are "deeper" into the game's progression logic.***

This bias is mostly warranted. For Prime, this weighting is directly responsible for the prevention of Chozo Ruins being backfilled with the majority of progression items simply because it contains the majority of pickup locations.

There is an oversight, however. This preference works in our favor when it places items which are part of the prototypical progression chain: beams, ball upgrades, suits, etc. Critically - ***This preference works against us for inert items like Artifacts. Biasing the placement of an item which does not unlock another serves no purpose other than to reduce seed variety.***

# Solution

Keep the recency bias, but readjust the weights such that the recency term is nullified when considering inert pickups. Define inert pickups as ones which do not unlock any additional if collected.

### Rejected Solutions

- Scaling the recency bias instead of treating as a boolean. I found this comment which describes something I saw evidence of in my experimentation:
```py
# this used to weigh actions according to *how many* resources were unlocked, but we've determined that the results are more fun if we only care about something being unlocked at all
```
- Implement a floor on the weights assigned by `calculate_uncollected_index_weights`. Adding a `+ 0.08` term to the weight calculation solved this problem, but that solution does not scale to other presets or games.
- Reduce the missile count required to access ToL. Most tempting, but hard to do in a way that doesn't require extra player education or feel like a cheap hack.
- Increase missile count requirements for other standard preset locations. This would definitely help - I just don't if it makes sense to do.

# Caveat

I think setting `index_age_impact = 0.0` for artifacts would have handled this for just Prime1. I didn't see this knob until after I finished my implementation. My implementation is generic and makes for better default behavior for new games. My implementation also increases computational complexity of seed generation. This is the tradeoff.

# Validation

- Generate 300 seeds on `main`
- Re-generate those same 300 seeds on this PR branch
- Calculate placement probabilities for each location and calculate the before/after delta
- Upload as [Google Sheet](https://docs.google.com/spreadsheets/d/1u26IafxPXuKb_3G_36vF7RsO8HGfk0UaH9pljSnVSoA/edit?usp=sharing)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Major** - Added: Multiworld worlds can now be abandoned; abandoning a world is irreversible. An abandoned world can be played automatically by your own Randovania instance like a bot, via a dedicated game connection: it collects one location that is in logic for it every few seconds, and keeps doing so as it receives more items from other players.
 - Added: You can now disable game connections, preventing them from connecting without having to be removed entirely.
+- Changed: Improved average location distribution of "inert" pickups such as Prime Artifacts and Echoes STKs.
 - Changed: Automatic Item Tracker: The selected theme for the default game is now always shown while the selected game connection is not connected.
 - Changed: The "History" tab in the "Multiworld session" window now displays pickups where the provider and receiver are the same world, even in non-coop sessions.
 - Fixed: Automatic Item Tracker: The text "Not currently connected to any games" is now always shown when the selected game connection is not connected and no default game is set.
--- a/randovania/games/am2r/game_data.py
+++ b/randovania/games/am2r/game_data.py
@@ -91,7 +91,7 @@ def _test_data() -> randovania.game.game_test_data.GameTestData:
     from randovania.layout.base.trick_level import LayoutTrickLevel
 
     return randovania.game.game_test_data.GameTestData(
-        expected_seed_hash="N7MNABZW",
+        expected_seed_hash="W5LKIYAB",
         # Some items require shinesparking to reach in vanilla,
         # which due to varying difficulty has been made into a trick
         database_collectable_include_tricks=(("Shinesparking", LayoutTrickLevel.ADVANCED),),
--- a/randovania/games/blank/game_data.py
+++ b/randovania/games/blank/game_data.py
@@ -87,7 +87,7 @@ def _hash_words() -> list[str]:
 
 def _test_data() -> randovania.game.game_test_data.GameTestData:
     return randovania.game.game_test_data.GameTestData(
-        expected_seed_hash="BURBIEUO",
+        expected_seed_hash="KX665YWG",
     )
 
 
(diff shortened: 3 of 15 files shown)


Miepee:
> I think setting index_age_impact = 0.0 for artifacts would have handled this for just Prime1.

I remember experimenting with this in the past, but thought that I didn't get any good results with it? Has been a long while since then, so I don't remember it all that well tbh. But you not seeing it maybe also means that we should probably start having documentation for all the "generation knobs"

henriquegemignani (commented):
I have a feeling that there's a way to avoid doing that reach calculation. Mostly because it'll be done after assigning the pickup, so maybe delay it until then

toasterparty:
@henriquegemignani

> I have a feeling that there's a way to avoid doing that reach calculation. Mostly because it'll be done after assigning the pickup, so maybe delay it until then

Would you accept a cached answer solution?

toasterparty:
@henriquegemignani fixed

| commit | average seed time (n=10) |
| --- | --- |
| main | 12.44s |
| 8fba628859efc80f22685a49a0b6a4437bb23e63 | 14.03s |
| ceef18a3241d66be4f84b03ff0371ee11d37b672 | 12.47s |

henriquegemignani (approved):

PR #9433, opened 2026-07-25
Generator/fix front of dock with forced two way lock

Tyranisaur:


randovania/graph/world_graph_factory.py
```diff
@@ -273,6 +267,22 @@ def _should_create_front_node(database_view: GameDatabaseView, original_node: Do
...
+    """
+    Decide if we should wrap the dock node with an extra node.
+    Important since crossing ResourceNodes can be problematic in the generator.
+    """
+    area = database_view.area_from_node(original_node)
+
+    may_have_lock = _can_dock_have_lock(database_view, original_node)
+
+    if not may_have_lock:
+        other_dock = database_view.node_by_identifier(original_node.default_connection)
```
henriquegemignani, line 283: If the correct answer depends on the connection, checking only the default value isn't correct. And this would mean the answer depends on every possible value so it's kinda always true then?

test/generator/test_generator_reach.py
```diff
@@ -340,14 +340,11 @@ def test_graph_module(blank_world_graph, blank_game_description, empty_patches):
     g = GeneratorDiGraph.new(blank_world_graph)
 
     g.add_node(1)
-    g.add_node(3)
```
henriquegemignani, line 343: Kinda my fault for not describing well the test originally, but I don't have a clue what the change is doing to this test case

randovania/graph/world_graph_factory.py
Tyranisaur: I'm not exactly sure what you mean here. The point is that some docks can be randomized to get locks when the dock normally is a type that can't be randomized, but the other side of the dock can. Cases like sensor locks with open passage on the other side, or access closed/access locked.

test/generator/test_generator_reach.py
Tyranisaur: This is the revert of the change I did previously. The point was originally to show the difference between a pickup node and a dock node in terms of whether you are allowed to path through them before collecting them.

I was originally trying to fix a bug in that change, where I also created some new nodes to even better show this difference. But as a result I inadvertently created a "Front of" node that currently doesn't get created for this node. That is when I realized that fixing the creation of the "Front of" nodes is a better resolution, because I believe the first bug is exclusive to randomized doors and these funny docks that are affected by forced two way.

PR #9442, opened 2026-07-29
Randomprime 1.33.0

toasterparty:
*Largest prime patcher update in years. Many features with high regression risk so this (ideally) should go in shortly after the August release followed by a Beta Tester role ping.*

# Prerequisites

- #9443
- ~~#9446~~ (*recommended but no longer required*)

# Change Summary

- Update py-randomprime from `v1.31.3` to `v1.32.0`: [py-randomprime release notes](https://github.com/randovania/py-randomprime/releases/tag/v1.32.0)
- Game-runtime stability improvements
- Overhaul visorless and beamless behavior + logic
- Overhaul suit color customization features + GUI
- QoL patches which utilize save-slot data: "location-based completion %", "named saves" and "deny loading wrong save"

# Database Reviewers

If you are only here to review database changes, here's what you need to know:
- Commit [4bc39ae](https://github.com/toasterparty/randovania/commit/4bc39ae974971a087a4fd9ef135ccfc494806221) renames several templates and replaces `Shoot X Beam` templates with just the item (`X Beam`).
- Commits [c186bff..6fd9a62](https://github.com/toasterparty/randovania/compare/4bc39ae974971a087a4fd9ef135ccfc494806221...6fd9a6250f4b04f8dd9ae26f98c89d17eeeb81f8) contain the actually pertinent logic changes which need reviewing.

# Media

## Cosmetic Options
Screenshot_1
Screenshot_2

## Named Saves
image

## Test Seed

The seed I played a few minutes of for validation. All ticks intermediate, start with no visors/beams, only Morph + Bombs.

Screenshot_4

# Follow-up

Should make some better preview assets for the new suit customization GUI (remove background, etc.)


CHANGELOG.md
```diff
@@ -5,6 +5,39 @@ All notable changes to this project will be documented in this file.
...
+
+- Fixed: Less vulnerable to "Alloc Failed" crashes.
+- Fixed: Transport to Magmoor Caverns North: Longer load time when approached from Sun Tower Access.
+- Fixed: Research Lab Hydra: Battle music behaving differently than vanilla.
+- Fixed: In some minimal inventory starts, players could use Power Beam and Scan Visor out of logic and experience other unintuitive behavior.
+- Fixed: Gravity Suit HD model looking more blue/purple than the low-poly counterpart when color customized.
+- Fixed: Power Suit and Varia Suit color customization both affecting shared chest-piece texture.
+- Fixed: Ruined Shrine: Large beetle does not aggro player until cutscene duration elapses regardless the cutscene was skipped.
+- Fixed: Life Grove: Boostless spinner manipulation trick does not work if pillar raised cutscene skipped.
+- Added: Replaced vanilla's inventory-based completion percent calculation with an algorithm based on location count. Completion percent now always reflects the progress towards collection of all non-nothing pickups in the player's local game.
```
henriquegemignani: Maybe it should just all pickups, instead of the non-nothing?

Also "world" instead of "local game"

randovania/games/prime1/exporter/patch_data_factory.py
```diff
@@ -214,6 +216,47 @@ def _create_results_screen_text(description: LayoutDescription) -> str:
     return f"{randovania.VERSION} | Seed Hash - {description.shareable_word_hash} ({description.shareable_hash})"
 
 
+def _join_shortened_words(words: list[str], lengths: list[int]) -> str:
+    result = ""
+    for i, (word, length) in enumerate(zip(words, lengths)):
```
henriquegemignani: strict=True?

randovania/games/prime1/exporter/patch_data_factory.py
```diff
@@ -214,6 +216,47 @@ def _create_results_screen_text(description: LayoutDescription) -> str:
...
+    lengths = [len(word) for word in words]
+    if len(_join_shortened_words(words, lengths)) <= max_length:
+        return _join_shortened_words(words, lengths)
+
+    cap = max(lengths)
+    while cap > 1 and len(_join_shortened_words(words, [cap] * len(words))) > max_length:
+        cap -= 1
+    lengths = [min(cap, length) for length in lengths]
+
+    grew = True
```
henriquegemignani: Maybe a comment of what this while block is doing, and another for the one above?

henriquegemignani (changes requested):

randovania/games/prime1/exporter/patch_data_factory.py
```diff
@@ -1097,7 +1098,11 @@ def get_random_size(minimum: float, maximum: float) -> float:
                 self.configuration.enemy_attributes.as_json if self.configuration.enemy_attributes is not None else None
             ),
             "uuid": list(
-                self.worlds_config.get_own_uuid().bytes,
+                (
+                    self.worlds_config.get_own_uuid()
+                    if self.worlds_config.is_multiworld
+                    else self.description.seed_uuid
```
Miepee: won't this cause issues in multis? i thought there was code that did something along the lines of "if uuid is not "0000-1111-0000" (or whatever our default empty uuid is) then assume its a solo world and don't bother checking the clients joined game sessions for a world that matches the id"

randovania/games/am2r/exporter/patch_data_factory.py
```diff
@@ -272,7 +272,7 @@ def _create_hash_dict(self, rdv_meta: PatcherDataMeta) -> dict:
             "contains_spoiler": not rdv_meta["in_race_setting"],
             "word_hash": self.description.shareable_word_hash,
             "hash": self.description.shareable_hash,
-            "session_uuid": str(self.worlds_config.get_own_uuid()),
+            "session_uuid": str(self.world_uuid),
```
Miepee, line 275: not a fan of a PR named "bump the randomprime version" causing the expected output of the PDF to get updated for all other games.

randovania/games/prime1/logic_database/Phazon Mines.txt
```diff
@@ -237,6 +237,8 @@ Extra - aabb: [131.38985, -30.599165, 12.26794, 156.20804, 7.0214443, 21.498444]
   * Extra - world_position: [151.119934, -6.224174, 17.407558]
   > Door to Main Quarry
       After Phazon Mines Save Station A Barrier
+  > Event - Save Station A Barrier Opened
```
Miepee, line 240: is this worth a DB comment to indicate the change compared to vanilla?

randovania/games/prime1/logic_database/header.txt
```diff
@@ -1,68 +1,57 @@
...
-      All of the following:
-          Can Use Arm Cannon
-          Ice Beam or Plasma Beam or Power Beam or Wave Beam
-
-* Shoot Power Beam:
-      Power Beam and Can Use Arm Cannon
+* Any Visor:
+      Combat Visor or Scan Visor or Thermal Visor or X-Ray Visor
 
-* Shoot Wave Beam:
```
Miepee, line 28: I don't understand, why are the templates removed? From my understanding, you can't just shoot wave beam, you still need *some* visor. E.g. if you do want to deal with a scatter bombu by shooting it dead, you cant just kill it with wave and scan.

randovania/games/prime1/logic_database/header.txt
```diff
@@ -1,68 +1,57 @@
...
-      Wave Beam and Can Use Arm Cannon
+* Any Beam:
```
Miepee: I prefer templates having a verb at the beginning as it makes them more distinct. Its also more in line how we use templates in other games.

randovania/games/prime1/logic_database/header.txt
```diff
@@ -14,6 +14,44 @@ Templates
 * Any Visor:
       Combat Visor or Scan Visor or Thermal Visor or X-Ray Visor
 
+* Fight (Trivial):
```
Miepee, line 17: I get why, but I'm not really a fan of having trick names with parantheses in template names :/

randovania/games/prime1/logic_database/header.txt
```diff
@@ -14,6 +14,44 @@ Templates
...
+      All of the following:
```
Miepee, line 18: Can you write comments or something indicating where all of these *should* be used? Theres probably a decent amount you can generalize, but I feel like sometimes there are outliers where these shouldnt be used instead (e.g. killing scarabs so you dont take damage from them doesnt sound like it needs combat intermediate in order to be able to lock on)

randovania/games/prime1/logic_database/header.txt
```diff
@@ -14,6 +14,44 @@ Templates
...
+          Any Beam
+          Any of the following:
+              Combat (Ludicrous)
+              Combat (Expert) and Can Target Enemy
+
+* Fight (Ludicrous):
+      All of the following:
+          Any Beam
+          Combat (Ludicrous) and Can Target Enemy
```
Miepee: can it really be generalized like this? i would intuitively assume that in most situations its perfectly adequate for ludicrous to not be able to lock on.

CHANGELOG.md
Miepee: > Maybe it should just all pickups, instead of the non-nothing?

Agree with this. Makes it easy at the end screen to see "i definitely did collect every location", whereas with the non-nothing behaviour you need to calculate how many of your collected locations were nothings/for other players.

This would be very confusing in coop seeds for example.

CHANGELOG.md
```diff
@@ -5,6 +5,37 @@ All notable changes to this project will be documented in this file.
...
+- Added: Power Suit and Varia Suit color customization now affect the textures of Samus' gunship.
```
Miepee: power **and** varia?

Miepee (commented):
gave this a quick commit-by-commit look. kinda wish that the cosmetic changes and `fight (x)`-changes would be split up into other PRs for easier reviewing, but that'd probably be really annoying conflict wise.

CHANGELOG.md
```diff
@@ -5,6 +5,39 @@ All notable changes to this project will be documented in this file.
...
+- Added: Rainbow Ball cosmetic option.
+- Changed: Organized cosmetic options into collapsable sections, suit customization now features preview images and Fusion Suit color preferences are now saved independently from non-Fusion.
+- Changed: Morph Ball now always renders the "inner glow" color appropriate for the current highest suit instead of the vanilla behavior which renders the "wrong" color when Spider, Gravity or Phazon are collected out of the vanilla-intended order.
+- Changed: It is now impossible for players to loading save slots which belong to another rdvgame or multiworld slot.
```
MrMiguel211: to load*

CHANGELOG.md
```diff
@@ -5,6 +5,39 @@ All notable changes to this project will be documented in this file.
...
+### Metroid Prime
+
+- Fixed: Less vulnerable to "Alloc Failed" crashes.
+- Fixed: Transport to Magmoor Caverns North: Longer load time when approached from Sun Tower Access.
+- Fixed: Research Lab Hydra: Battle music behaving differently than vanilla.
+- Fixed: In some minimal inventory starts, players could use Power Beam and Scan Visor out of logic and experience other unintuitive behavior.
+- Fixed: Gravity Suit HD model looking more blue/purple than the low-poly counterpart when color customized.
+- Fixed: Power Suit and Varia Suit color customization both affecting shared chest-piece texture.
+- Fixed: Ruined Shrine: Large beetle does not aggro player until cutscene duration elapses regardless the cutscene was skipped.
+- Fixed: Life Grove: Boostless spinner manipulation trick does not work if pillar raised cutscene skipped.
```
MrMiguel211: Better to say "not working" in past tense here instead of "does not work"

MrMiguel211 (changes requested):

randovania/games/prime1/logic_database/Phazon Mines.txt
JeffGainsNGames: Save Station Mines A missing `spawn point` attribute

randovania/games/prime1/logic_database/header.txt
Uncle-Reggie: I see no issues with this.

randovania/games/prime1/logic_database/header.txt
Uncle-Reggie: I've always found this annoying and was happy to see the leading verbs go. You're right that including them matches other game behavior, and now I'm realizing I'm annoyed at all the other DBs.

Miepee: Having them phrased like that makes it very easy to intuitively see "this is a template".

CHANGELOG.md
Uncle-Reggie: I prefer keeping the trailing noun. If it must be changed "Power and Varia Suits..."

Miepee: Thats not what i was getting at.
I was confused at the fact that varia suit is mentioned at all. the GUI and randomprime change only lists power suit. So either this is a typo, or other places have typos.

randovania/games/prime1/logic_database/header.txt
duncathan: this is a case where we really would rather have parameterized templates

randovania/games/prime1/logic_database/header.txt
Uncle-Reggie: There should be an accompanying Fight req in rooms with them signifying the ability to attack them with said beams/valid visor combinations.

Uncle-Reggie (approved):
DB bits looks good. Nothing stands out. The Fight reqs were something I'd forgotten about and seeing their prevalence reminded me they existed. Definitely a big change so I'd love to have a second DB maintainer review.

CHANGELOG.md
Uncle-Reggie: Imagine being a clown and being in editor mode and not context mode. Couldn't be me.

CHANGELOG.md
toasterparty: Can you elaborate on the coop use case? I'm not too familiar.

The idea behind excluding nothings is that "I will check (scout) every location" is a more pragmatic goal for players than "I will collect every location". The nothing model is unique and therefore an uncontested signal to the player that there will be zero consequences if left uncollected. This currently includes zero consequence to completion percent, but your suggestion would change that.

CHANGELOG.md
toasterparty: I made the logic for this somewhat convoluted and am starting to regret it.

If Power Suit is left at default, then the ship is recolored by the Varia Suit value, otherwise it is recolored by Power Suit value.

It should have just been tied to Power Suit full stop.

CHANGELOG.md
duncathan: > The nothing model is unique and therefore an uncontested signal to the player that there will be zero consequences if left uncollected. 

What about when item models are hidden?

Miepee: > Can you elaborate

In coop, all pickups are remote pickups, even if theyre for your world. Meaning the percentage will always stay at 0%. You could theoretically change this in the prime connector, where if the pickup ks sent from the same world that receives jr, that we also send memops to increase the percentage.

But that wouldnt fix the issue for highish world multis being primarily composed of remote pickups. For multis especially, its much more useful to know "did i collect every location" because thats way easier to track than "did i collect and remember every local and remote pickup"

For solo purposes, its also just kinda weird to see a non-100% number for collecting everything. In MSR we got frequentish reports about exactly that.  And while the cause of that was slightly different, I'm fairly certain we still wouldve received those reports if we had excluded nothings(/remotes)

toasterparty: > Meaning the percentage will always stay at 0%

This is the misunderstanding. A non-nothing pickup includes remote pickups which are "Unknown Item 1". Non-nothing means the pickup type passed to randovania is literally `Nothing`, not that it has no affect on the local world.

This is how randovania places a nothing in multiworld:
```json
"Storage Depot A": {
    "pickups": [
        {
            "currIncrease": 0,
            "hudmemoText": "Nothing acquired!",
            "maxIncrease": 0,
            "respawn": false,
            "scanText": "Your Nothing.",
            "showIcon": true,
            "type": "Nothing",
            "model": "Nothing"
        }
    ]
```

This is a remote item:
```json
"Chapel of the Elders": {
    "pickups": [
        {
            "currIncrease": 42,
            "hudmemoText": "Sent Missile Expansion to Player 2!",
            "maxIncrease": 42,
            "respawn": false,
            "scanText": "Player 2's Missile Expansion. Provides 5 Missiles.",
            "showIcon": true,
            "type": "Unknown Item 1",
            "model": "Missile"
        }
    ]
```

Miepee: Ah good to know.
In which case it does seem to me even stranger that nothings arent counted.

toasterparty: > What about when item models are hidden?

The only way to check/scout an item is to collect it, but the "pragmatic goal" is still valid.

I still prefer to exclude nothings in this scenario for consistency with how the rest of the game behaves.

Think of a large, long ongoing async multi.

If I see a 100% on my save file, it means there is no possible way that I am still blocking someone's game progression. If I see 99%, it means a single item for myself or another is still uncollected.

Including Nothing items in that gap between the current % and 100% is just noise that people have to filter out. I don't want people going back into games just to collect nothings, especially since they can't be tracked in-game (e.g. no `X out of N Nothing Collected` screen in the pause menu).

Miepee: Oh, i was under the impression that the dividor for a normal preset was always 100, not (everything except nothings)

duncathan: Does Prime currently report all collected locations in multiworld, or only offworld locations? I don't recall off the top of my head. I normally use the "collected locations" number in the multiworld window as my source of truth for whether there's anything left in my world for other games. It's not out of the question for those to diverge, of course - that happens any time the player soft resets after sending an item.

duncathan: Another thing to consider: this does mean that collecting a non-nothing item provides the player information about exactly how many nothings are in their world. Inconsequential for single player, of course. Is that desirable for you in multiworld?

randovania/games/prime1/logic_database/header.txt
Samuel6710: I agree with Miepee here, 'Any Beam' could be non obvious at a glance whereas something like 'Have Any Beam' or 'Shoot Any Beam' removes any room for confusion.

toasterparty: The extra specificity makes it sound like you can have it without shooting it, but in the case of beams and visors, having always allows you to shoot/use.

It would be like having a resource or template for `Use Morph Ball`. Morph Ball is already a single item whose ability to use is self-evident and not conditional. The template isn't `Have Any Heat-Resisting Suit`.

If a templates need to be uniquely identifiable in the editor, style the text differently or something.

I will not die on the hill so I will change my mind if we get a tiebreaker opinion. It's 2v2 as I read the situation currently.

CHANGELOG.md
toasterparty: > Does Prime currently report all collected locations in multiworld, or only offworld locations?

Reports offworld only.

> This does mean that collecting a non-nothing item provides the player information about exactly how many nothings are in their world.

I thought you only get your own nothings? I thought if you removed an item from the shuffle pool it always results in one additional dead check for you.

randovania/games/prime1/exporter/patch_data_factory.py
toasterparty: Fixed in #9443, a stated prequisite to this PR.

CHANGELOG.md
henriquegemignani: They read as if they're your own nothing, but you can have your world be full of pickups even if your pool is half empty. (just unlikely)

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1382,9 +1384,9 @@ Extra - aabb: [279.84372, -45.27577, 55.823124, 300.77795, 16.10456, 70.44933]
   > Door to Sunchamber Access
       Any of the following:
           # Deal with Weed and Vines
-          Shoot Any Beam
-          Combat (Beginner) and Normal Damage ≥ 32
-          Combat (Expert) and Normal Damage ≥ 9
+          Fight (Trivial)
+          Normal Damage ≥ 32 and Fight (Beginner)
+          Normal Damage ≥ 9 and Fight (Expert)
```
Tyranisaur: This Fight (Expert) branch seems redundant, since the Fight (Trivial) branch with no damage cost would be satisfied..

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1396,9 +1398,9 @@ Extra - aabb: [279.84372, -45.27577, 55.823124, 300.77795, 16.10456, 70.44933]
   > Door to Arboretum
       Any of the following:
           # Deal with Weed and Vines
-          Shoot Any Beam
-          Combat (Beginner) and Normal Damage ≥ 26
-          Combat (Expert) and Normal Damage ≥ 6
+          Fight (Trivial)
+          Normal Damage ≥ 26 and Fight (Beginner)
+          Normal Damage ≥ 6 and Fight (Expert)
```
Tyranisaur: Same here.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1670,9 +1672,9 @@ Extra - aabb: [275.33786, 16.10455, 59.97493, 293.06842, 76.59607, 79.606064]
   > Door to Sunchamber Lobby
       Any of the following:
           # Dealing with the Weed and Vines
-          Shoot Any Beam
-          Combat (Beginner) and Normal Damage ≥ 26
-          Combat (Advanced) and Normal Damage ≥ 2
+          Fight (Trivial)
+          Normal Damage ≥ 26 and Fight (Beginner)
+          Normal Damage ≥ 2 and Fight (Advanced)
```
Tyranisaur: Same here.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1786,7 +1788,7 @@ Extra - aabb: [-22.001799, 235.39693, -1.5920587, 41.40501, 292.60748, 11.864963
   * Extra - world_rotation: [0.01903269, 0.12722364, -89.99848]
   > Door to Hive Totem
       # Deal with Plazmites
-      Morph Ball or Combat (Intermediate) or Shoot Any Beam
+      Morph Ball or Fight (Intermediate) or Fight (Trivial)
```
Tyranisaur: Fight (Intermediate) or Fight (Trivial) Seems like redundant combination, because the intermediate one is never satisfied in any cases that the trivial one is not.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -1799,7 +1801,7 @@ Extra - aabb: [-22.001799, 235.39693, -1.5920587, 41.40501, 292.60748, 11.864963
   * Extra - world_rotation: [-8.872797, -0.0, 0.0]
   > Door to Ruined Gallery
       # Deal with Plazmites
-      Morph Ball or Combat (Intermediate) or Shoot Any Beam
+      Morph Ball or Fight (Intermediate) or Fight (Trivial)
```
Tyranisaur: Case 2

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -2291,35 +2296,35 @@ Extra - aabb: [397.9558, -8.934963, 4.7323303, 483.05167, 117.17189, 59.302334]
...
+              Fight (Trivial)
               All of the following:
                   # Tank Blastcap and get back by jumping through water
-                  Combat (Beginner) and Normal Damage ≥ 13
+                  Normal Damage ≥ 13 and Fight (Beginner)
                   After Flaahgra or Poisoned Water Damage
               All of the following:
                   # Destroy Blastcap with bombs and get back by jumping through water
-                  Combat (Beginner)
+                  Fight (Beginner)
```
Tyranisaur: These fight templates are redundant when there's a trivial one on another branch of any of.

randovania/games/prime1/logic_database/Chozo Ruins.txt
```diff
@@ -2334,8 +2339,8 @@ Extra - aabb: [397.9558, -8.934963, 4.7323303, 483.05167, 117.17189, 59.302334]
   > Behind Gate
       Any of the following:
           # Deal with Eyons
-          Shoot Any Beam
-          Combat (Beginner) and Normal Damage ≥ 9
+          Fight (Trivial)
+          Normal Damage ≥ 9 and Fight (Beginner)
```
Tyranisaur: Redundant

randovania/games/prime1/logic_database/Frigate Orpheon.txt
```diff
@@ -708,9 +708,9 @@ Extra - aabb: [139.20834, -377.1841, -156.8139, 225.37779, -271.66705, -19.18690
   * Extra - nonstandard: False
   > Event - Parasite Queen
       Any of the following:
-          Shoot Any Beam
+          Fight (Trivial)
           All of the following:
-              Morph Ball and Power Bomb ≥ 4 and Combat (Advanced) and Can Use Power Bombs
+              Morph Ball and Power Bomb ≥ 4 and Can Use Power Bombs and Fight (Advanced)
```
Tyranisaur: Redundant

randovania/games/prime1/logic_database/header.txt
```diff
@@ -1,111 +1,135 @@
...
-* Shoot Any Beam:
+* Any Visor:
+      Combat Visor or Scan Visor or Thermal Visor or X-Ray Visor
+
+* Fight (Trivial):
       All of the following:
-          Can Use Arm Cannon
-          Ice Beam or Plasma Beam or Power Beam or Wave Beam
+          Any Beam
+          Combat (Intermediate) or Can Target Enemy
```
Tyranisaur: I'm going to point out that it's inconsistent for the trivial template to use intermediate for the branch without targeting, since all the others use only one trick level higher than the name of the template. However, it looks like the usage is a mixed bag, where you're also accounting for this difference.

toasterparty:
Identified a regression: When clicking the `Save as Database` button, the application freezes indefinitely.

Something about rooms with a lot of `Fight (X)` templates causing `find_database_errors...check_for_resources_to_use_together` to calculate exponentially too many things.

Will have to come back to this later in the week to optimize.

randovania/games/prime1/exporter/patch_data_factory.py
toasterparty: Refactored these functions to be more readable and avoid `while` loops with tricky exit conditions. Let me know if it's still confusing and I'll drop a comment.

CHANGELOG.md
Tyranisaur: "Your Nothing" is a misnomer. Nothings don't have owners. They can end up in any world in the multi. They will always be placed during backfill.

randovania/games/prime1/logic_database/header.txt
duncathan: I do generally prefer it when template names are semantic, describing an *action* rather than just a set of resources. I find it makes reading unfamiliar logic a lot easier, since I don't need to make as many assumptions about what a specific requirement involves.

An excellent example of this imo is Dread's `Shoot Beam` template, which is a template for a trivial requirement! I think the semantics of indicating the action are just that valuable that it's worth using a template for Trivial lol

duncathan: To be clear I don't feel super strongly about this and wouldn't block it myself, but since toaster was looking for preferences I figured I'd mention mine

CHANGELOG.md
toasterparty: To recap:

There was a misunderstanding about the calculation. The denominator is the number of non-nothing items placed in the world. The numerator is the count of which of those items have been collected so far. "Nothing item" here is defined are the one randovania writes as "Your Nothing" in the scan text/hudmemo and use the glitchy metroid model.

There is a very obscure counting technique you can do to calculate an estimation of the count of nothings are in your game in a competitive context. It requires the player to save-and-quit to check their completion percentage after every item to see when the percentage integer increments by 2 instead of 1, and then cross-referencing that with an integer floor division table. I aknowledge this, but dismiss it because of how obscure and inconsequential it is in a real competitive context. In my opinion, it does not outweigh the convenience a non-nothing completion percent provides casual multiworld play.

randovania/games/prime1/logic_database/header.txt
toasterparty: Renamed templates to start with verbs

randovania/games/prime1/logic_database/header.txt
toasterparty: Out of scope for this PR

randovania/games/prime1/logic_database/header.txt
toasterparty: I updated so fights are never considered impossible just because they are made harder at ludicrous without the ability to combat lock-on.

This change made sense because previously it was making stuff like the low-missile wavebuster flaahgra fight out of logic without the ability to target flaahgra.

CHANGELOG.md
duncathan: Fine by mii. Just wanted to be sure it was considered :+1:

MrMiguel211 (approved):

CHANGELOG.md
toasterparty: Will address this discrepancy with the next patcher release.

henriquegemignani (approved):
I have not checked the database changes.

toasterparty:
Note that there is a known visual bug with the patcher:
https://github.com/randovania/randomprime/issues/176

I plan on making a new py-randomprime release by the end of the week to address this and a few other minor issues.

toasterparty:
I pushed a few changes. I think this PR is ready for it's last review cycle. Thanks for bearing with me:
- Updated randomprime [v1.32 -> v1.33](https://github.com/randovania/py-randomprime/releases/tag/v1.33.0). Critically, this fixes the blast shield regression.
- Replace png with webp
- Make use of the additional 7 named save slot characters (`saveName`)
- Add gunship preview to cosmetic options GUI:
Screenshot_1

While this PR adds the *gameplay* functionality of the Gravity Suit toggle, I will open a separate PR which updates the logic database.

randovania/interface_common/persisted_options.py
```diff
@@ -224,6 +230,7 @@ def _split_tracker_theme(options: dict) -> None:
     _only_new_fields,  # dread: show_dna_in_hud/enable_debug_logging
     _only_new_fields,  # msr: option to use fusion models and reveal map
     _only_new_fields,  # enabled field for builders
+    _prime1_fusion_suit_color_rotations,  # prime1: suit colors are now remembered per Fusion Suit state
```
henriquegemignani: add to the end of the list

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #9453, opened 2026-08-01
Fusion: Enable OpenHatchLockRando misc  resource

Miepee:
Aparently we just never added the code in the bootstrap lol
There's some logic in fusion that expects some cross-room stuff to work if DLR is enabled but open hatches are not shuffled. Which originally i think got added because fusion had a few issues patching wise with open hatches.

### DIFF
--- a/test/games/fusion/generator/test_fusion_bootstrap.py
+++ b/test/games/fusion/generator/test_fusion_bootstrap.py
@@ -10,6 +10,7 @@ from randovania.game_description.resources.pickup_index import PickupIndex
 from randovania.games.fusion.generator import FusionBootstrap
 from randovania.games.fusion.layout.fusion_configuration import FusionArtifactConfig
 from randovania.generator.pickup_pool import pool_creator
+from randovania.layout.base.dock_weakness_distributor_configuration import DockWeaknessDistributorMode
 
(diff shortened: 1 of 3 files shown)


test/games/fusion/generator/test_fusion_bootstrap.py
```diff
@@ -62,3 +63,70 @@ def test_patch_resource_database(
...
+        fusion_configuration = dataclasses.replace(fusion_configuration, adjusted_geron_weaknesses=True)
+        expected_resources.add("NerfGerons")
+    else:
+        fusion_configuration = dataclasses.replace(fusion_configuration, adjusted_geron_weaknesses=False)
+
+    enabled_resources = FusionBootstrap()._get_enabled_misc_resources(
+        fusion_configuration, fusion_game_description.get_resource_database_view()
+    )
+
+    print(enabled_resources)
```
henriquegemignani: print :p

henriquegemignani (approved):

test/games/fusion/generator/test_fusion_bootstrap.py
Miepee: oops

PR #9459, opened 2026-08-02
Generator/fix one sided doors with not randomized side

Tyranisaur:


randovania/game_description/db/dock.py
```diff
@@ -111,6 +111,7 @@ class WeaknessDistributorSettings:
 
     change_from: set[DockWeakness]
     change_to: set[DockWeakness]
+    indirect_change_from: set[DockWeakness]
```
henriquegemignani: Please document this field. Even the other two could use some explanation, but this one is considerably more cryptic.

randovania/game_description/integrity_check.py
```diff
@@ -320,6 +320,41 @@ def find_inconsistent_dock_configuration(dock_db: DockTypeDatabase) -> Iterator[
...
+            if not isinstance(other_dock, DockNode):
+                yield (
+                    f"{dock.identifier}"
+                    f" weakness can be randomized, must become a two sided dock, but the default connection is to"
+                    f" {other_dock.identifier}, which is not a DockNode"
+                )
+                continue
+            if dock.dock_type != other_dock.dock_type:
+                yield (
+                    f"{dock.identifier}"
```
henriquegemignani: I'd say add a `base_message = f"{dock.identifier} weakness can be randomized, must become a two sided dock, the default connection is to {other_dock.identifier}"` which you use for these messages to avoid duplicated text.

test/test_files/patcher_data/dread/dread/all_settings/world_1.json
```diff
@@ -7812,9 +7812,9 @@
             }
         ]
     },
-    "layout_uuid": "05243515-6203-8931-9cd5-4434c8bd79ff",
+    "layout_uuid": "b7a4f8c7-8a2a-8010-b916-626080ae64bf",
```
henriquegemignani, line 7815: not for this PR but I'm not sure I like this churn. Maybe this uuid generation needs to be mocked like the seed hashes are too.

henriquegemignani (changes requested):

randovania/game_description/db/dock.py
```diff
@@ -112,6 +112,13 @@ class WeaknessDistributorSettings:
     change_from: set[DockWeakness]
     change_to: set[DockWeakness]
 
+    """
```
henriquegemignani, line 116: docstrings for attributes comes after the attribute 😅

henriquegemignani (approved):

PR #9461, opened 2026-08-03
Change Nintendont executor to API 2

Miepee:
Requires a new Nintendont release still and bumping that in our release script


henriquegemignani (commented):
Maybe fix `tools/fake_nintendont_server.py` along

Miepee:
Done. Although i'd rather deprecate that and use some better tooling that can actually interact with the game instead. 
Maybe (something like) DMEd? Not sure.

henriquegemignani:
> Done. Although i'd rather deprecate that and use some better tooling that can actually interact with the game instead. Maybe (something like) DMEd? Not sure.

In that case, I see you've volunteered to go add this new version to https://github.com/randovania/dolphin-memory-engine-daemon :p

henriquegemignani (approved):

PR #9462, opened 2026-08-03
Split structure and theme for the auto tracker

henriquegemignani:
image
image

I didn't bother splitting the current user choice for which theme they were using, it'll fall back to default again.


JeffGainsNGames (approved):
work reviewed, everything functions as expected for built in trackers
this is a very welcome change to user interface imo
please fix changelog when ready

PR #9475, opened 2026-08-06
Fusion Lava Lake Logic Update

M4RKERSM4SK:
Helps #9031, starting with Lava Lake first to get a bearing to base future changes off of.

Based on duncathan's advice on writing the logic from the discord discussion on damage runs, I have updated the logic in lava lake to reflect what is needed for the calculations to be accurate.

Updated the vidoes on the damage runs that were needed or missing.

~~Updated the shine spark trick which helps with #9068, though I am not sure if this is the correct way to do this, so would like a double check there pretty please~~ Scraped this change


randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -678,32 +675,35 @@ Hint Features - Climbable Surface, Multiple Pickups
...
   * L2 Hatch to Lava Lake Annex/Door to Lava Lake
   * Extra - door_idx: [36]
   > Door to Zebesian Zig
       Any of the following:
-          # Intended
+          # Lava Immunity
           Gravity Suit and Varia Suit
-          # Damage run suitless: https://youtu.be/RF2_8Y0PaVQ
+          # Damage Run with Gravity Suit: https://youtu.be/ovdmde7pdCs
+          Gravity Suit and Damage Runs (Beginner) and Lava Damage ≥ 65
```
cluekitty: The video shows you taking 69 Damage here, should this really be `>= 65`?

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -713,25 +713,25 @@ Hint Features - Climbable Surface, Multiple Pickups
       Any of the following:
           Space Jump
           # Shinespark across from Moto Motorway - https://www.youtube.com/watch?v=HyXBjdbLfDs
-          Level 0 Keycard and Speed Booster and Shinespark Tricks (Beginner) and Disabled Door Lock Rando and Disabled Entrance Rando
+          Speed Booster and Shinespark Tricks (Beginner) and Disabled Entrance Rando
```
cluekitty: Why would the L0 Keycard and Disabled DLR options be acceptable to remove here? I understand adding Disabled Entrance Rando, but with DLR it is still possible to have a permanently locked door preventing the shinespark. The L0 keycard is still there in vanilla and just because the built-in presets provide it as a starting item does not mean that other presets will shuffle L0 keycard.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -713,25 +713,25 @@ Hint Features - Climbable Surface, Multiple Pickups
...
-              # Deal with Lava
+              # Lava Immunity
               Gravity Suit and Varia Suit
-              # Suitless Damage Run: https://youtu.be/GUfsbWoRpzA
-              Damage Runs (Intermediate) and Lava Damage ≥ 150
-              Varia Suit and Damage Runs (Beginner) and Lava Damage ≥ 55
+              # Damage Run with Gravity Suit: https://youtu.be/dykEhsdnpA4
               Gravity Suit and Damage Runs (Beginner) and Lava Damage ≥ 80
+              # Damage Run suitless: https://youtu.be/GUfsbWoRpzA
+              Damage Runs (Intermediate) and Lava Damage ≥ 150
```
cluekitty: The video showcases taking roughly 120 damage here. Is this extra leeway intended? Based on other changes in this room I almost expected something like 125 or 130 as the value.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -645,26 +645,23 @@ Hint Features - Climbable Surface, Multiple Pickups
...
   > Door to Lava Lake Annex
       Any of the following:
-          # Deal with Lava
+          # Lava Immunity
           Gravity Suit and Varia Suit
-          # Suitless Damage Run: https://youtu.be/TC4s0dx6lg4
+          # Damage Run with Gravity Suit: https://youtu.be/0ik0A0k1Hho
+          Gravity Suit and Damage Runs (Beginner) and Lava Damage ≥ 30
+          # Damage Run suitless: https://youtu.be/TC4s0dx6lg4
           Damage Runs (Intermediate) and Lava Damage ≥ 60
```
Miepee, line 653: once again throwing this in here with "excluding the amount of damage you take, does having gravity really make this easier?"

cluekitty: Honestly, I don't think it make any difference on overall difficulty if you ignore damage taken. Both are pretty comparable movement if you assume that the player is immune to lava damage.

CHANGELOG.md
```diff
@@ -20,6 +20,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed: Some situations where the generator thinks pathing via certain doors is not possible with randomized doors.
 
+### Metroid Fusion
```
StalledStorm: Metroid Fusion should go above Metroid Samus Returns in the changelog, to be alphabetically ordered

StalledStorm:
As per a previous discussion in discord (https://discordapp.com/channels/914291389293027329/1215675687936196608/1503495486949294252) regarding S3 Lava Maze, should the damage runs to the pickup in lava be changed to put all of the required damage just in the connection to the pickup?

randovania/games/fusion/logic_database/Sector 1 SRX.txt
M4RKERSM4SK: Thanks for the catch, will change for 75

randovania/games/fusion/logic_database/Sector 1 SRX.txt
M4RKERSM4SK: To be completely honest I forgot about permanently  locked doors and that L0 can still be shuffled. Will revert and keep this PR more for fixing the damage logic

randovania/games/fusion/logic_database/Sector 1 SRX.txt
M4RKERSM4SK: I had kept the original value from the DB due to not knowing if there was previous discussion. When looking up old threads, it seemed there was discussion on this one have some margin of error for climbing the platforms? 

I would rather bring this down to 130 to stay consistent though, do you have any rememberance since you recorded the video on it?

CHANGELOG.md
M4RKERSM4SK: Thank you will fix

randovania/games/fusion/logic_database/Sector 1 SRX.txt
M4RKERSM4SK: In this exact case and limited to this room, I would agree it doesn’t seem like much, but then it would make the whole room inconsistent. If gravity doesn’t make it easier here, then suitless in this case should be brought down to beginner and the gravity portion removed.

However, what constitutes lava run difficulty? Is it movement that makes lava runs harder? Is it the amount of damage? If its both, what is the threshold for the balance of the two that justifies it from beginner to intermediate? I think gravity does give a inherit confidence due to movement alone to justify it being added for beginner even in the small cases here (and especially in this case where in the next room it does make a huge difference lava run wise which is where this connection would come in importance overall)

M4RKERSM4SK:
> As per a previous discussion in discord (https://discordapp.com/channels/914291389293027329/1215675687936196608/1503495486949294252) regarding S3 Lava Maze, should the damage runs to the pickup in lava be changed to put all of the required damage just in the connection to the pickup?

Good point, and I still dont know the correct answer to be honest. I was trying to follow what was already established in this room originally with the damage being in the connection back to stay consistent. If its better for the generator, I can make the return trips trivial and place the total damage in the connections getting there. If someone with extensive generator knowledge could confirm if this is the correct way for this case, I would be happy to change it. I just dont know have enough knowledge on the generator.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
Miepee: > then it would make the whole room inconsistent.

I don't mind. I would rather have a few rooms stay inconsistent for a while, where we're trying to fix things, than trying to handle a monster of a PR that touches *everything*. 

> However, what constitutes lava run difficulty?

IMO, it's complexity of movement and length. Not just for a lava run, but also for hazard runs in general.
E.g. "only hold left for 5 seconds" vs "hold left for 50 seconds" vs "do parcour including morphing for 5 seconds" vs "do parcour including morphing and space jumping for 50 seconds". 
We have 8 rooms with lava:
- Lava Lake
- Lava Lake Annex
- Lava Maze
- Sova Processing
- Garbage Chute
- Lava Reservoirs
- Boiler Access
- Pyrochamber

We could probably open up a thread and discuss how a sorta general guideline for difficulty judging, but we'd have to keep in mind the other types of hazard runs too (heat, cold, electric, acid).


> I think gravity does give a inherit confidence due to movement alone 

Not so sure about a confidence boost tbh. Gravity *does* make the runs a lot shorter, but that comes at the cost of loosing precision, since without it youre essentially playing in "slow motion".

randovania/games/fusion/logic_database/Sector 1 SRX.txt
```diff
@@ -713,25 +713,25 @@ Hint Features - Climbable Surface, Multiple Pickups
       Any of the following:
           Space Jump
           # Shinespark across from Moto Motorway - https://www.youtube.com/watch?v=HyXBjdbLfDs
-          Level 0 Keycard and Speed Booster and Shinespark Tricks (Beginner) and Disabled Door Lock Rando and Disabled Entrance Rando
+          Speed Booster and Shinespark Tricks (Beginner) and Disabled Entrance Rando
   > Pickup (Missile Tank in Lava)
       All of the following:
           Morph Ball
           Any of the following:
-              # Deal with Lava
```
Miepee, line 721: FWIW the "Deal with lava" was a comment to indicate that the whole ANY block was only requirements regarding the lava and nothing anymore regarding normal movement of the area.

randovania/games/fusion/logic_database/Sector 1 SRX.txt
Miepee: > When looking up old threads, it seemed there was discussion on this one have some margin of error for climbing the platforms?

From what I recall, It was less about the platforms, and moreso having a small margin of error:
- to have nicer rounded numbers
- to compensate in case we did a measuring/calculation mistake somewhere
- to compensate in case the db's suit reduction are slightly inaccurate.

Miepee:
> should the damage runs to the pickup in lava be changed to put all of the required damage just in the connection to the pickup

Thats really annoying/weird to do here, because of the different trips you can do. In lava maze it is a pure one way path. East reservoir is similar, where you only have one path (going in, and going out the same way)

Here you have *two* paths you can take. Either to the top or to the left. And damage wise if youd want to go annex while picking up the item, it would be cheaper to go top door -> item -> bottom door than top door -> item -> top door -> bottom door

Miepee:
Looking at this room, I wondered whether we should maybe model it like this:
image

It'd let us not duplicate the health requiements from both the doors that want to get to the item. But I think that gets kinda funky because theres no nice perfect for that middle generic node. Since if you want to go bottom door -> top door, you should jump a few tiles further to the left, and if you want to go from pickup -> top door you should jump a few tiles further to the right.
Unless we just say "screw it" and remodel our new damage runs in such a way that they *have* to first be underneath the platform before going another path. But doing that would add a decent amount of damage to it. From  my quick and rough testing:
- Top door -> item: 122
- Top door -> middle node -> item: 142
- Top door -> bottom door: 73
- Top door -> middle node -> bottom door: 118
(Although we could also probably overcomplicate it, by making the middle node -> top door a one way, and keeping the top door -> bottom door and top door -> item as one ways too....)

randovania/games/fusion/logic_database/Sector 1 SRX.txt
M4RKERSM4SK: Will change to 130 then, as 150 might be too large considering those factors

randovania/games/fusion/logic_database/Sector 1 SRX.txt
M4RKERSM4SK: > We could probably open up a thread and discuss how a sorta general guideline for difficulty judging, but we'd have to keep in mind the other types of hazard runs too (heat, cold, electric, acid).

If we had solid participation on a thread like this then that makes sense to me. I would rather nail down a foundation\understanding that everyone agrees on so I don't have to redo changes for future items and we finally have a good overall justification for damage run difficulty.

> IMO, it's complexity of movement and length. 

Gravity helps both of these factors considerably, even in small cases.

> Not so sure about a confidence boost tbh. Gravity does make the runs a lot shorter, but that comes at the cost of loosing precision, since without it youre essentially playing in "slow motion".

I heavily disagree here and I think this comment is being overcritical. Overall lava runs with gravity on movement:
- Eliminates panicking with "slow motion" movement while taking damage. Have watched countless times spin jumping up the Lava Lake platforms go wrong or jumping too early due to this panic or missing jumps in other cases.
- Liquid based movement is supposed to feel uncomfortable, you have more "precision" with familiar movement with gravity.
- Wall jumps, space jump, speedboost movement mechanics comes back with gravity.

All of this contributes to the confidence of movement while trying the lava damage run with gravity. Casually, I don't see a case for toggle off gravity for lava runs to make it "more precise" or "easier".

M4RKERSM4SK:
> Thats really annoying/weird to do here, because of the different trips you can do. In lava maze it is a pure one way path. East reservoir is similar, where you only have one path (going in, and going out the same way)

That makes sense to me will keep it that way then.

> kinda funky because theres no nice perfect for that middle generic node

Yeah it just seems to add more complexity to add a generic node too without much benefit and it seems goofy having the center be in a hazard. The missile pickup really is the true center but then no requirements from the pickup to the top right door makes logically the top right door the true "center" I guess. I don't know, there doesn't seem to be a nice way to divide this room nicely than what's already there.

PR #9484, opened 2026-08-08
Add multiworld for async race rooms

ThanatosGit:
Would be nice to get some more general feedback before a real code review.

Workflow:
- Create an async race room via the known dialog
- in the dialog you can add one world or multiple worlds
- as soon as you have more than one world or set "allow co-op", the concept of a "team" is used
- the team size must be set because whatever people can come up with is variable ("coop one world, one player race" to have every item offworld, "coop race", "multiple worlds but only one player" etc.)

image

- in the async race room itself you will now see "teams". teams can be created, you can only be part of one team
- you can join a team by a specific `join code`. just copy and paste it to your team members.
- there is also a "captain" who is in charge and the only person who can start or stop the timer for the team

image

- for each team there will be a hidden and not listed multiplayer (multiworld) session
- it can be opened by the team members via right click menu
image

- the options set by the async race room are immutable, so every multiworld session for each team has the same settings like "allow abandoned world", "enable co-op" etc.
- in the multiworld session the player must claim their world, then they should export and when everyone is ready the captain should hit the "start" button in the async race room

My ToDos:
- [ ] Checkbox for "abandon" does not make sense if no multiworld session (no team mode) is used
- [ ]  Race room: window of others does not update "live". (Team still showed 1 member but the other RDV instance already joined)

Upcoming PR:
- Should split the code. With the additions, some of the files (like `room_api.py`) have too much LoC for one file but I have to think where and what should be moved.

### DIFF
--- a/randovania/gui/dialog/async_race/async_race_creation_dialog.py
+++ b/randovania/gui/dialog/async_race/async_race_creation_dialog.py
@@ -12,8 +12,11 @@ from randovania.gui.lib import async_dialog
 from randovania.gui.lib.background_task_mixin import BackgroundTaskMixin
 from randovania.gui.lib.generation_failure_handling import GenerationFailureHandler
 from randovania.gui.widgets.generate_game_mixin import GenerateGameMixin
+from randovania.network_common.async_race_room import race_uses_teams
 
 if TYPE_CHECKING:
+    from collections.abc import Iterator
+
     from randovania.gui.lib.window_manager import WindowManager
     from randovania.interface_common.options import Options
     from randovania.layout.layout_description import LayoutDescription
@@ -23,7 +26,7 @@ if TYPE_CHECKING:
 
 class AsyncRaceCreationDialog(QtWidgets.QDialog, GenerateGameMixin, BackgroundTaskMixin):
     ui: Ui_AsyncRaceCreationDialog
-    selected_preset: VersionedPreset | None = None
+    selected_presets: list[VersionedPreset]
     _preset_selection_dialog: SelectPresetDialog | None = None
     layout_description: LayoutDescription | None = None
 
@@ -45,6 +48,7 @@ class AsyncRaceCreationDialog(QtWidgets.QDialog, GenerateGameMixin, BackgroundTa
         self._options = options
         self.failure_handler = GenerationFailureHandler(self)
         self._background_task = self
+        self.selected_presets = []
 
         self.progress_update_signal.connect(self.update_progress)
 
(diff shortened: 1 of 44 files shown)


randovania/network_client/network_client.py
```diff
@@ -641,6 +641,44 @@ async def async_race_join_and_export(
...
 
+    async def async_race_create_team(self, room: AsyncRaceRoomEntry, team_name: str) -> AsyncRaceRoomEntry:
+        """
+        Creates a new team in a room played in teams, with yourself as its first member.
+        :param room: The room's data from get_async_race_room
+        :param team_name: Name for the new team
+        :return: Updated room details
+        """
+        return AsyncRaceRoomEntry.from_json(
+            await server_signals.AsyncRace.CreateTeam.call_server(self)(room.id, room.auth_token, team_name)
```
henriquegemignani: Here I come yet again asking for new server APIs to be REST :)

randovania/gui/dialog/async_race/async_race_creation_dialog.py
```diff
@@ -45,18 +46,107 @@ def __init__(
...
+        return [
+            game
+            for game in RandovaniaGame.sorted_all_games()
+            if game.data.development_state.can_view()
+            and (game.data.defaults_available_in_game_sessions or not uses_teams)
+        ]
+
+    def _multiworld_messages(self) -> list[str]:
+        """
+        Reasons the selected worlds can't be raced in a multiplayer session. Only the first world
```
henriquegemignani: > Only the first world can be at fault

This seems wrong

Edit: ah, because the widget doesn't let you select a preset that'd cause issues after. This could be worded better. Maybe this fits better as a comment in the for loop?

randovania/network_common/async_race_room.py
```diff
@@ -93,17 +111,30 @@ class AsyncRacePauseEntry(JsonDataclass):
 @dataclasses.dataclass
 class AsyncRaceEntryData(JsonDataclass):
     """
-    All data about a user's entry to a race. Should only be available to admins.
+    All data about one participant of a race: a single user in a room played individually,
+    or a whole team in a room played in teams. Should only be available to admins.
     """
 
-    user: RandovaniaUser
+    user: RandovaniaUser | None
```
henriquegemignani, line 112: maybe it should just be inside `members`?

randovania/server/async_race/room_api.py
```diff
@@ -411,33 +478,188 @@ async def join_and_export(
...
+
+async def get_team_join_code(sa: ServerApp, sid: str, room_id: int) -> str:
+    """
+    Returns the code that lets someone else join the current user's team.
+    """
+    user = await sa.get_current_user(sid)
+    room = AsyncRaceRoom.get_by_id(room_id)
+    team = _get_own_team(room, user)
+
+    return sa.encrypt_and_b85_dict(
```
henriquegemignani: I believe the join code should be something short and easy to share, say 10 ascii letters+numbers.
It probably means we need something like a database of created join codes. Not the worst thing, we can definitely reuse it for say, sharing links for joining a multiworld session or a non-multi async race that aren't public.

henriquegemignani (changes requested):

randovania/gui/dialog/async_race/async_race_creation_dialog.py
ThanatosGit: Right. In Retrospect it's maybe just better to disable the "Add World" button if the first world is not compatible with a multiplayer session?

henriquegemignani: ... Yeah that is a good idea

ThanatosGit:
Update:
Create a room now contains a field for the timer "shared" or "accumulated".
image

For a finished race with accumulated timer, it shows the time of every player in the "Teams" section and your team's time in the "Participation" section
image

For an ongoing race it only shows information of your own team:
image

randovania/server/async_race/room_api.py
ThanatosGit: A serious question: Is there really a difference between a string of 10 random characters and a string of x random characters in this context? I mean, as soon as the word "random" comes up, everyone just copies and pastes it, right?

henriquegemignani: being short and pure ascii makes it viable to say to someone else if needed. It also makes it easier to copy paste and not be truncated and so on.

If going to make a huge thing, gotta at least make sure to use only characters that will cause double-click-to-select to work, which doesn't work with standard base64 (much less base85...)

ThanatosGit: Changed it to 8 random ascii uppercase letters and digits. It must be unique per room.
In the event of a collision, there is a loop with 10 retry attempts. Even though the probability of a collision is very low, since it only needs to be unique within a single room.

henriquegemignani (approved):

PR #9486, opened 2026-08-09
Prime1 - Gravity Suit logic is always safe

toasterparty:
## Changes

- Add FAQ entry
- Remove `allow_underwater_movement_without_gravity` preset option and add a migration to remove it
- Remove checkbox/label from preset tab
- Remove all before gravity and dangerous gravity instances in the logic database

Stale DZ review from the [botched 🥞 PR](https://github.com/toasterparty/randovania/pull/110):

image


randovania/games/prime1/game_data.py
```diff
@@ -154,6 +154,12 @@ def _test_data() -> randovania.game.game_test_data.GameTestData:
             "- Storage Cavern\n"
             "- Every room after Twin Fires Tunnel\n",
         ),
+        (
+            "I picked up Gravity Suit too early and it is inhibiting my ability to perform gravityless "
+            "slope jumps. How do I proceed?",
```
JeffGainsNGames: Would this be better with generic wording of "gravityless movement" since things like CFLTDG with morph is also affected

JeffGainsNGames (commented):
one wording comment, otherwise logic and ui changes look good

randovania/games/prime1/game_data.py
toasterparty: image

Updated to reference the trick category by full name

JeffGainsNGames (approved):

PR #9496, opened 2026-08-11
Fix test_persist_logical_pickup_placement causing GUI to show up

henriquegemignani:


duncathan (approved):
wtf lol does this actually work

henriquegemignani:
Apparently it's because pressing space causes the GUI to show up. We'll see if it works

PR #9499, opened 2026-08-13
Convert Prime Hunters images to webp

henriquegemignani:
Before: 11M
After: 1.4M


dyceron:
It looks like there's some quality loss but I guess it's not the biggest deal in the world.

henriquegemignani:
I could've been less aggressive with the quality, but yeah as long as you can tell what is what in the room you don't need this to be high resolution.

dyceron (approved):

PR #9510, opened 2026-08-15
Make `bytes` serializable

ThanatosGit:
Tried to switch to REST API endpoints in my async race room PR and failed because `AsyncRaceRoomEntry` contains `presets_raw: list[bytes]` which fails when sending as a `JSONResponse`

### DIFF
--- a/randovania/bitpacking/json_dataclass.py
+++ b/randovania/bitpacking/json_dataclass.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import base64
 import dataclasses
 import datetime
 import inspect
@@ -76,6 +77,9 @@ def _decode_with_type(arg: typing.Any, type_: type, extra_args: dict, metadata: 
     elif type_ is uuid.UUID:
         return uuid.UUID(arg)
 
+    elif type_ is bytes:
+        return base64.b64decode(arg)
+
     elif type_ is datetime.datetime:
         return datetime.datetime.fromisoformat(arg)
 
@@ -107,6 +111,9 @@ def _encode_value(value: typing.Any, metadata: Metadata) -> typing.Any:
 
     elif isinstance(value, uuid.UUID):
         return str(value)
+
+    elif isinstance(value, bytes):
+        return base64.b64encode(value).decode("ascii")
 
     elif type_lib.is_named_tuple(type(value)):
         if metadata.get("store_named_tuple_without_names"):
--- a/test/bitpacking/test_json_dataclass.py
+++ b/test/bitpacking/test_json_dataclass.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 import datetime
+import json
 import uuid
 from enum import Enum
 from typing import NamedTuple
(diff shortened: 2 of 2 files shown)


henriquegemignani:
Does it work better if you make it a pydantic model instead?

randovania/bitpacking/json_dataclass.py
```diff
@@ -108,6 +112,9 @@ def _encode_value(value: typing.Any, metadata: Metadata) -> typing.Any:
     elif isinstance(value, uuid.UUID):
         return str(value)
 
+    elif isinstance(value, bytes):
+        return base64.b64encode(value).decode("ascii")
```
henriquegemignani, line 116: Can use `base64.b85encode` to be slightly more efficient

henriquegemignani (changes requested):

randovania/bitpacking/json_dataclass.py
ThanatosGit: Do you want to optimise for size or for speed?
Afaik b64 is faster but larger than b85

ThanatosGit:
> Does it work better if you make it a pydantic model instead?

The `AsyncRaceRoomEntry`? Can be done like this. Must define the `model_config` because it defaults to UTF-8 encoding.
Then I have too touch a few places who are calling "as_json" etc. on the dataclass.

```python
class AsyncRaceRoomEntry(BaseModel):
    """
    Contains all data a client can receive about a AsyncRaceRoom.
    """
    model_config = ConfigDict(ser_json_bytes="base64", val_json_bytes="base64")

    id: int
    name: str
    creator: str
    creation_date: datetime.datetime
    start_date: datetime.datetime
    end_date: datetime.datetime
    visibility: "MultiplayerSessionVisibility"
    race_status: AsyncRaceRoomRaceStatus
    auth_token: str
    game_details: GameDetails
    presets_raw: list[bytes]
    is_admin: bool
    self_status: AsyncRaceRoomUserStatus
    allow_pause: bool

    @property
    def presets(self) -> list[VersionedPreset]:
        return [VersionedPreset.from_bytes(s) for s in self.presets_raw]
```

henriquegemignani:
It's nice when the return type of fastapi is a pydantic model because it auto-generates docs with that. I do have some stuff I started looking for having the client call fastapi routes and decode stuff a bit more automatically.

randovania/bitpacking/json_dataclass.py
henriquegemignani: Never really considered it being slower 🤔

ThanatosGit: Output on my machine:
```
=== Data size: 100 bytes, 1000 repetitions ===
b64encode              0.0003 s total         0.28 us/call
b85encode              0.0045 s total         4.52 us/call
-> b64encode is 15.85x faster (encoding)
b64decode              0.0003 s total         0.26 us/call
b85decode              0.0073 s total         7.28 us/call
-> b64decode is 27.70x faster (decoding)
Output size b64: 136 bytes
Output size b85: 125 bytes

=== Data size: 10,000 bytes, 1000 repetitions ===
b64encode              0.0201 s total        20.09 us/call
b85encode              0.4095 s total       409.45 us/call
-> b64encode is 20.38x faster (encoding)
b64decode              0.0116 s total        11.56 us/call
b85decode              0.6260 s total       625.95 us/call
-> b64decode is 54.13x faster (decoding)
Output size b64: 13,336 bytes
Output size b85: 12,500 bytes

=== Data size: 1,000,000 bytes, 100 repetitions ===
b64encode              0.2263 s total      2262.93 us/call
b85encode              5.2438 s total     52438.36 us/call
-> b64encode is 23.17x faster (encoding)
b64decode              0.1101 s total      1100.64 us/call
b85decode              7.3784 s total     73784.50 us/call
-> b64decode is 67.04x faster (decoding)
Output size b64: 1,333,336 bytes
Output size b85: 1,250,000 bytes
```

henriquegemignani (approved):

PR #9512, opened 2026-08-16
Change room api to REST

ThanatosGit:
Changes on the client side are ToDo.


randovania/server/async_race/room_api.py
```diff
@@ -92,7 +90,8 @@ def _fast_get_games_list_from_raw_layout(layout_description_json: bytes) -> list
     return [g for g in RandovaniaGame.sorted_all_games() if g.value in present_games]
 
 
-async def list_rooms(sa: ServerApp, sid: str, limit: int | None) -> Sequence[TypedJsonObject[AsyncRaceRoomListEntry]]:
+@router.get("/async-race-room/rooms")
```
henriquegemignani: This could be just `GET /async-race-room`, creating would be `POST /async-race-room` and get rest would be `<> /async-race-room/{room_id}`.

And maybe just `/async-race`? Unsure about this one.

ThanatosGit: I chose `async-race-room` because it is already used by the live split integration.

henriquegemignani:
Looks good so far

ThanatosGit:
I have no idea why I have to make the changes to the datetime. It feels wrong but I guess it comes from the JSON serialisation?
The `if now < start:` was failing because `now` was timezone aware with UTC but `start` and `end` were without a timezone.

How do we want to make this work?

```python
@router.post(endpoints.room_join_and_export_template)
async def join_and_export(
    sa: ServerAppDep, user: UserDep, room_id: int, auth_token: str, cosmetic_patches: BaseCosmeticPatches
) -> dict:
```
We never sent `BaseCosmeticPatches` but only any subclass of it, so the serialisation fails on validating for every new field e.g. for `MSRCosmeticPatches` because of `use_laser_color` etc.

Do we want to define it as a plain `dict` or something? The other solution I have not evaluated completly is to add some sort of discriminator with all possible classes like:
```python
CosmeticPatches = typing.Annotated[
    CosmeticPatchesA | CosmeticPatchesB | CosmeticPatchesC,
    Field(discriminator="type"),
]
```


and finally: we have nearly no tests for the client side :(

henriquegemignani:
> I have no idea why I have to make the changes to the datetime. It feels wrong but I guess it comes from the JSON serialisation?
> The if now < start: was failing because now was timezone aware with UTC but start and end were without a timezone.

The ruff update had some new rules regarding datetime and non-UTC aware and it did trip on some of our code. Which is likely the case that we have bugs!

> We never sent BaseCosmeticPatches but only any subclass of it, so the serialisation fails on validating for every new field e.g. for MSRCosmeticPatches because of use_laser_color etc.

Hmmm. Not sure, would need to look at it myself. wouldn't mind it being just dict or json I guess.

ThanatosGit:
> 
> The ruff update had some new rules regarding datetime and non-UTC aware and it did trip on some of our code. Which is likely the case that we have bugs!

I was the one who reviewed the code and I was not aware and also cannot find any change. The  DTZ rules are still a fixme.
But after some sleep I solved the mystery: It's the fact that our `JsonDataclass` adds timezone information when calling `as_json`
```python
    elif isinstance(value, datetime.datetime):
        return value.astimezone(datetime.UTC).isoformat()
```
but the new `as_json` from `JsonBaseModel` sents it as is with `return self.model_dump(mode="json")`.

That took me way too long to realize.

> 
> > We never sent BaseCosmeticPatches but only any subclass of it, so the serialisation fails on validating for every new field e.g. for MSRCosmeticPatches because of use_laser_color etc.
> 
> Hmmm. Not sure, would need to look at it myself. wouldn't mind it being just dict or json I guess.

Will try it.

ThanatosGit:
Should be good for a review now.
I had to correct a few things here and there, since serialization and deserialization are handled by pydantic.
Unfortunately, we didn’t have any tests for this at all, and it was quite tedious to fix some of these issues. 
I simply had an LLM handle correcting and creating tests. I also added end-to-end tests, since serialization and deserialization can easily lead to errors when changing something in the models.

henriquegemignani (approved):
lil concerned about the performance of these tests that starts a server with uvicorn but we'll see

ThanatosGit:
> lil concerned about the performance of these tests that starts a server with uvicorn but we'll see

Slightly changed the setup in the conftest which cut the time in half on my machine. 
I merged main because of conflicts. Nothing special, but the `AuditEntry` is located in `randovania\network_common\audit.py` and is a Pydantic model. I didn’t add a third `ruff.toml` file, just because it’s only relevant in a single line.

henriquegemignani:
>  I didn’t add a third ruff.toml file, just because it’s only relevant in a single line.

Isn't even needed, because there should never be a non-typechecking fastapi import in that folder as it'd fail on clients.

henriquegemignani (approved):

PR #9514, opened 2026-08-16
Dread/set items as bulk placement

Tyranisaur:
https://docs.google.com/spreadsheets/d/1jYTs3Y2d-pnZ2dCxrHYjkGweJmM5PFT9Zj3hpb0YxXM/edit?usp=sharing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Dread
 
+- Changed: The following pickups will now be placed in bulk at the same step during generation when they multiple pickups are necessary to progress:
+  - Missile Tank
+  - Missile+ Tank
+  - Energy Part
+  - Energy Tank
+  - Power Bomb Tank
+  - Flash Shift Uprade
+  - Speed Booster Upgrade
 - Fixed: Some situations where the generator thinks pathing via certain doors is not possible with randomized doors.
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Dread
 
+- Changed: The following pickups will now be placed in bulk at the same step during generation when they multiple pickups are necessary to progress:
```
Logarhythmic: > same step during generation when they multiple pickups are necessary 

extra "they" here

Logarhythmic (commented):
What does changing the step in which these items are placed accomplish?

Miepee:
The generator can encounter situations, where it needs multiple items together to unlock new resource nodes. E.g. an example from fusion that unlocks some boss: `[P: Energy Tank, P: Energy Tank, P: Energy Tank, P: Wide Beam]`
Usually, in multi-item requirements like these, the generator would only ever place one pickup at a time. So in this case, first the first Etank, then the second etank then the third Etank and finally the wide beam.
Bulk placement for one pickup allows placing all those pickups at once. E.g. having Etanks marked as bulk placement would change it to first placing 3 Etanks, and then the second action placing wide beam.

Logarhythmic (approved):
Seems like a relatively small change. As best I understand it, this seems OK to me.

Miepee (approved):
core approval from db maintainer

PR #9515, opened 2026-08-17
Add the TC rule for ruff

henriquegemignani:
This apparently has been broken for a while.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -228,26 +228,28 @@ extend-exclude = [
 
 [tool.ruff.lint]
 extend-select = [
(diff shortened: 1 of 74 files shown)


ThanatosGit:
pre-commit says there a still TC rule violations

henriquegemignani:
Fixed. Turns out I forgot to run ruff for the tests, and then it broke a test, and then investigating I found that FastAPI on the server stuff definitely shouldn't be moved to typechecking. Nor any pydantic model!

ThanatosGit (approved):

PR #9516, opened 2026-08-17
Add BaseSocketExecutor and SignalSocketExecutor

henriquegemignani:


randovania/game_connection/builder/socket_connector_builder.py
```diff
@@ -0,0 +1,64 @@
...
+    def pretty_text(self) -> str:
+        return f"{super().pretty_text}: {self.ip}"
+
+    def create_executor(self) -> Any:
+        """Builds the executor for this game. Imports are delayed to avoid too many early imports in startup."""
+        raise NotImplementedError
+
+    def create_connector(self, executor: Any) -> RemoteConnector:
+        """Builds the connector wrapping a connected executor."""
+        raise NotImplementedError
```
ThanatosGit: I don't really like typing it as `Any` (actually I never like using it all unless you really have to)
Following the new structure and verify that the old and new logic is the same is hard. I started from the `MercuryExecturor` to the parent Class `SignalSocketExecutor` to the partent class `BaseSocketExecutor` and searched for references to `connect` and only found test files because it is typed as `Any` here. Can we add a generic and bound its upperclass to something which ensure that it has a `connect` method?

henriquegemignani: Added a protocol. I did at first change all XConnectorBuilder to inherit from the specialised base class, but that caused the Executor type to be imported lazy and I don't want to figure out if that's bad.

ThanatosGit: Hmmm the protocol does not really help with static analysis to find references to a method call if there is nothing implementing the protocol.
It is very loose now.

About the lazy import: You mean you did something like `class MSRConnectorBuilder(SocketConnectorBuilder[MSRExecutor]):` and `class BaseSocketExecutor[HolderT: CommonSocketHolder](ExecutorProtocol):` ?
That shifts the import outside of the `TYPE_CHECKING` block for `MSRExecutor` to be eagerly loaded. The executor is already lazy loaded within the `create_executor` methods.
Or TL;DR: I think you meant it caused the executor type to be imported eager and not lazy anymore.

henriquegemignani: Doing that caused the executor to be imported eagerly, correct 
I see i missed a not in the other message

ThanatosGit (approved):

PR #9517, opened 2026-08-18
Door lock rando changes (draft)

MetroidSlayer:
Features a new randomization mode: "Individually (thorough)", or INDIVIDUAL_THOROUGH internally. (I'll probably need help brainstorming a better name, this one kinda sucks.) This mode is distinct from the current "Individually" mode in that it is completely blind to the player's equipment, and places doors regardless of whether they can presently be opened, instead checking whether the game is beatable at all with them there.

It takes longer to generate a seed with this mode, but door weaknesses have a much greater impact on progression, forcing more exploration and backtracking, and also often forcing the player to take alternate routes they usually wouldn't. Because doors are placed if a seed is beatable, this mode works best when critical items like DNA/artifacts/keys have duplicates that aren't required, as more docks will have the capability to have a lock. I find the dock locations are still interesting even if there are no extra keys, though.

When testing this mode, I recommend not adding permalocked doors to change_to when testing: there's a chance that certain items will just be unobtainable as a result of how doors are placed, so some doors might end up functionally permalocked anyway. I think you'll get a better feel for what this mode is going for without them.

Also features 2 important variables in dock_weakness_distributor_configuration.py:

- locked_percentage (int)
Door placement will stop entirely once the percentage of locked doors in the seed meets locked_percentage, significantly decreasing generation time while also making seeds more open and explorable. To account for this change, the method by which docks are placed has been changed. The unlocked weakness is not added to weighted_weaknesses unless absolutely necessary, so doors are placed as often as possible so as to reach locked_percentage faster. This is probably the biggest change I've made, as it fundamentally changes the normal INDIVIDUAL_DOCK mode as well. That wasn't something I set out to do, but given the improvement to generation time and openness I think it's worth it.
The game, trick settings, and randomization mode will effect what value feels "right" here. As a general rule, when using INDIVIDUAL_THOROUGH, 20% seems about right for the Primes, and 25% seems about right for the 2D games _(edit: I've now added code in distribute_post_fill_weaknesses to force this, so you don't have to change it yourself)_. When using INDIVIDUAL_DOCK, I expect a higher value like 50% to feel a bit better. Finding the right percentage to be the default for each game will likely require some more testing. I'd like for players to be able to choose their own value for this.

- attempt_similar_quantities (bool)
For each dock, the weights of each weakness will be adjusted so that weaknesses that haven't been placed as much will be exponentially more likely. When using INDIVIDUAL_THOROUGH, this forces the items required to satisfy the weaknesses to each have an equal chance of being important in a given seed. When using INDIVIDUAL_DOCK, this helps alleviate repetitive door placement and slightly helps items have a higher chance of being important, though it's limited by the fact that different weaknesses are available for each dock depending on item progression. In other words, there's only so much this method can do for INDIVIDUAL_DOCK, but it should help at least a little.
I'd like this to be toggleable by players, but if you don't think that's a good idea due to UI clutter or something, I'm content just baking this into INDIVIDUAL_THOROUGH and letting INDIVIDUAL_DOCK remain as it is.

Again, some of my changes have altered the way INDIVIDUAL_DOCK works, so if you want to play a seed that uses it as a control to compare the new mode against, you should probably generate it using a stable release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,7 @@
 <!-- The Begin and End comments throughout this document are used in order to pull specific sections of the readme into the main GUI window at runtime. -->
 Hello to anyone reading this!!!
 
-For a variety of reasons, I haven't enjoyed the individual door lock randomization for Randovania games. Some randomizers place doors before they place items. Randovania places items before it places doors. With the former, items must be placed in a way where the game is possible given the new doors; so, in theory, if you generated the exact same seed twice, with door locks on and off, the path of progression would likely be totally different.
-
-With the latter, though, the doors are placed such that the intended, logical path through the game is just as possible as it was without them. If you generated the same seed twice, with door locks on and off, the path would likely be barely different, if not completely identical. In other words, the current implementation of door locks functionally does nothing at all to progression.
-
-It's more complicated than that, and Randovania's method has a TON of things going for it! It's not useless, its functionality just lies more with speedrunning than exploration. I don't want to ramble too much, so please take my word for it when I say there are pros and cons to both approaches for sure.
-
-ANYWAY, I wrote up a hackey, hardcoded way of making the doors get placed before the items. I've tested it a few times with both Samus Returns and Prime 2, and it works great. (Disclaimer: haven't tested with any other game yet, but I see no reason it shouldn't work with all of them.) I'm personally having tons of fun with this implementation. I'll likely try to get it working as an actual selectable setting in the menu when I find the time.
-
-In order to get this to work, select "Individually" as the Randomization Mode on the Door Locks tab. Then, make sure EVERY door type in the left section is selected. You can pick and choose from the right section, though. (I'd recommend not adding Permanently Locked doors.) This implementation has a somewhat high failure rate, so just keep retrying the generation process until you get a seed through. It's a lot faster of a process than the post-item-placement implementation, so hopefully it's not too annoying.
+The goals of this fork have changed since last time. I hadn't yet realized what I was really trying to achieve - less guided, more exploration-based seeds - and I was under the assumption that doors-then-items was the only way to get results like that. After experimenting with many different approaches, I've settled on something that I think achieves what I'm going for while still using the items-then-doors approach.
 
 # Randovania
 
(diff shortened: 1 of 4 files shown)


duncathan:
Thanks. Having the code in front of mii will help a lot with assessing the feature and whether the new mode is worth the maintenance burden. I'm extremely busy this week all of a sudden, but I'll try to look through it when I have time and certainly I'll look at it next week!

MetroidSlayer:
Take all the time you need, and then some! I'm honestly looking forward to having a break, lol

MetroidSlayer:
After thinking about all this some more, I think I want to make some more adjustments to my approach. I'm not gonna completely overhaul things, so you can probably still review this version of the feature and offer feedback if you'd like, but if you want to hold off until I'm done, I totally understand

MetroidSlayer:
Okay, pretty sure I'm done with my adjustments! If you were waiting to review until I was done, you can go for it now.

...

You don't have to read the rest of this, but it explains the problem I was running into and what I did to solve it. I just can't help writing whole essays for everything I do :/

Since the locks were being placed specifically when the game was beatable with them there, the path forward sometimes became obvious due to long chains of unlocked doors. To make progression harder to guess, I've re-introduced non-blind door placement similar to INDIVIDUAL_DOCK that can happen under certain conditions. To avoid such doors clogging up the limited spots enforced by locked_percentage, and to stop them from becoming similarly repetitive and indicative of progression, only *one* guaranteed-beatable, non-blind instance can exist per door type.

This should generally stop players from being able to tell which doors are necessary to get past and which aren't. Of course, this means INDIVIDUAL_THOROUGH is no longer totally blind to the player's equipment. Very few doors per seed will be placed in this way, though. Additionally, I saw various other improvements to how progression felt, but I'm just about outta time to write this, so I won't go into them.

When I find time next, I'll probably also make some more changes to how locked_percentage and attempt_similar_quantities work, to hopefully stop them from affecting INDIVIDUAL_DOCK mode or non-door dock randomizers. (I still think they could help improve INDIVIDUAL_DOCK, but after thinking about it some more, this is probably a big maintenance burden thing, and I don't want my features to be more trouble than they're worth!)

PR #9518, opened 2026-08-19
Prime2-OPR: Hive Summit Logic

XenoWars:
Adds logic in the direction of Portal to Aerie to Portal from Aerie in preparation for two-way portals. Also lowers the difficulty of the Movement trick to Portal to Aerie since there is no further input needed after the boost. Could be argued it can even be trickless.


randovania/games/prime2_opr/logic_database/Ing Hive.txt
```diff
@@ -1832,6 +1832,14 @@ Extra - camera_data:
   * Layers: default
   * Light Portal to Aerie/Portal from Hive Summit
   * Extra - dock_name: VirtualArenaDock
+  > Portal from Aerie
```
Samuel6710, line 1835: What's the idea behind adding a connection here?

Samuel6710 (changes requested):
Needs changelog entry

JeffGainsNGames (approved):

PR #9521, opened 2026-08-20
Fusion: Make charge core and watery hole possible in DLR

Miepee:
I thank jeff in the past for this wonderful visualization.

grafik

Watery Hole is fairly simple to explain. There are only 2 doors in between, both doors cannot be permanently locked as this is a one-way dead-end, so once you can reach the save room, you are guaranteed to open the 2 doors in the way.
(Technically you can do it without reaching the same room and starting in hornoad housing but then you have to do a frame perfect shinespark crouch. Frame to early and you dont get the spark, frame too late and you fall off the clif)

The Charge Core Arena is similar. None of the doors cannot be permanently locked, because the whole charge area is a one way, so if you reached the save room and are now at the lower door, you were guaranteed to be able to open the 3 doors in between.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Logic Database
 
 ##### Sector 1 (SRX)
-- Changed: Watery Hole: Getting the item is now possible in Door Lock Rando.
+- Changed: Watering Hole: Getting the item is now possible in Door Lock Rando.
 - Changed: Charge Core Arena: Getting the upper item locked by a Shinespark is now possible in Door Lock Rando.
 
 ### Metroid: Samus Returns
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed: Some situations where the generator thinks pathing via certain doors is not possible with randomized doors.
 
+### Metroid Fusion
+
+#### Logic Database
+
+##### Sector 1 (SRX)
+- Changed: Watery Hole: Getting the item is now possible in Door Lock Rando.
```
cluekitty:
```suggestion
- Changed: Watering Hole: Getting the item is now possible in Door Lock Rando.
```

cluekitty (changes requested):

cluekitty (approved):

JeffGainsNGames (approved):

PR #9528, opened 2026-08-22
Fusion: reword phrase when metroid scan shows 0

Miepee:
Minor thing, but it happened in a seed recently and the text caused me some slight confusion.


Miepee:
> Can you describe what was the issue with the original wording?


The phrase "This terminal **was unable to scan**" made me think that a weird bug happened that caused it to not be able to scan / that it failed, rather than it having scanned and the scan results being empty.

JeffGainsNGames (approved):

PR #9535, opened 2026-08-24
Fusion: Sector 6 Make important item paths safe in DLR

Miepee:
A lot more work that could still be done there, just wanted the important ones out of the way


JeffGainsNGames (approved):

JeffGainsNGames:
Can't see what test it is on mobile, but needs a fix

PR #9541, opened 2026-08-26
Fusion: Remove more things from Main Deck to not require DLR

Miepee:
For next release

Worth noting that we should probably get a new/better video for Station Entrance that actually demonstrates it better. The current one is also having wide beam which makes shooting the blocks a little bit easier.


JeffGainsNGames (approved):
DB approval, noting changelog entry is dependent on next release section being ready

henriquegemignani:
Post-release ping.

Miepee:
changelog added

PR #9543, opened 2026-08-26
Add required event to make connection DLR safe

StalledStorm:


Miepee (approved):
thanks for fixing my mistake

PR #9548, opened 2026-08-29
Limit cython version we can use

henriquegemignani:


Miepee (approved):

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 requires = [
     "setuptools>=64",
     "setuptools_scm>=10",
-    "cython",
+    "cython<3.3",
```
Miepee, line 5: just realized, as usual, a comment as to why we limit one would be nice

henriquegemignani: bleh, I didn't even look why it failed. no matter the reason I call it out of scope for the release

PR #9552, opened 2026-08-31
Hunters - Fix Shield Keys using the wrong state bit

dyceron:
I had the wrong state bits set for a few keys which was causing issues with export the correct values.


Miepee (approved):
I have no way of confirming this, I just trust that this is correct.

PR #9559, opened 2026-09-02
Hunters - Bump OPHR to 0.13.4

dyceron:


ThanatosGit (approved):
Maybe release as 11.0.1 ?

dyceron:
If desired. The first fix is more relevant while the second is more of a safety.

PR #9561, opened 2026-09-02
Prime 1: Add required-main support for Missile and Power Bomb Expansions

reedwhaley:
Was asked to re-upload without using AI to polish.

Prime 1 currently has the resources for required-main Missile and Power Bomb ammo, but the exporter explicitly rejects presets that enable `requires_main_item`.

This finishes that support for:

- Missile Expansion -> Missile Launcher
- Power Bomb Expansion -> Main Power Bomb

Expansions collected before the main item are kept, but the ammo can't be used until the corresponding launcher is acquired. Once the launcher is collected, the capacity that was already picked up becomes usable.

The main thing that made this a little awkward is that Randovania and RandomPrime represent the locked state differently.

Randovania uses `LockedMissile` and `LockedPB` logically while the main item is missing. RandomPrime still stores that ammo in the normal Missile and Power Bomb inventory slots.

So the exporter keeps the locked resource on the Randovania side, but uses the unlocked resource set when determining what RandomPrime should physically give the game.

Missile Launcher and Main Power Bomb also now have separate ownership state from their ammo. The remote connector reads those ownership bits back from the game and handles remote delivery of the main item separately from adding ammo.

Starting inventory was updated for the same reason. If required mains are enabled, starting with Missile or Power Bomb ammo no longer implicitly means the launcher is owned. If required mains are disabled, the old behavior stays the same.

There is also matching pickup text support through RandomPrime's new optional `conditionalHudmemo`.

Example:

```json
{
  "hudmemoText": "Missile Expansion acquired!",
  "conditionalHudmemo": {
    "requiredItem": "Missile Launcher",
    "missingText": "Missile Expansion acquired, but the Missile Launcher is required to use it."
  }
}
```

If the launcher is already owned, the normal pickup text is used. If it isn't, the locked message is used instead.

The remote/Multiworld path was updated as well. I tested this against live Dolphin with the following sequence:

1. Missile Expansion before Missile Launcher
2. Missile Launcher
3. Power Bomb Expansion before Main Power Bomb
4. Main Power Bomb

That verified the locked ammo state, custom launcher ownership state, inventory readback, capacity becoming usable after the launcher arrives, remote HudMemo text, and the Multiworld magic counter.

The original remote locked-item messages were also shortened after testing because the longer versions could get cut off once the provider name was included.

Testing done on this branch:

```text
Focused Prime 1 tests:
136 passed

Focused feature tests:
69 passed

Acceptance update:
106 passed

Acceptance verification:
106 passed

Full test suite:
8042 passed, 4 skipped

mypy:
1173 source files, no issues

ruff:
clean

git diff --check:
clean

package build:
sdist and wheel both built successfully
```

I also tested the matching Randovania/RandomPrime changes in-game for the local pickup path and the live remote/Multiworld path.

AI use: GPT-5.6 Sol with reasoning enabled was used to trace the existing Prime 1 pickup/resource handling, work through the exporter and remote connector changes, generate portions of the implementation/tests, and debug the runtime behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [11.1.0] - 2026-10-01
 
+### Metroid Prime
+
+- Added: Missile Expansions and Power Bomb Expansions can now require the Missile Launcher and Main Power Bomb respectively. Expansions collected before their main item are retained and become usable once the main item is acquired.
+
 ### Metroid Prime Hunters
 
 - Fixed: The Alimbic Turrets in Celestial Archives - Data Shrine 03 will now always respawn on room reload to prevent a softlock if the door to Synergy Core was not unlocked prior to leaving.
--- a/randovania/game_connection/connector/prime1_remote_connector.py
+++ b/randovania/game_connection/connector/prime1_remote_connector.py
@@ -12,6 +12,7 @@ from randovania.game_connection.executor.memory_operation import (
     MemoryOperationException,
     MemoryOperationExecutor,
 )
+from randovania.game_description.resources.inventory import InventoryItem
 from randovania.game_description.resources.item_resource_info import ItemResourceInfo
 from randovania.games.prime1.patcher import prime_items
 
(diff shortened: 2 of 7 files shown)


toasterparty (changes requested):
Two critical things to mention upfront:

1. In it's current state, this PR is broken because it relies on a feature which does not yet exist in randomprime (Proposed in [Randomprime PR #190](https://github.com/randovania/randomprime/pull/190)). It must not merge until randomprime is updated or the assumed patcher feature is removed.

2. Missing a changelog entry

reedwhaley:
Yep, both are fair.

I moved the randomprime dependency up front in the description so it's clear this can't merge before #190 is available.

The changelog entry was supposed to be in my final amend, but I just caught that the PR branch was still pointing at the earlier commit. I've pushed the corrected branch with the changelog included.

PR #9569, opened 2026-09-04
Use a memory pool to make allocations less expensive

henriquegemignani:
~7% faster for Dread DLR.

### DIFF
--- a/randovania/lib/native/pool.hpp
+++ b/randovania/lib/native/pool.hpp
@@ -1,0 +1,204 @@
+// Size-class free-list allocator for the Cython "native" extension modules.
+//
+// One Pool per compiled module (each .pyd gets its own `pool()` static). Slabs are bump-allocated
+// and never freed; freed blocks go back onto a per-size free list. No locking: always used under
+// the GIL.
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <cstring>
+#include <deque>
+#include <memory>
+#include <new>
+#include <type_traits>
+#include <vector>
+
+namespace rdv {
+
+class Pool {
+public:
+    static constexpr size_t kGranule = 16;
+    static constexpr size_t kSmallMax = 1024;
+    static constexpr size_t kNumClasses = kSmallMax / kGranule;
+    static constexpr size_t kSlabBytes = 64 * 1024;
+
+    struct Stats {
+        size_t allocations = 0;
+        size_t deallocations = 0;
+        size_t freelist_hits = 0;
+        size_t slabs_allocated = 0;
+        size_t bytes_from_slabs = 0;
+        size_t large_allocations = 0;
+    };
+
+    Pool() = default;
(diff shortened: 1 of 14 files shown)


randovania/lib/native/pool.hpp
```diff
@@ -0,0 +1,204 @@
...
+        Slab& slab = slabs_.back();
+        void* p = slab.data.get() + slab.used;
+        slab.used += bytes;
+        stats_.bytes_from_slabs += bytes;
+        return p;
+    }
+
+    Node* free_lists_[kNumClasses] = {};
+    // Real std::deque: keeps slab pointers stable as more are pushed. Must not go through the pool.
+    std::deque<Slab> slabs_;
```
ThanatosGit, line 125: Couldn't this just be a `std::vector` because I don't see a reason why the `Slab` itselfs must have a stable pointer. The pointer returned by `bump_allocate` is stable because it points to an address within the `data` buffer and the `data` buffer itself is never moved

randovania/lib/native/pool.hpp
```diff
@@ -0,0 +1,204 @@
...
+using pvector = std::vector<T, PoolAllocator<T>>;
+
+// FIFO queue over a single pool-allocated vector plus a head index; replaces std::deque, which
+// on MSVC allocates a new block roughly every 4 pushes.
+template <class T>
+class IndexQueue {
+public:
+    void push_back(const T& value) { items_.push_back(value); }
+
+    void pop_front() {
```
ThanatosGit, line 171: Don't know how much is done here for optimisation but in theory that allows to `pop_front` without pushing anything and then the `size()` is negative.

henriquegemignani: In the case you mentioned, we'd have two cases of handling it:
- C++ exception, converted into Python code by Cython
- bool return for success/failure
- debug mode assertion

The use case for `pop_front` is in main loop of the generator and resolver, to get the next node/path to check - a big hotspot for both, especially the resolver. It's also provably not empty there.

C++ exception catching is supposedly free when an exception doesn't happen, though it'd be very expensive when it does. Claude's suggestion is it'd break inlining by the compiler, which is significant given how small this function is.

The bool return is interesting, but very liable to just not bothering to check since it becomes a thing the caller needs to do.

Debug mode assertion is almost ideal, but we have zero workflows that'd enable that right now.

Given the performance-critical of where it is, I prefer to assert but I'm willing to do differently.

randovania/lib/native/pool.hpp
henriquegemignani: The reason is so you can take a pointer to a Slab and use it freely, like what `bump_allocate` does.
But it's also true that no current code uses the pointer in an unsafe way. It's that way to be safe against future changes as it can be tricky to catch errors.
Given the usage (always fetch the last one, few elements, infrequent append) deque is as good as vector.

All that said, I don't mind changing to vector.

PR #9582, opened 2026-09-08
Nintendont: simplify splitting ops

Miepee:


henriquegemignani (approved):
tests?

Miepee:
What tests are you referring to? The ones you wrote ages ago still exist and actually caused me to find issues with the first few commits.
Codecov says that only one line is missing tests which looks like is the case of `op.byte_count == 0`

henriquegemignani:
> What tests are you referring to? The ones you wrote ages ago still exist and actually caused me to find issues with the first few commits. Codecov says that only one line is missing tests which looks like is the case of `op.byte_count == 0`

Nah, just wondering if it needs any haha

PR #9587, opened 2026-09-13
Hunters - Add support for shuffling Octoliths anywhere

dyceron:
Also added some FAQ entries that are useful
image
image
image
image


dyceron:
Needs another patcher update hence the draft. But what's done so far doesn't rely on that at least except for bumping the version again.

randovania/games/prime_hunters/generator/bootstrap.py
```diff
@@ -45,6 +45,8 @@ def create_damage_state(self, game: GameDatabaseView, configuration: HuntersConf
     def assign_pool_results(
         self, rng: Random, configuration: HuntersConfiguration, patches: GamePatches, pool_results: PoolResults
     ) -> GamePatches:
+        if configuration.octoliths.prefer_anywhere:
+            return super().assign_pool_results(rng, configuration, patches, pool_results)
```
ThanatosGit, line 49: I have no clue about the game but I have to ask here: With `prefer_anywhere` active, it immediately returns with the `return super().assign_pool_results`.
This will completly skip your code below for the `pickups_to_preplace`, which looks correct to me, but it also skips the `if not configuration.shuffle_shield_keys` which is a different setting and it looks odd to me because that means the settings has no effect anymore.

CHANGELOG.md
```diff
@@ -26,10 +26,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Metroid Prime Hunters
 
+- **MAJOR** - Added: It is now possible to shuffle Octoliths anywhere in the game instead of being forced to only be in the Biodefense Chambers.
 - Fixed: The Alimbic Turrets in Celestial Archives - Data Shrine 03 will now always respawn on room reload to prevent a softlock if the door to Synergy Core was not unlocked prior to leaving.
 - Fixed: The boss door in all Biodefense Chambers will now unlock automatically when collecting the Octolith pickup.
 - Fixed: Sic Transit: The scan door will now unlock in the post-boss layers if the shuffled Shield Key is collected.
 - Fixed: Alinos Perch: The force field no longer despawns after defeating Cretaphid v3.
+- Fixed: It is no longer possible to accidentally skip certain triggers when collecting certain items.
```
ThanatosGit, line 34: That reads very vague but I assume it will be far too complicated or too lengthy to write it down more specific.

ThanatosGit (changes requested):

randovania/games/prime_hunters/generator/bootstrap.py
dyceron: Should I rearrange the order of this so the shield keys come first but leave that return where it is, then do the preplace?

ThanatosGit: I would just remove the early return and nest your code for preplacement of the octoliths like the one for the keys with a 
```python
if not configuration.octoliths.prefer_anywhere:
    do preplacement octoliths

if not configuration.shuffle_shield_keys:
   do preplacement shield keys

return super().assign....
```

Other games just do the early return for "anywhere mcguffins" in this method because they do nothing besides preplacing them or not. But Hunters can preplace two things.

CHANGELOG.md
dyceron: Yeah it's vague on purpose. It's technically the same thing as the biodefense chamber comment with the doors unlocking, so I could just combine those two into one entry.

randovania/games/prime_hunters/layout/preset_describer.py
```diff
@@ -18,11 +18,15 @@
...
 def describe_objective(octoliths: HuntersOctolithConfig) -> list[dict[str, bool]]:
     has_octoliths = octoliths.placed_octoliths > 0
+    placement = "Prefers Bosses" if octoliths.prefer_bosses else "Place at any item location"
     if has_octoliths:
         return [
             {
                 f"{octoliths.placed_octoliths} Octoliths": True,
             },
+            {
+                placement: octoliths.prefer_bosses or True,
```
ThanatosGit: lgtm besides that tiny code thing: A boolean `or True` is always just `True`.

ThanatosGit (approved):

randovania/games/prime_hunters/layout/preset_describer.py
dyceron: Yeah I was thinking that it could just say true. I'll fix it

